PR #4, opened 2021-12-08
Place all astro generated tables into a temporary schema. 

dimberman:
Since the `astro` library creates a lot of temporary tables, we want to place all of these tables into a single schema that would make it easy for a DBA to bulk delete unneeded tables. With this PR we can create a V1 of this feature, in which all tables are stored in a schema that is either user defined, or just called `tmp_astro`. In later iterations we hope to create the ability to clear temp tables after a DAG run completes or even create temporal dags (e.g. delete after a week).


bolkedebruin:
Just saying here that the title of this PR really confuses me. User generated tables temporary or not should never reside in the 'airflow' schema that's just stuff waiting to happen for a "OMG I deleted the Airflow database".

dimberman:
@bolkedebruin changed the name

pgzmnk:
Reviewed code and ran Postgres unit tests locally. Verified that unit tests created expected table under the `tmp_astro` schema (`\dt tmp_astro.*`).

This approach is great to segment and clean-up temporary tables. Superb job!

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -65,14 +67,22 @@ def execute(self, context):
         df = self._load_dataframe(self.path)
 
         # Retrieve conn type
+        conn = BaseHook.get_connection(self.output_table.conn_id)
+        if type(self.output_table) == TempTable:
```
kaxil, line 71: `isinstance(self.output_table, TempTable)` might be better

kaxil (commented):
looks good

PR #9, opened 2021-12-17
Add documentation around schemas and temptables

dimberman:


README.md
```diff
@@ -187,10 +188,60 @@ with dag:
     my_second_sql_transformation(my_table)
 ```
 
+### The output_table parameter
```
kentdanas, line 191: Are we planning on adding more here, or was this intended to be a heading one level higher?

dimberman: Ah good catch thank you

PR #22, opened 2021-12-27
Smart open

utkarsharma2:


pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "s3fs",
     "snowflake-sqlalchemy ==1.2.0",
     "snowflake-connector-python[pandas]",
+    "smart-open[all]>=5.2.1",
```
dimberman, line 26: Can you test if we can remove boto3 from these requirements now? smart open might handle that.

.github/workflows/ci.yaml
```diff
@@ -62,5 +62,6 @@ jobs:
           SNOWFLAKE_ACCOUNT: gp21411
           SNOWFLAKE_REGION: us-east-1
           SNOWFLAKE_ROLE: TRANSFORMER
+          GCP_CREDENTIALS: ${{ secrets.GCP_CREDENTIALS }}
```
dimberman, line 65: We should make this `AIRFLOW__ASTRO__GCP_CREDENTIALS` s.t. people can put this in their `airflow.cfg

dimberman (changes requested):

.github/workflows/ci.yaml
utkarsharma2: Sounds great will update it.

utkarsharma2: I think we need to replace the GOOGLE_APPLICATION_CREDENTIALS with AIRFLOW__ASTRO__GCP_CREDENTIALS, since env var GOOGLE_APPLICATION_CREDENTIALS is used while creating sessions in the code and GCP_CREDENTIALS is just used as a workaround for the json file content.

dimberman (approved):

PR #25, opened 2022-01-06
Stats check

utkarsharma2:


tests/operators/test_agnostic_stats_check.py
```diff
@@ -0,0 +1,235 @@
...
+import astro.sql as aql
+from astro.sql.table import Table
+
+log = logging.getLogger(__name__)
+DEFAULT_DATE = timezone.datetime(2016, 1, 1)
+
+
+class TestStatsCheckOperator(unittest.TestCase):
+    """
+    Test Boolean Check Operator.
```
dimberman, line 27:
```suggestion
    Test Stats Check Operator.
```

integration-tests/test_load.py
```diff
@@ -1,14 +0,0 @@
-import timeit
```
dimberman, line 1: Why are you deleting this?

src/astro/sql/operators/agnostic_stats_check.py
```diff
@@ -0,0 +1,392 @@
...
+        accepted_std_div: int,
+        threshold: float,
+    ) -> None:
+        self.name = name
+        self.columns_map = columns_map
+        self.accepted_std_div = accepted_std_div
+        self.threshold = threshold
+
+
+class ChecksHandler:
```
dimberman, line 34: Placing this into its own class is a really great call. Will make it easier if we decide to change how we handle checks in the future.

src/astro/sql/operators/agnostic_stats_check.py
```diff
@@ -0,0 +1,392 @@
...
+        )
+
+    def prepare_main_stats_snowflake_sql(self, main_table: Table):
+        select_expressions = []
+        for check in self.checks:
+            for key in check.columns_map:
+                stats_query = "stddev({key}) as {check_name}_{key}_stddev, avg({key}) as {check_name}_{key}_avg"
+                replacements = {"{key}": key, "{check_name}": check.name}
+                for key, val in replacements.items():
+                    stats_query = stats_query.replace(key, val)
```
dimberman, line 67: Do we need to worry about the column names being SQL escapes as well?

dimberman (changes requested):

src/astro/sql/operators/agnostic_stats_check.py
utkarsharma2: I'm not sure entirely, but they are used with 'AS' keyword anything besides an identifier should raise an syntax error.

integration-tests/test_load.py
utkarsharma2: Accidentally deleted it. Adding it back.

utkarsharma2: Also, In current state this fails on my local. Does this work for you while running UT?

dimberman: This doesn't need to run, it's just a tool I use for scale testing. I can fix it in a different PR

dimberman (approved):

PR #27, opened 2022-01-13
Improve GCS authentication setup

tatiana:
Improve GCS authentication setup.

Before, to use **astro** with GCS, we forced users to:
1. Have a GCP service account with the `Basic GCS role`
2. Declare `AIRFLOW__ASTRO__GOOGLE_APPLICATION_CREDENTIALS`

Although these can be seen as good practices, users might prefer to declare their GCP credentials using other standard ways supported by the `google-cloud-storage` package, including, but not exclusively, the ones documented in:
https://cloud.google.com/storage/docs/reference/libraries#setting_up_authentication

The current version allows GCP users to use **astro** from their local dev environments without setting up service accounts, by running:
```
gcloud auth application-default login
```

### DIFF
--- a/src/astro/sql/operators/agnostic_load_file.py
+++ b/src/astro/sql/operators/agnostic_load_file.py
@@ -140,10 +140,13 @@ class AgnosticLoadFile(BaseOperator):
         """
         get GCS credentials for storage.
         """
-        service_account_path = os.environ[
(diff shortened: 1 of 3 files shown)


utkarsharma2 (approved):
looking good.

PR #29, opened 2022-01-17
Add support to load_file from json/ndjson

tatiana:
This PR changes the approach towards testing in `test_agnostic_load_file.py`.

Before, all the tests would heavily rely on `unittest.TestCase`. Since we are using Pytest, this can be a bit limiting (https://docs.pytest.org/en/6.2.x/unittest.html#pytest-features-in-unittest-testcase-subclasses), since it doesn't allow us to use features such as [parametrization](https://docs.pytest.org/en/6.2.x/example/parametrize.html).

I believe we can benefit by using parametrization, since it has the perspective of allowing us to have less code to maintain, while having a pattern to validate our combinations of parameters (eg. multiple input file types, databases, multiple transport layers, etc).

Work in progress: I'll extent the scope of this PR so we also cover writing to `json` / `ndjson`.

### DIFF
--- a/tests/operators/test_agnostic_load_file.py
+++ b/tests/operators/test_agnostic_load_file.py
@@ -23,13 +23,14 @@ Run test:
     python3 -m unittest tests.operators.test_agnostic_load_file.TestAgnosticLoadFile.test_aql_local_file_to_postgres
 
 """
-
+import copy
 import logging
 import os
 import pathlib
 import unittest.mock
 
 import pandas as pd
+import pytest
 import utils as test_utils
 from airflow.exceptions import DuplicateTaskIdFound
 from airflow.models import DAG, DagRun
@@ -50,6 +51,29 @@ from astro.sql.table import Table, TempTable
 
 log = logging.getLogger(__name__)
 DEFAULT_DATE = timezone.datetime(2016, 1, 1)
+OUTPUT_TABLE_NAME = test_utils.get_table_name("load_file_test_table")
+OUTPUT_SCHEMA = os.getenv("SNOWFLAKE_SCHEMA")
+CWD = pathlib.Path(__file__).parent
+
+DEFAULT_SCHEMA = "tmp_astro"
+SQL_SERVER_HOOK_PARAMETERS = {
+    "snowflake": {
+        "snowflake_conn_id": "snowflake_conn",
+        "schema": os.getenv("SNOWFLAKE_SCHEMA"),
+        "database": os.getenv("SNOWFLAKE_DATABASE"),
+        "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
+    },
+    "postgres": {"postgres_conn_id": "postgres_conn"},
+}
+SQL_SERVER_CONNECTION_KEY = {
+    "snowflake": "snowflake_conn_id",
+    "postgres": "postgres_conn_id",
+}
+
+SQL_SERVER_HOOK_CLASS = {
+    "snowflake": SnowflakeHook,
+    "postgres": TempPostgresHook,
+}
 
 
 def drop_table_postgres(table_name, postgres_conn):
@@ -65,7 +89,6 @@ class TestAgnosticLoadFile(unittest.TestCase):
     Test agnostic load file.
     """
 
-    cwd = pathlib.Path(__file__).parent
     gcs_creds_filename = "gcp_credentials.json"
     bucket_name = "dag-authoring"
     blob_file_name = "homes.csv"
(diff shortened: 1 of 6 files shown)


tests/operators/test_agnostic_load_file.py
```diff
@@ -466,47 +468,94 @@ def test_aql_gcs_file_to_postgres(self):
...
+        POSTGRES_TABLE_NAME = "test_dag_load_file_homes_csv_2"
+        hook = TempPostgresHook(postgres_conn_id="postgres_conn", schema="pagila")
+        hook.run(f"DROP TABLE IF EXISTS tmp_astro.{OUTPUT_TABLE_NAME}")
+        yield hook
+        hook.run(f"DROP TABLE IF EXISTS tmp_astro.{OUTPUT_TABLE_NAME}")
+    else:
+        raise ValueError(f"Unsupported database {request.param}")
+
+
+@pytest.mark.parametrize("hook", ["snowflake", "postgres"], indirect=True)
```
utkarsharma2, line 521: Would it make sense to parametrize on "output table" param instead of hooks? that way when we introduce multiple DB support we just need to add to the list of output table?

tests/operators/test_agnostic_load_file.py
```diff
@@ -466,47 +468,94 @@ def test_aql_gcs_file_to_postgres(self):
         )
         assert df.iloc[0].to_dict()["sell"] == 142.0
 
-    def test_aql_local_file_to_snowflake(self):
+
+@pytest.fixture
+def sample_dag():
+    yield DAG("test_dag", default_args={"owner": "airflow", "start_date": DEFAULT_DATE})
```
dimberman, line 474: Huh... today I learned you can yield in a function and then keep going. Is there a benefit to doing it this way?

tests/operators/test_agnostic_load_file.py
```diff
@@ -466,47 +468,94 @@ def test_aql_gcs_file_to_postgres(self):
...
+            "file_conn_id": "",
+            "output_table": Table(
+                table_name=OUTPUT_TABLE_NAME,
+                database=os.getenv("SNOWFLAKE_DATABASE"),
+                schema=OUTPUT_SCHEMA,
+                warehouse=os.getenv("SNOWFLAKE_WAREHOUSE"),
+                conn_id="snowflake_conn",
+            ),
         }
+        schema = OUTPUT_SCHEMA
```
dimberman, line 547: There seems to be a lot of repeated code here. Are we saving much if we still have to separate the logic in the function on a per db basis?

dimberman (changes requested):

tests/operators/test_agnostic_load_file.py
tatiana: My understanding is that this is the Pytest native way of doing fixtures (setup and teardown):
https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended
The first time I saw this, I thought it was strange. I guess I have gotten used to it since then - which may be good or not :joy: 

I believe the benefit, when compared to using `unittest.TestCase` traditional `setUp` and `tearDown` is only being able to use other Pytest features, as described in:
https://docs.pytest.org/en/6.2.x/unittest.html#pytest-features-in-unittest-testcase-subclasses

tests/operators/test_agnostic_load_file.py
tatiana: Excellent point, @dimberman; I just tried to improve the repetition - please, let me know your thoughts on the latest version!

tests/operators/test_agnostic_load_file.py
```diff
@@ -50,6 +51,29 @@
...
+
+DEFAULT_SCHEMA = "tmp_astro"
+SQL_SERVER_HOOK_PARAMETERS = {
+    "snowflake": {
+        "snowflake_conn_id": "snowflake_conn",
+        "schema": os.getenv("SNOWFLAKE_SCHEMA"),
+        "database": os.getenv("SNOWFLAKE_DATABASE"),
+        "warehouse": os.getenv("SNOWFLAKE_WAREHOUSE"),
+    },
+    "postgres": {"postgres_conn_id": "postgres_conn"},
```
dimberman, line 66: This looks much better. Much easier to add future DBs. I'd love to know if there was a way for us to set a flag with pytest to only run postgres or only run snowflake?

tatiana: @dimberman, this is an example if we want to run the `ndjson` + `snowflake` combination:
```
pytest  -v -k "test_load_file[ndjson-snowflake]"
```
And this is another example if we want to run all the tests which have `snowflake` - either in their name, or as a parameter from parametrize:
```
pytest  -v -k "test_load_file" -k snowflake
```
https://docs.pytest.org/en/6.2.x/example/markers.html#using-k-expr-to-select-tests-based-on-their-name

PR #43, opened 2022-01-19
README edits

jwitz:
Some suggestions and questions for the `astro` README. Would eventually love to get some of this info into an Airflow Guide or an Astronomer doc!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@
 The `astro` library is a collection of SQL and dataframe modules that you can use to write tasks without needing to define Airflow connections and metadata for each task. By design, `astro` modules automatically pass database contexts to your tasks, meaning that you can focus on writing Python and leave metadata definitions for load time.
 
 - **Without `astro`**: Database connections and context are defined in operators. This metadata does not pass automatically from task to task, meaning that you have to redefine it for each new task/ operator. SQL queries must be static. 
-- **With `astro`**: Databse connections and context are defined at load time. Metadata is automatically passed into functions, meaning that you can write generic queries in your task an run these queries in various contexts without knowing the context ahead of time. SQL queries can be templated with JINJA.
+- **With `astro`**: Database connections and context are defined at load time. Metadata is automatically passed into functions, meaning that you can write generic queries in your task and run these queries without knowing the context ahead of time. SQL queries can be templated with JINJA.
 
 # Philosophy
 
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ sql_schema=<new-temp-schema>
 ## Setting up SQL files
 
 When writing out a SQL DAG using `astro`, you can think of each SQL file as its own Airflow task. For example, if you wanted
-to aggregate orders, aggregate customers, and then join customers and orders you could havesh
+to aggregate orders, aggregate customers, and then join customers and orders, you could have the following directory of files:
 
 ```
 |
--- a/README.md
+++ b/README.md
@@ -295,7 +295,163 @@ with dag:
 For those who don't want to store their transformations in external SQL files or who want to create tranformation
 functions that are extendable and importable, we offer a rich python API that simplifies the SQL experience for the python engineer!
 
-Here is an example DAG of a SQL + Python workflow using `astro`. We'll break down how this DAG works in the subsequent sections.
+## Setting Input and Output Tables
+
+Before we can complete any transformations, we need to define a way to get our tables in and out of Airflow. We can do this by defining either `Table` or `TempTable` objects in the `input_table` and `output_table` parameters of our table instantiations.
+
+### The Table class
+
+To instantiate a table or bring in a table from a database into the `astro` ecosystem, you can pass a `Table` object into the class. This Table object will contain all of the metadata that's necessary for handling table creation between tasks. Once you define at in the beginning of your pipeline, `astro` can automatically pass that metadata along to downstream tasks.
+
+In the following example, we define the our table in the DAG instantiation. In each subsequent task, we only pass in an input table argument because `astro` automatically passes in the additional context from our original `input_table` parameter.
+
+```python
+from astro import sql as aql
+from astro.sql.table import Table
+
+
+@aql.transform
+def my_first_sql_transformation(input_table: Table):
+    return "SELECT * FROM {input_table}"
+
+
+@aql.transform
+def my_second_sql_transformation(input_table_2: Table):
+    return "SELECT * FROM {input_table_2}"
+
+
+with dag:
+    my_table = my_first_sql_transformation(
+        input_table=Table(table_name="foo", database="bar", conn_id="postgres_conn")
+    )
+    my_second_sql_transformation(my_table)
+```
+
+### The TempTable Class
+
+Following the traditional dev ops concept of [pets vs. cattle](http://cloudscaling.com/blog/cloud-computing/the-history-of-pets-vs-cattle/), you can decide whether the result of a function is a "pet" (e.g. a named table that you would want to reference later), or a "cattle" that can be deleted at any time for garbage collection. 
+
+If you want to ensure that the output of your task is a cattle, you can declare it as a nameless `TempTable`. This places the output into your temp schema, 
+which can be later bulk deleted. By default, all `aql.transform` functions will output to TempTables unless a `Table` object is used in the `output_table` 
+argument.
+
+In the following example DAG, we set an `output_table` to a nameless `TempTable` meaning that any output from this DAG will be deleted once the DAG completes. If we wanted to keep our output, we would simply update the parameter to instantiate a `Table` instead.
+
+
+```python
+from astro import sql as aql
+from astro.sql.table import Table, TempTable
+
+
+@aql.transform
+def my_first_sql_transformation(input_table: Table):
+    return "SELECT * FROM {input_table}"
+
+
+@aql.transform
+def my_second_sql_transformation(input_table_2: Table):
+    return "SELECT * FROM {input_table_2}"
+
+
+with dag:
+    my_table = my_first_sql_transformation(
+        input_table=Table(table_name="foo", database="bar", conn_id="postgres_conn"),
+        output_table=TempTable(database="bar", conn_id="postgres_conn"),
+    )
+    my_second_sql_transformation(my_table)
+```
+
+## Loading Data
+
+To create an ELT pipeline, users can first load CSV or parquet data from either local, S3, or GCS into a SQL database with the `load_sql` function. 
+To interact with S3, you must set an S3 Airflow connection in the `AIRFLOW__ASTRO__CONN_AWS_DEFAULT` environment variable.
+
+In the following example, we load data from S3 by specifying the path and connection ID for our S3 database in `aql.load_file`: 
+
+```python
+from astro import sql as aql
+from astro.sql.table import Table
+
+raw_orders = aql.load_file(
+    path="s3://my/s3/path.csv",
+    file_conn_id="my_s3_conn",
+    output_table=Table(table_name="my_table", conn_id="postgres_conn"),
+)
+```
+
+## Transform
+
+The `transform` function of the SQL decorator is the "T" of the ELT system. Each step of the transform pipeline creates a new table from the
+`SELECT` statement and enables tasks to pass those tables as if they were native Python objects. The following example DAG shows how we can quickly pass tables between tasks when completing a data transformation.
+
+
+```python
+@aql.transform
+def get_orders():
+    ...
+
+
+@aql.transform
+def get_customers():
+    ...
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,87 +41,94 @@
 </h3>
 <br/>
 
+# Overview
```
jwitz, line 44: I think that we need a more straightforward introduction to what `astro` is. I took my best shot at creating an Overview section here, but please add more / less detail depending on how you view this project. I also thought including a before/ after section would make it even more clear what value `astro` adds to your Airflow project.

README.md
```diff
@@ -41,87 +41,94 @@
...
+
+The `astro` library is a collection of SQL and dataframe modules that you can use to write tasks without needing to define Airflow connections and metadata for each task. By design, `astro` modules automatically pass database contexts to your tasks, meaning that you can focus on writing Python and leave metadata definitions for load time.
+
+- **Without `astro`**: Database connections and context are defined in operators. This metadata does not pass automatically from task to task, meaning that you have to redefine it for each new task/ operator. SQL queries must be static. 
+- **With `astro`**: Databse connections and context are defined at load time. Metadata is automatically passed into functions, meaning that you can write generic queries in your task an run these queries in various contexts without knowing the context ahead of time. SQL queries can be templated with JINJA.
```
jwitz, line 49:
```suggestion
- **With `astro`**: Database connections and context are defined at load time. Metadata is automatically passed into functions, meaning that you can write generic queries in your task an run these queries in various contexts without knowing the context ahead of time. SQL queries can be templated with JINJA.
```

README.md
```diff
@@ -146,13 +153,16 @@ SELECT c.customer_id, c.source, c.region, c.member_since,
 
 ### Defining dependencies
 
-As these SQL files are related to Airflow DAGs, there is an expectation that there should be dependencies so you can break up your SQL into
-multiple reproduceable steps. We offer two ways to define dependencies within an `astro` sql file.
+When running SQL queries in Airflow DAGs, you need to define dependencies that break up your SQL into
+multiple, reproduceable steps. We offer two ways to define dependencies within an `astro` SQL file:
```
jwitz, line 157:
```suggestion
multiple, reproducible steps. We offer two ways to define dependencies within an `astro` SQL file:
```

README.md
```diff
@@ -146,13 +153,16 @@ SELECT c.customer_id, c.source, c.region, c.member_since,
...
 
-The first way to define a dependency is by defining a variable via the `template_vars` option.
+The first way to define a dependency is by defining a variable via the `template_vars` frontmatter argument. We set `template_vars` so that our `join_customers_and_orders` query cannot run until both our `customers_table` and `agg_orders` queries finish. This data dependecy is equivalent task dependency in our DAG. The only difference is that we're defining it directly in our SQL instead of using Airflow's dependency operators.
```
jwitz, line 159:
```suggestion
The first way to define a dependency is by defining a variable via the `template_vars` frontmatter argument. We set `template_vars` so that our `join_customers_and_orders` query cannot run until both our `customers_table` and `agg_orders` queries finish. This data dependency is equivalent to a task dependency in our DAG. The only difference is that we're defining it directly in our SQL instead of using Airflow's dependency operators. 
```

README.md
```diff
@@ -288,110 +295,15 @@ with dag:
 For those who don't want to store their transformations in external SQL files or who want to create tranformation
```
jwitz, line 295:
```suggestion
For those who don't want to store their transformations in external SQL files or who want to create transformation
```

README.md
```diff
@@ -288,110 +295,15 @@ with dag:
...
-```
-
-
-## The output_table parameter
+Before we can complete any transformations, we need to define a way to get our tables in and out of Airflow. We can do this by defining either `Table` or `TempTable` objects in the `input_table` and `output_table` parameters of our table instantiations.
 
 ### The Table class
 
-To instantiate a table or bring in a table from a database into the `astro` ecosystem, you can pass a `Table` object into the class. This Table object will contain all necessary metadata to handle table creation between tasks. Once you define it in the beginning of your pipeline, `astro` can automatically pass that metadata along.
+To instantiate a table or bring in a table from a database into the `astro` ecosystem, you can pass a `Table` object into the class. This Table object will contain all of the metadata that's necessary for handling table creation between tasks. Once you define at in the beginning of your pipeline, `astro` can automatically pass that metadata along to downstream tasks.
```
jwitz, line 304:
```suggestion
To instantiate a table or bring in a table from a database into the `astro` ecosystem, you can pass a `Table` object into the class. This Table object will contain all of the metadata that's necessary for handling table creation between tasks. After you define a Table's metadata in the beginning of your pipeline, `astro` can automatically pass that metadata along to downstream tasks.
```

README.md
```diff
@@ -288,110 +295,15 @@ with dag:
...
+
+In the following example, we define the our table in the DAG instantiation. In each subsequent task, we only pass in an input table argument because `astro` automatically passes in the additional context from our original `input_table` parameter.
```
jwitz, line 306:
```suggestion
In the following example, we define our table in the DAG instantiation. In each subsequent task, we only pass in an input table argument because `astro` automatically passes in the additional context from our original `input_table` parameter.
```

README.md
```diff
@@ -146,13 +153,16 @@ SELECT c.customer_id, c.source, c.region, c.member_since,
 
 ### Defining dependencies
 
-As these SQL files are related to Airflow DAGs, there is an expectation that there should be dependencies so you can break up your SQL into
-multiple reproduceable steps. We offer two ways to define dependencies within an `astro` sql file.
+When running SQL queries in Airflow DAGs, you need to define dependencies that break up your SQL into
+multiple, reproducible steps. We offer two ways to define dependencies within an `astro` SQL file:
 
-The first way to define a dependency is by defining a variable via the `template_vars` option.
+The first way to define a dependency is by defining a variable via the `template_vars` frontmatter argument. We set `template_vars` so that our `join_customers_and_orders` query cannot run until both our `customers_table` and `agg_orders` queries finish. This data dependency is equivalent to a task dependency in our DAG. The only difference is that we're defining it directly in our SQL instead of using Airflow's dependency operators.
```
jwitz, line 159: We never state a second way?

README.md
```diff
@@ -41,87 +41,94 @@
...
+# join_customers_and_orders.sql
 SELECT c.customer_id, c.source, c.region, c.member_since,
         CASE WHEN purchase_count IS NULL THEN 0 ELSE 1 END AS recent_purchase
         FROM orders c LEFT OUTER JOIN customers p ON c.customer_id = p.customer_id
 ```
 
 ### Defining metadata
 
-Once your SQL is working as expected, you might want to define the database and schema for this query when running. 
+Once your SQL is working as expected, you might want to define the database and schema for the query when the query itself runs.
```
jwitz, line 126:
```suggestion
Once your SQL is working as expected, you might want to define the query's database and schema during its runtime.
```

README.md
```diff
@@ -41,87 +41,94 @@
...
 To expose this functionality while keeping your SQL easy to run in your favorite SQL notebook,
```
jwitz, line 127:
```suggestion
To configure this for Airflow while keeping your SQL easy to run in your favorite SQL notebook, 
```

README.md
```diff
@@ -41,87 +41,94 @@
...
 [astro]
-sql_schema=<your schema here>
+sql_schema=<new-temp-schema>
 ```
 
 ## Setting up SQL files
 
-When writing out a SQL DAG using `astro`, you can think of each SQL file as its own task. So for example if you wanted
-to aggregate orders, aggregate customers, and then join customers and orders you could have
+When writing out a SQL DAG using `astro`, you can think of each SQL file as its own task. For example, if you wanted
```
jwitz, line 105:
```suggestion
When writing out a SQL DAG using `astro`, you can think of each SQL file as its own Airflow task. For example, if you wanted
```

README.md
```diff
@@ -41,87 +41,94 @@
...
+to aggregate orders, aggregate customers, and then join customers and orders you could havesh
```
jwitz, line 106:
```suggestion
to aggregate orders, aggregate customers, and then join customers and orders, you could have the following directory of files:
```

README.md
```diff
@@ -41,87 +41,94 @@
...
+
 ```
 |
```
jwitz, line 109: Expose top level directory here to match up with the examples later on?

README.md
```diff
@@ -162,24 +172,21 @@ SELECT c.customer_id, c.source, c.region, c.member_since,
         FROM orders c LEFT OUTER JOIN customers p ON c.customer_id = p.customer_id
 ```
 
-In this example, we are setting the value `customers` to tie to the `customers_table.sql` file, 
-essentially setting up a task dependency by creating a data dependency.
+Here is a list of currently supported frontmatter argument:
```
jwitz, line 175:
```suggestion
Here is a list of supported frontmatter arguments:
```
Might make sense to have this in a "Reference" section outside of the guide component of this doc.

README.md
```diff
@@ -41,87 +41,94 @@
...
 ```
 |
 -- customers_table.sql
 -- orders_table.sql
 -- join_customers_and_orders.sql
 ```
 
-In each of these SQL files, you can create a table by having a standard `SELECT` statement (we will handle all of the
-temporary tables for you).
+In each of these SQL files, you can create a table by having a standard `SELECT` statement. `astro` handles creating all of the temporary tables.
```
jwitz, line 115: Not clear what table you're creating here. Is it the table that becomes our `output_table` in later suggestions?

README.md
```diff
@@ -215,12 +222,12 @@ with dag:
...
 
-### Passing on tables to subsequent tasks
-
 Once the `render` function completes, it returns a dictionary of all of the models based on the sql file name.
 If you have a sql file named `join_orders_and_customers.sql`, then the result would be stored in `model["join_orders_and_customers]`.
 
-Here is an example of passing a single model to a subsequent rendering (perhaps you want to separate your ingest and transforms).
+### Passing on tables to subsequent tasks
+
+The following DAG is an example of passing a single model to a subsequent rendering (perhaps you want to separate your ingest and transforms):
```
jwitz, line 230: The `**` example seems to be an exact copy of the previous example, pretty sure that's a typo.

README.md
```diff
@@ -256,7 +263,7 @@ with dag:
     )
 ```
 
-Finally, you can even pass the resulting tables into a python function that uses the `astro.dataframe` function and we'll
+You can even pass the resulting tables into a python function that uses the `astro.dataframe` to
 automatically convert your table into a dataframe (but we'll go into more detail on that further down).
```
jwitz, line 267:
```suggestion
automatically convert your table into a dataframe. We'll discuss how to do this more in "Using Astro as a Python Engineer".
```

README.md
```diff
@@ -521,26 +439,126 @@ with self.dag:
...
 
 ```python
 @aql.run_raw_sql
 def drop_table(table_to_drop):
     return "DROP TABLE IF EXISTS {table_to_drop}"
 ```
 
+## Putting it All Together 
+
+The following is a full example DAG of a SQL + Python workflow using `astro`.
```
jwitz, line 454: Can we add some more explanation to what this DAG is doing? What features from before are we using here?

dimberman:
cc: @kentendanas

dimberman (approved):

README.md
```diff
@@ -41,87 +41,94 @@
 </h3>
 <br/>
 
+# Overview
+
+The `astro` library is a collection of SQL and dataframe modules that you can use to write tasks without needing to define Airflow connections and metadata for each task. By design, `astro` modules automatically pass database contexts to your tasks, meaning that you can focus on writing Python and leave metadata definitions for load time.
```
jwitz, line 46: The `astro` library is a suite of tools for writing ETL and ELT workflows in Airflow. It lets SQL engineers focus on writing SQL, Python engineers focus on writing Python, and lets all engineers focus on data engineering instead of configuration. By design, `astro` modules automatically pass database contexts to your tasks, meaning that you can focus on writing code and leave metadata definitions for load time.

README.md
jwitz: ingest-models

README.md
jwitz: Make it clear that the table itself is automatically generated and can be referenced in other SQL queries by the data dependency alone

README.md
jwitz, line 159: Leave as "the one way"

README.md
```diff
@@ -256,8 +263,8 @@ with dag:
     )
```
jwitz, line 263: **ingest_models,

README.md
jwitz: Pulling data from S3, running SQL transformations to merge with existing data, and moving that result into a dataframe so that you can do complex work on it with Python / ML

PR #44, opened 2022-01-20
Add output table option to SQL files

dimberman:


src/astro/sql/table.py
```diff
@@ -37,7 +37,7 @@ def __str__(self):
 
 
 class TempTable(Table):
-    def __init__(self, conn_id, database, warehouse=""):
+    def __init__(self, conn_id=None, database=None, warehouse=""):
```
tatiana, line 40: Which is the use-case to have `conn_id=None`?

tatiana (commented):
Hi @dimberman , that's great - with this proposal, you address one of Roger's concerns, allowing us to have a plain SQL statement - and handling templating in the header.
I have a minor question inline.

src/astro/sql/table.py
dimberman: @tatiana the case where conn_id is none is the case where you inherit the conn_id from the previous task. We might want to decide at runtime what conn_id to use.

PR #45, opened 2022-01-20
Add the ability to use load_file from a YAML file

dimberman:


dimberman:
After discussing with @tatiana, we are temporarily closing this line of work as the python API should be "good enough" to start (We can revisit this at a future date).

PR #48, opened 2022-01-21
Readme fixes

dimberman:


README.md
```diff
@@ -41,87 +41,94 @@
 </h3>
 <br/>
 
+# Overview
+
+The `astro` library is a collection of SQL and dataframe modules that you can use to write tasks without needing to define Airflow connections and metadata for each task. By design, `astro` modules automatically pass database contexts to your tasks, meaning that you can focus on writing Python and leave metadata definitions for load time.
```
jwitz, line 46:
```suggestion
The astro library is a suite of tools for writing ETL and ELT workflows in Airflow. It lets SQL engineers focus on writing SQL, Python engineers focus on writing Python, and all engineers focus on data engineering instead of configuration. By design, `astro` modules automatically pass database contexts to your tasks, meaning that you can focus on writing code and leave metadata definitions for load time.
```

README.md
```diff
@@ -41,87 +41,94 @@
...
 ```
 
 ## Setting up SQL files
 
-When writing out a SQL DAG using `astro`, you can think of each SQL file as its own task. So for example if you wanted
-to aggregate orders, aggregate customers, and then join customers and orders you could have
+When writing out a SQL DAG using `astro`, you can think of each SQL file as its own Airflow task. For example, if you wanted
+to aggregate orders, aggregate customers, and then join customers and orders, you could have the following directory of files:
+
 ```
```
jwitz, line 108: Add `ingest-models` to this folder structure

README.md
```diff
@@ -41,87 +41,94 @@
...
 |
 -- customers_table.sql
 -- orders_table.sql
 -- join_customers_and_orders.sql
 ```
 
-In each of these SQL files, you can create a table by having a standard `SELECT` statement (we will handle all of the
-temporary tables for you).
+In each of these SQL files, you can create a table by having a standard `SELECT` statement. `astro` handles creating all of the temporary tables.
```
jwitz, line 115:
```suggestion
In each of these SQL files, your standard `SELECT` statement automatically creates a table that can be referenced in downstream SQL files via a data dependency. `astro` handles creating all of the temporary tables required for this process.
```

README.md
```diff
@@ -146,13 +153,16 @@ SELECT c.customer_id, c.source, c.region, c.member_since,
 
 ### Defining dependencies
 
-As these SQL files are related to Airflow DAGs, there is an expectation that there should be dependencies so you can break up your SQL into
-multiple reproduceable steps. We offer two ways to define dependencies within an `astro` sql file.
+When running SQL queries in Airflow DAGs, you need to define dependencies that break up your SQL into
+multiple, reproducible steps. We offer two ways to define dependencies within an `astro` SQL file:
 
-The first way to define a dependency is by defining a variable via the `template_vars` option.
+The first way to define a dependency is by defining a variable via the `template_vars` frontmatter argument. We set `template_vars` so that our `join_customers_and_orders` query cannot run until both our `customers_table` and `agg_orders` queries finish. This data dependency is equivalent to a task dependency in our DAG. The only difference is that we're defining it directly in our SQL instead of using Airflow's dependency operators.
```
jwitz, line 159:
```suggestion
You can define your dependency via the `template_vars` frontmatter argument. In the following example, we set `template_vars` so that our `join_customers_and_orders` query cannot run until both our `customers_table` and `agg_orders` queries finish. This data dependency is equivalent to a task dependency in our DAG. The only difference is that we're defining it directly in our SQL instead of using Airflow's dependency operators. 
```

README.md
```diff
@@ -256,8 +263,8 @@ with dag:
     )
```
jwitz, line 263: Can't edit this branch, but `orders_and_customers=ingest_models["join_orders_and_customers"],` should be replaced with `**ingest_models`.

README.md
```diff
@@ -521,26 +439,126 @@ with self.dag:
...
 ```
 
-## Raw SQL
-Most ETL use-cases can be addressed by cross-sharing Task outputs, as shown above with `@aql.transform`. For SQL operations that _don't_ return tables but might take tables as arguments, there is `@aql.run_raw_sql`. 
+### Raw SQL
+
+Most ETL use cases can be addressed by cross-sharing task outputs, as shown above with `@aql.transform`. If you need to perform SQL operation that doesn't return a table but might take a table as an argument, you can use `@aql.run_raw_sql`.
```
jwitz, line 444:
```suggestion
Most ETL use cases can be addressed by cross-sharing task outputs, as shown above with `@aql.transform`. If you need to perform a SQL operation that doesn't return a table but might take a table as an argument, you can use `@aql.run_raw_sql`. 
```

README.md
```diff
@@ -521,26 +439,126 @@ with self.dag:
...
 
 ```python
 @aql.run_raw_sql
 def drop_table(table_to_drop):
     return "DROP TABLE IF EXISTS {table_to_drop}"
 ```
 
+## Putting it All Together 
+
+The following is a full example DAG of a SQL + Python workflow using `astro`.
```
jwitz, line 454:
```suggestion
The following is a full example DAG of a SQL + Python workflow using `astro`. We pull data from S3, run SQL transformations to merge our pulled data with existing data, and move the result of that merge into a dataframe so that we can complete complex work on it using Python / ML.
```

jwitz (commented):
@dimberman Just added suggestions based on what we discussed last friday! Can't merge/ edit this PR because I'm not authorized, but you should be able to apply the suggestions directly in a batch. This is now ready for a tech review pass.

README.md
dimberman: Wow I absolutely love this

Issue #38, opened 2022-01-19
Add basic support for Google BigQuery


dimberman:
Acceptance criteria:

aql.load_file
aql.save_file
aql.transform
aql.truncate
astro.dataframe

not required;
aql.transform_file
aql.append
aql.merge
All validation checks

For this first round, if the user attempts any of the "not required" functions, they will receive a warning that these functions are not yet supported

Users should able to load, save, and run queries against BigQuery using the AQL library. For this first round, we do not need to support validation checks or merge/append functions.

For this ticket, we expect implementaiton, test coverage, and documentation

PR #50, opened 2022-01-24
#38 Added support for bigQuery in load_file() operator

utkarsharma2:
1. Added support for bigQuery in load_file() operator.
2. Replaced to_sql() with to_gbq() due to a bug within pandas to_sql() function.
3. Fixed BigQueryHook issue with a work around Temp hook.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ dependencies = [
     "apache-airflow >=2.0",
     "apache-airflow-providers-postgres",
     "apache-airflow-providers-snowflake",
+    "apache-airflow-providers-google",
     "python-frontmatter",
     "pandas >=1.3.4",
     "s3fs",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
     "snowflake-connector-python[pandas]",
     "smart-open[all]>=5.2.1",
     "SQLAlchemy == 1.4.28",
-    "snowflake-sqlalchemy"
 ]
 
 keywords = ["airflow", "provider", "astronomer", "sql", "decorator"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,8 @@ dependencies = [
     "snowflake-connector-python[pandas]",
     "smart-open[all]>=5.2.1",
     "SQLAlchemy == 1.4.28",
+    "snowflake-sqlalchemy",
+    "sqlalchemy-bigquery"
 ]
 
 keywords = ["airflow", "provider", "astronomer", "sql", "decorator"]
--- a/src/astro/sql/operators/sql_decorator.py
+++ b/src/astro/sql/operators/sql_decorator.py
@@ -26,6 +26,7 @@ from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
 from airflow.utils.db import provide_session
 from sqlalchemy.sql.functions import Function
 
+from astro.sql.operators.temp_hooks import TempBigQueryHook
 from astro.sql.table import Table, create_table_name
 from astro.utils import postgres_transform, snowflake_transform
 from astro.utils.load_dataframe import move_dataframe_to_sql
@@ -220,6 +221,13 @@ class SqlDecoratoratedOperator(DecoratedOperator):
             session_parameters=None,
         )
 
+    def get_bigquery_hook(self):
+        return TempBigQueryHook(
+            bigquery_conn_id=self.conn_id,
+            use_legacy_sql=False,
+            gcp_conn_id=self.conn_id,
+        )
+
     def get_postgres_hook(self):
         return PostgresHook(postgres_conn_id=self.conn_id, schema=self.database)
 
@@ -260,6 +268,7 @@ class SqlDecoratoratedOperator(DecoratedOperator):
             "snowflake": self.get_snow_hook(),
             "postgresql": self.get_postgres_hook(),
             "postgres": self.get_postgres_hook(),
+            "bigquery": self.get_bigquery_hook(),
         }[self.conn_type]
         return hook.get_sqlalchemy_engine()
 
--- a/src/astro/sql/operators/temp_hooks.py
+++ b/src/astro/sql/operators/temp_hooks.py
@@ -13,8 +13,13 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import warnings
+from typing import Dict, Optional, Sequence, Union
 from urllib.parse import quote_plus
 
+from airflow.hooks.dbapi import DbApiHook
+from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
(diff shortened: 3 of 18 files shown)


src/astro/sql/operators/temp_hooks.py
```diff
@@ -56,3 +61,46 @@ def get_uri(self) -> str:
         if self.schema:
             uri += self.schema
         return uri
+
+
+class TempBigQueryHook(BigQueryHook):
```
dimberman, line 66: @utkarsharma2 can you please document the bug you are working around so we can remove this once the upstream is fixed?

src/astro/utils/load_dataframe.py
```diff
@@ -44,6 +49,12 @@ def move_dataframe_to_sql(
             schema=schema,
             warehouse=warehouse,
         ),
+        "google_cloud_platform": BigQueryHook(
+            gcp_conn_id=conn_id, use_legacy_sql=False
+        ),
+        "bigquery": TempBigQueryHook(
+            bigquery_conn_id=conn_id, use_legacy_sql=False, gcp_conn_id=conn_id
+        ),
```
dimberman, line 57: What is the difference here?

dimberman (changes requested):

pyproject.toml
```diff
@@ -18,13 +18,13 @@ dependencies = [
...
     "apache-airflow-providers-postgres",
     "apache-airflow-providers-snowflake",
+    "apache-airflow-providers-google",
     "pandas >=1.3.4",
     "s3fs",
     "snowflake-sqlalchemy ==1.2.0",
     "snowflake-connector-python[pandas]",
     "smart-open[all]>=5.2.1",
     "SQLAlchemy == 1.4.28",
-    "snowflake-sqlalchemy"
```
dimberman, line 27: Why did you remove this:?

pyproject.toml
```diff
@@ -18,13 +18,13 @@ dependencies = [
     "apache-airflow >=2.0",
     "apache-airflow-providers-postgres",
     "apache-airflow-providers-snowflake",
+    "apache-airflow-providers-google",
     "pandas >=1.3.4",
     "s3fs",
     "snowflake-sqlalchemy ==1.2.0",
     "snowflake-connector-python[pandas]",
     "smart-open[all]>=5.2.1",
     "SQLAlchemy == 1.4.28",
```
dimberman, line 27: @ashb do we need to downgrade this?

dimberman (changes requested):

pyproject.toml
dimberman: Oh I see the double import nvm

src/astro/utils/load_dataframe.py
utkarsharma2: Ohh we don't need 'google_cloud_platform' removing it.

src/astro/sql/operators/temp_hooks.py
utkarsharma2: Sure, will add a comment.

pyproject.toml
utkarsharma2: I think we need to downgrade it to Airflow's sqlAlchemy version - 1.3, because with latest pip we cannot have conflicting dependencies.

pyproject.toml
```diff
@@ -18,11 +18,16 @@ dependencies = [
...
+    "apache-airflow-providers-google",
+    "python-frontmatter",
     "pandas >=1.3.4",
     "s3fs",
     "snowflake-sqlalchemy ==1.2.0",
     "snowflake-connector-python[pandas]",
     "smart-open[all]>=5.2.1",
+    "SQLAlchemy == 1.4.28",
+    "snowflake-sqlalchemy",
+    "sqlalchemy-bigquery"
```
tatiana, line 30: @utkarsharma2 @dimberman I'd love to get the view of both of you on our approach towards Astro's dependencies

I just created isolated virtualenvs for both `main` and `SQLAlchemy-demo` branches by running:
```
pip install .[dependencies]
```

Some of the differences were that in `SQLAlchemy-demo` we downgraded a few packages:
- `apache-airflow==2.2.3` -> `apache-airflow==2.0.1`
- `apache-airflow-providers-postgres==2.4.0` -> `apache-airflow-providers-postgres==1.0.2 `
- `apache-airflow-providers-postgres==2.4.0` -> `apache-airflow-providers-snowflake==1.3.0`

Was this downgrade be intentional? Could this affect Astro's behaviour/adoption in any sense?

 Also, in `main`, `pip freeze` lists 145 dependencies, and the branch   `SQLAlchemy-demo` has 192. Are we happy for Astro to have this amount of dependencies, or do we value fewer dependencies? There are successful open source projects which followed both paths of lots of dependencies / minimal dependencies (e.g. Plone and Django) - it would be great to understand where we stand.

src/astro/sql/operators/sql_decorator.py
```diff
@@ -94,23 +95,24 @@ def execute(self, context: Dict):
         self.run_id = context.get("run_id")
         self.convert_op_arg_dataframes()
         self.convert_op_kwarg_dataframes()
-        if not self.sql:
+        if self.sql is None or self.sql == "":
```
tatiana, line 98: @utkarsharma2 it may be worth adding a comment why we are not doing `not self.sql` - people less familiar with SQLAlchemy might be tempted to write as it was before

src/astro/sql/operators/temp_hooks.py
```diff
@@ -28,7 +33,7 @@ def get_uri(self) -> str:
         """Override DbApiHook get_uri method for get_sqlalchemy_engine()"""
         conn_config = self._get_conn_params()
         uri = (
-            "snowflake://{user}:{password}@{account}.{region}/{database}/{schema}"
```
tatiana, line 31: Does dropping `region` have any implications?

tatiana (commented):
Great progress on supporting BQ, @utkarsharma2 !
I left a few questions/comments inline.
What do you think if we had this PR based on `main` as opposed to `SQLAlchemy` branch?

tests/operators/test_agnostic_load_file.py
```diff
@@ -466,47 +497,84 @@ def test_aql_gcs_file_to_postgres(self):
...
+    },
+)
+def test_aws_decode():
+    from astro.utils.cloud_storage_creds import parse_s3_env_var
+
+    k, v = parse_s3_env_var()
+    assert v == "@#$%@$#ASDH@Ksd23%SD546"
+
+
+@pytest.mark.parametrize("sql_server", ["snowflake", "postgres"], indirect=True)
```
tatiana, line 551: @utkarsharma2 what do you think of adding BQ to this test?

dimberman: I think that sounds like a good idea.

README.md
```diff
@@ -42,19 +42,25 @@
 </h3>
 <br/>
 
+# Overview
```
dimberman, line 45: @utkarsharma2 can you fix this so that there aren't conflicts with the REDME on main?

dimberman (changes requested):

PR #51, opened 2022-01-25
Run a better way of testing DAGs

dimberman:
As we build out astro, we need to consider testing more complex flows
(e.g. tasks with multiple dependencies), and the current DAG running
system is extremely ad-hoc and unscalable. Running dag.run will allow us
to test all kinds of DAG configurations while reducing code.


tatiana (approved):
This is a great improvement, thank you very much, @dimberman !

PR #65, opened 2022-01-31
Fix bug where schema is not set for postgres

dimberman:
Fixes and tests for a bug where we do not set schema in the query for
postgres based SQL queries


tatiana (approved):
Thanks for fixing this, @dimberman !

PR #68, opened 2022-02-04
#69 Fix issue with inheritance in aql.render

dimberman:
There were three primary issues with aql.render where the DAGs were
showing up in the graph view properly but were not running properly when
tasks inherit from eachother.

1. We were not properly returning the result which meant that the Table
   object was not being saved to XCOM
2. (snowflake specific) There was an issue in how we would parse
   snowflake parameters that prevented them from properly overwriting
   parameters in the SQL statement.
3. There was a bug in snowflake when attempting to handle multiple
   schemas at the same time.

### DIFF
--- a/src/astro/sql/operators/agnostic_boolean_check.py
+++ b/src/astro/sql/operators/agnostic_boolean_check.py
@@ -120,7 +120,7 @@ class AgnosticBooleanCheck(SqlDecoratoratedOperator):
         self.handler = lambda curr: curr.fetchall()
 
         self.sql, self.parameters = execute_boolean_checks(
-            self.table.table_name, self.checks, self.conn_id  # type: ignore
+            self.table, self.checks, self.conn_id  # type: ignore
         )
 
         results = super().execute(context)
@@ -161,7 +161,9 @@ class AgnosticBooleanCheck(SqlDecoratoratedOperator):
         return failed_check_name, failed_check_index
 
     @staticmethod
-    def execute_postgres_boolean_checks(table: str, checks: List[Check], conn_id: str):
+    def execute_postgres_boolean_checks(
+        table: Table, checks: List[Check], conn_id: str
+    ):
         statement = (
             "SELECT {results} FROM (SELECT {expressions} From {table}) as temp_table"
         )
@@ -175,7 +177,9 @@ class AgnosticBooleanCheck(SqlDecoratoratedOperator):
         statement = (
             sql.SQL(statement)
             .format(
-                results=results, expressions=expressions, table=sql.Identifier(table)
+                results=results,
+                expressions=expressions,
+                table=sql.Identifier(table.table_name),
             )
             .as_string(hook.get_conn())
         )
@@ -183,9 +187,10 @@ class AgnosticBooleanCheck(SqlDecoratoratedOperator):
 
     @staticmethod
     def execute_snowflake_boolean_checks(table: str, checks: List[Check], conn_id: str):
-        statement = "SELECT {results} FROM (SELECT {expressions} From Identifier(%(table)s) as temp_table)"
+        statement = (
+            "SELECT {results} FROM (SELECT {expressions} From %(table)s as temp_table)"
+        )
 
-        statement = statement.replace("{table}", table)
         statement = statement.replace(
             "{expressions}",
             ",".join([check.get_snowflake_expression() for check in checks]),
@@ -211,9 +216,7 @@ class AgnosticBooleanCheck(SqlDecoratoratedOperator):
         return statement, {}
 
     def snowflake_prep_results(self, results):
-        statement = (
-            "SELECT * from Identifier(%(table)s) WHERE {expressions} LIMIT {limit}"
-        )
+        statement = "SELECT * from %(table)s WHERE {expressions} LIMIT {limit}"
         statement = statement.replace(
             "{expressions}",
             "AND".join([self.checks[index].expression for index in results]),
--- a/src/astro/sql/operators/agnostic_boolean_check.py
+++ b/src/astro/sql/operators/agnostic_boolean_check.py
@@ -65,8 +65,8 @@ class AgnosticBooleanCheck(SqlDecoratoratedOperator):
         **kwargs,
     ):
         """
-        :param table: table name
-        :type table: str
+        :param table: table to check
+        :type table: Table
         :param checks: check class object, which represent boolean expression
         :type checks: Check
         :param max_rows_returned: number of row returned if the check fails.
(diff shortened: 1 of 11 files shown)


src/astro/utils/snowflake_transform.py
```diff
@@ -20,7 +20,11 @@
 
 
 def handle_table(t: Table):
-    return t.schema + "." + t.table_name if t.schema else t.table_name
+    return (
+        t.database + "." + t.schema + "." + t.table_name
+        if t.schema and "." not in t.table_name
+        else t.table_name
```
tatiana, line 26: In which circumstances do we expect `table_name` to contain  `.`?

tests/operators/test_snowflake_decorator.py
```diff
@@ -142,7 +152,7 @@ def tearDown(self):
     def test_snowflake_query(self):
         @aql.transform
         def sample_snow(input_table: Table):
-            return f"SELECT * FROM {input_table.table_name} LIMIT 10"
+            return "SELECT * FROM {input_table} LIMIT 10"
```
tatiana, line 155: Why don't we need the `f-string` anymore?

tests/parsers/test_sql_directory_parser.py
```diff
@@ -98,3 +98,10 @@ def test_parse_frontmatter(self):
             new_customers_table.operator.sql
             == "SELECT * FROM {customers_table} WHERE member_since > DATEADD(day, -7, '{{ execution_date }}')"
         )
+
+    def test_parse_creates_xcom(self):
+        with self.dag:
+            rendered_tasks = aql.render(dir_path + "/single_task_dag")
+        from tests.operators import utils as test_utils
```
tatiana, line 105: What's the benefit in having the import statement from within the test function?

tests/parsers/test_sql_directory_parser.py
```diff
@@ -98,3 +98,10 @@ def test_parse_frontmatter(self):
...
+
+        test_utils.run_dag(self.dag)
```
tatiana, line 107: How are we confirming the table object was recorded in XCOM?
Are we missing any asserts in this test?

src/astro/sql/operators/agnostic_boolean_check.py
```diff
@@ -120,7 +120,7 @@ def execute(self, context: Dict):
         self.handler = lambda curr: curr.fetchall()
 
         self.sql, self.parameters = execute_boolean_checks(
-            self.table.table_name, self.checks, self.conn_id  # type: ignore
+            self.table, self.checks, self.conn_id  # type: ignore
```
tatiana, line 123: The docstring in lines 68-69 also needs to be updated

tests/operators/test_snowflake_decorator.py
```diff
@@ -166,10 +176,20 @@ def sample_snow(input_table: Table):
...
+                    schema=os.getenv("SNOWFLAKE_SCHEMA"),
+                    database=os.getenv("SNOWFLAKE_DATABASE"),
+                    warehouse=os.getenv("SNOWFLAKE_WAREHOUSE"),
+                ),
+            )
         test_utils.run_dag(self.dag)
 
         df = hook.get_pandas_df(
-            f'SELECT * FROM "{os.getenv("SNOWFLAKE_DATABASE")}"."{os.getenv("SNOWFLAKE_SCHEMA")}"."{self.snowflake_table}"'
+            f'SELECT * FROM "{os.getenv("SNOWFLAKE_DATABASE")}"."{os.getenv("SNOWFLAKE_SCHEMA")}"."{self.snow_inherit_table}"'
```
tatiana, line 192: Do we still have other tests covering the non-inheritance, more basic, use case?

tatiana (commented):
Thank you very much for identifying and fixing all those issues at the speed of light, @dimberman ! :rocket: 

Some questions inline

src/astro/sql/operators/sql_decorator.py
```diff
@@ -166,6 +166,10 @@ def handle_schema(self, output_table_name):
         """
         if self.conn_type == "postgres" and self.schema:
             output_table_name = self.schema + "." + output_table_name
+        elif self.conn_type == "snowflake" and self.schema and "." not in self.sql:
```
utkarsharma2, line 169: Why are we checking based on  - "." not in self.sql?

src/astro/utils/snowflake_append.py
```diff
@@ -34,11 +34,11 @@ def wrap_identifier(inp):
 
     if columns or casted_columns:
         statement = (
-            "INSERT INTO Identifier(%(main_table)s) ({main_cols}{sep}{main_casted_cols})"
-            "(SELECT {fields}{sep}{casted_fields} FROM Identifier(%(append_table)s))"
+            "INSERT INTO %(main_table)s ({main_cols}{sep}{main_casted_cols})"
```
utkarsharma2, line 37: why are we removing - Identifier()? this may lead to SQL injections right?

src/astro/utils/snowflake_transform.py
dimberman: @tatiana if the user provides a full tablename path. I've seen multiple test users in snowflake choose to identify the entire table name "DATABASE"."SCHEMA"."TABLE NAME"

tests/operators/test_snowflake_decorator.py
dimberman: @tatiana these were never supposed to be f-strings and by identifying them as f-strings users are actually opening themselves up to SQL injections. This raises a huge question over whether our current system looks too much like f-strings and therefore might promote bad habits.

tests/parsers/test_sql_directory_parser.py
dimberman: None I can move that

src/astro/sql/operators/sql_decorator.py
dimberman: Because if the user has already identified the schema and database in the table_name then we don't need to add it.

see here: https://github.com/astro-projects/astro/pull/68#discussion_r799533367

src/astro/utils/snowflake_append.py
dimberman: The identifier is actually now being added in in the sql_decorator.py. We now have it set that it can detect tables and automatically do that. We don't have that same detection for columns which is why those are still there.

src/astro/utils/snowflake_transform.py
dimberman: @tatiana I added a docstring and made it more explicit

dimberman: Actually let's just always do fully qualified name. I'm worried there will be edge cases with table names that have period in them.

tests/parsers/test_sql_directory_parser.py
tatiana: @dimberman I would feel more comfortable if the test confirmed that one task depends on the other and had some more explicit checks. If someone inadvertently changed the files within `dir_path + "/single_task_dag"`, this test would likely continue working - but we would not be testing the original problem.

PR #72, opened 2022-02-05
Fix converting SQL to DF if there is a '.' in table name.

dimberman:
When converting a postgres SQL table to a dataframe, there is a bug
where if the table has a . in it (e.g. if the task lives within a
taskgroup), then the way that psycopg2 reads the name with the schema
creates a string that sqlalchemy can not parse.

### DIFF
--- a/src/astro/sql/table.py
+++ b/src/astro/sql/table.py
@@ -36,8 +36,21 @@ class Table:
     def identifier_args(self):
         return (self.schema, self.table_name) if self.schema else (self.table_name,)
 
-    def qualified_name(self):
-        return self.schema + "." + self.table_name if self.schema else self.table_name
+    def fully_qualified_name(self, conn_type, schema=None):
+        """
+        In postgres, we set the schema in the query itself instead of as a query parameter.
+        This function adds the necessary {schema}.{table} notation.
+        :param output_table_name:
+        :param schema: an optional schema if the output_table has a schema set. Defaults to the temp schema
+        :return:
+        """
+        table_name = self.table_name
+        schema = self.schema or schema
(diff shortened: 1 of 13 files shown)


tests/parsers/test_sql_directory_parser.py
```diff
@@ -109,3 +109,22 @@ def test_parse_creates_xcom(self):
             rendered_tasks = aql.render(dir_path + "/single_task_dag")
 
         test_utils.run_dag(self.dag)
+
+    def test_parse_to_dataframe(self):
```
tatiana, line 113: Is the goal for this test to cover the original issue (a `task` defined within a `task_group` should still be loaded into a data frame)?

If the internal implementation of `render` stops using task group, we would no longer be covering the original problem. Would it be worth to have an explicit test using task groups without `render`?

src/astro/sql/table.py
```diff
@@ -29,8 +29,22 @@ def __init__(
...
+        schema.table in postgres and DATABASE.SCHEMA.TABLE in snowflake).
+        :param conn_type: What type of connection are we creating.
+        :param schema: an optional schema if the output_table has a schema set. Defaults to the temp schema
+        :return:
+        """
+        table_name = self.table_name
+        schema = self.schema or schema
+        if (conn_type == "postgres" or conn_type == "postgresql") and schema:
+            table_name = schema + "." + self.table_name
+        elif conn_type == "snowflake" and schema and "." not in self.table_name:
```
utkarsharma2, line 45: Can we validate "." case when someone creates table object? and raise an exception if someone adds it? That way we can standardize this part of our code?

dimberman:
Closing in favor of https://github.com/astro-projects/astro/pull/110

PR #90, opened 2022-02-08
fix: Chunking in load_file

tatiana:
* Enable `load_file` to use `chunksize`
* Set default chunk size to 2000, both to `load_file` and `@dataframe`

Without chunking, we are unable to load CSV files that are ~300MB into Postgres (it starves memory)

Solves: https://github.com/astro-projects/astro/issues/103


src/astro/constants.py
```diff
@@ -0,0 +1 @@
+DEFAULT_CHUNK_SIZE = 2000
```
tatiana, line 1: Change this to 1,000,000 based on the numbers:
https://github.com/astro-projects/astro/pull/105#discussion_r806605527

dimberman: @tatiana why does this need to be changed in a separate PR?

dimberman, line 1:
```suggestion
DEFAULT_CHUNK_SIZE = 1000000
```

PR #94, opened 2022-02-08
Convert templating system to feel more jinja native

dimberman:


tatiana (approved):
That looks really nice, thanks, @dimberman ! If possible, please, update the documentation so it uses the new templating standard.

utkarsharma2 (approved):

PR #113, opened 2022-02-16
Manage dependencies

### LINKED ISSUE #82: Only install dependencies the user needs
Currently, we make users install a LOT of dependencies (aws, google cloud, snowflake, etc.)

Users should have the ability to download dependencies 
e.g  "pip install astro[aws, snowflake]"

Acceptance criteria:
- Should support sub-modules: gcs, aws, snowflake, sqlite (if supported)
- Should fail gracefully if a user attempts to use a library that isn't installed (potentially including example install)

utkarsharma2:
1. Added conditional imports to manage import errors due potential package absence. 
2. Updated pyproject.toml to move packages to optional packages. Choose Postgres as default db for time being till we have support for sqlite. 

Resolves  #82


utkarsharma2:
@dimberman @tatiana  - Do you guys have any suggestion to test this with UT?

src/astro/utils/postgres_transform.py
```diff
@@ -15,7 +15,10 @@
 """
 import inspect
 
-from airflow.providers.postgres.hooks.postgres import PostgresHook
+try:
+    from airflow.providers.postgres.hooks.postgres import PostgresHook
+except ModuleNotFoundError:
+    from astro.utils.schema_util import RaiseException as PostgresHook
```
uranusjr, line 21: I think you need something like this instead?

```suggestion
try:
    from airflow.providers.postgres.hooks.postgres import PostgresHook
except ModuleNotFoundError:
    from astro.utils.schema_util import RaiseException

    PostgresHook = RaiseException("airflow.providers.postgres.hooks.postgres")
```

Also I’d change the name from `RaiseException`; at first glance I thought the class is subclassing `Exception` (it’s not).

pyproject.toml
```diff
@@ -16,17 +16,10 @@ license = {file = "LICENSE"}
...
-    "apache-airflow-providers-google",
     "python-frontmatter",
     "pandas >=1.3.4",
-    "s3fs",
-    "snowflake-sqlalchemy ==1.2.0",
-    "snowflake-connector-python[pandas]",
-    "smart-open[all]>=5.2.1",
     "SQLAlchemy==1.3.24",
-    "sqlalchemy-bigquery==1.3.0"
+    "apache-airflow-providers-postgres",
```
uranusjr, line 22: If a user needs Postgres support, they’d install `astro[postgres]` and pick up the Postgres provider from the extra, so this shouldn’t be needed.

utkarsharma2: So Postgre is default DB until we have support for sqlite.

src/astro/utils/postgres_transform.py
utkarsharma2: Thanks TP, I have made the changes you suggested.

src/astro/utils/dependencies.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class MissingPackage(object):
+    def __init__(self, module_name):
+        self.module_name = module_name
+
+    def __getattr__(self, item):
+        raise RuntimeError(
+            f"Error loading the module {self.module_name},"
+            f" please make sure all the dependencies are installed"
```
tatiana, line 25: What do you think if we gave a more helpful message, telling the users how to install the missing packages?

pyproject.toml
```diff
@@ -46,6 +39,46 @@ tests = [
...
+    "sqlalchemy-bigquery==1.3.0",
+    "smart-open[all]>=5.2.1",
+    "apache-airflow-providers-snowflake",
+    "snowflake-sqlalchemy ==1.2.0",
+    "snowflake-connector-python[pandas]",
+    "apache-airflow-providers-postgres",
+    "s3fs",
+    "smart-open[s3]>=5.2.1",
+]
+gcs = [
```
tatiana, line 52: Do these dependencies need to be within a `[project.optional-dependencies]` block?
https://www.python.org/dev/peps/pep-0631/#optional-dependencies

tatiana, line 52: it may be worth renaming `gcs` to `gcp`

tatiana:
@utkarsharma2 it's great to see progress in this stream!

1. Docs: we probably want to update the docs on how the installation changed, explaining how users installing Astro with pip can benefit from this feature. It's straightforward if we are installing locally, having the `pyproject.toml`, but it may be worth documenting to users downloading from PyPI using `pip install astro-projects[snowflake]` etc.

2. Tests: I agree we need to find a way of testing things before merging this feature. 
A strategy could be to mock the value of `some_module.__import__` so it raises `ModuleNotFoundError`, have you tried this?

pyproject.toml
utkarsharma2: @uranusjr can you please confirm this?

src/astro/utils/dependencies.py
utkarsharma2: Yes, I agree, I think we can point out the command to run in error message itself.
Like then they can do `pip install astro-project[aws]` if package related to AWS was missing.

pyproject.toml
uranusjr: Having a group in `optional-dependencies` (as is structured now) is enough to expose those as extras. The `pip install` should work.

src/astro/utils/dependencies.py
tatiana: That would be awesome, @utkarsharma2 !

pyproject.toml
```diff
@@ -47,6 +40,46 @@ tests = [
...
+    "snowflake-connector-python[pandas]",
+]
+postgres = [
+    "apache-airflow-providers-postgres",
+]
+aws = [
+    "s3fs",
+    "smart-open[s3]>=5.2.1",
+]
+all = [
```
kaxil, line 70: @uranusjr Does `all` keyword work with flit -- https://github.com/pypa/flit/issues/488

I wanted to use `flit` for https://github.com/astronomer/astronomer-providers/ but the following two issues made me think otherwise:

- https://github.com/pypa/flit/issues/525 (Minor, yet annoying)
- https://github.com/pypa/flit/issues/488

uranusjr: The keys here can be arbitrary, everything should work. (Well, if it is a valid extra name.)

uranusjr: Flit currently only puts license information in trove classifiers. The license field is pending standardisation. (If we really want to we can probably extend Flit to add that field to the wheel, but the value wouldn’t be very useful due to its lack of standardisation.)

The second issue you linked is requesting a “magic” extra that’s added automatically without declaring. IMO that’s not that good an idea, but manually declaring `all` should work just fine.

tatiana: Thank you very much, @kaxil & @uranusjr , I always learn a lot from your comments!

tatiana (approved):
It looks great, thank you very much @utkarsharma2 !

tatiana (approved):

Issue #100, opened 2022-02-10
Make Example DAGs work


dimberman:
Acceptance criteria:

* There should be CI/CD job that launches an airflow, and runs all example dags, and only passes if all DAGs pass. If there is a failure, it should surface the logs for the failed task. 
* There should be at least 4 example dags, showcasing a variety of use-cases (python-only, python with SQL files, loading from s3, running validation checks).

PR #114, opened 2022-02-17
#100 add integration tests around example dags

dimberman:


tests/test_example_dags.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.mark.parametrize(
+    "dag_id",
+    [
+        "astro_test_dag",
+        "sql_file_dag",
+        "demo_with_s3_and_csv",
+        "snowflake_animal_adoption_example",
+    ],
+)
+def test_example_dag(session, dag_id):
```
utkarsharma2, line 38: We were planning to run this in a periodic manner like a once a day or so, but this looks like part of CI?

tests/test_example_dags.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.fixture()
+def session():
+    return get_session()
+
+
+@pytest.mark.parametrize(
+    "dag_id",
+    [
```
tatiana, line 31: Would it be worth listing all the dags in the examples folder  (e.g. `glob.glob("../../example_dags/*.py`) instead of having a hard-coded list?

example_dags/demo_parse_directory/sql-directory.py
```diff
@@ -1,9 +1,11 @@
 import os
```
tatiana, line 1: Why do some of our example DAGs have hyphened names?

example_dags/demo_parse_directory/ingest_models/adoption_center_1.sql
```diff
@@ -0,0 +1,7 @@
+---
+conn_id: snowflake_conn
+database: DWH_LEGACY
```
tatiana, line 3: This doesn't work when I try to run the tests locally.
Would there be a way of us not having this in the SQL header?

tests/test_example_dags.py
utkarsharma2: I think we might want to pick and choose instead of running all dags?

tatiana: I believe all the DAGs in our examples folders should be functional - and my understanding is that this is the way we have to make sure they'll be working. My concern with having a hard-coded list here is that we may forget to update this test when new example DAGs are added - and they may stop working if we don't have an automatic way of validating them.

If we want to run specific tests for local development, we can always do that with `pytest -k`

example_dags/steven-dag.py
```diff
@@ -37,7 +37,7 @@
 }
 
 dag = DAG(
-    dag_id="pagila_dag",
+    dag_id="steven_dag",
```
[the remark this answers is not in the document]
dimberman: Oooo good call!

example_dags/demo_parse_directory/ingest_models/adoption_center_1.sql
dimberman: @tatiana I've added in the ability to place "default" values in the render function to address this. I've also changed this DAG to work on postgres. Now the only one that won't work locally is the snowflake one, but that will still pass in CI

tests/test_example_dags.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+
+    dag.clear(start_date=DEFAULT_DATE, end_date=DEFAULT_DATE, dag_run_state=State.NONE)
+
+    dag.run(
+        executor=DebugExecutor(),
+        start_date=DEFAULT_DATE,
+        end_date=DEFAULT_DATE,
+        run_at_least_once=True,
+    )
```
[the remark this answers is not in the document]
dimberman: The way these DAGs work I think it's sufficient for now. We have unit tests around DAG structures/run tests against DAG outcomes elsewhere.

tests/test_example_dags.py
dimberman: I've set the tests to run against all DAGs, but I think having this list of DAG IDs is the easiest way to parameterize. I'm not sure how we would go about getting these DAG ids otherwise.

dimberman: @tatiana I have added all DAGs to the test

tests/test_example_dags.py
dimberman: @utkarsharma2 yes this runs fairly quickly. I'm gonna attempt to split up the CI soon which should speed up everything considerably.

tests/test_example_dags.py
dimberman: @tatiana I tried to make it grab all DAGs from the dagbag which would've allowed this to automatically pick up new dags but for some reason that caused the CI to infinitely loop. I suppose we can just make sure that any new dags get added to the list?

tatiana: That's a shame. Please, let me know if you'd like to pair on this at some point - with a rubber duck, we may be able to find a way that works.
It is not a blocker, though. You did great work which deserves to be in our main branch! :tada:

tatiana (approved):

tests/test_example_dags.py
tatiana: Actually, it seems you solved it in this otherPR:

PR #119, opened 2022-02-18
Added support for SQLite

### LINKED ISSUE #86: Support for SQLite
To support easier local development and to create simpler quickstart tutorials, we should support SQLite.

Acceptance criteria:
- All current functions work against a SQLite database

utkarsharma2:
Resolves #86

### DIFF
--- a/ci-test-connections.yaml
+++ b/ci-test-connections.yaml
@@ -31,7 +31,7 @@ connections:
     schema: null
   - conn_id: sqlite_conn
     conn_type: sqlite
-    host: ./sqlite.db
+    host: ////sqlite.db
     schema:
     login:
     password:
--- a/ci-test-connections.yaml
+++ b/ci-test-connections.yaml
@@ -29,3 +29,9 @@ connections:
     password: null
     port: null
     schema: null
+  - conn_id: sqlite_conn
+    conn_type: sqlite
+    host: ./sqlite.db
+    schema:
+    login:
+    password:
--- a/ci-test-connections.yaml
+++ b/ci-test-connections.yaml
@@ -31,7 +31,7 @@ connections:
     schema: null
   - conn_id: sqlite_conn
     conn_type: sqlite
-    host: ////sqlite.db
+    host: ////tmp/sqlite.db
     schema:
     login:
     password:
--- a/src/astro/utils/load_dataframe.py
+++ b/src/astro/utils/load_dataframe.py
@@ -15,8 +15,8 @@ limitations under the License.
 """
 import os
 from typing import Optional, Union
-from airflow.hooks.sqlite_hook import SqliteHook
 
+from airflow.hooks.sqlite_hook import SqliteHook
 from pandas import DataFrame
 from pandas.io.sql import SQLDatabase
 from snowflake.connector import pandas_tools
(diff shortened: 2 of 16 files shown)


tests/operators/test_agnostic_save_file.py
```diff
@@ -290,6 +290,40 @@ def test_save_bigquery_table_to_local_file_exists_overwrite_false(self):
         # Delete output file after run
         os.remove(OUTPUT_FILE_PATH)
 
+    def test_save_sqlite_table_to_local_file_exists_overwrite_false(self):
```
tatiana, line 293: An alternative to adding 22 new lines of code could be to add 'sqlite'  as a database in the `test_save_file`? I noticed it is also missing BQ, it would be great if it could be covering all the databases supported! I already had adjusted the load_file one to include BQ, so it should be simple :crossed_fingers: 
https://github.com/astro-projects/astro/pull/119/files#diff-5692c92b876e45ee085b44ef344a19a0c40b9d29af9c03de199e98dff0fe8c0eR380

tests/operators/test_agnostic_load_file.py
```diff
@@ -270,6 +271,39 @@ def test_aql_local_file_to_postgres_no_table_name(self):
             "taxes": 3167.0,
         }
 
+    def test_aql_local_file_to_sqlite_no_table_name(self):
```
tatiana, line 274: How important is it to have a SQLite-specific test without a table name? Do we need this type of test for all types of DB?

tatiana, line 274: We could avoid these new 22 lines of code if we extended the `test_load_file` method to also validate `sqlite`:
https://github.com/astro-projects/astro/pull/119/files#diff-811d2b3637fa84cdde5a2e2ab5c85d15b16b80bf39f845dc082913cda5a5ea57R559

It already supports all the other DBs!

tatiana (commented):
I'm super excited about this feature, @utkarsharma2 ! It will allow us to have some very simple examples which will definitely help with onboarding users.

Some minor comments in-line

ci-test-connections.yaml
```diff
@@ -28,4 +28,10 @@ connections:
     login: null
     password: null
     port: null
-    schema: null
\ No newline at end of file
+    schema: null
+  - conn_id: sqlite_conn
+    conn_type: sqlite
+    host: ////sqlite.db
```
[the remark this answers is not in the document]
utkarsharma2: //// - means absolute path. 

So, the final path should be - ////tmp/sqlite.db

I was trying with ////sqlite.db, since that's the only dir I knew exist, seems like CI user don't have permission to create a file there.

tests/operators/test_agnostic_load_file.py
utkarsharma2: I was thinking of the same, but the queries passed to different DB are not same, we probably need to put that in a fixture as well.

tests/operators/test_agnostic_save_file.py
utkarsharma2: Sure, I'll update the PR

src/astro/utils/load_dataframe.py
```diff
@@ -83,6 +87,15 @@ def move_dataframe_to_sql(
             chunksize=chunksize,
             project_id=hook.project_id,
         )
+    elif conn_type == "sqlite":
```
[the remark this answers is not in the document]
utkarsharma2: schema is not required for sqlite, since it is only meant to be accessed by one app/user.

tests/operators/test_agnostic_load_file.py
tatiana: How different are they?

dimberman (approved):

PR #122, opened 2022-02-18
try splitting tests

dimberman:


tatiana (approved):
Hey @dimberman , it looks great. Thank you!
It would be great if you could add a comment to the PR giving the context of why we're making this change - reducing the time to run the tests - and, if possible the numbers :) 
If the commit message could also include what problem you're solving (reducing the test run time, if possible with how much), that would be great!

pyproject.toml
```diff
@@ -44,6 +45,7 @@ classifiers = [
 [project.optional-dependencies]
 tests = [
     "pytest >=6.0",
+    "pytest-split",
```
tatiana, line 48: How does `pytest-split` relate to `pytest-xdist` (https://github.com/pytest-dev/pytest-xdist)?

PR #130, opened 2022-02-23
Fix optional dependencies installation

### LINKED ISSUE #127: Allow users to specify their version of `snowflake_sqlalchemy`
**Context**
As of Astro 0.5.1, we pinned the `snowlflake_sqlalchemy` [dependency to 1.2.0](https://github.com/astro-projects/astro/blob/a11545e42bc47c871838a648fa3a0d0b77dd289c/pyproject.toml#L25)

This conflicts with users who want to use `snowflake_sqlalchemy==1.2.4` 

**Acceptance criteria**
* Users should be able to use `snowflake_sqlalchemy>=1.2.0,<=1.2.4` without having conflicts with Astro.
* Review Astro dependencies and identify those that we may be able to have a range of versions supported, as opposed to pinning a specific one.
* Ask Maggie (bug reporter) to review the PR

### LINKED ISSUE #135: Installation issue on release 0.5.2
How to reproduce:
```
pip install  astro-projects==0.5.2 (without snowflake)
```

Error message:
```
File "/Users/p/code/astronomer/astrobuild-repos/astrobuild-api-rest/.nox/pytest/lib/python3.9/site-packages/astro/utils/load_dataframe.py", line 21, in <module>
    from snowflake.connector import pandas_tools
ModuleNotFoundError: No module named 'snowflake'
```

tatiana:
Fix #135 
Fix #127 

Bonus:
* Includes some minor improvements to the package metadata.
* Removed duplicated dependencies in the project `tests` section.
* Normalised the name of examples, so their names highlight which dependencies they use and which features they illustrate
* Automated in the CI the validation of the installation of some of our optional dependencies

**Disclaimer: The individual commits of this branch should be kept when merging to the main branch.**

Note: some legacy tests rely on a pre-populated Snowflake table. Recently there were migrations in the Snowflake instance used by the CI, which led tests to fail. Part of this issue is being solved by #132 . Once that PR is merged, I'll rebase the current PR.


noxfile.py
```diff
@@ -15,17 +15,42 @@ def dev(session: nox.Session) -> None:
...
     session.install("-e", ".[tests]")
 
 
 @nox.session(python=["3.7", "3.8", "3.9"])
 def test(session: nox.Session) -> None:
     """Run unit tests."""
+    session.install("-e", ".[all]")
     session.install("-e", ".[tests]")
     session.run("airflow", "db", "init")
     session.run("pytest", *session.posargs)
```
dimberman, line 28: @tatiana so is the idea that we're running the example dags twice? Once with all deps and one with a subset of deps? this command will run those same examples.

dimberman (approved):

noxfile.py
tatiana: Exactly, @dimberman !
We still care about making sure things work when we have all the dependencies.

PR #132, opened 2022-02-23
fix our creds to use more secure snowflake DB

dimberman:


tatiana (approved):
Thank you very much for updating this @dimberman. The repo will soon be fully operational on @astronaut-chris new sandbox! :tada: 

I believe we should make our tests self-contained, if possible. If we could have fixtures to pre-populate Snowflake, we would probably not be running into the `ADOPTION_CENTER_1` issue - and this would allow users who may not have access to a specific Snowflake instance to be able to run the test suite. What are your thoughts, @dimberman ?

tatiana (changes requested):
@dimberman since I approved the scope of this PR changed a bit.

I believe we could benefit from separating the fix to issue #134 from fixing our CI credentials/tests.

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -65,7 +65,7 @@ def setUpClass(cls):
         cls.aggregate_table_bigquery = Table(
             "aggregate_check_test",
             conn_id="bigquery",
-            schema="tmp_astro",
+            schema="ASTROFLOW_CI",
```
tatiana, line 68: We can use this opportunity to remove the hard-coded schema from (at least the python) tests - so people from outside Astronomer who may not have the ability to create this schema may still run the tests.

tests/test_example_dags.py
```diff
@@ -29,6 +29,7 @@ def session():
 @pytest.mark.parametrize(
     "dag_id",
     [
+        "astro_homes_etl_dag",
```
tatiana, line 32: As part of #130 , I changed how we name the example DAGs. I suggest we use the same pattern. Also, it can help significantly debugging if the DAG filename matches the `dag_id` it creates.

PR #140, opened 2022-02-24
Allow users to define defaults on a per-DAG basis

dimberman:
Not all DAGs will want to have the same default database, schema, or
conn_id. This feature will allow users to define these features using
the dag_defaults

### DIFF
--- a/src/astro/sql/operators/agnostic_load_file.py
+++ b/src/astro/sql/operators/agnostic_load_file.py
@@ -73,6 +73,7 @@ class AgnosticLoadFile(BaseOperator):
         """
 
         # Read file with Pandas load method based on `file_type` (S3 or local).
+        self._set_default_values_from_dag_default(context)
         df = self._load_dataframe(self.path)
 
         # Retrieve conn type
@@ -99,6 +100,23 @@ class AgnosticLoadFile(BaseOperator):
         )
         self.log.info(f"returning table {self.output_table}")
         return self.output_table
+
+    def _set_default_values_from_dag_default(self, context):
+        if type(self.output_table) == TempTable:
+            self.output_table = self.output_table.to_table(
+                create_table_name(context=context), get_schema(context)
+            )
+        self.output_table.schema = self.output_table.schema or get_schema(context)
+        if context.get("dag"):
+            self.output_table.conn_id = self.output_table.conn_id or context[
+                "dag"
+            ].default_args.get("conn_id")
+            self.output_table.database = self.output_table.database or context[
+                "dag"
+            ].default_args.get("database")
+            self.output_table.warehouse = self.output_table.warehouse or context[
+                "dag"
+            ].default_args.get("warehouse")
 
     @staticmethod
     def validate_path(path):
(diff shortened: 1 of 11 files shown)


src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -97,6 +98,16 @@ def execute(self, context):
         self.log.info(f"returning table {self.output_table}")
         return self.output_table
 
+    def _set_default_values_from_dag_default(self, context):
+        self.schema = self.schema or get_schema(context)
```
tatiana, line 102: Why don't we currently try to retrieve the schema from the `context["dag"]` as well, like the other parameters?

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -130,6 +129,16 @@ def file_exists(self, output_file_path, output_conn_id=None):
         except IOError:
             return False
 
+    def _set_default_values_from_dag_default(self, context):
```
tatiana, line 132: Is this code the same as the one in `load_file`? Would it be worth it for them to define this method in a single place and have both reusing it?

tests/operators/test_postgres_decorator.py
```diff
@@ -362,10 +362,11 @@ def test_sql_file(self):
 
         cwd = pathlib.Path(__file__).parent
 
+        self.dag.default_args["conn_id"] = "postgres_conn"
+
         with self.dag:
             f = aql.transform_file(
                 sql=str(cwd) + "/test.sql",
-                conn_id="postgres_conn",
```
tatiana, line 368: do we still have a test covering the scenario where `conn_id` is defined? It might be worth renaming this test so people can be aware one of its goals is to validate using `transform_file` without `conn_id`

src/astro/utils/schema_util.py
```diff
@@ -35,8 +35,15 @@ def create_schema_query(conn_type, hook, schema_id, user):
         return f"CREATE SCHEMA IF NOT EXISTS {schema_id}"
 
 
-def get_schema():
-    return os.getenv("AIRFLOW__ASTRO__SQL_SCHEMA") or "tmp_astro"
+def get_schema(context=None):
```
tatiana, line 38: it feels we could have a unit test covering the new scenarios for this function

src/astro/sql/operators/sql_decorator.py
```diff
@@ -160,11 +161,21 @@ def execute(self, context: Dict):
...
 
+    def _set_default_values_from_dag_default(self, context):
+        self.schema = self.schema or get_schema(context)
+        if context.get("dag"):
+            self.conn_id = self.conn_id or context["dag"].default_args.get("conn_id")
+            self.database = self.database or context["dag"].default_args.get("database")
+            self.warehouse = self.warehouse or context["dag"].default_args.get(
+                "warehouse"
+            )
+            self.role = self.role or context["dag"].default_args.get("role")
```
tatiana, line 177: How are we testing this new behavior?

src/astro/sql/operators/agnostic_save_file.py
dimberman: @tatiana hte code is now different as load file changes the "output_table" and save file uses the "input table"

tatiana:
@dimberman what do you think of moving this to a draft PR, until we get back to this work?

Issue #124, opened 2022-02-21
Make database & schema optional in SQL files when using aql.render()


tatiana:
**Context**

At the moment, `database` and `schema` need to be specified in all SQL files used by aql.render(); otherwise, Astro (0.5.1) raises the following exception:

```
  File "/usr/local/lib/python3.9/site-packages/astro/utils/snowflake_transform.py", line 30, in _handle_table
        return t.database + "." + t.schema + "." + t.table_name
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

This means that the user has to specify the database and schema in all upstream and downstream tasks.

**Acceptance criteria**
* Evaluate the possible ways of solving this problem, and chose an approach. Some possible solutions could involve using a configuration file, an environment variable, or inheriting from the upstream task.
* Astro users are able to use `aql.render` without specifying `database` and `schema` headers in all SQL files.
* Ask for Maggie's review

PR #144, opened 2022-02-25
#124 Fix inheritance from partially defined output tables

dimberman:
This fixes an issue where the sql decorator doesn't pass on credentials
for partially defined or undefined output tables.


dimberman:
cc: @mag3141592

PR #147, opened 2022-02-28
Python and SQL SDK example DAGs

kentdanas:
@tatiana here are a couple of example DAGs to add to the project that can also be used for testing. They use the `homes` datasets in the `tests/` directory. We'll need to update the Snowflake connection/schema/database info, but I'm not sure what we use for this repo. Feel free to update, or let me know what we should change them to and I'll update :)


dimberman:
closing in favor of https://github.com/astro-projects/astro/pull/148

PR #148, opened 2022-03-01
Python and SQL SDK example DAGs

dimberman:


example_dags/example_snowflake_partial_table.py
```diff
@@ -8,39 +8,73 @@
 from astro.sql import append, load_file, transform
 from astro.sql.table import Table
```
tatiana, line 10: Perhaps this filename could include "append", so it is clear for users looking at example filenames a feature that is tested here.

tests/test_example_dags.py
```diff
@@ -30,6 +30,7 @@ def session():
         "example_amazon_s3_snowflake_transform",
         "example_postgres_render",
         "example_snowflake_partial_table",
+        "astroflow_homes_sql_dag",
```
tatiana, line 33: It would be interesting to follow the new naming convention  - so that the `nox` runs this example using a subset of the optional dependencies!

example_dags/astroflow_render_example.py
```diff
@@ -0,0 +1,58 @@
+import os
```
tatiana, line 1: What are your thoughts on naming this as `example_snowflake_render.py`, so we follow the current naming convention?

tatiana (approved):
Great work at spotting and fixing this issue, @kentdanas @dimberman !
I made some minor comments inline

example_dags/example_snowflake_render.py
```diff
@@ -0,0 +1,58 @@
...
+FILE_PATH = dir_path + "/data/"
+
+"""
+This DAG highlights using the render function to execute SQL queries.
+Here the render function results in three sequential tasks that run 
+queries in the /include/sql directory. The queries, combine, clean, and 
+aggregate data in Snowflake.
+
+To use the DAG, you must have two tables called "Homes" and "Homes2" in 
+your database. Data to populate these tables are located in csv's in the
```
kentdanas, line 21:
```suggestion
your database. The `load_file` tasks in this DAG load the data from homes csv's in the `data/` directory.
```

example_dags/example_snowflake_render.py
```diff
@@ -0,0 +1,58 @@
...
+
+"""
+This DAG highlights using the render function to execute SQL queries.
+Here the render function results in three sequential tasks that run 
+queries in the /include/sql directory. The queries, combine, clean, and 
+aggregate data in Snowflake.
+
+To use the DAG, you must have two tables called "Homes" and "Homes2" in 
+your database. Data to populate these tables are located in csv's in the 
+/include directory.
```
kentdanas, line 22:
```suggestion
```

kentdanas (approved):
@dimberman thanks for updating these!! I made on suggested change to a comment in the DAG based on your updates, but otherwise looks good to me.

PR #151, opened 2022-03-01
Fix capitalized cols

utkarsharma2:
Handled uppercase col name issue with SnowflakeDB.


dimberman:
@utkarsharma2 please add a fix to make the columns lowercase in sql_dataframe.py using `df.columns=df.colums.lower`

example_dags/example_snowflake_partial_table_with_append.py
```diff
@@ -111,7 +111,13 @@ def example_snowflake_partial_table_with_append():
     append(
         append_table=filtered_data,
         columns=["sell", "list", "variable", "value"],
-        main_table=Table(table_name="homes_reporting"),
+        main_table=Table(
```
utkarsharma2, line 114: @dimberman, I think we need to infer connection details from 1st table and pass it explicitly?

tatiana:
@utkarsharma2 does this address issue: #154?

utkarsharma2:
@tatiana - this is one of the solution, but the solution mentioned in ticket(#154) is generic one. @dimberman - can we merge this for the time being?

tatiana:
@utkarsharma2 the generic implementation was merged (#215), how should we proceed with this PR?

PR #160, opened 2022-03-04
Create release documentation

dimberman:
Create release guidelines s.t. other team members can craft releases.


dimberman:
cc: @kaxil

RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+git commit hashes and messages). Depending on the fix there might be some git conflicts to resolve. If you run into conflicts, please 
+resolve said conflicts, run `git add .`, and then `git cherry-pick --continue` to continue the merge.
+
+After cherry-picking all needed fixes, follow the instructions to [create a release from the release branch](#creating-a-release-from-the-release-branch)
+
+## Handling minor releases
+### When should I create a minor release?
+
+Minor release bumps should be for adding new features (as well as fixing any bugs or security issues). Notably, the difference between
+a minor and major release is that a minor release _should not break existing functionality_. A notable exception to this is that while
```
tatiana, line 31: @dimberman I think this fully makes sense once we have stabilized Astro's interface.
In my opinion, since Astro is in the alpha development phase, however, I believe it is still fine for us to introduce breaking changes. I believe that once we release 1.0.0, that's when we need to be more careful about breaking changes.

RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+Astro is in alpha we will allow breaking minor releases, however any release 1.0+ should only break functionality at major releases.
+
+It is, however, ok to put deprecation warnings on existing features in minor releases to alert users that a feature will go away in a future
+major release.
+
+### Creating a minor release
+
+To create a minor release, first create a new release branch based on main under the new minor release number. 
+If we are currently releasing from `release-0.6` then you should create the branch `release-0.7`. Since this branch is
```
tatiana, line 40: Just double-checking: so we'll have branches that are prefixed with `release` - but the tags and releases themselves won't have this prefix, right?

tatiana (approved):
That's very nice, @dimberman, I haven't used this strategy before, and I'd love to try it out!
Some minor comments are inline.

RELEASE.md
dimberman: @tatiana it says this ```A notable exception to this is that while
Astro is in alpha we will allow breaking minor releases, however any release 1.0+ should only break functionality at major releases.```

RELEASE.md
dimberman: Correct.

docs/RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Handling patch releases
+
+### When should I create a patch release?
+Patch releases are exclusively for bugs and and security fixes. Patch releases can have minor _enhancements_ to 
+existing features but really should not affect the API in any noticeable way (unless it is adding an argument). For any
+functionality changes or new endpoints, please create a minor or major release.
+
+A patch release should increment the third digit e.g. `0.6.9 -> 0.6.10`. This indicates to users that the release
```
utkarsharma2, line 11: @dimberman, `This indicates to users that the release` is this complete statement?

docs/RELEASE.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+### Creating a patch release
+
+When looking through the branches on the main repo, you will notice a series of `release` branches labeled
+`release-<major version>.<minor version>`. So if we are currently working on the 0.6 release, the branch should be labeled 
+`release-0.6`. Before creating a minor release please check the [milestones](https://github.com/astro-projects/astro/milestones)
+page to ensure that all relevent bug fixes have been PRed and merged. 
+
+Once you merge all expected fixes, the next step is to ensure that all fixes have been cherry-picked to the release branch from main (please use `git cherry-pick -x <commit id>` to retain
```
utkarsharma2, line 21: @dimberman, since we are referring to only commits that are fixes/security fix, would it be worth to add commits with certain category markers to indicate if they are normal dev commit or fix/security fix? or we gonna infer it from associated ticket?

PR #161, opened 2022-03-04
Directly save pandas Dataframes without requiring SQL database

### LINKED ISSUE #77: load_file & save_file from GCS/S3 to a Pandas data-frame
Acceptance criteria
- A user should be able to load a file from GCS/S3 into a dataframe without needing any SQL database connection.
- does not need to test with large scale data
- user should also be able to save the dataframe directly

dimberman:
Users who only want to work in pandas dataframes should be able to
save/load those dataframes from the local/cloud storage of their choice
Resolve  #77

### DIFF
--- a/example_dags/example_amazon_s3_postgres_load_and_save.py
+++ b/example_dags/example_amazon_s3_postgres_load_and_save.py
@@ -51,4 +51,6 @@ def example_amazon_s3_postgres_load_and_save():
     )
 
 
-dag = example_amazon_s3_postgres_load_and_save()
+example_amazon_s3_postgres_load_and_save_dag = (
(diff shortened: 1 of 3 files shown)


dimberman:
@tatiana @utkarsharma2 If we are going to allow people to save dataframes as well as tables, should we still call that variable `input_table`? Or should we have to variables `input_table/input_df` and just require that only one of them is populated?

tatiana:
@dimberman another option could be to have a parameter `input` which would handle both `tables` and `dataframes`, what are you thoughts about this possiblity?

tatiana (approved):
Awesome! :100:

PR #162, opened 2022-03-04
remove transform file

dimberman:
Now that we have aql.render, the transform_file function is redundant and should be removed as part of 1.0

### DIFF
--- a/src/astro/sql/__init__.py
+++ b/src/astro/sql/__init__.py
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import warnings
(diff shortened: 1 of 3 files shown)


src/astro/sql/__init__.py
```diff
@@ -63,6 +64,10 @@ def transform_file(
     warehouse: Optional[str] = None,
     output_table: Table = None,
 ):
+    warnings.warn(
+        "astro.sql.transform_file is now deprecated and will be taken out in version 1.0. Please use astro.sql.render instead"
+    )
```
tatiana, line 69: @dimberman since we are in an alpha version, and we have a disclaimer saying that the interface can change, wouldn't it make sense to just remove this feature now - as opposed to wait a stable release? This seems a less disruptive change than changing how we do query templating - and for that change we did not wait release `1.0.x`

dimberman: @tatiana fair enough, I'll just remove it

Issue #126, opened 2022-02-21
Allow users to customise the task name when using @df


tatiana:
**Context**

As of Astro 0.5.1, if a user calls multiple times a function decorated with `@df` (example below), the tasks will be automatically named `row_to_gcs_files__1`. At the moment, the user cannot override the default `task_id` name when using `@df`.

```
@df
def rows_to_gcs_files()
```
This behavior is different from the `@task` decorator, which allows users to specify `task_id`. We should make them consistent. This works:
```
@task(task_id = 'load_json')
```

**Acceptance criteria**
* Users are able to override the default `task_id` name when using the `df` decorator, similar to the `@task` decorator
* Ask @mag3141592 (bug reporter) to review the PR

PR #168, opened 2022-03-05
Allow users to customise the task name when using @df #126

dimberman:
Context

As of Astro 0.5.1, if a user calls multiple times a function decorated with @df (example below), the tasks will be automatically named row_to_gcs_files__1. At the moment, the user cannot override the default task_id name when using @df.

@df
def rows_to_gcs_files()
This behavior is different from the @task decorator, which allows users to specify task_id. We should make them consistent. This works:

@task(task_id = 'load_json')
Acceptance criteria

Users are able to override the default task_id name when using the df decorator, similar to the @task decorator
Ask @mag3141592 (bug reporter) to review the PR

### DIFF
--- a/src/astro/dataframe/__init__.py
+++ b/src/astro/dataframe/__init__.py
@@ -28,12 +28,14 @@ def dataframe(
     database: Optional[str] = None,
     schema: Optional[str] = None,
     warehouse: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


tatiana (approved):
That's great, thank you very much @dimberman !  :rocket:

PR #172, opened 2022-03-07
Fix Minor Doc & Code Issues

kaxil:
- Fixes Typos, incomplete sentences
- Missing `__init__.py` in `src/astro/sql/parsers`


.pre-commit-config.yaml
```diff
@@ -1,19 +1,3 @@
-# Licensed to the Apache Software Foundation (ASF) under one
```
kaxil, line 1: This project isn't under ASF

docs/RELEASE.md
```diff
@@ -4,21 +4,23 @@
 ## Handling patch releases
 
 ### When should I create a patch release?
-Patch releases are exclusively for bugs and and security fixes. Patch releases can have minor _enhancements_ to 
+Patch releases are exclusively for bugs and security fixes. Patch releases can have minor _enhancements_ to 
 existing features but really should not affect the API in any noticeable way (unless it is adding an argument). For any
 functionality changes or new endpoints, please create a minor or major release.
 
-A patch release should increment the third digit e.g. `0.6.9 -> 0.6.10`. This indicates to users that the release
```
kaxil, line 11: Was an incomplete sentence

.github/workflows/ci.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
         with:
           path: ~/.cache/pip
           key: ${{ hashFiles('pyproject.toml') }}
-      - run: cat ci-test-connections.yaml > test-connections.yaml
+      - run: cat .github/ci-test-connections.yaml > test-connections.yaml
```
tatiana, line 98: Since we moved it to be inside the `.github` folder, we could also rename the file to be `test-connections.yaml`

.pre-commit-config.yaml
tatiana: Should we need to add the Astronomer header here? How has it been done in the providers' package, @kaxil ?

tatiana (approved):
That's great, @kaxil , thank you very much for contributing to Astro! :1st_place_medal:

.pre-commit-config.yaml
kaxil: No, there is no IP in this file, so no point in adding license to it

kaxil: We also don’t need to add license to all files since we already have a top-level LICENSE file

.github/workflows/ci.yaml
tatiana: @kaxil I noticed the CI is breaking, it may be because of this change :thinking: 
I had a quick look at `conftest.py` and I couldn't spot why this is not working - @dimberman  could you give us a hand?

kaxil:
Given https://github.com/astro-projects/astro/pull/176 failed with the same error, I believe secrets are not propagated to the PRs from forks. Created https://github.com/astro-projects/astro/pull/178 as a result.

We should fix this so that Community members can contribute

tatiana:
@kaxil, absolutely, this is something we need to fix. I logged a ticket for us to address this:
https://github.com/astro-projects/astro/issues/179

It has a very high priority!

PR #173, opened 2022-03-08
Fix load_file support arg file_conn_id

### LINKED ISSUE #137: Unable to use `load_file` when using the argument `file_conn_id`
**Context**
Astro version: 0.5.1 and before

One of the arguments for `load_file` is `file_conn_id`, as we can see in the [code](
https://github.com/astro-projects/astro/blob/2b7c626cf741aa29b010e55028926fed58a44ec5/src/astro/sql/operators/agnostic_load_file.py#L57). However, if a user attempts to use it with either GCS or S3 at the moment, it does not work.

Within Astro, we are not using Airflow Hooks to interface with these - and the part of the code which [retrieves credentials for S3 and GCS](https://github.com/astro-projects/astro/blob/2b7c626cf741aa29b010e55028926fed58a44ec5/src/astro/utils/cloud_storage_creds.py#L38) does not take into account `file_conn_id`.

**Acceptance criteria**
* Have tests covering this scenario
* Fix `load_file` so it uses `file_conn_id`, if declared.
* Ask @guohui-gao (bug reporter) to review the change

tatiana:
Fixes  #137

This PR consists of two independent commits, which are expected to be kept when merging to main:
1. a refactor on the tests
2. the actual fix for #137

### DIFF
--- a/src/astro/utils/cloud_storage_creds.py
+++ b/src/astro/utils/cloud_storage_creds.py
@@ -8,6 +8,7 @@ from astro.utils.dependencies import (
     AwsBaseHook,
     BotoSession,
     GCSClient,
+    GCSHook,
     google_service_account,
 )
 
@@ -28,9 +29,8 @@ def s3fs_creds(conn_id=None):
     s3fs enables pandas to write to s3
     """
     if conn_id:
-        BaseHook.get_connection(
-            conn_id
-        )  # This line helps to raise a friendly exception
+        # The following line raises a friendly exception
+        BaseHook.get_connection(conn_id)
         aws_hook = AwsBaseHook(conn_id, client_type="S3")
         session = aws_hook.get_session()
     else:
@@ -46,14 +46,10 @@ def gcs_client(conn_id=None):
     """
     get GCS credentials for storage.
     """
-    credentials = None
     if conn_id:
-        extra = BaseHook.get_connection(conn_id).extra_dejson
-        json_file_path = extra["extra__google_cloud_platform__key_path"]
-        with open(json_file_path) as fp:
-            json_account_info = json.load(fp)
-            credentials = google_service_account.Credentials.from_service_account_info(
-                json_account_info
-            )
-    client = GCSClient(credentials=credentials)
+        gcs_hook = GCSHook(conn_id)
+        client = gcs_hook.get_conn()
+    else:
+        client = GCSClient()
+
     return dict(client=client)
--- a/src/astro/utils/dependencies.py
+++ b/src/astro/utils/dependencies.py
@@ -85,3 +85,8 @@ try:
     from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
 except ModuleNotFoundError:
     AwsBaseHook = MissingPackage("apache-airflow-providers-amazon", "amazon")
+
+try:
+    from airflow.providers.google.cloud.hooks.gcs import GCSHook
+except ModuleNotFoundError:
+    GCSHook = MissingPackage("apache-airflow-providers-google", "google")


tests/operators/test_agnostic_load_file.py
```diff
@@ -526,6 +531,86 @@ def test_load_file_templated_filename(sample_dag, sql_server):
...
+
+    task_params = {
+        "path": file_uri,
+        "file_conn_id": file_conn_id,
+        "output_table": Table(table_name=OUTPUT_TABLE_NAME, **sql_server_params),
+    }
+
+    test_utils.create_and_run_task(sample_dag, load_file, (), task_params)
+    df = sql_hook.get_pandas_df(f"SELECT * FROM {OUTPUT_TABLE_NAME}")
+    assert len(df) == 3
```
tatiana, line 611: Also, add a unit test to validate this behavior. Mark this test as integration.

src/astro/utils/cloud_storage_creds.py
```diff
@@ -15,24 +18,36 @@ def parse_s3_env_var():
...
-def s3fs_creds():
-    # To-do: reuse this method from sql decorator
+def s3fs_creds(conn_id=None):
     """Structure s3fs credentials from Airflow connection.
     s3fs enables pandas to write to s3
     """
-    # To-do: clean-up how S3 creds are passed to s3fs
+    if conn_id:
+        extra = BaseHook.get_connection(conn_id).extra_dejson
+        key = extra["aws_access_key_id"]
```
utkarsharma2, line 27: @tatiana  - Are we assuming below connection type for aws connection?
https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/connections/aws.html

src/astro/utils/cloud_storage_creds.py
```diff
@@ -15,24 +18,36 @@ def parse_s3_env_var():
...
-def gcs_client():
+def gcs_client(conn_id=None):
     """
     get GCS credentials for storage.
     """
-    client = GCSClient()
+    credentials = None
+    if conn_id:
+        extra = BaseHook.get_connection(conn_id).extra_dejson
+        json_file_path = extra["extra__google_cloud_platform__key_path"]
```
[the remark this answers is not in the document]
tatiana: https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html

src/astro/utils/cloud_storage_creds.py
tatiana: @utkarsharma2 Since the [S3Hook inherits](https://airflow.apache.org/docs/apache-airflow/1.10.6/_modules/airflow/hooks/S3_hook.html#S3Hook) from the base AWS Hook,  I tried to make it consistent with examples I found and [one of the currently supported ways of authenticating](https://airflow.apache.org/docs/apache-airflow-providers-amazon/1.1.0/_modules/airflow/providers/amazon/aws/hooks/base_aws.html) - so we could tie it to `smart_open`. Given the scope of this ticket, do you think we should change this implementation?

utkarsharma2: @tatiana, can we have something like this - https://github.com/apache/airflow/blob/99cf75d5fd07aad7db0a4290b1c7663485e999af/airflow/providers/amazon/aws/hooks/base_aws.py#L200 

or better if we can use this somehow?

utkarsharma2: If we just finalize approach here I can pick in my end of this task.

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -121,10 +121,10 @@ def _load_dataframe(self, path):
         file_type = path.split(".")[-1]
         file_scheme = urlparse(path).scheme
 
-        default_params_getter = lambda: None
+        default_params_getter = lambda conn_id: None
```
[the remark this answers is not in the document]
tatiana: At the moment, this line is used to create a function that returns `transport_params` if the file location is either the local filesystem or `http`.

We are not explicitly calling the functions, unless necessary, so we avoid raising the dependency exceptions.

src/astro/utils/cloud_storage_creds.py
tatiana: @utkarsharma2 the `BaseSessionFactory` isn't available in Airflow `2.2.4`, but I found another way - please, check it out!

utkarsharma2: LGTM.

utkarsharma2 (approved):

PR #174, opened 2022-03-08
Add full integration test DAG

dimberman:
To ensure full feature parity between data systems, we should develop integration DAGs that test all known scenarios, this is beyond just example DAGs as the DAGs will make it easy to switch databases by changing the input table params.

### DIFF
--- a/src/astro/sql/__init__.py
+++ b/src/astro/sql/__init__.py
@@ -89,7 +89,7 @@ def append(
         columns=columns,
         casted_columns=casted_columns,
         **kwargs,
-    )
+    ).output
 
 
 def merge(
--- a/tests/integration_test_dag.py
+++ b/tests/integration_test_dag.py
@@ -121,6 +121,13 @@ def run_dataframe_funcs(input_table: Table):
     compare(table_counts, df_counts)
 
 
+@aql.run_raw_sql
+def add_constraint(table: Table):
+    if table.conn_type == "sqlite":
+        return "CREATE UNIQUE INDEX unique_index ON {{table}}(list,sell)"
+    return "ALTER TABLE {{table}} ADD CONSTRAINT airflow UNIQUE (list,sell)"
+
+
 @task_group
 def run_append(output_specs: TempTable):
     load_main = aql.load_file(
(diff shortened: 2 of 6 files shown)


src/astro/sql/__init__.py
```diff
@@ -89,7 +89,7 @@ def append(
         columns=columns,
         casted_columns=casted_columns,
         **kwargs,
-    )
+    ).output
```
tatiana, line 92: @dimberman why we didn't have `output` before? Did it work?

tests/integration_test_dag.py
```diff
@@ -0,0 +1,210 @@
...
+    elif isinstance(hook, PostgresHook):
+        return TempTable(conn_id=hook.postgres_conn_id, database=hook.schema)
+    elif isinstance(hook, SqliteHook):
+        return TempTable(conn_id=hook.sqlite_conn_id, database="sqlite")
+    # elif isinstance(hook, BigQueryHook):
+    #     return TempTable(conn_id=hook.gcp_conn_id, database=)
+
+
+@pytest.fixture
+def merge_keys(sql_server):
```
tatiana, line 79: We might be overusing Pytest fixtures here - would it be possible to use a normal function?

tests/integration_test_dag.py
```diff
@@ -0,0 +1,210 @@
...
+    sql_name, _ = sql_server
+
+    if sql_name == "snowflake":
+        return {"list": "list", "sell": "sell"}
+    else:
+        return ["list", "sell"]
+
+
+@aql.transform
+def do_a_thing(input_table: Table):
```
tatiana, line 95: We could give more meaningful names to the functions, e.g.:
`apply_transform`

tests/integration_test_dag.py
```diff
@@ -0,0 +1,210 @@
+import pathlib
```
tatiana, line 1: Did you consider also adding `truncate`. `render` and `stats_check` the `integration_test_dag.py`?

tatiana (approved):
This is brilliant, @dimberman. :clap:  :clap: :clap: 

It will give us much more confidence in keeping things working in Astro.

I made some questions, but they are not a blocker for approving this ticket. We can have to follow up on dedicated issues to address them.

tests/integration_test_dag.py
```diff
@@ -0,0 +1,210 @@
...
+        return TempTable(
+            conn_id=hook.snowflake_conn_id,
+            database=hook.database,
+            warehouse=hook.warehouse,
+        )
+    elif isinstance(hook, PostgresHook):
+        return TempTable(conn_id=hook.postgres_conn_id, database=hook.schema)
+    elif isinstance(hook, SqliteHook):
+        return TempTable(conn_id=hook.sqlite_conn_id, database="sqlite")
+    # elif isinstance(hook, BigQueryHook):
```
tatiana, line 74: Shall we have a ticket to re-enable this?

src/astro/sql/__init__.py
dimberman: @tatiana without that `.output` there is no way to pass the result of the append to a validation function. Before the class was simply returning nothing. I'll make a ticket for this feature add, but this won't break any existing workflows.

tests/integration_test_dag.py
dimberman: @tatiana I want to add those in future PRs. I want to start with a working integration test DAG and then as we fix the test cases we will add all of those features in.

tests/integration_test_dag.py
dimberman: Fair enough, converted to function :)

tests/integration_test_dag.py
dimberman: Done

tests/integration_test_dag.py
dimberman: Bigquery now works with recent merges! We'll just need to keep checking as we add new features to this DAG.

PR #180, opened 2022-03-09
Fix merge tests

dimberman:
In order to shore up our integration testing abilities, and ensure that future databases can be added easily, I've created a test generator for our merge functionality. This will use pytest fixtures to generate scenarios that can be run across all databases.


tests/operators/test_agnostic_merge.py
```diff
@@ -0,0 +1,177 @@
...
+        return ""
+    return (
+        "ALTER TABLE {{table}} ADD CONSTRAINT airflow UNIQUE"
+        + f" ({','.join(columns)})"
+    )
+
+
+@adf
+def validate_results(df: pd.DataFrame, mode, sql_type):
+    # make columns lower and reverse due to snowflake defaulting to uppercase
```
tatiana, line 103: @dimberman what is the advantage of checking over columns as opposed to rows?

tests/operators/test_agnostic_merge.py
```diff
@@ -0,0 +1,177 @@
...
+        assert set_compare(df.taxes.to_list()[:-1], [3167.0, 4033.0, 1471.0, 3204.0])
+        assert set_compare(df.taxes.to_list()[:-1], [3167.0, 4033.0, 1471.0, 3204.0])
+        assert set_compare(df.list.to_list(), [160, 180, 132, 140, 240])
+        assert set_compare(df.sell.to_list()[:-1], [142, 175, 129, 138])
+    elif mode == "mixed":
+        assert df.taxes.to_list() == [1, 1, 1, 1, 1]
+        assert set_compare(df.age.to_list()[:-1], [60.0, 12.0, 41.0, 22.0])
+
+
+@task_group
```
tatiana, line 129: That's a neat way of testing them! :love_you_gesture:

conftest.py
```diff
@@ -32,3 +39,37 @@ def sample_dag():
...
+    hook_class = test_utils.SQL_SERVER_HOOK_CLASS.get(sql_name)
+    if hook_parameters is None or hook_class is None:
+        raise ValueError(f"Unsupported SQL server {sql_name}")
+    hook = hook_class(**hook_parameters)
+    schema = hook_parameters.get("schema", test_utils.DEFAULT_SCHEMA)
+    if not isinstance(hook, BigQueryHook):
+        hook.run(f"DROP TABLE IF EXISTS {schema}.{OUTPUT_TABLE_NAME}")
+    yield (sql_name, hook)
+    if not isinstance(hook, BigQueryHook):
+        hook.run(f"DROP TABLE IF EXISTS {schema}.{OUTPUT_TABLE_NAME}")
```
tatiana, line 75: This code seems to be the same as:
https://github.com/astro-projects/astro/blob/df8492fe3d40574cba72e56f7ab23f6725ad923d/tests/operators/test_agnostic_load_file.py#L473-L486

@dimberman what do you think of refactoring this, so both tests can share the same fixture?

tatiana (commented):
@dimberman this looks great!
I'm very happy the integration test pattern adopted in `load_file` is being spread across the project :tada: This will make Astro more reliable!
I have added some questions inline and one request - so we can avoid code duplication.
It would be great to link this PR to an issue, that details the bugs found.

conftest.py
dimberman: @tatiana great call, now both tests use the same code

tests/operators/test_agnostic_merge.py
dimberman: Thanks @tatiana! I'd love to eventually set up these task_group functions to be something we can mix and match for larger integration tests!

tests/operators/test_agnostic_merge.py
dimberman: @tatiana the advantage is that it's easier to track columns than rows in dataframes. The whole reason we're doing this is that unfortunately comparing dataframes with nan values is not supported.

PR #181, opened 2022-03-09
Updated validation tests

utkarsharma2:
Updated testcase to use dags for running unit test. 
1. Aggregate Checks 
2. Boolean Checks


tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -79,6 +82,16 @@ def setUpClass(cls):
...
+    def get_dag(self):
+        very_high_number = 99999999
+        return DAG(
+            "test_dag_" + str(random.randint(0, very_high_number)),
+            default_args={
+                "owner": "airflow",
+                "start_date": DEFAULT_DATE,
+            },
+        )
+
```
dimberman, line 94: So instead of using this, you should look into [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html).

If you look [here](https://github.com/astro-projects/astro/blob/main/conftest.py#L36-L41), you'll see that we have a fixture that generates a DAG AND deletes it once the test is completed. I want us to move towards pytest fixtures so you should read up on these.

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -92,194 +105,188 @@ def setUp(self):
...
+                table=table,
                 check="select count(*) FROM {{table}}",
                 greater_than=2,
                 less_than=6,
             )
-            a.execute({"run_id": "foo"})
-            assert True
-        except ValueError:
-            assert False
+        run_dag(dag)
```
dimberman, line 176: I'm noticing that a LOT of this code is being repeated, and I think we can cut down on a lot of the code using [pytest parameterize](https://docs.pytest.org/en/6.2.x/parametrize.html)

I think the two main cases here are 1) DAGs that pass and 2) DAGs that fail. You can parameterize the parameters like we have [here](https://github.com/astro-projects/astro/blob/main/tests/operators/test_agnostic_merge.py#L48-L80). This will also make it really easy for us to run these tests against all databases.

dimberman (changes requested):

src/astro/sql/operators/sql_decorator.py
```diff
@@ -218,29 +218,22 @@ def handle_output_table_schema(self, output_table_name, schema=None):
         :return:
         """
         schema = schema or get_schema()
-        if self.conn_type == "postgres" and self.schema:
+        if self.conn_type in ["postgres", "bigquery"] and self.schema:
```
tatiana, line 221: I get the impression we are doing this type of manipulation in several parts of the code. Could we replace this function by using `Table.qualified_name` (https://github.com/astro-projects/astro/blob/ab9efd134799129985622846236a50bbd8e6add2/src/astro/sql/table.py#L57)

src/astro/sql/operators/sql_decorator.py
```diff
@@ -218,29 +218,22 @@ def handle_output_table_schema(self, output_table_name, schema=None):
...
         schema_statement = ""
-        if self.conn_type == "postgres":
+        if self.conn_type in ["postgres", "snowflake", "bigquery"]:
             schema_statement = create_schema_query(
                 conn_type=self.conn_type,
                 hook=self.hook,
                 schema_id=self.schema,
                 user=self.user,
             )
-        elif self.conn_type == "snowflake":
```
tatiana, line 236: Good refactor! :tada: 
I wonder if we could be abstracting some of this logic as part of the `Table` class, and have some method e.g. `has_schema`, so we avoid having to hard-code this type of list in multiple places.

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -33,253 +36,213 @@
...
 import astro.sql as aql
 from astro.sql.table import Table
-from tests.operators.utils import DEFAULT_SCHEMA
+from tests.operators.utils import DEFAULT_SCHEMA, get_dag, get_table_name, run_dag
 
 log = logging.getLogger(__name__)
 DEFAULT_DATE = timezone.datetime(2016, 1, 1)
+CWD = pathlib.Path(__file__).parent
+
+DATABASES_LIST = ["postgres", "bigquery", "sqlite", "snowflake"]
```
tatiana, line 45: I really like this constant!
This list could live somewhere else - so we can reuse it in multiple tests. What do you think of placing it in `constants.py`, and naming it `SUPPORTED_DATABASES`?

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -33,253 +36,213 @@
...
+
+
+@pytest.fixture(scope="session")
+def tables(request):
```
tatiana, line 49: @utkarsharma2 my feeling is that you adopted a strategy different from what we've been using so far:
https://github.com/astro-projects/astro/blob/48b09347170afc0e644a7db131b7ba11d01dfa53/conftest.py#L62-L75

In other integration tests, we have a fixture that varies depending on the DB (passed using `request.params`) - and returns a single instance. In this case, you changed the approach to return a list of databases. What do you think if we were consistent?

It also feels that a significant part of the logic here (parameters specified by the database) is available in other places, and perhaps could be reused/refactored - so we have less code duplication:
https://github.com/astro-projects/astro/blob/48b09347170afc0e644a7db131b7ba11d01dfa53/tests/operators/utils.py#L17-L29

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -33,253 +36,213 @@
...
-            "aggregate_check_test",
-            conn_id="bigquery",
-            schema=DEFAULT_SCHEMA,
+    with sample_dag:
+        aggregate_table = get_table(tables[database])
+        aql.aggregate_check(
+            table=aggregate_table,
+            check="select count(*) FROM {{table}}",
+            greater_than=4,
+            less_than=4,
```
tatiana, line 97: It would be great to have some sort of assert.  Do we expect this task to log something? Does it add anything Xcom?

tatiana (commented):
Great work, @utkarsharma2 , it's great to see us moving towards simplifying and making our tests more reliable!

I made a few comments related to `test_agnostic_aggregate_check.py`, but the same principles and questions apply to `test_agnostic_boolean_check.py`.

tests/operators/test_agnostic_aggregate_check.py
utkarsharma2: Done

tests/operators/test_agnostic_aggregate_check.py
utkarsharma2: Done

tests/operators/test_agnostic_aggregate_check.py
utkarsharma2: Done, make sense.

tests/operators/test_agnostic_aggregate_check.py
utkarsharma2: The 'aggregate_check' op will raise a value error,  so we expect the dag to fail if the test fails.

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -33,253 +36,213 @@
...
+    with pytest.raises(BackfillUnfinished):
+        with sample_dag:
+            aggregate_table = get_table(tables[database])
+            aql.aggregate_check(
+                table=aggregate_table,
                 check="select count(*) FROM {{table}}",
-                greater_than=4,
-                less_than=4,
+                greater_than=10,
+                less_than=20,
```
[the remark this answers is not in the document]
utkarsharma2: Yes, the operator logs message for the user.

src/astro/sql/table.py
```diff
@@ -52,6 +55,19 @@ def qualified_name(self):
     def __str__(self):
         return f"Table(table_name={self.table_name}, database={self.database}, schema={self.schema}, conn_id={self.conn_id}, warehouse={self.warehouse})"
 
+    def drop(self):
```
dimberman, line 58: Oh this is a nice touch! I like this idea of treating tables as object that have functionality

tests/operators/test_agnostic_boolean_check.py
```diff
@@ -53,194 +53,114 @@ def drop_table_snowflake(
...
-    @classmethod
-    def tearDownClass(cls):
-        drop_table_snowflake(
-            table_name=cls.snowflake_table,
-            schema=os.getenv("SNOWFLAKE_SCHEMA"),
-            database=os.getenv("SNOWFLAKE_DATABASE"),
-            warehouse=os.getenv("SNOWFLAKE_WAREHOUSE"),
-            conn_id="snowflake_conn",
-        )
+# DROP TABLE IF EXISTS {table_name} CASCADE;
```
dimberman, line 56: What's the point of this comment?

tests/operators/test_agnostic_boolean_check.py
```diff
@@ -53,194 +53,114 @@ def drop_table_snowflake(
...
-            ),
-            checks=[Check("test_1", "{{table}}.rooms > 3")],
+@pytest.fixture(scope="module")
+def table(request):
+    boolean_check_table = Table(
+        "boolean_check_test",
+        database="pagila",
+        conn_id="postgres_conn",
+        schema="airflow_test_dag",
+    )
```
dimberman, line 66: Does it make sense to make this a single table name? What happens if multiple people are running this test at the same time?

tests/operators/test_agnostic_boolean_check.py
```diff
@@ -53,194 +53,114 @@ def drop_table_snowflake(
...
+    boolean_check_table_bigquery = Table(
+        "boolean_check_test",
+        conn_id="bigquery",
+        schema=SCHEMA,
+    )
```
dimberman, line 71: What happens if a user who doesn't have bigquery credentials tries to run this test?

tests/operators/test_agnostic_boolean_check.py
```diff
@@ -53,194 +53,114 @@ def drop_table_snowflake(
...
+            marks=pytest.mark.xfail(
+                reason="bigquery don't expect table name before cols."
+            ),
+        ),
+        pytest.param(
+            "snowflake",
+            marks=pytest.mark.xfail(
+                reason="Binding data in type (table) is not supported."
+            ),
+        ),
```
dimberman, line 149: Can you please explain further why this test is expected to fail in snowflake?

tests/operators/utils.py
```diff
@@ -105,3 +107,14 @@ def run_dag(dag):
...
+
+
+def get_dag():
+    very_high_number = 99999999
+    return DAG(
+        "test_dag_" + str(random.randint(0, very_high_number)),
+        default_args={
+            "owner": "airflow",
+            "start_date": DEFAULT_DATE,
+        },
```
dimberman, line 119: Why are you using this instead of the sample_dag fixture?

dimberman (changes requested):

tests/operators/utils.py
utkarsharma2: It's not getting used now, was part of old code. Removing it.

tests/operators/test_agnostic_boolean_check.py
utkarsharma2: I think it's a bug, I"ll explore it as part of a new ticket.

src/astro/constants.py
```diff
@@ -3,3 +3,8 @@
 DEFAULT_CHUNK_SIZE = 1000000
 PYPI_PROJECT_NAME = "astro-projects"
 DEFAULT_SCHEMA = "tmp_astro"
+POSTGRES = "postgres"
```
tatiana, line 6: This is looking great, what do you think about using an `Enum` to group them?

src/astro/sql/operators/sql_decorator.py
```diff
@@ -204,29 +208,25 @@ def handle_output_table_schema(self, output_table_name, schema=None):
         :return:
         """
         schema = schema or SCHEMA
-        if self.conn_type == "postgres" and self.schema:
+        if self.conn_type in ["postgres", "bigquery"] and self.schema:
```
tatiana, line 211: should we use the newly created constants here?

src/astro/sql/operators/sql_decorator.py
```diff
@@ -204,29 +208,25 @@ def handle_output_table_schema(self, output_table_name, schema=None):
...
         schema_statement = ""
-        if self.conn_type == "postgres":
-            schema_statement = create_schema_query(
-                conn_type=self.conn_type,
-                hook=self.hook,
-                schema_id=self.schema,
-                user=self.user,
-            )
-        elif self.conn_type == "snowflake":
+        if self.conn_type in ["postgres", "snowflake", "bigquery"]:
```
tatiana, line 219: Should we use the constants here, to avoid string comparison?

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -20,264 +20,204 @@
...
+        conn_id="snowflake_conn",
+    )
+    path = str(CWD) + "/../data/homes_merge_1.csv"
+    tables = {
+        "postgres": aggregate_table,
+        "bigquery": aggregate_table_bigquery,
+        "sqlite": aggregate_table_sqlite,
+        "snowflake": aggregate_table_snowflake,
+    }
+    aql.load_file(
```
tatiana, line 76: Are we also dropping this table?

tests/operators/test_agnostic_aggregate_check.py
```diff
@@ -20,264 +20,204 @@
...
+        path=path,
+        output_table=tables[request.param],
+    ).operator.execute({"run_id": "foo"})
+
+    yield tables[request.param]
```
tatiana, line 81: It can help readability to:
1. at the beginning of the fixture, have something like:
`database = request.param`
2. to explicitly use an if statement with three clauses in the fixture to highlight the fact the fixture does not aim to create three instances of table - but only one.

tests/operators/test_agnostic_boolean_check.py
```diff
@@ -53,194 +53,116 @@ def drop_table_snowflake(
...
-        self.clear_run()
-        self.addCleanup(self.clear_run)
-        self.dag = DAG(
-            "test_dag",
-            default_args={
-                "owner": "airflow",
-                "start_date": DEFAULT_DATE,
-            },
-        )
+@pytest.fixture(scope="module")
```
tatiana, line 56: This fixture is very similar to [test_agnostic_aggregate_check.py](https://github.com/astro-projects/astro/pull/181/files#diff-880cb6c270aa05815d7b53526ceb0f68dbc1a1f88d521b041a36cfefacbc7150)

Should we move it to `conftest.py` so we can re-use it and avoid code duplicates?

tests/operators/test_agnostic_boolean_check.py
tatiana: @utkarsharma2 could you create and link the two tickets - one for `BigQuery` and one for `snowflake`, before merging this PR?

tatiana (approved):
@utkarsharma2 it's super exciting to see these improvements! :tada: 
I made some last comments, and I'd appreciate it if you could address them before merging. I think we're super close to merge them, so I've just approved!

tests/operators/test_agnostic_boolean_check.py
utkarsharma2: Created it - https://github.com/astro-projects/astro/issues/219

tests/operators/test_agnostic_boolean_check.py
utkarsharma2: This fixture closely resembles the tmp_table fixture, but without the upload capability. I want to make tmp_tables generic enough so that it can accept the file_path(s) as param and can also load a file into DB, should be simple enough.

Also, the default behavior should be able to just return tmp_table object without uploading a file to db, so it can be used in testcases like load_file where we just need a tmp_table object.

src/astro/constants.py
utkarsharma2: I tried to update it, and got carried away with this refactor updated all the instances of string dbs was raising lot of issues on my end, I'll address this as part of the a refactor ticket.

src/astro/sql/operators/sql_decorator.py
utkarsharma2: Will address this in refactor

src/astro/sql/operators/sql_decorator.py
utkarsharma2: Will address this in refactor

PR #184, opened 2022-03-10
File pattern

utkarsharma2:
Added a feature to handle file path patterns
1. For s3  ang gcs - we currently supports only prefixs
2. For local files we support the rules used by the Unix shell to find the pattern.


src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -72,31 +73,39 @@ def execute(self, context):
...
-            conn_id=self.output_table.conn_id,
-            database=self.output_table.database,
-            warehouse=self.output_table.warehouse,
-            schema=self.output_table.schema,
-            df=df,
-            conn_type=conn.conn_type,
-            user=conn.login,
-            chunksize=self.chunksize,
-        )
+        if_exists = "replace"
```
dimberman, line 90: What if the user wants to append everything?

tests/operators/test_agnostic_load_file.py
```diff
@@ -50,9 +51,11 @@
 from google.cloud import bigquery, storage
 from pandas.util.testing import assert_frame_equal
 
+from astro.constants import DEFAULT_SCHEMA
```
dimberman, line 54: @tatiana do you agree with us having this DEFAULT_SCHEMA variable?

tests/operators/test_agnostic_load_file.py
```diff
@@ -477,6 +512,143 @@ def test_aql_gcs_file_to_postgres(self):
...
+        # },
+        # {
+        #     "path": str(CWD) + "/../data/homes_merge_*",
+        #     "file_conn_id": "",
+        #     "load_data": lambda *args: None,
+        #     "files": [
+        #         {"filename": str(CWD) + "/../data/homes_pattern_1.csv"},
+        #         {"filename": str(CWD) + "/../data/homes_pattern_2.csv"},
+        #     ],
+        # },
```
dimberman, line 611: Why this giant comment?

tests/operators/test_agnostic_load_file.py
```diff
@@ -477,6 +512,143 @@ def test_aql_gcs_file_to_postgres(self):
...
+                    "bucket_id": "dag-authoring",
+                    "object_name": "a.csv",
+                    "filename": str(CWD) + "/../data/homes.csv",
+                },
+                {
+                    "bucket_id": "dag-authoring",
+                    "object_name": "aa.csv",
+                    "filename": str(CWD) + "/../data/homes.csv",
+                },
+            ],
```
dimberman, line 567: Please export this to a fixture.

dimberman (changes requested):
@utkarsharma2 This is a fantastic start! Got a few comments, but overall looking really good!

.github/ci-test-connections.yaml
```diff
@@ -35,3 +35,19 @@ connections:
...
     login:
     password:
+  - conn_id: gcp_conn
+    conn_type: google_cloud_platform
+    description: null
+    extra: null
+  - conn_id: google_cloud_default
+    conn_type: google_cloud_platform
+    description: null
+    extra: null
```
tatiana, line 45: Why are we adding this connection as opposed to using the default Airflow one which has the same name?

.github/ci-test-connections.yaml
```diff
@@ -35,3 +35,19 @@ connections:
...
+  - conn_id: aws_conn
+    conn_type: aws
+    description: null
+    extra: null
+  - conn_id: aws_default
```
tatiana, line 50: The same question on `google_cloud_default` applies here

src/astro/constants.py
```diff
@@ -1,2 +1,5 @@
+import os
+
 DEFAULT_CHUNK_SIZE = 1000000
 PYPI_PROJECT_NAME = "astro-projects"
+DEFAULT_SCHEMA = os.getenv("AIRFLOW__ASTRO__SQL_SCHEMA") or "astroflow_ci"
```
tatiana, line 5: I feel that we should not be defaulting the schema to `astroflow_ci` in the application - this should only happen in the test context. The previous value `tmp_astro` seemed to make more sense from a user perspective.

I also believe `constants.py` should contain constants - and not user-defined values through environment variables. We could have `settings.py` or something similar for this value type.

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -141,6 +150,38 @@ def _load_dataframe(self, path):
...
             )
 
+    def get_paths(self, path, file_conn_id):
+        url = urlparse(path)
+        file_location = url.scheme
+        return {
+            "s3": self.get_paths_from_s3,
+            "gs": self.get_paths_from_gcs,
+            "http": lambda url, file_conn_id: [urlunparse(list(url))],
+            "https": lambda url, file_conn_id: [urlunparse(list(url))],
```
tatiana, line 160: Why are we using `urlparse` on line 154 and then `urlunparse` here?

tests/operators/test_agnostic_load_file.py
tatiana: We already have a similar fixture:
https://github.com/astro-projects/astro/blob/b1d9b2947be0d594f5c107579a961887f40b6c4f/tests/operators/test_agnostic_load_file.py#L518-L570

tests/operators/test_agnostic_load_file.py
tatiana: @dimberman I commented here:
https://github.com/astro-projects/astro/pull/184/files#r825188964

.github/ci-test-connections.yaml
utkarsharma2: Fixed.

src/astro/sql/operators/agnostic_load_file.py
utkarsharma2: We have to get the scheme to identify the file location to do that we are parsing url via urlparse and then all the methods are accepting the same set of params, so to keep the pattern consistent.

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -141,6 +150,38 @@ def _load_dataframe(self, path):
...
+    def get_paths_from_s3(self, url, file_conn_id=None):
+        bucket = url.netloc
+        prefix = url.path
+        hook = s3.S3Hook(aws_conn_id=file_conn_id) if file_conn_id else s3.S3Hook()
+        return [
+            urlunparse((url.scheme, url.netloc, keys, "", "", ""))
+            for keys in hook.list_keys(bucket_name=bucket, prefix=prefix[1:])
+        ]
+
+    def get_paths_from_gcs(self, url, file_conn_id=None):
```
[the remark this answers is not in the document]
utkarsharma2: Refactored it.

tests/operators/test_agnostic_load_file.py
utkarsharma2: Used the existing fixture

dimberman (approved):

PR #185, opened 2022-03-10
#154 Add case-sensitive param to df

### LINKED ISSUE #154: Add "case_sensitive" option to snowflake and default to all lowercase
the `get_dataframe` function of the SnowflakeHook has a somewhat annoying feature where it will return all uppercase values for columns that are unquoted. We want to create an easier dataframe experience for astro users, but at the same time we want to ensure that those users can still have 1:1 matching if that is something they care about (e.g. if they have a column like "Customers").

The idea here is to have a `case_sensitive` option

```
@df(identifiers_as_lower =False)
def my_case_sensitive_snow(df: DataFrame):
    return df["Customers"]

#by default we turn identifiers into lower
@df
def my_case_non_sensitive_snow(df: DataFrame):
    return df["customers"]

```

bhavaniravi:
closes #154 

Adds `identifiers_as_lower` to `get_dataframe` function

### DIFF
--- a/example_dags/example_amazon_s3_postgres.py
+++ b/example_dags/example_amazon_s3_postgres.py
@@ -42,7 +42,7 @@ def sample_create_table(input_table: Table):
     return "SELECT * FROM {{input_table}} LIMIT 10"
 
 
-@df
+@df(identifiers_as_lower=False)
 def my_df_func(input_df: DataFrame):
     print(input_df)
 
--- a/example_dags/example_amazon_s3_snowflake_transform.py
+++ b/example_dags/example_amazon_s3_snowflake_transform.py
@@ -40,7 +40,7 @@ def clean_data(input_table: Table):
     """
 
 
-@df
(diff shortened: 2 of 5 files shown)


dimberman (approved):

dimberman:
@bhavaniravi until we figure out our story around external users running the CI I will run a copy of your branch here https://github.com/astro-projects/astro/pull/195

tatiana:
@bhavaniravi that looks great, thank you very much for contributing to Astro!

There are a few tests failing - you should be able to see them now! I logged a ticket for us to improve our [CI, so community pull requests can run the tests](https://github.com/astro-projects/astro/issues/179). Please, do reach out if we can support you in this.

In the meantime, would it be possible for you to recreate the PR by cloning `astro-projects/astro` and creating a branch in the original project? This should allow you to see the tests failing/passing at each iteration.

bhavaniravi:
Closing this:: duplicate of #215

PR #209, opened 2022-03-14
Fix running tests in parallel

### LINKED ISSUE #208: Fix concurency issue on `main` branch
**Describe the bug**
Since we introduced some parallelism in running our tests, we have seen the CI getting red more often than desired, many times breaking tests with parts of the code that were not changed.

This is happening because some tests do not use unique table names and do not implement tear down effectively.

**Version**
* Astro: 0.7.0

**To Reproduce**
Create a pull request / or merge it and - intermittently - see the tests failing:
Screenshot from 2022-03-14 09-30-23

**Expected behavior**
The tests should not break unless there is a bug in the code.

**Acceptance criteria**
* [ ] Make the tests more reliable, by avoiding table name clashes.

tatiana:
Fix #208 by giving unique dag_ids for test DAGs

Acceptance criteria
* [x] Make the tests more reliable, by avoiding table name clashes.

### DIFF
--- a/src/astro/sql/table.py
+++ b/src/astro/sql/table.py
@@ -70,11 +70,17 @@ class Table:
 
 class TempTable(Table):
     def __init__(self, conn_id=None, database=None, warehouse=""):
+        self.table_name = ""
         super().__init__(
-            table_name="", conn_id=conn_id, database=database, warehouse=warehouse
+            table_name=self.table_name,
+            conn_id=conn_id,
+            database=database,
+            warehouse=warehouse,
         )
 
     def to_table(self, table_name: str, schema: str) -> Table:
+        self.table_name = table_name
+        self.schema = schema
         return Table(
             table_name=table_name,
             conn_id=self.conn_id,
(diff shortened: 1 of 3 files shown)


src/astro/sql/table.py
```diff
@@ -87,7 +95,7 @@ def create_table_name(context) -> str:
     dag_run: DagRun = ti.get_dagrun()
     table_name = f"{dag_run.dag_id}_{ti.task_id}_{dag_run.id}".replace(
         "-", "_"
-    ).replace(".", "__")
+    ).replace(".", "__")[:MAX_TABLE_NAME_SIZE]
```
tatiana, line 98: This approach can have side effects, particularly if the first 63 characters of the table name are the same across different tasks. @dimberman @utkarsharma2 what do you think would be a good strategy?

We probably should improve this as part of a separate ticket.

dimberman: @tatiana hard to say, I believe that's a restriction that is given to us from elsewhere (e.g. certain dbs have table name size limits)

tatiana: @dimberman https://github.com/astro-projects/astro/issues/218

PR #215, opened 2022-03-15
154 case sensitive identifiers

### LINKED ISSUE #154: Add "case_sensitive" option to snowflake and default to all lowercase

bhavaniravi:
closes #154 

Adds `identifiers_as_lower` to `get_dataframe` function

### DIFF
--- a/example_dags/example_amazon_s3_postgres.py
+++ b/example_dags/example_amazon_s3_postgres.py
@@ -42,7 +42,7 @@ def sample_create_table(input_table: Table):
     return "SELECT * FROM {{input_table}} LIMIT 10"
 
 
-@df
+@df(identifiers_as_lower=False)
 def my_df_func(input_df: DataFrame):
     print(input_df)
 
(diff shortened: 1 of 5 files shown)


tatiana (approved):
Thanks for addressing all the coments from https://github.com/astro-projects/astro/pull/185, @bhavaniravi !
Please, feel free to merge it :tada:

PR #217, opened 2022-03-15
193 append test refactor

dimberman:
**Description**
As a step towards "maturing" the astro DAG authoring project, we must rewrite our tests to ensure that every integration test runs against every database.

This step will simultaneously reduce the number of tests we need to maintain, make testing much simpler as we add new databases, and will make a future refactor much simpler as we can ensure proper coverage.

To do this, we will take advantage of two features in pytest, fixtures and parameterize.

For this ticket, we will update the `append` function.

**Acceptance criteria**
Have a single test file validating `append` across all databases.

1. Integration tests should be marked with `pytest.marker.integration`:

- [x] Each test should work across all databases
- [ ]  Validating both `Table` and `TempTable` as inputs
- [x] Use `test_utils.run_dag`
- [x] Use PyTest `fixtures` and `parameterize` to have a single main test that will validate transform across multiple databases

The tests should validate these scenarios:
- [x] appending two tables against a single column
- [x]  appending two tables against multiple columns
- [x]  appending against all fields by not specifying fields
- [x] appending with casting
- [x]  append with some casted fields and some uncasted fields
- [x] test with two different databases (should fail)


tests/operators/test_agnostic_append.py
```diff
@@ -0,0 +1,116 @@
...
+
+@adf
+def validate_caste_only(df: pd.DataFrame):
+    assert len(df) == 6
+    assert not df["age"].hasnans
+    assert df["sell"].hasnans
+
+
+@pytest.fixture
+def append_params(request):
```
utkarsharma2, line 36: I think this should be expanded to testcases instead of fixtures?

tests/operators/test_agnostic_append.py
```diff
@@ -0,0 +1,116 @@
...
+        "postgres",
+        "sqlite",
+    ],
+    indirect=True,
+)
+def test_append(sql_server, sample_dag, tmp_table, append_params):
+    app_param, validate_append = append_params
+
+    with sample_dag:
+        load_main = aql.load_file(
```
utkarsharma2, line 72: can we move loading files into tmp_table? is there any usecase with just tmp_table without loading it to db?

tests/operators/test_agnostic_append.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.mark.parametrize(
+    "sql_server",
+    [
+        "postgres",
+    ],
+    indirect=True,
+)
+def test_append_on_tables_on_different_db(sample_dag, sql_server):
```
utkarsharma2, line 99: Should this test case be renamed or we should add more DBs?

utkarsharma2:
+116 −586, nicely done!

tests/operators/test_agnostic_append.py
dimberman: For now using load_file ensures that the actual function is being passed info from a previous task instead of a table object

tests/operators/test_agnostic_append.py
dimberman: I think this will make it much easier to add more testcases with less boilerplate. We can see if it becomes a problem but this ultimately makes it easier to create test grids

tests/operators/test_agnostic_append.py
dimberman: Why would it need to be renamed? It's testing a piece of code that's not DB dependent

tests/operators/test_agnostic_append.py
tatiana: I agree with @utkarsharma2 on this one. I don't think we should aim to have a single test per operator. We already have a few dimensions we are using fixtures:
- multiple databases, when relevant
- multiple files, when relevant
- multiple file locations, when relevant
- multiple file types, when relevant

I strongly recommend we do not use parametrization for groups of parameters sent to our tasks/operators.

tests/operators/test_agnostic_append.py
```diff
@@ -0,0 +1,116 @@
...
+@pytest.fixture
+def append_params(request):
+    mode = request.param
+    if mode == "basic":
+        return {
+            "columns": ["sell", "living"],
+        }, validate_basic
+    if mode == "all_fields":
+        return {}, validate_append_all
+    if mode == "with_caste":
```
tatiana, line 44: If this test runs against all the databases, we probably do not need to run "basic" and "cast_only" against all the other databases.

tests/operators/test_agnostic_append.py
```diff
@@ -0,0 +1,116 @@
...
+@pytest.mark.parametrize(
+    "sql_server",
+    [
+        "postgres",
+    ],
+    indirect=True,
+)
+def test_append_on_tables_on_different_db(sample_dag, sql_server):
+    tmp_table_1 = TempTable(conn_id="postgres_conn")
+    tmp_table_2 = TempTable(conn_id="sqlite_conn")
```
tatiana, line 101: @dimberman Are we tearing down these tables?

Issue #98, opened 2022-02-10
Improve the speed of load_file and save_file in Postgres


dimberman:
During testing, we found that the load speed of dataframe -> postgres to be fairly slow. We believe that there should be a faster native way to load data into postgres.

Acceptance criteria:
- A notable speed improvement on postgres load and save numbers in the load and save file
- detailed instructions for users to add any needed extensions to their database that would enable these features (e.g. s3 support)

bhavaniravi:
We can use Postgres `COPY`. We can write a custom callable for `method` in `to_sql` and modify it to use Postgres COPY

References: 
https://stackoverflow.com/a/64883481/6340775
https://towardsdatascience.com/upload-your-pandas-dataframe-to-your-database-10x-faster-eb6dc6609ddf

Can I go ahead with this approach?

tatiana:
In order to implement this feature, I ended up refactoring several parts of the code, so we can iterate and move forward in the next tickets related to `load`:
1. Add constants to represent file type and file location: #223
2. Add a module to handle file functions # 224
3. Add a module to handle path functions # 225
4. Add a module to handle database functions, extending to have a constant to represent databases #228

The following PRs are still being cleaned up and will soon be ready for review:
5. Add a module to handle load operations
6. Add a module to handle deletion
7. Refactor load file to use the new functions

tatiana:
7. Add a module to handle deletion: #246

And one last PR to wrap up this task...!

tatiana:
Today I learned that  *psycopg* `copy_from` does not parse correctly column values that have double quotes and commas inside them, raising:
```
psycopg2.errors.BadCopyFileFormat: extra data after last expected column
```

I am now using the method `copy_expert` and running into other edge cases while attempting to process the Github timeline dataset. I have the hope to find a configuration that will work for this dataset.

tatiana:
We split into two new tasks:
* https://github.com/astronomer/astro-sdk/issues/431
* https://github.com/astronomer/astro-sdk/issues/428

PR #225, opened 2022-03-17
Create utils.path module for path-related operations

tatiana:
Relates to: #98 (part 3)

### DIFF
--- a/src/astro/utils/path.py
+++ b/src/astro/utils/path.py
@@ -1,0 +1,113 @@
+import glob
+import os
+import pathlib
+from urllib.parse import urlparse, urlunparse
+
+from astro.constants import FileLocation, FileType
+from astro.utils.cloud_storage_creds import gcs_client, s3fs_creds
+from astro.utils.dependencies import gcs, s3
+
+
+def get_location(path):
+    """
+    Identify where a file is located.
+
+    :param path: Path to a file in the filesystem
+    :type path: str
+    :return: Location of the file
+    :rtype: astro.constants.FileLocation (enum) constant
+    """
+    file_scheme = urlparse(path).scheme
+    if file_scheme == "":
+        location = FileLocation.LOCAL
+    else:
+        try:
+            location = getattr(FileLocation, file_scheme.upper())
+        except (UnboundLocalError, AttributeError):
+            raise ValueError(
+                f"Unsupported scheme '{file_scheme}' from path '{path}'"
+            )  # TODO: Use string interpolation as opposed to fstring
+    return location
+
+
+def validate_path(path):
+    """
+    Check if the give path is either a valid URI or a local file.
+
+    :param path: Either local filesystem path or remote URI
+    :type path: str
+    :return: if fails, raise `ValueError`
+    :rtype: `ValueError`
+    """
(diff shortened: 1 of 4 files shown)


src/astro/utils/path.py
```diff
@@ -0,0 +1,113 @@
...
+    :type path: str
+    :return: Location of the file
+    :rtype: astro.constants.FileLocation (enum) constant
+    """
+    file_scheme = urlparse(path).scheme
+    if file_scheme == "":
+        location = FileLocation.LOCAL
+    else:
+        try:
+            location = getattr(FileLocation, file_scheme.upper())
```
utkarsharma2, line 25: Can we change this to
location = FileLocation(file_scheme.upper())

src/astro/utils/path.py
```diff
@@ -0,0 +1,113 @@
...
+    :rtype: `ValueError`
+    """
+    result = urlparse(path)
+    if not (all([result.scheme, result.netloc]) or os.path.isfile(path)):
+        raise ValueError(f"Invalid path: '{path}'")
+
+
+def get_transport_params(path, conn_id):
+    """
+    Given a filesystem, GCS or S3 prefix, return files within that path.
```
utkarsharma2, line 49: I think this description string is not accurate, since it return creds and not file.
"Given a filesystem, GCS or S3 prefix, return files within that path."

utkarsharma2: Probably a copy paste error.

src/astro/utils/path.py
```diff
@@ -0,0 +1,113 @@
...
+    :param path: Either local filesystem path or remote URI
+    :param conn_id: Airflow connection ID, if connecting to S3 or GCS
+    :type path: str
+    :type conn_id: str
+    :return: A list with URIs or local paths
+    :rtype: List(str)
+    """
+    url = urlparse(path)
+    location = get_location(path)
+    if location == FileLocation.LOCAL:
```
utkarsharma2, line 80: Do you think we can split this function in smaller function each for Local, Http, GS and S3 locations, that way they are easy to write unit test for?

src/astro/utils/path.py
```diff
@@ -0,0 +1,113 @@
...
+def is_local(path):
+    """
+    Check if a path is local to the filesystem or not.
+
+    :param path: Either a filesystem path or remote URI
+    :type path: str
+    :return: True or False
+    :rtype: bool
+    """
+    file_location = get_location(path)
```
utkarsharma2, line 112: we can simplify this - 
return get_location(path) == FileLocation.LOCAL

tests/utils/test_path.py
```diff
@@ -0,0 +1,152 @@
...
+@pytest.fixture()
+def local_dir():
+    os.mkdir(LOCAL_DIR)
+    open(LOCAL_DIR_FILE_1, "a").close()
+    open(LOCAL_DIR_FILE_2, "a").close()
+    yield
+    shutil.rmtree(LOCAL_DIR)
+
+
+def describe_get_location():
```
utkarsharma2, line 49: This is really nice, grouping a function by its cases.

PR #228, opened 2022-03-21
Create utils.database module for db operations

tatiana:
Relates to: #98

### DIFF
--- a/src/astro/utils/database.py
+++ b/src/astro/utils/database.py
@@ -1,0 +1,82 @@
+from typing import Union
+
+from airflow.hooks.base import BaseHook
+from airflow.providers.sqlite.hooks.sqlite import SqliteHook
+from sqlalchemy import create_engine, text
+from sqlalchemy.engine import Engine
+from sqlalchemy.engine.result import ResultProxy
+
+from astro.constants import Database
+from astro.utils.dependencies import BigQueryHook, PostgresHook, SnowflakeHook
+
+
+def get_database_name(interface: Union[Engine, BaseHook, SqliteHook]) -> Database:
+    """
+    Given a hook or a SQL engine, identify the database name.
+
+    :param interface: interface to the database
+    :type interface: SQLAlchemy engine or Airflow Hook (BigQueryHook, PostgresHook, SnowflakeHook, SqliteHook)
+    :return: the database this interface relates to (e.g. Database.SQLITE)
+    :rtype: astro.constants.Database enum item
+    """
+    if isinstance(interface, BaseHook):
+        hook_to_database = {
+            BigQueryHook: Database.BIGQUERY,
+            PostgresHook: Database.POSTGRES,
+            SnowflakeHook: Database.SNOWFLAKE,
+            SqliteHook: Database.SQLITE,
+        }
+        try:
+            database_name = hook_to_database[type(interface)]
+        except KeyError:
+            raise ValueError(f"Unsupported database {type(interface)}")
+    else:  # SqlAlchemy engine
+        database_name = getattr(Database, interface.name.upper())
+    return database_name
+
+
+def get_sqlalchemy_engine(hook: Union[BaseHook, SqliteHook]) -> Engine:
+    """
+    Given a hook, return a SQLAlchemy engine for the target database.
+
+    :param hook: Airflow Hook used to access a SQL-like database
+    :type hook: (BigQueryHook, PostgresHook, SnowflakeHook, SqliteHook)
+    :return: SQLAlchemy engine
+    :rtype: sqlalchemy.Engine
+    """
+    database = get_database_name(hook)
+    engine = None
+    if database == Database.SQLITE:
+        uri = hook.get_uri()
+        if "////" not in uri:
+            uri = hook.get_uri().replace("///", "////")
+        engine = create_engine(uri)
+    if engine is None:
+        engine = hook.get_sqlalchemy_engine()
+    return engine
+
+
+def run_sql(
+    engine: Engine,
+    sql_statement: Union[str, text],
+    parameters: Union[None, dict] = None,
+) -> ResultProxy:
+    """
+    Run a SQL statement using the given engine.
+
+    :param engine: SQLAlchemy engine
+    :type engine: sqlalchemy.Engine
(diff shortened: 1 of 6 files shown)


src/astro/utils/database.py
```diff
@@ -0,0 +1,74 @@
+from airflow.hooks.base import BaseHook
+from airflow.providers.sqlite.hooks.sqlite import SqliteHook
+from sqlalchemy import create_engine, text
+
+from astro.constants import Database
+from astro.utils.dependencies import BigQueryHook, PostgresHook, SnowflakeHook
+
+
+def get_database_name(interface):
```
dimberman, line 9: Is there a type hint you can put here?

tests/utils/test_database.py
```diff
@@ -0,0 +1,109 @@
...
+        db = get_database_name(engine)
+        assert db == Database.BIGQUERY
+
+    def with_snowflake_hook():
+        hook = SnowflakeHook(snowflake_conn_id="snowflake_conn")
+        db = get_database_name(hook)
+        assert db == Database.SNOWFLAKE
+        engine = hook.get_sqlalchemy_engine()
+        db = get_database_name(engine)
+        assert db == Database.SNOWFLAKE
```
dimberman, line 29: Seems like these can be shortened with some functional magic, but maybe that would make things hard to read?

Maybe using behaves_like?

Something like:

```
def get_db_func(hook, db_name):
    db = get_database_name(hook)
    assert db == db_name
    engine = hook.get_sqlalchemy_engine()
    db = get_database_name(engine)
    assert db == db_name


def describe_get_database_name():
    @behaves_like(get_db_func)
    def with_snowflake_hook():
        @pytest.fixture
        def hook():
            return SnowflakeHook(snowflake_conn_id="snowflake_conn")

        @pytest.fixture
        def db_name():
            return Database.SNOWFLAKE
        
    @behaves_like(get_db_func)
    def with_postgres_hook():
        @pytest.fixture
        def hook():
            return PostgresHook()

        @pytest.fixture
        def db_name():
            return Database.POSTGRES
```

I realized it might not save code lines in this specific case, but it at least would help ensure that all the tests are run the same way

dimberman (changes requested):

src/astro/utils/database.py
tatiana: @dimberman sure, I can add - the code has been inconsistent in this way and it was not in our definition of done, but I'll adopt them from now onwards.

tests/utils/test_database.py
tatiana: I believe tests should be kept simple and with the least logic as possible - I'd rather keep the previous implementation.

src/astro/utils/database.py
```diff
@@ -0,0 +1,86 @@
...
+from sqlalchemy import create_engine, text
+from sqlalchemy.engine import Engine
+from sqlalchemy.engine.result import ResultProxy
+
+from astro.constants import Database
+from astro.utils.dependencies import BigQueryHook, PostgresHook, SnowflakeHook
+
+
+def get_database_name(
+    interface: Union[Engine, SqliteHook, BigQueryHook, PostgresHook, SnowflakeHook]
```
tatiana, line 14: I attempted to have a granular type hint here, using:
`interface: Union[Engine, SqliteHook, BigQueryHook, PostgresHook, SnowflakeHook]`
However, this conflicts with our current strategy on handling dependency:
```
TypeError: Union[arg, ...]: each arg must be a type. Got <astro.utils.dependencies.MissingPackage object at 0x7f350c80e790>.
```
I'll keep `SqliteHook` and  `BaseHook`, removing the others.

dimberman: Oh yikes good to know

tests/utils/test_database.py
dimberman: Ok sounds good

dimberman (approved):

tests/utils/test_database.py
tatiana: Absolutely! This pattern can certainly simplify other test cases!

src/astro/utils/database.py
```diff
@@ -0,0 +1,82 @@
...
+from airflow.providers.sqlite.hooks.sqlite import SqliteHook
+from sqlalchemy import create_engine, text
+from sqlalchemy.engine import Engine
+from sqlalchemy.engine.result import ResultProxy
+
+from astro.constants import Database
+from astro.utils.dependencies import BigQueryHook, PostgresHook, SnowflakeHook
+
+
+def get_database_name(interface: Union[Engine, BaseHook, SqliteHook]) -> Database:
```
utkarsharma2, line 13: Why do we have sqlitehook separately mentioned, since we already have BaseHook  here?

src/astro/utils/database.py
```diff
@@ -0,0 +1,82 @@
...
+            PostgresHook: Database.POSTGRES,
+            SnowflakeHook: Database.SNOWFLAKE,
+            SqliteHook: Database.SQLITE,
+        }
+        try:
+            database_name = hook_to_database[type(interface)]
+        except KeyError:
+            raise ValueError(f"Unsupported database {type(interface)}")
+    else:  # SqlAlchemy engine
+        database_name = getattr(Database, interface.name.upper())
```
utkarsharma2, line 34: can we rewrite it - 
database_name = Database(interface.name.upper())

utkarsharma2: Also I think we should handle ValueError.

src/astro/utils/database.py
```diff
@@ -0,0 +1,82 @@
...
+    :type hook: (BigQueryHook, PostgresHook, SnowflakeHook, SqliteHook)
+    :return: SQLAlchemy engine
+    :rtype: sqlalchemy.Engine
+    """
+    database = get_database_name(hook)
+    engine = None
+    if database == Database.SQLITE:
+        uri = hook.get_uri()
+        if "////" not in uri:
+            uri = hook.get_uri().replace("///", "////")
```
utkarsharma2, line 52: we are not supporting windows?
https://docs.sqlalchemy.org/en/14/core/engines.html#sqlite

Issue #194, opened 2022-03-10
Refactor load tests


dimberman:
**Description**
As a step towards "maturing" the astro DAG authoring project, we must rewrite our tests to ensure that every integration test runs against every database.

This step will simultaneously reduce the number of tests we need to maintain, make testing much simpler as we add new databases, and will make a future refactor much simpler as we can ensure proper coverage.

To do this, we will take advantage of two features in pytest, fixtures and parameterize.

For this ticket, we will update the `append` function.

**Acceptance criteria**
Have a single test file validating `append` across all databases.

1. Integration tests should be marked with `pytest.marker.integration`:

* Each test should work across all databases
* Validating both `Table` and `TempTable` as inputs
* Use `test_utils.run_dag`
* Use PyTest `fixtures` and `parameterize` to have a single main test that will validate transform across multiple databases

The tests should validate these scenarios:
1. loading against all formats (csv, parquet, avro, etc.)
2. loading to a temp_table or named table
3. loading to default schema and to named schema

tatiana:
@dimberman the description mentions `append`, but the title says `load_file`. Which one is this?

Also, I think it is important to mention that `load_file` was the first place where we started introducing the new style of integration tests.

These lines were the first place where we adopted the new integration tests approach:
https://github.com/astro-projects/astro/blob/main/tests/operators/test_agnostic_load_file.py#L532-L801

What is missing, IMPOV is to refactor these lines:
https://github.com/astro-projects/astro/blob/a83308a32f513a1e4049bc6dd98730c4e12de9f5/tests/operators/test_agnostic_load_file.py#L68-L529

Finally, I don't think we should be aiming to have a single test function per operator. I'd vote for us to have three independent tests than this type of fixture:
https://github.com/astro-projects/astro/blob/main/tests/operators/test_agnostic_merge.py#L47-L79

We are at risk of overusing fixtures, which can compromise the readability of the tests.

utkarsharma2:
@dimberman, I think I can partially work on this, refactoring testcases which are having conflicting table names.

PR #229, opened 2022-03-21
Refactored Load_file Op. testcases

utkarsharma2:
Closes: #194 

- [x] Each test should work across all databases
- [x] Use test_utils.run_dag
- [x] Use PyTest fixtures and parameterize to have a single main test that will validate transform across multiple databases
- [x] Loading against all formats (csv, parquet, avro, etc.)  -- **Partially done, since it's not required for all tests.**
- [x] Loading to a temp_table or named table -- **Partially done, since it's not required for all tests.**
- [x] Loading to default schema and to named schema -- **Partially done, since it's not required for all tests.**

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -64,32 +64,46 @@ def sample_dag():
 
 
 @pytest.fixture
-def tmp_table(sql_server):
+def tmp_table(request, sql_server):
+    table_type = True
+    table_options = {}
+
+    if getattr(request, "param", None):
+        table_type = request.param.get("is_temp", True)
+        table_options = request.param.get("param", {})
     sql_name, hook = sql_server
 
     if isinstance(hook, SnowflakeHook):
-        temporary_table = TempTable(
-            conn_id=hook.snowflake_conn_id,
-            database=hook.database,
-            warehouse=hook.warehouse,
-        )
+        params = {
+            "conn_id": hook.snowflake_conn_id,
+            "database": hook.database,
+            "warehouse": hook.warehouse,
+        }
     elif isinstance(hook, PostgresHook):
-        temporary_table = TempTable(conn_id=hook.postgres_conn_id, database=hook.schema)
+        params = {"conn_id": hook.postgres_conn_id, "database": hook.schema}
     elif isinstance(hook, SqliteHook):
-        temporary_table = TempTable(conn_id=hook.sqlite_conn_id, database="sqlite")
+        params = {"conn_id": hook.sqlite_conn_id, "database": "sqlite"}
     elif isinstance(hook, BigQueryHook):
-        temporary_table = TempTable(conn_id=hook.gcp_conn_id)
-    yield temporary_table
+        params = {
+            "conn_id": hook.gcp_conn_id,
+        }
+    else:
+        raise ValueError("Unsupported Database")
+
+    params.update(table_options)
+    table = TempTable(**params) if table_type else Table(**params)
+
+    yield table
 
     if isinstance(hook, SqliteHook):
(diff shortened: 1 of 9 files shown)


dimberman (changes requested):

tests/operators/test_agnostic_load_file.py
```diff
@@ -44,470 +44,175 @@
...
-                "file_conn_id": "",
-                "output_table": Table(
-                    OUTPUT_TABLE_NAME,
-                    conn_id="bigquery",
-                    schema=SCHEMA,
-                ),
-            },
-        )
-
+def get_dataframe_from_table(sql_name: str, tmp_table: Table, hook):
```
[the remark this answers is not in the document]
utkarsharma2: I think you mean `Union[Table, TempTable]`, if so I have added it.

conftest.py
```diff
@@ -64,32 +64,46 @@ def sample_dag():
 
 
 @pytest.fixture
-def tmp_table(sql_server):
+def tmp_table(request, sql_server):
+    table_type = True
```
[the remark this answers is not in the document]
utkarsharma2: Make sense, updated it.

conftest.py
```diff
@@ -64,32 +64,46 @@ def sample_dag():
 
 
 @pytest.fixture
-def tmp_table(sql_server):
+def tmp_table(request, sql_server):
```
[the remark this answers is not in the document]
utkarsharma2: Make sense, updated it.

tests/operators/test_agnostic_load_file.py
```diff
@@ -644,34 +333,71 @@ def test_load_file_using_file_connection_fails_inexistent_conn(caplog, sample_da
...
+    database_name, sql_hook = sql_server
+
+    with sample_dag:
+        load_file(
+            path=str(pathlib.Path(CWD.parent, f"data/sample.{file_type}")),
+            file_conn_id="",
+            output_table=test_table,
+        )
+    test_utils.run_dag(sample_dag)
+    df = sql_hook.get_pandas_df(f"SELECT * FROM {test_table.qualified_name()}")
```
dimberman, line 400: @utkarsharma2 we've run into issues with hanging tests in the past when people run `get_pandas_df` like this. Instead, could you please create an @adf decorated function that validates the test? You should see an example of this in the `test_transform` file

dimberman (changes requested):
One change otherwise looks good!

tests/operators/test_agnostic_load_file.py
utkarsharma2: @dimberman, I think that would be because of conflicting table names, but since now we are using fixture to get tables, it should not reoccur. But we are adopting this as best practice I'll change it.

tatiana: I believe it is worth using the `get_pandas_df` - it keeps the tests simple enough.

tests/operators/test_agnostic_load_file.py
```diff
@@ -44,468 +45,176 @@
...
-                    conn_id="postgres_conn",
-                ),
-            },
-        )
-
-    def test_aql_s3_file_to_postgres(self):
-        OUTPUT_TABLE_NAME = "expected_table_from_s3_csv"
-
-        self.hook_target = PostgresHook(
-            postgres_conn_id="postgres_conn", schema="pagila"
```
tatiana, line 333: It's so satisfying to see all these lines deleted :raised_hands:

tatiana (approved):
Looks great, thanks, @utkarsharma2 ! :rocket:

dimberman (approved):

Issue #125, opened 2022-02-21
Support `params` argument in `aql.render` to declare values to SQL Jinja template


tatiana:
**Context**
At the moment, Astro 0.5.1 does not use the `params` keyword argument. This argument would allow us to overwrite SQL Jinja template values.

Example:
```
SELECT
   t.table_schema,
   t.table_name,
   CONCAT(t.table_catalog, '.', t.table_schema, '.', t.table_name) table_path
FROM dwh_legacy.information_schema.tables t
WHERE
       t.table_schema IN ({{ params.schemas }})
   AND t.table_type = 'BASE TABLE';
```

```
aql.render(path='/usr/local/airflow/include/sql/schemas',
                                      params={'schemas': dwh_schemas})
```

Currently raises the error:
```
  File "/sr/local/lib/python3.9/site-packages/jinja2/runtime.py", line 903, in _fail_with_undefined_error
      raise self._undefined_exception(self._undefined_message)
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'schemas'
```

**Acceptance criteria**
* Users of `aql.render` can override Jinja template values by using the `params`  argument
* Have an `example_dag` which uses `params`
* Add at least one unit test which validate this behaviour
* Update the docstring so it details this feature
* Ask @mag3141592 (bug reporter) to review the PR

dimberman:
Upon further testing I have found that the following works

```
SELECT {{ params['foo'] }} FROM actor
```

I will update the documentation to reflect this.

dimberman:
further addressed in https://github.com/astro-projects/astro/pull/254

PR #230, opened 2022-03-21
#125 added parameters to example

dimberman:
Clarifies in example dags that users can use `param` dictionary in their SQL files when using aql.render


tatiana (approved):
LGTM :rocket:

Issue #123, opened 2022-02-21
Allow `aql.render` to use the DAG's `template_searchpath`


tatiana:
**Context**
When instantiating Airflow DAGs, users can declare `template_searchpath`, which is a list of folders (non-relative) that define where jinja will look for templates. By default, it contains the DAGs folder.

Many SQL-like operators, such as the `SnowflakeOperator` and `BigQueryOperator`, support retrieving SQL files from this path, as described in the [doc](https://www.astronomer.io/guides/airflow-sql-tutorial).

At the moment, Astro 0.5.1 expects users to declare the path to the SQL files as an argument of the `aql.render` method.

**Acceptance criteria**
* `aql.render` tries to fetch SQL files from the DAG's `template_searchpath`.
* If both `aql.render` path and the `template_searchpath` are declared, the priority should be given to the `path` value.
* Ask @mag3141592 (bug reporter) to review the PR

dimberman:
@ashb @tatiana I think there are a few "gotchas" in this one.

1. What do we do if there are multiple values in the template_searchpath? Wouldn't this only work once? Like we wouldn't be able to run render mutliple times unless we're point to subdirectories? If we're pointing to subdirectories how do we know what order to do them in?

sunank200:
Is this still relevant @tatiana ?

PR #231, opened 2022-03-22
aql.render can now use the dag's template_searchpath

dimberman:
**Context**
When instantiating Airflow DAGs, users can declare `template_searchpath`, which is a list of folders (non-relative) that define where jinja will look for templates. By default, it contains the DAGs folder.

Many SQL-like operators, such as the `SnowflakeOperator` and `BigQueryOperator`, support retrieving SQL files from this path, as described in the [doc](https://www.astronomer.io/guides/airflow-sql-tutorial).

At the moment, Astro 0.5.1 expects users to declare the path to the SQL files as an argument of the `aql.render` method.

**Acceptance criteria**
* `aql.render` tries to fetch SQL files from each of the DAG's `template_searchpath` variables.
* If a user supplies a "path" variable, `aql.render` will check each template searchpath for that subpath (so if template_searchpath =['foo/bar', 'fizz/buzz'] and path='baz', we will search ['foo/bar/baz', 'fizz/buzz/baz']
* If there are no matches for directories in any of those template_searchpaths, we will return the original path and assume it is an absolute path.
* Ask @mag3141592 (bug reporter) to review the PR


tatiana:
@dimberman just confirming, does this ticket relate to #123 ?
Regarding the acceptance criteria, perhaps `template_searchpath` should be the default. Still, if the user specifies a path in `render,` that would take priority - so a user can have a default location but customize in case of need. What are your thoughts on this?

tatiana (commented):
It's looking good, @dimberman ! I think it would be just interesting to align on the default/specific behaviour - and where we implement each of them.

tatiana:
@dimberman, I have another question: how does this work relate to #140 ?

I also remember some discussion about not being able to use `default_args` to set a default schema per DAG due to Task Flow API limitations. Could the approach adopted in this PR be also used in that context?

dimberman:
> @dimberman, I have another question: how does this work relate to #140 ?
> 
> I also remember some discussion about not being able to use `default_args` to set a default schema per DAG due to Task Flow API limitations. Could the approach adopted in this PR be also used in that context?

@tatiana the difference here is that template_searchpath is an existing Airflow functionality, so it wouldn't make sense for us to override that. I think that default_args for other more `astro` specific things would make sense (e.g. conn_ids and databases), but we can defer to the standard here.

example_dags/example_postgres_render/example_postgres_render.py
```diff
@@ -42,7 +44,6 @@ def print_results(df: pd.DataFrame):
     print(df.to_string())
 
 
-dir_path = os.path.dirname(os.path.realpath(__file__))
 with dag:
-    models = aql.render(dir_path + "/models")
+    models = aql.render(path="models")
```
utkarsharma2, line 48: @dimberman, what it means to have "path=models"?

src/astro/sql/parsers/sql_directory_parser.py
```diff
@@ -6,22 +7,91 @@
...
+    """
+    template_path = DagContext.get_current_dag().template_searchpath
+    if template_path:
+        for t in template_path:
+            try:
+                os.listdir(t + "/" + path)
+                logging.info("Template_path found. rendering %s", t + "/" + path)
+                return t + "/" + path
+            except FileNotFoundError:
+                logging.info("Could not find template_path %s", t + "/" + path)
```
utkarsharma2, line 33: Shouldn't we continue with the loop, if this errors out?

src/astro/sql/parsers/sql_directory_parser.py
```diff
@@ -6,22 +7,91 @@
...
+    dependencies. The result of this function is a dictionary of tasks that you can reference in downstream tasks.
+
+    e.g.:
+
+    dag = DAG(
+        dag_id="example_postgres_render",
+        start_date=datetime(2019, 1, 1),
+        max_active_runs=3,
+        schedule_interval=timedelta(minutes=30),
+        default_args=default_args,
```
tatiana, line 58: We probably can remove these lines from the example, to keep it as simple as possible

src/astro/sql/parsers/sql_directory_parser.py
```diff
@@ -6,22 +7,91 @@
...
+        template_searchpath=dir_path,
+    )
+
+
+    @adf
+    def print_results(df: pd.DataFrame):
+        print(df.to_string())
```
tatiana, line 65: This forces the user who is learning to use `render` to also have to learn about the Dataframe operator. Would there be a simpler way of showing the benefit of `render`?

src/astro/sql/parsers/sql_directory_parser.py
```diff
@@ -6,22 +7,91 @@
...
+def get_path_for_render(path):
+    """
+    Generate a path using either the DAG's template_searchpath with a relative path, or with the relative path
+    to the DAG file
+
+    :param path:
+    :return:
+    """
+    template_path = DagContext.get_current_dag().template_searchpath
+    if template_path:
```
tatiana, line 26: @dimberman in the original ticket description (#123 ), we had:
>  If both `aql.render` path and the template_searchpath are declared, the priority should be given to the path value.

Why did you change this behavior?
Also, since  @mag3141592 was the one who reported this originally, I think it would be great to have her reviewing the PR/feature.

tatiana (commented):
@dimberman a few minor comments, and one major concern: the fact the implementation does not match the behavior suggested in the ticket #123

src/astro/sql/parsers/sql_directory_parser.py
dimberman: @tatiana because choosing one or the other doesn't match the existing behavior. 

If you look at [this example](https://stackoverflow.com/questions/50761084/using-airflow-template-files-and-template-searchpath-in-google-cloud-composer), the template_searchpath is treated as a base directory, not as the whole path. People still point to the file or subdirectory they want to use.

src/astro/sql/parsers/sql_directory_parser.py
dimberman: @tatiana I'd imagine the only two functions that would make sense here are transform and dataframe. No  preference on which. The dataframe function IMO is cooler because it shows you can move the SQL into python

src/astro/sql/parsers/sql_directory_parser.py
dimberman: @tatiana I discussed with Maggie and have changed the ticket to reflect our discussion.

tatiana (approved):
HI @dimberman , thanks a lot! Since Maggie is supportive of this approach - and it is consistent with Airflow :shipit:

PR #237, opened 2022-03-22
attempt to split by database

dimberman:


.github/workflows/ci.yaml
```diff
@@ -71,7 +71,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        group: [ 1, 2, 3, 4, 5, 6, 7, 8 ]
+        database: ["postgres", "snowflake", "bigquery", "sqlite", "not postgres and not snowflake and not sqlite and not bigquery"]
+        group: [ 1, 2]
```
utkarsharma2, line 75: @dimberman How is this grouping working? 

For example if postgres have 4 test:
Will it create two groups 
postgres-1
postgres-2


Also, is it still deterministic?

PR #241, opened 2022-03-24
Add CodeSpell as pre-commit check to avoid typos

kaxil:
Uses https://github.com/codespell-project/codespell to check code for common misspellings


utkarsharma2 (approved):
Much needed! :)

PR #242, opened 2022-03-24
Fix commands in `README.md`

kaxil:
This PR fixes 4 issues in our `README.md`:
(1) It changes the `date` command to make it run on MacOS. Otherwise MacOS users will see the following error:

```
❯ date --iso-8601=seconds
date: illegal option -- -
usage: date [-jnRu] [-r seconds|file] [-v[+|-]val[ymwdHMS]]
            [-I[date | hours | minutes | seconds]]
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]

```

(2) The `dag_id` was wrong in `airflow dags test` command
(3) Fix URL for Airflow homepage
(4) The sqlite_default host is different for MAC vs. Linux so I have fixed that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 **astro** allows rapid and clean development of {Extract, Load, Transform} workflows using Python.
 It helps DAG authors to achieve more with less code.
-It is powered by [Apache Airflow](https://www.airflow.apache.org) and maintained by [Astronomer](https://astronomer.io).
+It is powered by [Apache Airflow](https://airflow.apache.org) and maintained by [Astronomer](https://astronomer.io).
 
 > :warning: **Disclaimer** This project's development status is alpha. In other words, it is not production-ready yet.
 The interfaces may change. We welcome alpha users and brave souls to test it - any feedback is welcome.
@@ -40,7 +40,7 @@ pip install astro-projects[amazon,google,snowflake,postgres]
 
 ## Quick-start
 
-After installing Astro, copy the following example dag `calculate_top_animations.py` to a local directory named `dags`:
+After installing Astro, copy the following example dag `calculate_popular_movies.py` to a local directory named `dags`:
 
 ```Python
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,12 +94,25 @@ airflow db init
 Create an SQLite database for the example to run with and run the DAG:
 ```
 sqlite3 /tmp/sqlite_default.db "VACUUM;"
-airflow dags test calculate_top_animations `date --iso-8601=seconds`
+airflow dags test calculate_popular_movies `date -Iseconds`
```
tatiana, line 97: It might be worth updating on line 43 to also have the same name

README.md
```diff
@@ -94,12 +94,25 @@ airflow db init
...
 ```
 
 Check the top five animations calculated by your first Astro DAG by running:
-```commandline
-sqlite3 /tmp/sqlite_default.db "select * from top_animation;" ".exit"
+```shell
+# The sqlite_default host is different for MAC vs. Linux
```
tatiana, line 102: :see_no_evil: I didn't realise that was the case

tatiana (approved):

README.md
kaxil: L43 is the name of the dag file though, and this is the dag_id

README.md
kaxil: I didn’t know too 😄 - It is because of the following line:

https://github.com/apache/airflow/blob/f06b3955b1d937138fb38021a6a373b94ae8f9e8/airflow/utils/db.py#L560

Which returns a different dir /var/random_nums  on my mac🤦‍♂️

README.md
tatiana: I know - but it can simplify to the end-user keeping them with the same value :)

README.md
```diff
@@ -94,12 +94,25 @@ airflow db init
 Create an SQLite database for the example to run with and run the DAG:
 ```
 sqlite3 /tmp/sqlite_default.db "VACUUM;"
```
utkarsharma2, line 96: Why do we need this line? isn't it redundant on mac?

README.md
kaxil: fixed :)

README.md
kaxil: Updated to use env var to get file name.

PR #243, opened 2022-03-25
Simplify Code Coverage

kaxil:
This PR/commit:

- Moves codecov settings to `pyproject.toml` so we don't need to set it at each place in `ci.yaml`
- Explicitly sets `--cov=src` so we don't need to exclude "tests" or any other directory
- Also fixes issue where CodeCov comments and fails early after a single report is uploaded as we run our tests in parallel

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,7 +3,8 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 2%
-ignore:
-  - "test_*.py"
-  - "*/**/tests/*"
+        threshold: 5%
+        only_pulls: true
+  notify:
+    wait_for_ci: yes
+    after_n_builds: 8
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,8 @@ Home = "https://astronomer.io/"
 provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
+minversion = "6.0"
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -86,6 +86,8 @@ Home = "https://astronomer.io/"
 provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
+minversion = "6.0"
+addopts = "--cov=src --cov-report=xml --cov-branch"
```
tatiana, line 90: @kaxil locally, we probably would instead use `--cov-report=report-missing` - how do you suggest we handle both different use cases (local x CI)?

kaxil: Yeah that should still work, for example

```
❯ nox -s test -- --cov-report=term-missing
```

codecov.yml
```diff
@@ -4,6 +4,3 @@ coverage:
       default:
         target: auto
         threshold: 2%
-ignore:
-  - "test_*.py"
-  - "*/**/tests/*"
```
dimberman, line 9: @kaxil why did you remove this? We don't want test files counting in code coverage otherwise we'll get incorrect reports that coverage is going down when we write tests. Is there somewhere else where this is being handled?

dimberman (dismissed):

PR #244, opened 2022-03-25
Refactor stats check

utkarsharma2:
Closes: #236

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,22 +2,27 @@ import os
 import pathlib
 import uuid
 
+import pandas as pd
 import pytest
 import yaml
 from airflow import settings
+from airflow.hooks.base import BaseHook
 from airflow.models import DAG, Connection, DagRun
 from airflow.models import TaskInstance as TI
 from airflow.utils import timezone
 from airflow.utils.db import create_default_connections
 from airflow.utils.session import create_session, provide_session
 
-from astro.constants import Database
+from astro.constants import DEFAULT_CHUNK_SIZE
 from astro.settings import SCHEMA
 from astro.sql.table import Table, TempTable, create_unique_table_name
+from src.astro.constants import FileType
+from astro.sql.table import Table, TempTable
 from astro.utils.cloud_storage_creds import parse_s3_env_var
 from astro.utils.database import get_database_name
-from astro.utils.dependencies import BigQueryHook, gcs, s3
-from src.astro.constants import FileType
+from astro.utils.dependencies import BigQueryHook, PostgresHook, SnowflakeHook, gcs, s3
+from astro.constants import FileType, Database
+from astro.utils.load_dataframe import move_dataframe_to_sql
 from tests.operators import utils as test_utils
 
 DEFAULT_DATE = timezone.datetime(2016, 1, 1)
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -76,46 +80,81 @@ def sample_dag():
...
+            }
+        elif database == Database.SQLITE:
+            default_table_options = {
+                "conn_id": hook.sqlite_conn_id,
+                "database": "sqlite",
+            }
+        elif database == Database.BIGQUERY:
+            default_table_options = {"conn_id": hook.gcp_conn_id, "schema": SCHEMA}
+        else:
+            raise ValueError("Unsupported Database")
```
tatiana, line 134: Since we're not covering this line, and this is part of the test suite, we may want to delete it - what are your thoughts?

src/astro/sql/operators/agnostic_stats_check.py
```diff
@@ -230,7 +228,19 @@ def handler_func(results):
...
-        metadata = MetaData()
+
+        valid_db = {
+            "postgres": Database.POSTGRES,
+            "postgresql": Database.POSTGRES,
+            "bigquery": Database.BIGQUERY,
+            "google_cloud_platform": Database.BIGQUERY,
+        }
+
+        metadata_params = {}
```
tatiana, line 239: An alternative to line 229 and these would be to capture the DB using:
```
from astro.utils.database import get_database_from_conn_id
database = get_database_from_conn_id(self.conn_id)
```

src/astro/sql/operators/agnostic_stats_check.py
```diff
@@ -170,8 +170,6 @@ def prepare_failed_checks_results(
 
 
 class AgnosticStatsCheck(SqlDecoratedOperator):
-    template_fields = ("table",)
```
tatiana, line 173: Why is this no longer necessary?

tests/operators/test_agnostic_stats_check.py
```diff
@@ -28,154 +29,94 @@
...
+                main_table=test_table[0],
+                compare_table=test_table[1],
                 checks=[aql.OutlierCheck("room_check", {"rooms": "rooms"}, 2, 0.0)],
                 max_rows_returned=10,
             )
-            a.execute({"run_id": "foo"})
-            assert True
-        except ValueError as e:
-            assert False
+        test_utils.run_dag(sample_dag)
```
tatiana, line 122: Is there any other type of meaningful assert we could add to this test?

tests/operators/test_agnostic_stats_check.py
```diff
@@ -28,154 +29,94 @@
...
+)
+def test_stats_check_postgres_outlier_exists(sample_dag, sql_server, test_table):
+    with sample_dag:
+        aql.stats_check(
+            main_table=test_table[0],
+            compare_table=test_table[1],
+            checks=[aql.OutlierCheck("room_check", {"rooms": "rooms"}, 2, 0.0)],
+            max_rows_returned=10,
+        )
+        test_utils.run_dag(sample_dag)
```
tatiana, line 75: Is there any other type of meaningful assert we could add to this test?

tatiana (commented):
Great work, @utkarsharma2 , some minor comments!

conftest.py
```diff
@@ -76,46 +80,81 @@ def sample_dag():
         session.query(TI).delete()
 
 
+def populate_table(path: str, table: Table, hook: BaseHook) -> None:
+    """
+    Populate a csv file into a sql table
+    :param path: path to csv file
```
kaxil, line 86:
```suggestion
    Populate a csv file into a sql table

    :param path: path to csv file
```

We need a blank line between docstring description/summary and the param docs

kaxil: This is needed by Sphinx

conftest.py
```diff
@@ -76,46 +80,81 @@ def sample_dag():
...
+    :param table: Astro table object
+    :return: None
```
kaxil, line 88: We can exclude this since we have it in TypeHints

src/astro/sql/operators/agnostic_stats_check.py
utkarsharma2: I think was a bug, since we don't have a parameter with this name. this was not caught since the Operator was not running as a task in a DAG.

tests/operators/test_agnostic_stats_check.py
utkarsharma2: Yes, we should I'll update.

tests/operators/test_agnostic_stats_check.py
utkarsharma2: Yes, I'll update.

src/astro/sql/operators/agnostic_stats_check.py
utkarsharma2: Done.

conftest.py
utkarsharma2: Done

conftest.py
tatiana: That's fine, @utkarsharma2 , it's a one-liner - in the tests :)

tatiana (approved):
:rocket:

PR #246, opened 2022-03-25
Add utils.delete module

tatiana:
Note: after this, there will be only one last PR  so we can finally close the ticket :raised_hands: 
Relates to: #98

### DIFF
--- a/src/astro/sql/table.py
+++ b/src/astro/sql/table.py
@@ -1,7 +1,10 @@
+import random
+import string
+
 from airflow.hooks.base import BaseHook
 from airflow.models import DagRun, TaskInstance
 
-from astro.constants import Database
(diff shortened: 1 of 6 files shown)


src/astro/sql/table.py
```diff
@@ -110,3 +113,16 @@ def create_table_name(context) -> str:
     if not table_name.isidentifier():
         table_name = f'"{table_name}"'
     return table_name
+
+
+def create_unique_table_name(size: int = UNIQUE_TABLE_NAME_SIZE) -> str:
```
kaxil, line 118: Should we rename `size` to `length`?

kaxil (approved):
A minor suggestion but LGTM otherwise

src/astro/sql/table.py
tatiana: I'll do this, @kaxil , thank you!

PR #253, opened 2022-03-29
Add quickstart guide to documentation

dimberman:
This PR will add a basic quickstart guide for users to use our python SDK. To begin we will only go over transform, dataframe, and render, and we can add more functions as we go.


README.md
```diff
@@ -151,7 +151,10 @@ A summary of the currently available operations in **astro**. More details are a
 ## Documentation
 
 The documentation is a work in progress, and we aim to follow the [Diátaxis](https://diataxis.fr/) system:
-* **Tutorial**: a hands-on introduction to **astro**
+* **[Python-only Tutorial](docs/QUICK_START_PYTHON.md)**: a hands-on introduction to **astro**'s python SDK.
+We'll go over basic concepts and functions for writing entire sql+python flows in your DAG file.
+* **[Python-with-sql Tutorial](docs/QUICK_START_SQL.md)**: a hands-on introduction to **astro**'s SQL rendering engine.
```
tatiana, line 156: @dimberman  My understanding is that the `render` is a feature that is part of the Python SDK , what are your thoughts?

If we agree,  perhaps we could make this more explicit by saying `SQL Python rendering engine`, what do you think? Or perhaps referring the Python SDK in this second quickstart as well?

docs/QUICK_START_SQL.md
```diff
@@ -0,0 +1,251 @@
+**NOTE: Please read the [python quickstart guide](QUICK_START_PYTHON.md) before reading this section. We will assume
```
tatiana, line 1: Given some of the ideas we've discussed recently, I think it may be misleading to call this document QUICK_START_SQL - since it is Python with SQL. Someone only with knowledge of SQL would be unlike to follow this path.

PR #254, opened 2022-03-29
Support `params` argument in `aql.render` to declare values to SQL Jinja template #125

dimberman:
**Context**
At the moment, Astro 0.5.1 does not use the `params` keyword argument. This argument would allow us to overwrite SQL Jinja template values.

Example:
```
SELECT
   t.table_schema,
   t.table_name,
   CONCAT(t.table_catalog, '.', t.table_schema, '.', t.table_name) table_path
FROM dwh_legacy.information_schema.tables t
WHERE
       t.table_schema IN ({{ params.schemas }})
   AND t.table_type = 'BASE TABLE';
```

```
aql.render(path='/usr/local/airflow/include/sql/schemas',
                                      params={'schemas': dwh_schemas})
```

Currently raises the error:
```
  File "/sr/local/lib/python3.9/site-packages/jinja2/runtime.py", line 903, in _fail_with_undefined_error
      raise self._undefined_exception(self._undefined_message)
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'schemas'
```

**Acceptance criteria**
* Users of `aql.render` can override Jinja template values by using the `params`  argument
* Have an `example_dag` which uses `params`
* Add at least one unit test which validate this behaviour
* Update the docstring so it details this feature
* Ask @mag3141592 (bug reporter) to review the PR

This PR addresses one part of the previous PR that was not correctly implemented, namely the adding params at a task level as well as at the DAG level

addresses https://github.com/astro-projects/astro/issues/125

### DIFF
--- a/src/astro/sql/parsers/sql_directory_parser.py
+++ b/src/astro/sql/parsers/sql_directory_parser.py
@@ -61,6 +61,7 @@ def render_single_path(
     schema: Optional[str] = None,
     warehouse: Optional[str] = None,
     role: Optional[str] = None,
+    **kwargs,
 ):
     # Parse all of the SQL files in this directory
     current_files = [
@@ -86,6 +87,8 @@ def render_single_path(
             operator_kwargs = set_kwargs_with_defaults(
                 front_matter_opts, conn_id, database, role, schema, warehouse
             )
+            if kwargs.get("params"):
+                operator_kwargs["params"] = kwargs.get("params")
 
             p = ParsedSqlOperator(
                 sql=sql,
@@ -169,8 +172,8 @@ def render(
             schema=schema,
             warehouse=warehouse,
             role=role,
+            **kwargs,
         )
-    print("Sdafasd")
 
     # Add the XComArg to the parameters to create dependency
     for filename in all_file_names:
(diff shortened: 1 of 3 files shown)


tests/parsers/test_sql_directory_parser.py
```diff
@@ -121,7 +120,9 @@ def test_parse_creates_xcom(self):
                     warehouse=os.getenv("SNOWFLAKE_WAREHOUSE"),
                 ),
             )
-            aql.render("single_task_dag", input_table=input_table)
+            aql.render(
+                "single_task_dag", input_table=input_table, params={"col": "rooms"}
+            )
```
kaxil, line 125: Please add a separate unit test to check that using `{{ params.col }}` produces `room`

src/astro/sql/parsers/sql_directory_parser.py
```diff
@@ -86,6 +87,8 @@ def render_single_path(
             operator_kwargs = set_kwargs_with_defaults(
                 front_matter_opts, conn_id, database, role, schema, warehouse
             )
+            if kwargs.get("params"):
+                operator_kwargs["params"] = kwargs.get("params")
```
kaxil, line 91: How would a user know that they can pass `params` dictionary to this function?

dimberman: @kaxil I made it an explicit parameter with docstring.

kaxil:
Cool, only 1 item remaining - https://github.com/astro-projects/astro/pull/254#discussion_r837823048

kaxil (approved):

PR #255, opened 2022-03-29
Fix issue with pycharm debugging

dimberman:
This PR fixes an issue where pycharm breakpoints were not working
because we were running codecoverage during debugging. This has been a
documented issue with other libraries (e.g. https://github.com/HypothesisWorks/hypothesis/issues/985)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,6 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--cov=src --cov-report=xml --cov-branch"
 env_files = [".env"]
 testpaths = ["tests"]
 markers = [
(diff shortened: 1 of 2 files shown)


kaxil (changes requested):
Let me try to fix it or document how we can bypass this, moving "addopts" for Pycharm's limitations/bug isn't great IMO. 

I will try to find a solution, if I don't find one then will happily merge this PR :)

pyproject.toml
```diff
@@ -85,7 +85,6 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--cov=src --cov-report=xml --cov-branch"
```
tatiana, line 88: @kaxil are you happy with reverting this?

kaxil (approved):
Adding `--no-cov` in the "Additional Arguments" field in the pytest run configuration template makes it work again.

https://youtrack.jetbrains.com/issue/PY-20186#focus=Comments-27-5470118.0-0

Although I will merge it so devs don't need to put the additional args on it

Issue #73, opened 2022-02-07
Unable to load_file when using nested NDJSON


tatiana:
Version: Astro 0.3.3

Astro is currently unable to load_file if the origin NDJSON is nested.

How to reproduce:
```
    github_nested_table = aql.load_file(
        path="gs://dag-authoring/github/github_nested_000000000007.ndjson",
        task_id="load_ndjson",
        #file_conn_id="google_cloud_default",
        output_table=Table(
            table_name="github_projects_counts_by_language",
            database="postgres",
            conn_id="postgres_conn"
        )
```

Output:
```
Traceback (most recent call last):
  File "/home/tati/.virtualenvs/astro-private/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1802, in _execute_context
    self.dialect.do_execute(
  File "/home/tati/.virtualenvs/astro-private/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
psycopg2.ProgrammingError: can't adapt type 'dict'

The above exception was the direct cause of the following exception:
```

tatiana:
Some options are detailed in:
https://www.psycopg.org/docs/extras.html#json-adaptation

dimberman:
We believe that a good first solution would be to flatten the ndjson and prefix the path to each key

utkarsharma2:
PR - https://github.com/astro-projects/astro/pull/257 Closes part of the ticket. We are handling only single-level nesting. 
Ndjson - 
{"a": {"b": ["a", "b"]}}


But not, 

Ndjson - 
{"a": {"b": [{"a":  "a"}, {"c": "a"}]}}

Above Json results in - 
sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) can't adapt type 'dict'
[SQL: INSERT INTO tmp_astro.e23vd9fl2gkcn4zb_load_file__1 ("a.b") VALUES (%(a.b_m0)s)]
[parameters: {'a.b_m0': [{'a': 'a'}, {'c': 'a'}]}]

kaxil:
@utkarsharma2 Are you planning to tackle the second case in a different PR?

utkarsharma2:
@kaxil yes since we need an interface that is completely agnostic of the underline lib used to flatten Ndjson. My earlier assumption was to just expose the pandas json_normalize method to do this.

PR #257, opened 2022-03-30
Normalize ndjson 2

utkarsharma2:
Closes: #73


src/astro/utils/load.py
```diff
@@ -58,14 +62,77 @@ def load_file_into_dataframe(
...
+     :type normalize_config: dict
+     :return: return updated config
+     :rtype: `dict`
+    """
+
+    replacement = "_"
+    illegal_char = "."
+
+    meta_prefix = normalize_config.get("meta_prefix")
+    if meta_prefix and meta_prefix == illegal_char:
```
tatiana, line 122: What are we gaining by exposing this parameter to end-uses?

src/astro/utils/load.py
```diff
@@ -58,14 +62,77 @@ def load_file_into_dataframe(
...
+        normalize_config["meta_prefix"] = replacement
+
+    record_prefix = normalize_config.get("record_prefix")
+    if record_prefix and record_prefix == illegal_char:
```
tatiana, line 126: What are we gaining by exposing this parameter to end-users?

tatiana (commented):
Thank you very much for addressing the previous comments, @utkarsharma2 !

I noticed the coverage of `load` went down, would it be possible to write some unit tests to cover the missing parts?
https://app.codecov.io/gh/astro-projects/astro/compare/232/tree/src/astro/utils/load.py

I added a few other comments inline as well

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         file_conn_id="",
         chunksize=DEFAULT_CHUNK_SIZE,
         if_exists="replace",
+        normalize_config=None,
```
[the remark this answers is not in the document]
utkarsharma2: Yes, I think it's required since we cannot make any assumptions about the field the user wants to add to flattened ndjson.

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -161,6 +169,9 @@ def load_file(
     :type file_conn_id: str
     :param task_id: task id, optional.
     :type task_id: str
+    :param normalize_config: config dict for pandas json_normalize method
+        https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html
```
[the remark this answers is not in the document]
utkarsharma2: I agree it's an implementation detail but since we are just using their function I think it would be wise to expose it since we are not providing a subset of features or modifying them in any way.

src/astro/utils/load.py
utkarsharma2: if we rely on the Pandas method to flatten the NDJSON, we do expose the interface to the user.

kaxil (commented):
Can you update the PR title and description, changelogwise I won't understand what "Normalize ndjson" means

src/astro/sql/operators/agnostic_load_file.py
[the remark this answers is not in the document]
tatiana: We don't need to give the user all the control over how to generate the JSON. Just an alternative for the default `_` we'll be using.

How do we as a library can verify/test if we expose the entire Pandas interface? We'd need to support more options, have more tests, and more code to maintain.

Let’s trim this down to only expose the param/arg which needs to be exposed.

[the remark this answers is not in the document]
tatiana: We don't need to give the user all the control over how we normalize the JSON. Just an alternative for the default `_` we'll be using should be enough.

How do we as a library can verify/test if we expose the entire Pandas interface? We'd need to support more options, have more tests, and more code to maintain. If we decide to move out of Pandas, we would still need to support all these features.

Let’s trim this down to only expose the arg which needs to be exposed.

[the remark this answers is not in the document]
utkarsharma2: @tatiana  I think there is some gap in my understanding of the expected result:

Consider this example of JSON file:
{
	“a”: {
		“b”: {
			c: “some string”,
			d: [“e”, ‘f’, ‘g’]
		}
	}
}
{
	“a”: {
		“b”: {
			c: “some string”,
			d: [“e”, ‘f’, ‘g’]
		}
	}
}

Can you please tell me what is expected according you?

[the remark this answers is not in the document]
utkarsharma2: what if the user doesn't want a certain col to be populated into a table?

utkarsharma2:
@kaxil added the description and updated a PR Title.

src/astro/utils/load.py
```diff
@@ -58,14 +62,71 @@ def load_file_into_dataframe(
...
+    """
+    Flatten the nested ndjson/json.
+     :param normalize_config: parameters in dict format of pandas json_normalize() function.
+        https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html
+     :param stream: io.TextIOWrapper object for the file
+     :type normalize_config: dict
+     :type stream: io.TextIOWrapper
+     :return: return dataframe containing the loaded data
+     :rtype: `pandas.DataFrame`
+    """
```
kaxil, line 85:
```suggestion
    """
    """
    Flatten the nested ndjson/json.

    :param normalize_config: parameters in dict format of pandas json_normalize() function.
        https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html
    :param stream: io.TextIOWrapper object for the file
    :type normalize_config: dict
    :type stream: io.TextIOWrapper
    :return: return dataframe containing the loaded data
    :rtype: `pandas.DataFrame`
    """
```

You can also ignore type since you are using Type hints

src/astro/utils/load.py
```diff
@@ -58,14 +62,73 @@ def load_file_into_dataframe(
...
         else:
             raise ValueError(f"Unable to load file '{filepath}' of type '{filetype}'")
         return dataframe
 
 
+def flatten_ndjson(
+    normalize_config: Union[None, dict], stream: io.TextIOWrapper
+) -> pd.DataFrame:
+    """
+    """
```
kaxil, line 77:
```suggestion
    """
```

src/astro/utils/load.py
```diff
@@ -58,14 +62,73 @@ def load_file_into_dataframe(
...
+    Flatten the nested ndjson/json.
+
+    :param normalize_config: parameters in dict format of pandas json_normalize() function.
```
kaxil, line 80:
```suggestion
    """
    Flatten the nested ndjson/json.

    :param normalize_config: parameters in dict format of pandas json_normalize() function.
```

src/astro/utils/load.py
```diff
@@ -58,14 +62,72 @@ def load_file_into_dataframe(
...
+
+def check_normalize_config(normalize_config, database: Database) -> dict:
+    """
+    Validate pandas json_normalize() parameter for databases, since default params result in
+     invalid column name. Default parameter result in the columns name containing '.' char.
+     :param normalize_config: :param normalize_config: config in dict format of pandas json_normalize() function.
+        https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html
+     :type normalize_config: dict
+     :return: return updated config
+     :rtype: `dict`
```
kaxil, line 109: Formatting (tabs/spaces/newlines) is off, it should be same as L77-L85

utkarsharma2: Done.

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -41,6 +46,7 @@ def __init__(
         file_conn_id="",
         chunksize=DEFAULT_CHUNK_SIZE,
         if_exists="replace",
+        ndjson_normalize_sep="_",
```
kaxil, line 49: Can you add this to docstring in L24-L33

src/astro/utils/load.py
```diff
@@ -58,14 +62,74 @@ def load_file_into_dataframe(
...
             raise ValueError(f"Unable to load file '{filepath}' of type '{filetype}'")
         return dataframe
 
 
+def flatten_ndjson(
+    normalize_config: Union[None, dict], stream: io.TextIOWrapper
+) -> pd.DataFrame:
+    """
+    Flatten the nested ndjson/json.
+    :param normalize_config: parameters in dict format of pandas json_normalize() function.
```
kaxil, line 78:
```suggestion
    Flatten the nested ndjson/json.

    :param normalize_config: parameters in dict format of pandas json_normalize() function.
```

src/astro/utils/load.py
```diff
@@ -58,14 +62,74 @@ def load_file_into_dataframe(
...
+            )
+        rows = stream.readlines(DEFAULT_CHUNK_SIZE)
+    return df
+
+
+def populate_normalize_config(ndjson_normalize_sep, database: Database) -> dict:
+    """
+    Validate pandas json_normalize() parameter for databases, since default params result in
+    invalid column name. Default parameter result in the columns name containing '.' char.
+    :param ndjson_normalize_sep: separator used to normalize nested ndjson.
```
kaxil, line 103:
```suggestion
    invalid column name. Default parameter result in the columns name containing '.' char.

    :param ndjson_normalize_sep: separator used to normalize nested ndjson.
```

kaxil (approved):
2 minor suggestions, lgtm otherwise

src/astro/sql/operators/agnostic_load_file.py
utkarsharma2: Done.

tatiana (commented):
Looks great, thank you very much, @utkarsharma2 !

Since this PR does fully close #73, it is a partial fix, I strongly suggest we change the description from "closes" to "Relates".

And we can cover the  missing use-case in a separate PR:
```
{"key": {"value1": 1, "value2": [{"inner_key": 2}]}}
```

Issue #179, opened 2022-03-08
Allow community members to run the tests in a CI


tatiana:
**Describe the bug**
At the moment, pull requests created from the community fail to run the Github tests. This happened in two PRs in the last two days:
1. https://github.com/astro-projects/astro/pull/172
2. https://github.com/astro-projects/astro/pull/176

**Version**
* Astro: 0.6.0

**To Reproduce**
1. Fork `Astro-projects/astro`
2. Create a PR from the fork into `astro-projects/astro`
3. See most tests failing, with messages complaining that credentials are not set, e.g.:

```
   File "/home/runner/work/astro/astro/.nox/test-3-8/lib/python3.8/site-packages/botocore/signers.py", line 93, in handler
    return self.sign(operation_name, request)
  File "/home/runner/work/astro/astro/.nox/test-3-8/lib/python3.8/site-packages/botocore/signers.py", line 165, in sign
    auth.add_auth(request)
  File "/home/runner/work/astro/astro/.nox/test-3-8/lib/python3.8/site-packages/botocore/auth.py", line 378, in add_auth
    raise NoCredentialsError()
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```
https://github.com/astro-projects/astro/runs/5463617773?check_suite_focus=true

**Expected behavior**
Community team members should be able to fork and create PRs and see the tests passing with no secrets issues.

tatiana:
We are in the process of refactoring our integration tests - once this is accomplished, it will be much easier to address this.

kaxil:
There are 2 options here:

1) Only run unit tests (which is a larger work as currently not all integrations tests are marked as integration)
2) Committers (one with write access to repo) can add a certain label to the PR to mark the PR as "trusted" and hence allow the PR to run tests with secrets from the CI

PR #258, opened 2022-03-31
Generate valid unique table name with `df` arg

jlaneve:


tatiana:
Thank you very much, I had to create a new PR (#260) to run the tests in the CI (due to #179)
The feature is merged, thank you! :rocket:

PR #263, opened 2022-04-01
Refactored dataframe testcase

utkarsharma2:
Closes: #235 
Refactored test cases:

1. Use sample_dag Fixture and run them with test_utils.run_dag(sample_dag) function.
2. Use test_table Fixture for getting temp/named tables and loading data into them.
3. Run Test cases for all supported databases.

### DIFF
--- a/tests/operators/test_dataframe.py
+++ b/tests/operators/test_dataframe.py
@@ -10,8 +10,8 @@ import astro.sql as aql
 from astro import dataframe as df
 from astro.constants import SUPPORTED_DATABASES, Database
 from astro.settings import SCHEMA
-from tests.operators import utils as test_utils
 from astro.sql.table import Table
+from tests.operators import utils as test_utils
 
 # Import Operator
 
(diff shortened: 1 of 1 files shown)


tests/operators/test_dataframe.py
```diff
@@ -1,287 +1,223 @@
...
-                    conn_id="snowflake_conn",
-                    schema=os.getenv("SNOWFLAKE_SCHEMA"),
-                    database=os.getenv("SNOWFLAKE_DATABASE"),
-                    warehouse=os.getenv("SNOWFLAKE_WAREHOUSE"),
-                )
+        }
+    ],
+    indirect=True,
+)
+def test_dataframe_from_sql_custom_task_id(sample_dag, sql_server, test_table):
```
kaxil, line 73: Are we testing that tasks with `custom_task_id` runs fine? This feels like a case where we don't need to run the dag for this case, only assess if the `task_id` is applied correctly or not>

Testing dag can be run or not seems redundant, please correct me if I am wrong

kaxil: @utkarsharma2 ^^

utkarsharma2: I think this is test is to ensure we don't generate duplicate task_id. I see your point, the `assert` itself is before the run_dag() function. Also, I think we can get rid of the last `assert` statement since that functionality is tested in other test cases.

kaxil (approved):

PR #265, opened 2022-04-01
Add example DAG to load/export from Bigquery

kaxil:
This Example DAG:
 - Pulls a CSV file from Github and loads it into BigQuery.
 - Extracts the data from BigQuery and load into in-memory Pandas Dataframe
 - Finds the Top 5 movies based on Rating using pandas dataframe
 - And loads it into a Google Cloud Storage bucket in a CSV file

closes https://github.com/astro-projects/astro/issues/150

### DIFF
--- a/example_dags/example_google_bigquery_gcs_load_and_save.py
+++ b/example_dags/example_google_bigquery_gcs_load_and_save.py
@@ -1,0 +1,54 @@
+"""
+This Example DAG:
+ - Pulls a CSV file from Github and loads it into BigQuery.
+ - Extracts the data from BigQuery and load into in-memory Pandas Dataframe
+ - Finds the Top 5 movies based on Rating using pandas dataframe
+ - And loads it into a Google Cloud Storage bucket in a CSV file
+
+Pre-requisites:
+ - Install dependencies for Astro Python SDK with Google, refer to README.md
+ - Create an Airflow Connection to connect to Bigquery Table. Example:
+    export AIRFLOW_CONN_BIGQUERY="bigquery://astronomer-dag-authoring"
+ - You can either specify a service account key file and set `GOOGLE_APPLICATION_CREDENTIALS`
+    with the file path to the service account.
+"""
+
+import pandas as pd
+from airflow.models.dag import DAG
+from airflow.utils import timezone
+
+import astro.sql as aql
+from astro import dataframe
+from astro.sql.table import Table
+
+with DAG(
+    dag_id="example_google_bigquery_gcs_load_and_save",
+    schedule_interval=None,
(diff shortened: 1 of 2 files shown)


example_dags/example_google_bigquery_gcs_load_and_save.py
```diff
@@ -0,0 +1,53 @@
+"""
+This Example DAG:
+ - Pulls a CSV file from Github and loads it into BigQuery.
+ - Extracts the data from BigQuery and load into in-memory Pandas Dataframe
+ - Finds the Top 5 movies based on Rating using pandas dataframe
+ - And loads it into a Google Cloud Storage bucket in a CSV file
+
+Pre-requisites:
+ - Create an Airflow Connection to connect to Bigquery Table. Example:
```
tatiana, line 9: What are your thoughts on encouraging users to use Airflow default connections as opposed to creating a new one?

example_dags/example_google_bigquery_gcs_load_and_save.py
```diff
@@ -0,0 +1,53 @@
+"""
+This Example DAG:
+ - Pulls a CSV file from Github and loads it into BigQuery.
+ - Extracts the data from BigQuery and load into in-memory Pandas Dataframe
+ - Finds the Top 5 movies based on Rating using pandas dataframe
+ - And loads it into a Google Cloud Storage bucket in a CSV file
+
+Pre-requisites:
```
tatiana, line 8: It may be worth to add to the pre-requisites the fact the user needs to install astro either using `astro[all]` or `astro[google]`.

tatiana (approved):
@kaxil it looks great! I have a minor suggestion and a question, inline.

example_dags/example_google_bigquery_gcs_load_and_save.py
kaxil: We recommend in Production use-cases to run `airflow db upgrade` (not `airflow db init`. Ref: [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/production-deployment.html#database-backend) ) which does not create default connections.

example_dags/example_google_bigquery_gcs_load_and_save.py
kaxil: Done on L9: I didn't add the library name purposely, so we don't have to rename it when we change the package name

kaxil: https://github.com/astro-projects/astro/blob/caa57e17db01b402ce6ca33ff5e655590b684650/example_dags/example_google_bigquery_gcs_load_and_save.py#L9

PR #291, opened 2022-04-05
Cleanup: Remove unused code

kaxil:
This PR removes some of the unused/dead code from the project.

It delete:
- `src/astro/ml` dir as they just use `dataframe` decorator currently. When in future we add additional functionalities, we can add those back
- `create_table_name`, `get_table_name` & `create_sql_engine` & others functions were just unused

closes https://github.com/astro-projects/astro/issues/277


utkarsharma2 (approved):

tatiana:
@kaxil it's looking great! Another strong candidate to be removed is: `astro/src/__init__.py`. This `src` shouldn't be a package.

kaxil:
> @kaxil it's looking great! Another strong candidate to be removed is: `astro/src/__init__.py`. This `src` shouldn't be a package.

Oh yes, good callout

Issue #136, opened 2022-02-24
Refactor tests which rely on pre-populated Snowflake database


tatiana:
**Context**
At the moment (Astro 0.5.2), some Snowflake tests run using a specific table (`ADOPTION_CENTER_1`), which was pre-populated in the past. Recently there was a Snowflake migration which made our CI break because the table was not available in the new schema:
https://github.com/astro-projects/astro/runs/5308330863?check_suite_focus=true

**Acceptance criteria**
* All Snowflake tests should be self-contained. If they rely on previous data, we should define (`pytest`) fixtures or setup/teardown methods. The setup step would load the dataset into the desired Snowflake table, and the teardown would delete the table.

dimberman:
This ticket may be addressed by #139. The simplest solution to just not use adoption center data

tatiana:
We were able to fetch from a legacy Snowflake DB the fixtures which are still used by legacy tests:

Adoption center 1
```
"2021-01-01","Dog","Stanley","3"                                                
"2021-01-01","Dog","Barney","6"
"2021-02-01","Cat","Whiskers","10"
"2021-03-01","Dog","Lila","1"
"2021-03-01","Dog","Toby","5"
"2021-04-01","Cat","Gretchen","4"
"2021-05-01","Guinea Pig","Maverick","11"
"2021-06-01","Cat","Pie","8"
"2021-07-01","Dog","Butter","12"
"2021-08-01","Dog","Peanut","2"
"2021-03-01","Cat","Fluffy","3"
"2021-04-01","Dog","Tobias","1"
"2021-04-01","Dog","Buddy","1"
"2021-04-01","Cat","Roo","2"
"2021-05-01","Guinea Pig","Piglet","2"
"2021-05-01","Cat","Ghost","3"
"2021-06-01","Dog","Ollie","4"
"2021-06-01","Dog","Eta","12"
```

And Adoption center 2
```
"2021-05-01","Cat","Dancer","3"                                                 
"2021-07-01","Dog","Prancer","1"
"2021-08-01","Dog","Comet","1"
"2021-09-01","Cat","Cupid","2"
"2021-10-01","Guinea Pig","Rudolph","4"
"2021-11-01","Cat","Spider","3"
"2021-12-01","Dog","Dasher","4"
"2021-10-01","Dog","Ettie","12"
```

kaxil:
Is this still an issue?

utkarsharma2:
[...]

PR #293, opened 2022-04-06
Improve `example_amazon_s3_snowflake_transform` dag

utkarsharma2:
Closes: #136

example_amazon_s3_snowflake_transform.py was using prepopulated tables in the snowflake database.

Removed dependency from prepopulated tables by using the `load_file` operator.

### DIFF
--- a/example_dags/example_amazon_s3_snowflake_transform.py
+++ b/example_dags/example_amazon_s3_snowflake_transform.py
@@ -46,6 +46,9 @@ def aggregate_data(df: pd.DataFrame):
     catchup=False,
 )
 def example_amazon_s3_snowflake_transform():
+
+    s3_bucket = os.getenv("S3_BUCKET", "s3://tmp9")
+
     input_table_1 = Table(
         "ADOPTION_CENTER_1",
         database=os.environ["SNOWFLAKE_DATABASE"],
@@ -60,12 +63,12 @@ def example_amazon_s3_snowflake_transform():
     )
 
     temp_table_1 = aql.load_file(
-        path="s3://tmp9/ADOPTION_CENTER_1.csv",
+        path=f"{s3_bucket}/ADOPTION_CENTER_1.csv",
         file_conn_id="",
         output_table=input_table_1,
     )
     temp_table_2 = aql.load_file(
-        path="s3://tmp9/ADOPTION_CENTER_2.csv",
+        path=f"{s3_bucket}/ADOPTION_CENTER_2.csv",
         file_conn_id="",
         output_table=input_table_2,
     )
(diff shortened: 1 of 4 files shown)


example_dags/example_amazon_s3_snowflake_transform.py
```diff
@@ -46,19 +46,33 @@ def aggregate_data(df: pd.DataFrame):
...
+        path="s3://tmp9/ADOPTION_CENTER_1.csv",
+        file_conn_id="",
+        output_table=input_table_1,
+    )
+    temp_table_2 = aql.load_file(
+        path="s3://tmp9/ADOPTION_CENTER_2.csv",
+        file_conn_id="",
+        output_table=input_table_2,
+    )
+
```
tatiana, line 72: @utkarsharma2 would it be an option to use that table fixture, so we don't leave trash behind, after running the test?

utkarsharma2: @tatiana Yes, It's an issue that we leave the temp tables behind. But I think we need a different way to clear them since they are example dags that are meant to be used by end-Users and adding any kind of fixtures means they also need the same setup to run them.

example_dags/example_amazon_s3_snowflake_transform.py
```diff
@@ -46,19 +46,33 @@ def aggregate_data(df: pd.DataFrame):
...
+        conn_id="snowflake_conn",
+    )
+
+    temp_table_1 = aql.load_file(
+        path="s3://tmp9/ADOPTION_CENTER_1.csv",
+        file_conn_id="",
+        output_table=input_table_1,
+    )
+    temp_table_2 = aql.load_file(
+        path="s3://tmp9/ADOPTION_CENTER_2.csv",
```
kaxil, line 68: I see some inconsistencies with using env vars for certain fields and not for others.

We could do same as OSS so a dag can be run. For example:

```python
        s3_bucket = os.get_environ.get("S3_BUCKET", "s3://tmp9")
        path=f"{s3_bucket}/ADOPTION_CENTER_1.csv",
```

kaxil: This way users can too run them by just setting up environment variables without having to change the Example DAG, wdyt?

kaxil: This does not need to block this PR but can you create a backlog issue to make example DAGs consistent with that pattern if you agree

example_dags/example_amazon_s3_snowflake_transform.py
tatiana: Fair point, @utkarsharma2 !

example_dags/example_amazon_s3_snowflake_transform.py
utkarsharma2: @kaxil Yes agreed I'll make this change for the current dag, and I'll create a ticket for the remaining dags.

kaxil (approved):

PR #294, opened 2022-04-06
Fix issues with using BigQuery Connection created from UI

kaxil:
This PR creates a constant for `CONN_TYPE_TO_DATABASE` and adds `gcpbigquery` to it. It then uses this for other places .

closes https://github.com/astro-projects/astro/issues/252

### DIFF
--- a/src/astro/constants.py
+++ b/src/astro/constants.py
@@ -36,3 +36,13 @@ SUPPORTED_FILE_TYPES = [const.value for const in FileType]
 SUPPORTED_DATABASES = list({const.value for const in Database})
 
 UNIQUE_TABLE_NAME_LENGTH = 63
+
+CONN_TYPE_TO_DATABASE = {
+    "postgres": Database.POSTGRES,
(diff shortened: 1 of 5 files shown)


kaxil:
Would love reviews on this @utkarsharma2 @tatiana

src/astro/constants.py
```diff
@@ -36,3 +36,13 @@ class Database(Enum):
 SUPPORTED_DATABASES = list({const.value for const in Database})
 
 UNIQUE_TABLE_NAME_LENGTH = 63
+
```
utkarsharma2, line 39: can we use aenum package(https://pypi.org/project/aenum/). It will simplify the code.

utkarsharma2: TBH I'm not sure about the stability of the project.

kaxil: Naah, don't think there is enough value to add an external dependency

utkarsharma2 (approved):

Issue #281, opened 2022-04-04
Add TypeHints


kaxil:
Make sure all the methods and functions within the following packages have typehints:
```
├── astro
│   ├── __init__.py
│   ├── constants.py
│   ├── dataframe
│   │   └── __init__.py

^^^ part 1

│   ├── settings.py
│   ├── sql
│   │   ├── __init__.py
│   │   ├── operators
│   │   │   ├── agnostic_aggregate_check.py
│   │   │   ├── agnostic_boolean_check.py

^^^ part 2

│   │   │   ├── agnostic_load_file.py
│   │   │   ├── agnostic_save_file.py
│   │   │   ├── agnostic_sql_append.py
│   │   │   ├── agnostic_sql_merge.py

^^^ part 3

│   │   │   ├── agnostic_sql_truncate.py
│   │   │   ├── agnostic_stats_check.py
│   │   │   ├── __init__.py
│   │   │   ├── sql_dataframe.py
│   │   │   └── sql_decorator.py

^^^ part 4

│   │   ├── parsers
│   │   │   ├── __init__.py
│   │   │   └── sql_directory_parser.py
│   │   └── table.py

^^^ part 5

│   └── utils
│       ├── bigquery_merge_func.py
│       ├── cloud_storage_creds.py
│       ├── database.py
│       ├── delete.py
│       ├── dependencies.py
│       ├── file.py

^^^ part 6

│       ├── __init__.py
│       ├── load.py
│       ├── path.py
│       ├── postgres_merge_func.py
│       ├── postgres_transform.py

^^^ part 7

│       ├── schema_util.py
│       ├── snowflake_append.py
│       ├── snowflake_merge_func.py
│       ├── snowflake_transform.py

^^^ part 8

│       ├── sqlite_merge_func.py
│       ├── table_handler.py
│       └── task_id_helper.py
https://github.com/astro-projects/astro/blob/main/tests/operators/utils.py

^^^ part 9
```
We'll be addressing this ticket in multiple PRs, so it is easy to review the work:

- [ ] Part 1 @utkarsharma2 
- [x] Part 2 @tatiana 
- [ ] Part 3 @utkarsharma2 
- [ ] Part 4 @tatiana 
- [ ] Part 5 @utkarsharma2 
- [ ] Part 6 @tatiana 
- [ ] Part 7 @utkarsharma2 
- [ ] Part 8 @tatiana 
- [ ] Part 9 @utkarsharma2

tatiana:
[...]

PR #295, opened 2022-04-06
Enable MyPy on CI

kaxil:
closes https://github.com/astro-projects/astro/issues/280

This PR removes Mypy from pre-commit-hook as it is not reliable to run MyPy on local system as it depends on Python version and installed packages and pre-commit runs it in a virtualenv.

We can instead rely on CI and in future add Make command to run it inside a container same as https://github.com/astronomer/astronomer-providers/blob/main/Makefile#L54-L59

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,31 @@ jobs:
           path: ~/.cache/pip
           key: ${{ hashFiles('pyproject.toml') }}
       - uses: pre-commit/action@v2.0.3
+  Mypy:
+    runs-on: ubuntu-latest
+    name: Mypy
+    env:
+      MYPY_FORCE_COLOR: 1
+      TERM: xterm-color
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pip
+          key: mypy-${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
+      - name: Install Dependencies
+        run: |
+          pip install wheel
+          pip install -U -e .[all,mypy]
+      - name: Run Mypy
+        run: |
+          mypy --version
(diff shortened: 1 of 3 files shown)


kaxil:
This is blocked by https://github.com/astro-projects/astro/issues/281

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,31 @@ jobs:
...
+          python-version: 3.9
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pip
+          key: mypy-${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
+      - name: Install Dependencies
+        run: |
+          pip install wheel
+          pip install -U -e .[all,mypy]
```
tatiana, line 41: @kaxil what is the befit of installing `mypy` like this as opposed to adding to the `tests` section of the `pyproject.yml` file?

tatiana:
@kaxil, what are your thoughts on the following alternative approach toward running MyPy:
https://github.com/astro-projects/astro/pull/296
Which would allow us to enable it through the development of https://github.com/astro-projects/astro/issues/281?

.github/workflows/ci.yaml
kaxil: I did that mainly because once we start adding more modules we will have more "Typing" dependencies, for example:

https://github.com/astronomer/astronomer-providers/blob/main/setup.cfg#L81-L103

```
mypy =
    mypy>=0.800
    types-aiofiles
    types-boto
    types-certifi
    types-croniter
    types-Deprecated
    types-docutils
    types-freezegun
    types-paramiko
    types-protobuf
    types-python-dateutil
    types-python-slugify
    types-pytz
    types-redis
    types-requests
    types-setuptools
    types-termcolor
    types-tabulate
    types-toml
    types-Markdown
    types-PyMySQL
    types-PyYAML
```

kaxil:
> @kaxil, what are your thoughts on the following alternative approach toward running MyPy: #296 Which would allow us to enable it through the development of #281?

Yes I like your approach better, I will close this PR

PR #296, opened 2022-04-07
Enrich modules with TypeHints (part 2/9)

tatiana:
Part of #281

Add TypeHints to the following modules:
```
│   ├── settings.py
│   ├── sql
│   │   ├── __init__.py
│   │   ├── operators
│   │   │   ├── agnostic_aggregate_check.py
│   │   │   ├── agnostic_boolean_check.py
```

During the development of this PR, there were a few complaints from DeepSource, related to pre-conditions of the code.
Those are being addressed in a separate PR: https://github.com/astro-projects/astro/pull/299

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,5 @@
 [flake8]
+min_python_version = 3.7
 exclude = .git,__pycache__,docs/source/conf.py,old,build,dist
 max-complexity = 12
 max-line-length = 120
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,10 +22,10 @@ jobs:
   Type-Check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v3
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
         with:
-          python-version: '3.10'
+          python-version: '3.8'
           architecture: 'x64'
       - run: pip3 install nox
       - run: nox -s type_check
@@ -45,10 +45,10 @@ jobs:
         ports:
           - 5432:5432
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v3
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
         with:
-          python-version: '3.10'
+          python-version: '3.8'
           architecture: 'x64'
       - uses: actions/cache@v2
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,12 +8,12 @@ jobs:
   Pre-Commit:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
         with:
           python-version: '3.9'
           architecture: 'x64'
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         with:
           path: ~/.cache/pip
           key: ${{ hashFiles('pyproject.toml') }}
@@ -21,12 +21,21 @@ jobs:
 
   Type-Check:
     runs-on: ubuntu-latest
+    env:
+      MYPY_FORCE_COLOR: 1
+      TERM: xterm-color
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
         with:
           python-version: '3.8'
           architecture: 'x64'
+      - uses: actions/cache@v3
+        with:
+           path: |
+             ~/.cache/pip
+             .nox
+           key: ${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
       - run: pip3 install nox
       - run: nox -s type_check
 
@@ -45,12 +54,12 @@ jobs:
         ports:
           - 5432:5432
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
         with:
           python-version: '3.8'
           architecture: 'x64'
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         with:
           path: |
             ~/.cache/pip
@@ -108,12 +117,12 @@ jobs:
         ports:
           - 5432:5432
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
         with:
           python-version: '3.8'
           architecture: 'x64'
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         with:
           path: |
             ~/.cache/pip
@@ -158,9 +167,9 @@ jobs:
       - Run-Optional-Packages-tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python 3.8
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.8
       - name: Install coverage
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,18 @@ jobs:
           path: ~/.cache/pip
           key: ${{ hashFiles('pyproject.toml') }}
       - uses: pre-commit/action@v2.0.3
+
+  Type-Check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
+          architecture: 'x64'
+      - run: pip3 install nox
+      - run: nox -s type_check
+
   Run-Optional-Packages-tests:
     runs-on: ubuntu-latest
     services:
@@ -33,10 +45,10 @@ jobs:
         ports:
           - 5432:5432
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
         with:
-          python-version: '3.8'
+          python-version: '3.10'
           architecture: 'x64'
       - uses: actions/cache@v2
         with:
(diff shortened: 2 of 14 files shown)


tatiana:
@kaxil I noticed DeepSource caught an issue that was pre-existing to this PR:
https://deepsource.io/gh/astro-projects/astro/run/99bc601f-513a-4dbc-b771-65cf795aa2aa/python/PYL-W0223

Should I address this issue as part of this PR?

.flake8
```diff
@@ -1,4 +1,5 @@
 [flake8]
+min_python_version = 3.7
```
tatiana, line 2: Since Astro only supports Python 3.7, I've added this to avoid:
```
src/astro/utils/__init__.py:1:1: TYP001 guard import by `if False:  # TYPE_CHECKING`: Type (not in 3.5.0, 3.5.1)
```
More info:
https://github.com/asottile/flake8-typing-imports

pyproject.toml
```diff
@@ -94,3 +96,12 @@ markers = [
 
 [tool.flit.module]
 name = "astro"  # Or "astro.sql" if you just want this directory, not the entire 'astro'.
+
+[tool.mypy]
+files = [
+    "src/astro/settings.py",
+    "src/astro/sql/__init__.py",
+    "src/astro/sql/operators/agnostic_aggregate_check.py",
+    "src/astro/sql/operators/agnostic_boolean_check.py"
```
tatiana, line 105: Once we complete  #281 we can replace the `files` parameter by the `mypy_path` one:
https://mypy.readthedocs.io/en/stable/config_file.html#import-discovery

src/astro/sql/operators/agnostic_boolean_check.py
```diff
@@ -121,15 +121,15 @@ def prep_boolean_checks_query(
 
         temp_table = (
             select(sqla_checks_object)
-            .select_from(text(table.qualified_name()))
+            .select_from(text(table.qualified_name()))  # type: ignore
```
tatiana, line 124: I believe this is an issue with the Dropbox package which defines TypeHints for SQLAlchemy 1.3.24.

I attempted to change `text(table.qualified_name())` by an SQLAlchemy table, and the type checks were happy, but that made the table be represented with double quotes, breaking this feature in the Astro library:
https://github.com/astro-projects/astro/pull/296/commits/f4bd0d1a7d4e0167de025c9cfc9b303bc8baca16#diff-9b06451b9adcb789538c75845a87a78e4cbf63cd055870c659ea8715a18336f4R126

pyproject.toml
kaxil: Should we add more checks similar to the one I added in https://github.com/astro-projects/astro/pull/295/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R104 ?

```
namespace_packages = true
explicit_package_bases = true
pretty = true
show_error_codes = true
ignore_missing_imports = true
no_implicit_optional = true
warn_redundant_casts = true
show_error_context = true
color_output = true
disallow_any_generics = true
disallow_untyped_defs = true
disallow_incomplete_defs = true
warn_unused_configs = true
warn_unused_ignores = true
warn_return_any = true
strict_equality = true
```

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,18 @@ jobs:
...
+  Type-Check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
+          architecture: 'x64'
+      - run: pip3 install nox
+      - run: nox -s type_check
```
kaxil, line 31: Can we adding caching like the one we have for tests here too please, so the CI doesn't need to install everything again.

kaxil: Same as

https://github.com/astro-projects/astro/blob/c57209b8c8b0c28f0206b09d4c01eec6a69103db/.github/workflows/ci.yaml#L116-L121

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,18 @@ jobs:
...
           key: ${{ hashFiles('pyproject.toml') }}
       - uses: pre-commit/action@v2.0.3
+
+  Type-Check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
```
kaxil, line 28: Should we change this to Python 3.9 and consider it as default until Airflow releases a stable version that supports 3.10 officially - this will happen in Airflow 2.3, wdyt?

noxfile.py
```diff
@@ -29,7 +29,15 @@ def test(session: nox.Session) -> None:
...
 
 
-@nox.session()
+@nox.session(python=["3.10"])
+def type_check(session: nox.Session) -> None:
+    """Run MyPy checks."""
+    session.install("-e", ".[all]")
+    session.install("-e", ".[tests]")
+    session.run("mypy")
+
```
kaxil, line 38: Really like this this approach of reusing `nox` 👏

pyproject.toml
```diff
@@ -32,6 +32,7 @@ classifiers = [
     "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
kaxil, line 35: Should we keep using 3.10 separate and for the future? Once Airflow supports 3.10 we can add this, and start running our unit tests for a matrix for 3.7 to 3.10

src/astro/utils/__init__.py
```diff
@@ -1,10 +1,19 @@
...
+    BigQueryHook,
+    MissingPackage,
+    PostgresHook,
+    SnowflakeHook,
+)
 
 
-def get_hook(conn_id, database, role=None, schema=None, warehouse=None):
+def get_hook(
+    conn_id, database, role=None, schema=None, warehouse=None
```
kaxil, line 15: Should we add typehints to params too, for example as follows?

```
    conn_id: str, database:  Database
```

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,18 @@ jobs:
           path: ~/.cache/pip
           key: ${{ hashFiles('pyproject.toml') }}
       - uses: pre-commit/action@v2.0.3
+
+  Type-Check:
+    runs-on: ubuntu-latest
+    steps:
```
kaxil, line 24:
```suggestion
    env:
      MYPY_FORCE_COLOR: 1
      TERM: xterm-color
    steps:
```

This will enable colored output

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,18 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
+          architecture: 'x64'
+      - run: pip3 install nox
```
kaxil, line 30:
```suggestion
      - uses: actions/cache@v3
        with:
          path: |
            ~/.cache/pip
          key: mypy-${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
      - run: pip3 install nox
```

.github/workflows/ci.yaml
kaxil: added suggestion here: https://github.com/astro-projects/astro/pull/296/files#r844991016

kaxil:
> @kaxil I noticed DeepSource caught an issue that was pre-existing to this PR: https://deepsource.io/gh/astro-projects/astro/run/99bc601f-513a-4dbc-b771-65cf795aa2aa/python/PYL-W0223
> 
> Should I address this issue as part of this PR?

Separate please, we can ignore that for now.

src/astro/utils/__init__.py
tatiana: Absolutely, I didn't add further type hints to this module because that's Part 7
(`src/astro/utils/__init__.py` is out of the scope of the current PR)
https://github.com/astro-projects/astro/issues/281

I just added the minimum for the part I was working to work :)

pyproject.toml
tatiana: Oh, I didn't realise Airflow didn't support 3.10 - I'll remove this.

noxfile.py
tatiana: If we can use the same strategy to install (test) dependencies regardless of the type of check we do, it could simplify spotting issues

src/astro/utils/__init__.py
kaxil: Ohh yes :)

noxfile.py
```diff
@@ -29,6 +29,14 @@ def test(session: nox.Session) -> None:
     session.run("pytest", *session.posargs)
 
 
+@nox.session(python=["3.8"])
```
utkarsharma2, line 32: is it assured that the other python version won't have an issue if 3.8 is happy?

pyproject.toml
```diff
@@ -94,3 +96,27 @@ markers = [
...
+
+[tool.mypy]
+color_output = true
+#disallow_any_generics = true
+#disallow_incomplete_defs = true
+#disallow_untyped_defs = true
+#explicit_package_bases = true
+# TODO: Once we complete #281 we can replace the files parameter by the mypy_path
+files = [
+    "src/astro/settings.py",
```
utkarsharma2, line 108: I was worried about this, glad you found a way to restrict mypy to files we are currently working on.

noxfile.py
tatiana: No, that is not assured. But since we're restricting this for only this `mypy` check, and we only use `mypy` for testing, this should not affect end-users.

The Python API for typing has changed significantly recently, and I thought it was worth focusing on type-checking for one of them.

src/astro/utils/dependencies.py
```diff
@@ -1,86 +1,97 @@
...
-try:
-    from google.cloud.storage import Client as GCSClient
-except ModuleNotFoundError:
-    GCSClient = MissingPackage("apache-airflow-providers-google", "google")
 
-try:
-    from google.cloud import bigquery
-except ModuleNotFoundError:
-    bigquery = MissingPackage("apache-airflow-providers-google", "google")
+class SnowflakeMissingPackage(metaclass=MissingPackage):
```
utkarsharma2, line 35: How did creating subclasses help?

tatiana: It allowed the method `get_hook` from `astro.utils` to consistently return a class - as opposed to sometimes a class and sometimes an instance. This made MyPy happy

utkarsharma2 (approved):

PR #300, opened 2022-04-08
[DO-NOT-MERGE] Get basic interface

kaxil:


src/astronew/databases/base.py
```diff
@@ -0,0 +1,139 @@
...
+from astronew.table import Table
+
+
+class BaseDB(metaclass=ABCMeta):
+
+    # Connection types
+    conn_types = []
+    max_table_name_size = 63
+
+    def __init__(self, table: Table):
```
utkarsharma2, line 17: @kaxil  I think Tables are first-class citizens since all the operator/task takes them as inputs and not DB. Due to this, I think we are mixing two concepts here. The BaseDB class should not be aware of Table. It should be the other way around. WDYT?
 
Proposed Changes:
 1. def __init__(self, conn_id: str, database_name: str, schema: str, warehouse: str):
 2. We should move all the table-related operations with the Table class.

tatiana: I agree with @utkarsharma2 , as a first impression, it feels multiple tables should reference databases, but not the other way round. @kaxil do you see any particular benefits in having the database referencing tables?

Regarding the parameters of a DB, perhaps `conn_id`, `name`, and `warehouse` would be enough - different tables can have be associated to schemas/namespaces

dimberman: +1 to what @utkarsharma2 and @tatiana said

PR #305, opened 2022-04-11
Fix example DAG which led to 306GB Snowflake table

tatiana:


kaxil (approved):
😮 -- Thanks for fixing this, good work @tatiana

Issue #308, opened 2022-04-12
Create integration test for run_raw_sql


tatiana:
**Context**
At the moment many users (e.g. Astro Build) are using the feature `run_raw_sql` but, although we use it internally, we don't seem to have integration tests explicitly running it.

**Acceptance criteria**
* Have integration tests for all databases using parameterized pytest fixtures validating this functionally works fine
* Have a follow up with users to understand why they are using this feature, and if we could make it a private method

PR #316, opened 2022-04-12
Added Integration testcase for SqlDecoratedOperator.

utkarsharma2:
Closes: #308

Added a basic test case to check the raw SQL execution of SqlDecoratedOperator.

### DIFF
--- a/tests/operators/test_sql_decorator.py
+++ b/tests/operators/test_sql_decorator.py
@@ -1,0 +1,61 @@
+import pathlib
+
+import pytest
+
+from astro.settings import SCHEMA
+from astro.sql.operators.sql_decorator import SqlDecoratedOperator
+from tests.operators import utils as test_utils
+
+CWD = pathlib.Path(__file__).parent
+
+
+@pytest.mark.parametrize(
+    "sql_server",
+    [
+        "snowflake",
+        "postgres",
+        "bigquery",
+        "sqlite",
+    ],
+    indirect=True,
+)
+@pytest.mark.parametrize(
+    "test_table",
+    [
+        {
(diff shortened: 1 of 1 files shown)


tests/operators/test_sql_decorator.py
```diff
@@ -0,0 +1,60 @@
...
+            parameters={},
+            task_id="SomeTask",
+            op_args=(),
+            handler=handler_func,
+            python_callable=null_function,
+            conn_id=test_table.conn_id,
+            database=test_table.database,
+            sql=f"SELECT list FROM {test_table.qualified_name()} WHERE sell=232",
+        )
+    test_utils.run_dag(sample_dag)
```
tatiana, line 60: Is there any explicit meaningful assert we could have in this test?

tests/operators/test_sql_decorator.py
```diff
@@ -0,0 +1,60 @@
...
+                "table_name": test_utils.get_table_name("test"),
+            },
+        }
+    ],
+    indirect=True,
+)
+def test_sql_decorator_basic_functionality(sample_dag, sql_server, test_table):
+    """Test basic sql execution of SqlDecoratedOperator."""
+
+    def handler_func(result):
```
tatiana, line 40: I have a feeling this method may cease to exist in the refactoring, could we move this check to an assert, by the end of the test?

tests/operators/test_sql_decorator.py
utkarsharma2: I added an assert in the handler function, but if it's going away. I'll add it explicitly.

kaxil (approved):

Issue #279, opened 2022-04-04
Refactor Module Structure (including MyPy and Unit tests)


kaxil:
We should refactor our module structure so we don't dump many things in the `utils` directory.

We should also follow the same directory structure for the `tests` directory so navigating the `tests` directory is predictable.

Proposed Structure (details in [this link](https://docs.google.com/document/d/1coJwrqm3wbCcgop_9fQSiQDdZOMLZ9r7_ENDOyO8JrQ/edit?usp=sharing)):

```
astro/
├── __init__.py
├── constants.py  # mostly internal, but also exposes what we support
├── settings.py  # input from users to define the lib behavior
├── exceptions.py  # this would potentially be exposed to the end-user
├── databases/ # use `importlib.import_module` to avoid MissingPackage
│   ├── __init__.py
│   ├── base.py
│   ├── bigquery.py
│   ├── postgres.py
│   ├── sqlite.py
│   └── snowflake.py
|── files/
    ├── types
    │   ├── __init__.py
    │   ├── base.py
    │   ├── csv.py
    │   ├── json.py
    │   ├── …
    │   └── parquet.py
    ├── locations
    │   ├── __init__.py
    │   ├── base.py
    │   ├── local.py
    │   ├── s3.py  # would include de credentials
    │   ├── http.py
    │   └── gcs.py  # would include de credentials
├── sql/
    ├── __init__.py
    ├── base.py
    ├── tables.py
    ├── load/
    │   ├── __init__.py
    │   ├── load_file.py
    │   └── load_dataframe.py
    ├── export/
    │   ├── __init__.py
    │   ├── export_file.py
    │   └── export_dataframe.py
    ├── transform/
    │   ├── __init__.py
    │   ├── merge.py  # will include both merge and append
    │   ├── render.py
    │   ├── transform.py
    │   └── truncate.py
    └── check/
        ├── __init__.py
        ├── aggregate.py
        ├── boolean.py 
        └── stats.py

```

The following items should also be covered when refactoring (Moved to this ticket from https://github.com/astro-projects/astro/issues/290 because of overlap):

- Replace string constants by grouping them into Enums or Dataclasses
- Single `get_connection` method given a `conn_id` that replaces `BaseHook.get_connection(table.conn_id)`
- Remove shadowing of built-in names
[...]

PR #317, opened 2022-04-13
Locations - Refactor Module Structure (including MyPy and Unit tests)

utkarsharma2:
Closes: #279 
Added a Locations module under the files module, which will contain all data location-based code.

- [x] Restructure/Refactor code (partially done)
- [x] Add TypeHints (partially done)
- [x] Add unit testcases (partially done)
- [x] Replace existing usage (partially done)


src/astro/files/locations/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+        FileLocation.LOCAL: Local,
+        FileLocation.S3: S3,
+        FileLocation.GS: GS,
+        FileLocation.HTTP: Http,
+        FileLocation.HTTPS: Http,
+    }
+    return location_to_object[location_type(path)](path, conn_id)
+
+
+def location_type(path: str) -> FileLocation:
```
tatiana, line 27: since this is a function, what do you think of prefixing it with a verb (e.g. `get` or `extract`)?

tests/files/locations/test_locations.py
```diff
@@ -0,0 +1,163 @@
+import os
```
tatiana, line 1: What do you think about splitting the tests in a similar folder structure to the code?
E.g. to have a `test_s3.py` inside `tests/files/locations/`, etc?

src/astro/files/locations/__init__.py
utkarsharma2: Make sense, I'll add it.

src/astro/files/locations/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+from astro.constants import FileLocation
+from astro.files.locations.base import Location
+from astro.files.locations.gcs import GS
+from astro.files.locations.http import Http
+from astro.files.locations.local import Local
+from astro.files.locations.s3 import S3
+
+
+def location_factory(path: str, conn_id: Optional[str] = None) -> Location:
```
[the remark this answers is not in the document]
utkarsharma2: I'm not so sure, don't you think this takes away the readability/simplicity of the code?

tests/files/locations/test_locations.py
utkarsharma2: Sure

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,52 @@
+import glob
+import os
+from abc import ABC, abstractmethod
+from typing import List, Optional
+from urllib.parse import urlparse
+
+
+class Location(ABC):
```
kaxil, line 8: Should we call this `FileLocation`?

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,52 @@
...
+    """Base Location abstract class"""
+
+    def __init__(self, path: str, conn_id: Optional[str] = None):
+        """
+        Location class constructor
+        :param path: Path to a file in the filesystem/Object stores
```
kaxil, line 14: Sphinx needs a blank-line between class description and param strings. Also let's move docstrings from __init__ to Class method itself. We don't need to but we can standardize it

```suggestion
        Location class constructor

        :param path: Path to a file in the filesystem/Object stores
```

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @staticmethod
+    def is_valid_path(path: str) -> bool:
+        """
+        Check if the give path is either a valid URI or a local file
+        :param path: Either local filesystem path or remote URI
+        """
+        result = urlparse(path)
+        if not (
+            all([result.scheme, result.netloc])
```
kaxil, line 47: nit:Since we just have two items, following might work too:

```suggestion
            (result.scheme and result.netloc)
```

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,52 @@
...
+    @abstractmethod
+    def get_transport_params(self):
+        """Get credentials required by smart open to access files"""
+        raise NotImplementedError
+
+    @staticmethod
+    def is_valid_path(path: str) -> bool:
+        """
+        Check if the give path is either a valid URI or a local file
+        :param path: Either local filesystem path or remote URI
```
kaxil, line 43:
```suggestion
        Check if the give path is either a valid URI or a local file

        :param path: Either local filesystem path or remote URI
```

src/astro/files/locations/gcs.py
```diff
@@ -0,0 +1,34 @@
+from typing import Dict, List
+from urllib.parse import urlparse, urlunparse
+
+from astro.constants import FileLocation
+from astro.files.locations.base import Location
+from astro.utils.dependencies import GCSClient, GCSHook, gcs
+
+
+class GS(Location):
```
kaxil, line 9: `GCS` please as it is called Google Cloud Storage

```suggestion
class GCS(Location):
```

src/astro/files/locations/gcs.py
```diff
@@ -0,0 +1,34 @@
...
+        """get GCS credentials for storage"""
+        if self.conn_id:
+            gcs_hook = GCSHook(self.conn_id)
+            client = gcs_hook.get_conn()  # type: ignore
+        else:
+            client = GCSClient()
+
+        return {"client": client}
+
+    def get_paths(self) -> List[str]:
```
kaxil, line 24: Since we are using "resolve" in the description, do you think `resolve_path` might be better? (I am not sure myself)

src/astro/files/locations/gcs.py
```diff
@@ -0,0 +1,34 @@
...
+        """Resolve GS file paths with prefix"""
+        url = urlparse(self.path)
+        bucket_name = url.netloc
+        prefix = url.path[1:]
+        hook = gcs.GCSHook(gcp_conn_id=self.conn_id) if self.conn_id else gcs.GCSHook()
```
kaxil, line 29: Should we use this logic in L20 too and return a hook

src/astro/files/locations/gcs.py
```diff
@@ -0,0 +1,34 @@
...
+
+    location_type = FileLocation.GS
+
+    def get_transport_params(self) -> Dict:
+        """get GCS credentials for storage"""
+        if self.conn_id:
+            gcs_hook = GCSHook(self.conn_id)
+            client = gcs_hook.get_conn()  # type: ignore
+        else:
+            client = GCSClient()
```
kaxil, line 20: Do we need to build a client? Should we not fallback to default connection id?

```python
    hook = GCSHook(self.conn_id) if self.conn_id else GCSHook()
    client = gcs_hook.get_conn()
```

https://github.com/apache/airflow/blob/27ded333a9cb5452541bd4bd362f18c03b29cfeb/airflow/providers/google/cloud/hooks/gcs.py#L134

src/astro/files/locations/s3.py
```diff
@@ -0,0 +1,49 @@
...
+        """Structure s3fs credentials from Airflow connection.
+        s3fs enables pandas to write to s3
+        """
+        if self.conn_id:
+            # The following line raises a friendly exception
+            BaseHook.get_connection(self.conn_id)
+            aws_hook = AwsBaseHook(self.conn_id, client_type="S3")
+            session = aws_hook.get_session()  # type: ignore
+        else:
+            key, secret = S3._parse_s3_env_var()
```
kaxil, line 32:
```suggestion
            key, secret = cls._parse_s3_env_var()
```

kaxil: That way if we rename the Class name anytime, we don't need to change it everywhere.

You can also use:

```suggestion
            key, secret = self._parse_s3_env_var()
```

src/astro/files/locations/s3.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_transport_params(self) -> Dict:
+        """Structure s3fs credentials from Airflow connection.
+        s3fs enables pandas to write to s3
+        """
+        if self.conn_id:
+            # The following line raises a friendly exception
+            BaseHook.get_connection(self.conn_id)
+            aws_hook = AwsBaseHook(self.conn_id, client_type="S3")
+            session = aws_hook.get_session()  # type: ignore
```
kaxil, line 30: We could also get this from `S3Hook().conn` property

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -15,17 +16,15 @@
...
 from astro.utils.task_id_helper import get_task_id
 
 
 class AgnosticLoadFile(BaseOperator):
-    """Load S3/local table to postgres/snowflake database.
-
-    :param path: File path.
-    :param output_table_name: Name of table to create.
+    """Load S3/local table to postgres/snowflake database
+    :param path: File path
```
kaxil, line 24:
```suggestion
    """Load S3/local table to postgres/snowflake database

    :param path: File path
```

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -135,14 +138,12 @@ def load_file(
...
     **kwargs,
 ) -> XComArg:
-    """Convert AgnosticLoadFile into a function.
-
-    Returns an XComArg object.
-
-    :param path: File path.
+    """Convert AgnosticLoadFile into a function
+    Returns an XComArg object
+    :param path: File path
```
kaxil, line 143:
```suggestion
    """Convert AgnosticLoadFile into a function that Returns an XComArg object

    :param path: File path
```

tests/files/locations/test_gcs.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def describe_is_valid_path():
+    """test is_valid_path() method"""
+
+    def with_supported_filepaths():  # skipcq: PTC-W0065, PYL-W0612
+        """With supported file paths"""
+        filepath = "gs://bucket/some-file"
+        location = location_factory(filepath)
+        assert location.is_valid_path(filepath) is True
```
kaxil, line 13: I think we can use parameterize over here instead of describe to test both Positive and negative cases here

src/astro/files/locations/base.py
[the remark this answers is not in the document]
kaxil: +1 well done 👏

tests/files/locations/test_gcs.py
```diff
@@ -0,0 +1,42 @@
...
+    """test get_transport_params() method"""
+
+    def with_gcs():  # skipcq: PYL-W0612, PTC-W0065
+        """with GCS filepath"""
+        path = "gs://bucket/some-file"
+        location = location_factory(path)
+        credentials = location.get_transport_params()
+        assert "google.cloud.storage.client.Client" in str(
+            credentials["client"].__class__
+        )
```
kaxil, line 26: `isinstance` check would be better and you can mock out the call if it tries to call an API

tests/files/locations/test_gcs.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def with_supported_filepaths():  # skipcq: PTC-W0065, PYL-W0612
+        """With supported file paths"""
+        filepath = "gs://bucket/some-file"
+        location = location_factory(filepath)
+        assert location.is_valid_path(filepath) is True
+
+
+def describe_get_transport_params():
+    """test get_transport_params() method"""
```
kaxil, line 17: Can you add comment about what we are trying to test here for example "Test get_transport_params returns a GCS client"

kaxil: Since `get_transport_params` inside `tests/files/locations/test_gcs.py` will only ever test gcs`, do we need to use describe?

we can just call it `test_get_transport_params` and since this is `test_gcs` file that is fine, wdyt?

tests/files/locations/test_gcs.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "remote_file",
+    [{"name": "google", "count": 2}],
+    ids=["google"],
+    indirect=True,
+)  # skipcq: PY-D0003
+def with_remote_object_store_prefix(remote_file):
```
kaxil, line 36: I think we can mock the repsonse and convert this to a unit test instead of a integration test

tests/files/locations/test_local.py
```diff
@@ -0,0 +1,98 @@
...
+        location = location_factory(LOCAL_FILEPATH)
+        assert location.is_valid_path(LOCAL_FILEPATH) is True
+
+    def with_unsupported_path_raises_exception():  # skipcq: PYL-W0612, PTC-W0065
+        """With all the unsupported locations, should raise a valueError exception"""
+        nonexistent_file = "/tmp/nonexistent-file"
+        with pytest.raises(ValueError) as exc_info:
+            _ = location_factory(nonexistent_file)
+        expected_msg = "Invalid path: '/tmp/nonexistent-file'"
+        assert exc_info.value.args[0] == expected_msg
```
kaxil, line 64: This is testing `BaseLocation.__init__`. We should just `Local.is_valid_path` directly without even using `location_factory`. This can be paramterized under a single `test_isvalid_path` with both positive and negative cases

kaxil: we should have a separate `test_base.py` where we can test `BaseLocation.__init__` for supported an unsupported path

tests/files/locations/test_local.py
```diff
@@ -0,0 +1,98 @@
...
+    def with_local_dir(local_dir):  # skipcq: PYL-W0612, PTC-W0065
+        """with local filepath"""
+        path = LOCAL_DIR
+        location = location_factory(path)
+        assert sorted(location.get_paths()) == [LOCAL_DIR_FILE_1, LOCAL_DIR_FILE_2]
+
+    def with_local_prefix(local_dir):  # skipcq: PYL-W0612, PTC-W0065
+        """with local filepath having glob pattern"""
+        path = LOCAL_DIR + "file_*"
+        location = location_factory(path)
```
kaxil, line 89: I would propose we just directly instantiate `Local(path)` instead of using `location_factory`, it is easier to read and self contained

tests/files/locations/test_local.py
```diff
@@ -0,0 +1,98 @@
...
+    """test get_paths() method"""
+
+    def with_local_dir(local_dir):  # skipcq: PYL-W0612, PTC-W0065
+        """with local filepath"""
+        path = LOCAL_DIR
+        location = location_factory(path)
+        assert sorted(location.get_paths()) == [LOCAL_DIR_FILE_1, LOCAL_DIR_FILE_2]
+
+    def with_local_prefix(local_dir):  # skipcq: PYL-W0612, PTC-W0065
+        """with local filepath having glob pattern"""
```
kaxil, line 87: This two can clubbed into one with parameterized as we instantitate and assert in both

tests/files/locations/test_s3.py
```diff
@@ -0,0 +1,40 @@
...
+
+def describe_get_transport_params():
+    """test get_transport_params() method"""
+
+    def with_s3():  # skipcq: PYL-W0612, PTC-W0065
+        """with S3 filepath"""
+        path = "s3://bucket/some-file"
+        location = location_factory(path)
+        credentials = location.get_transport_params()
+        assert "botocore.client.S3" in str(credentials["client"].__class__)
```
kaxil, line 24: `isinstance` might be bettr

tests/files/locations/test_s3.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@pytest.mark.integration
```
kaxil, line 27: This can be converted to unit test and mock results, we don't need an integration test for it.

src/astro/files/locations/base.py
utkarsharma2: It's within the Files folder, so kinda in the right context. But I have no strong opinion on this one.

src/astro/files/locations/gcs.py
utkarsharma2: Done

src/astro/files/locations/gcs.py
utkarsharma2: I'm not sure if I understand this correctly. 

But I think we need to create a client and give users an option to create it with explicit conn_id if they want to.

utkarsharma2: Please let me know If I'm missing something.

src/astro/files/locations/gcs.py
utkarsharma2: I think Something like Uniform Resource Identifier(URI) makes more sense because we have a URL as well as a local path here. WDYT?

Proposal - resolve_uri()

src/astro/files/locations/gcs.py
utkarsharma2: Done.

src/astro/files/locations/s3.py
utkarsharma2: Done

src/astro/files/locations/s3.py
utkarsharma2: Thanks, nice! didn't know this.

tests/files/locations/test_gcs.py
utkarsharma2: Yes, this is kinda static check on paths and is implemented in base class, moved it back to it's own test file - test_base.py with parameterized cases.

tests/files/locations/test_gcs.py
utkarsharma2: Totally, Done.

tests/files/locations/test_gcs.py
utkarsharma2: Done.

tests/files/locations/test_local.py
utkarsharma2: Done.

tests/files/locations/test_local.py
utkarsharma2: Done

tests/files/locations/test_local.py
utkarsharma2: Done.

tests/files/locations/test_s3.py
utkarsharma2: Done.

tests/files/locations/test_s3.py
utkarsharma2: Done.

tests/files/locations/test_gcs.py
utkarsharma2: Done.

src/astro/files/locations/__init__.py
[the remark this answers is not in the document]
tatiana: I think it is not a matter of readability only, but to have a centralized place where we're importing all supported modules - without the need for that. How will we handle optional dependencies? Using the MissingPackage approach? 

We can still have the other approach having an explicit map to dot paths of modules.

src/astro/files/locations/gcs.py
tatiana: Naming is always so hard :see_no_evil: 
I'd vote for `list_uris` or `list_filepaths`

tests/files/locations/test_http.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def with_api():  # skipcq: PYL-W0612, PTC-W0065
+        """with API endpoint"""
+        path = "http://domain/some-file.txt"
+        location = location_factory(path)
+        assert location.get_paths() == [path]
+
+        path = "https://domain/some-file.txt"
+        location = location_factory(path)
+        assert location.get_paths() == [path]
```
kaxil, line 60: Can you use parameterized to test both the following items instead of duplicating L55-L56:

- http://domain/some-file.txt
- https://domain/some-file.txt

tests/files/locations/test_http.py
```diff
@@ -0,0 +1,60 @@
...
+        """with API endpoint"""
+        path = "http://domain/file"
+        location = location_factory(path)
+        credentials = location.get_transport_params()
+        assert credentials is None
+
+        path = "https://domain/file"
+        location = location_factory(path)
+        credentials = location.get_transport_params()
+        assert credentials is None
```
kaxil, line 46: same as https://github.com/astro-projects/astro/pull/317/files#r856524087

tests/files/locations/test_http.py
```diff
@@ -0,0 +1,60 @@
...
+sample_filepaths_per_location = [
+    (FileLocation.HTTP, "http://domain/some-file"),
+    (FileLocation.HTTPS, "https://domain/some-file"),
+]
+sample_filepaths = [items[1] for items in sample_filepaths_per_location]
+sample_filepaths_ids = [items[0].value for items in sample_filepaths_per_location]
+
+
+def describe_get_location_type():
+    """test get_location_type()"""
```
kaxil, line 20: This docstring does not add any value! Either we need add more details or remove it entirely

src/astro/files/locations/gcs.py
```diff
@@ -0,0 +1,37 @@
...
+    def get_transport_params(self) -> Dict:
+        """get GCS credentials for storage"""
+        if self.conn_id:
+            gcs_hook = gcs.GCSHook(self.conn_id)
+            client = gcs_hook.get_conn()  # type: ignore
+        else:
+            hook = (
+                gcs.GCSHook(gcp_conn_id=self.conn_id) if self.conn_id else gcs.GCSHook()
+            )
+            client = hook.get_conn()
```
kaxil, line 23: This does not make sense.. isn't it just the following:

```suggestion
        hook = gcs.GCSHook(gcp_conn_id=self.conn_id) if self.conn_id else gcs.GCSHook()
        client = hook.get_conn()
```

src/astro/files/locations/__init__.py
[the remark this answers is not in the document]
tatiana: I think there are three approaches:

1. Have a central module that imports all the dependencies and uses the `MissingPackage` approach to handle optional dependencies.

2. Maintain an explicit dictionary mapping `types` to dot string module paths, without actually importing them. The modules would only be loaded if and when the user tried to use them, using `importlib` (https://github.com/astro-projects/astro/pull/320/commits/f8b220740a22d6b1bd117e0ee73c73bae35aa435#diff-a1f816ae9cccff9de86f230ac28a5e34c815fe6f51b99c7c66404e60375a1525R7-R12)
* Pro: simplifies the code, since we no longer need to worry about Missing Packages - while still having an explicit dictionary that is easy to maintain

3. Have an automatically generated mapping of `types` to modules dot string paths, for the default case (e.g. when a type & the implementation module have the same name), only adding explicit values when we want to match non-default types matching import paths.
* Pro: when we add a new database, we could in theory only care about adding a new database module - without having to remember to update any dictionary

tests/files/locations/test_http.py
utkarsharma2: I know, but the deepsource check is failing if we don't add the docsting. I'll try to add more details in docstring which make sense.

src/astro/files/locations/__init__.py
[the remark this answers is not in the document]
utkarsharma2: Had a conversation with @kaxil, will implement this approach.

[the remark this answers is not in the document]
utkarsharma2: Done.

src/astro/files/locations/base.py
kaxil: But these will be used in SQL operators, where we _might_ have an API Location in future and we will run into name collision, so `FileLocation` might be better or even `BaseFileLocation` to be consistent with `BaseDatabase` in https://github.com/astro-projects/astro/pull/320/files#diff-5280fa37ed21b32d1f80cf6e5768fe10eccf88b0ea0a33b16a3601ca41f82de1

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,78 @@
...
+
+from astro.constants import FileLocation
+
+
+class Location(ABC):
+    """Base Location abstract class"""
+
+    def __init__(self, path: str, conn_id: Optional[str] = None):
+        """
+        Location class constructor
```
kaxil, line 15: This needs to be changed to something that explains what this is class is about  - "Location class constructor" doesn't tell much

src/astro/files/locations/http.py
```diff
@@ -0,0 +1,18 @@
...
+
+    location_type = FileLocation.HTTP
+
+    def get_paths(self) -> List[str]:
+        """Resolve patterns in path"""
+        return [self.path]
+
+    def get_transport_params(self) -> Union[Dict, None]:
+        """Dummy method"""
+        return None
```
kaxil, line 18: Why is this a dummy method, is this because transport params are not required for HTTP, in that case can we add a comment about it

kaxil: or `return None` ins BaseLocation -- no strong preference between them

src/astro/files/locations/local.py
```diff
@@ -0,0 +1,27 @@
...
+        path_object = pathlib.Path(url.path)
+        if path_object.is_dir():
+            paths = [str(filepath) for filepath in path_object.rglob("*")]
+        else:
+            paths = glob.glob(url.path)
+        return paths
+
+    def get_transport_params(self) -> Union[Dict, None]:
+        """Dummy method"""
+        return None
```
kaxil, line 27: Looks like we need to return None on more than one occasions, probably better to return a default of Noen on base class and adding docstrings about when to override it

src/astro/files/locations/s3.py
```diff
@@ -0,0 +1,44 @@
+import os
+from typing import Dict, List, Tuple
+from urllib.parse import urlparse, urlunparse
+
+from astro.constants import FileLocation
+from astro.files.locations.base import Location
+from astro.utils.dependencies import BotoSession, s3
+
+
+class S3(Location):
```
kaxil, line 10: We might want to make this consistent across Location, Types and Database.

For example, we call it `SqliteDatabase` in https://github.com/astro-projects/astro/pull/320/files

kaxil: i.e. should we call it `S3Location`, `S3` , `S3FileLoc` 🤷 -- no strong opinion but would like it to be consistent across all the three entities

src/astro/files/locations/s3.py
```diff
@@ -0,0 +1,44 @@
...
+        s3fs enables pandas to write to s3
+        """
+        if self.conn_id:
+            session = s3.S3Hook(aws_conn_id=self.conn_id).get_session()
+        else:
+            key, secret = self._parse_s3_env_var()
+            session = BotoSession(
+                aws_access_key_id=key,
+                aws_secret_access_key=secret,
+            )
```
kaxil, line 31: Should we do same as L39.

```suggestion
        hook = s3.S3Hook(aws_conn_id=self.conn_id) if self.conn_id else s3.S3Hook()
        session = hook.get_session()
```

kaxil: That way we can fallback to default conn_id and remove `_parse_s3_env_var(` method, wdyt?

src/astro/files/locations/s3.py
utkarsharma2: I think we S3Location gives it the right context it needs, otherwise, this needs to be inferred from the import path. Will make this change

src/astro/files/locations/base.py
utkarsharma2: Done.

src/astro/files/locations/http.py
utkarsharma2: Done.

src/astro/files/locations/local.py
utkarsharma2: Done.

src/astro/files/locations/s3.py
utkarsharma2: Done.

src/astro/files/locations/s3.py
utkarsharma2: Done.

src/astro/files/locations/base.py
tatiana: +1 `BaseFileLocation`

tatiana (approved):
@utkarsharma2 Thanks a lot for addressing the import strategy concerns - once you rename the Base class, please, merge this! :rocket:

PR #320, opened 2022-04-14
Refactor: base & SQLite database classes

tatiana:
Relates to: #279

This PR addresses the Base and SQLite database classes.

The only part which was not implemented was the append functionality, which we agreed would combine the existing merge & append.


src/astro/databases/sqlite.py
```diff
@@ -0,0 +1,70 @@
...
+        if "////" not in uri:
+            uri = uri.replace("///", "////")
+        return create_engine(uri)
+
+    def run_sql(
+        self, sql_statement: Union[text, str], parameters: Optional[dict] = None
+    ) -> ResultProxy:
+        """
+        Run given SQL statement in the database using the Sqlalchemy engine.
+        :param sql_statement: SQL statement to be run on the engine
```
kaxil, line 32:
```suggestion
        Run given SQL statement in the database using the Sqlalchemy engine.
        
        :param sql_statement: SQL statement to be run on the engine
```
```suggestion
        Run given SQL statement in the database using the Sqlalchemy engine.
        :param sql_statement: SQL statement to be run on the engine
```

kaxil, line 32:
```suggestion
        Run given SQL statement in the database using the Sqlalchemy engine.
        
        :param sql_statement: SQL statement to be run on the engine
```
```suggestion
        Run given SQL statement in the database using the Sqlalchemy engine.
        :param sql_statement: SQL statement to be run on the engine
```

kaxil: Sphinx needs a blank line for proper rendering

kaxil: Sphinx needs a blank line for proper rendering

kaxil (commented):
This looks really good

src/astro/databases/sqlite.py
```diff
@@ -0,0 +1,101 @@
...
+from sqlalchemy import create_engine, text
+from sqlalchemy.engine.base import Engine
+from sqlalchemy.engine.result import ResultProxy
+
+from astro.constants import LoadExistStrategy, MergeConflictStrategy
+from astro.databases.base import Database as BaseDatabase
+from astro.sql.tables import Table
+
+
+class Database(BaseDatabase):
```
kaxil, line 15: This should be called one of the above so we don't have name Collisions instead please

```suggestion
class Sqlite(BaseDatabase):
```

or 

```suggestion
class SqliteDatabase(BaseDatabase):
```
or

```suggestion
class SqliteDB(BaseDatabase):
```

tatiana: @kaxil I did think about potential name collision issues. However, given the current approach to load the database and the fact we don't aim to expose these classes to the end-users (as opposed to existing Airflow hooks), in which circumstances do you see collision issues?

kaxil: Yeah, I still feel it is confusing to have just `Database` within each module. In my opinion, it should be clear from the class name itself on what it is, which we won't in this case without modules path. That way we won't even need to have `from astro.databases.sqlite import Database as SqliteDatabase` in tests and other places and it is clear

tatiana: @kaxil thanks for your feedback, I addressed it :)

conftest.py
```diff
@@ -237,3 +239,54 @@ def remote_file(request):
 
     session.delete(new_connection)
     session.flush()
+
+
+@pytest.fixture
+def database_table_fixture(request):
```
kaxil, line 245: Would be good to add a. single comment on what this is for and when we should use this fixture

conftest.py
```diff
@@ -237,3 +239,54 @@ def remote_file(request):
...
+    database.drop_table(table)
+    if filepath:
+        database.load_file_to_table(filepath, table)
+    yield database, table
+
+    database.drop_table(table)
+
+
+@pytest.fixture
+def cloud_object_fixture(request):
```
kaxil, line 264: Same here, would be good to have a comment on what this fixture is for and when to use it

src/astro/databases/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    "google_cloud_platform": DEFAULT_CONN_TYPE_TO_MODULE_PATH["bigquery"],
+}
+CONN_TYPE_TO_MODULE_PATH = {
+    **DEFAULT_CONN_TYPE_TO_MODULE_PATH,
+    **CUSTOM_CONN_TYPE_TO_MODULE_PATH,
+}
+SUPPORTED_DATABASES = set(DEFAULT_CONN_TYPE_TO_MODULE_PATH.keys())
+
+
+def get_database_from_conn_id(conn_id: str = ""):
```
kaxil, line 22: Any reason for not making `conn_id` a required parameter?

kaxil: Otherwise it would just fail with the following with the default:

```
In [1]: from airflow.hooks.base import BaseHook

In [2]: BaseHook.get_connection('').conn_type
---------------------------------------------------------------------------
AirflowNotFoundException                  Traceback (most recent call last)
Input In [2], in <module>
----> 1 BaseHook.get_connection('').conn_type

File ~/opt/anaconda3/lib/python3.8/site-packages/airflow/hooks/base.py:68, in BaseHook.get_connection(cls, conn_id)
     60 """
     61 Get connection, given connection id.
     62
     63 :param conn_id: connection id
     64 :return: connection
     65 """
     66 from airflow.models.connection import Connection
---> 68 conn = Connection.get_connection_from_secrets(conn_id)
     69 if conn.host:
     70     log.info(
     71         "Using connection to: id: %s. Host: %s, Port: %s, Schema: %s, Login: %s, Password: %s, "
     72         "extra: %s",
   (...)
     79         redact(conn.extra_dejson),
     80     )

File ~/opt/anaconda3/lib/python3.8/site-packages/airflow/models/connection.py:410, in Connection.get_connection_from_secrets(cls, conn_id)
    403     except Exception:  # pylint: disable=broad-except
    404         log.exception(
    405             'Unable to retrieve connection from secrets backend (%s). '
    406             'Checking subsequent secrets backend.',
    407             type(secrets_backend).__name__,
    408         )
--> 410 raise AirflowNotFoundException(f"The conn_id `{conn_id}` isn't defined")

AirflowNotFoundException: The conn_id `` isn't defined
```

src/astro/databases/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Given a conn_id, return the associated Database class.
+    If null, the Database class defaults to the default Airflow connection, if it exists.
```
kaxil, line 25: I am not sure about this, how will this function know the default conn_type?

conftest.py
dimberman, line 245: Do we need to have the word `fixture` in the name?

conftest.py
```diff
@@ -237,3 +239,54 @@ def remote_file(request):
...
     session.flush()
+
+
+@pytest.fixture
+def database_table_fixture(request):
+    params = request.param
+    table = params.get("table", NewTable())
+    filepath = params.get("filepath", "")
+
+    database_name = params["database"]
```
dimberman, line 250: Should there be a default database?

conftest.py
```diff
@@ -237,3 +239,54 @@ def remote_file(request):
...
+    else:
+        bucket_name = os.getenv("AWS_BUCKET", "tmp9")
+        object_path = f"s3://{bucket_name}/{object_prefix}"
+        hook = s3.S3Hook()
+        hook.delete_objects(bucket_name, object_prefix)
+
+    yield object_path, hook
+
+    if provider == "google":
+        hook.exists(bucket_name, object_prefix) and hook.delete(
```
dimberman, line 288: Does hook.delete fail if the object doesn't exist? If so can you add a one line comment explaining this line?

src/astro/databases/__init__.py
dimberman: Agreed, this feels like something that should be explicit

src/astro/databases/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+    :param conn_id: Database connection ID in Airflow
+    """
+    conn_type = BaseHook.get_connection(conn_id).conn_type
```
dimberman, line 29: This might be a good place for a custom error and a try/catch

src/astro/databases/base.py
```diff
@@ -0,0 +1,259 @@
...
+    Base class to represent all the Database interactions.
+
+    The goal is to be able to support new databases by adding
+    a new module to the `astro/databases` directory, without the need of
+    changing other modules and classes.
+
+    The exception is if the Airflow connection type does not match the new
+    Database module name. In that case, we should update the dictionary
+    `CUSTOM_CONN_TYPE_TO_MODULE_PATH` available at `astro/databases/__init__.py`.
+    """
```
dimberman, line 35: I really like this idea. It's really clean!

src/astro/databases/base.py
```diff
@@ -0,0 +1,259 @@
...
+    # ---------------------------------------------------------
+    def create_table(self, table: Table) -> None:
+        """
+        Create a SQL table. For this method to work, the table instance must contain columns.
+
+        :param table: The table to be created.
+        """
+        metadata = table.sqlalchemy_metadata
+        sqlalchemy_table = sqlalchemy.Table(table.name, metadata, *table.columns)
+        metadata.create_all(self.sqlalchemy_engine, tables=[sqlalchemy_table])
```
dimberman, line 103: Might be worth it for this to return an `astro.sql.table.Table` object

src/astro/databases/base.py
```diff
@@ -0,0 +1,259 @@
...
+    def get_table_qualified_name(self, table: Table) -> str:
+        """
+        Return table qualified name. This is Database-specific.
+        For instance, in Sqlite this is the table name. In Snowflake, however, it is the database, schema and table
+
+        :param table: The table we want to retrieve the qualified name for.
+        """
+        # Initially this method belonged to the Table class.
+        # However, in order to have an agnostic table class implementation,
+        # we are keeping all methods which vary depending on the database within the Database class.
```
dimberman, line 89: This KIND OF makes sense though it feels kind of clunky. Perhaps there's some way where the table can still own the function but takes the database as an argument? I'm not gonna die on this hill but something feels odd about the inheritance here.

src/astro/databases/base.py
```diff
@@ -0,0 +1,259 @@
...
+    ):
+        """
+        Append the source table rows into a destination table.
+        The argument `if_conflicts` allows the user to define how to handle conflicts.
+
+        :param source_table: Contains the rows to be appended to the target_table
+        :param target_table: Contains the destination table in which the rows will be appended
+        :param if_conflicts: The strategy to be applied if there are conflicts.
+        """
+        raise NotImplementedError
```
dimberman, line 194: Why is the append function a part of the database class?

src/astro/databases/sqlite.py
```diff
@@ -0,0 +1,116 @@
...
+        self.load_pandas_dataframe_to_table(
+            pandas_dataframe, target_table, if_exists, chunk_size
+        )
+
+    def append_table(
+        self,
+        source_table: Table,
+        target_table: Table,
+        if_conflicts: AppendConflictStrategy = "exception",
+    ):
```
dimberman, line 97: Is there a reason we have append but not merge?

src/astro/sql/tables.py
```diff
@@ -0,0 +1,55 @@
...
+    database: Union[str, None] = None
+    warehouse: Union[str, None] = None
+    role: Union[str, None] = None
+
+
+@dataclass
+class Table:
+    # TODO: discuss alternative names to this class, since it contains metadata as opposed to be the
+    # SQL table itself
+    # Some ideas: TableRef, TableMetadata, TableData, TableDataset
```
dimberman, line 23: I think the hard part is that this class is front facing, so whatever we name it is gonna be something that users see. This might be something to discuss with Mike

src/astro/utils/path.py
```diff
@@ -111,3 +112,54 @@ def is_local(path):
     """
     file_location = get_location(path)
     return file_location == FileLocation.LOCAL
+
+
+def get_module_dot_notation(module_path: pathlib.Path) -> str:
+    """
```
dimberman, line 118: What is this function for?

dimberman (changes requested):

conftest.py
utkarsharma2: I think there was another fixture dealing with the 'remote_file'. Can we have one fixture to cater to all remote file needs?

src/astro/databases/base.py
```diff
@@ -0,0 +1,259 @@
...
+        :param statement: SQL query statement
+        :param target_table: Destination table where results will be recorded.
+        :param parameters: (Optional) parameters to be used to render the SQL query
+        """
+        statement = self._create_table_statement.format(
+            self.get_table_qualified_name(target_table), statement
+        )
+        self.hook.run(statement, parameters)
+
+    def drop_table(self, table: Table) -> None:
```
utkarsharma2, line 123: Would it be better to move this method to Table Class?

src/astro/databases/__init__.py
tatiana: @kaxil, I believe it would be great if we could default to the default `conn_id` if those exist in Airflow.

As a DAG author, I find it convenient to instantiate a GCSHook, for instance, in both Composer & local env setup with `gcloud,` and do not have to worry about connections and their names. I believe AWS offers the same flexibility to users who use their managed platform.

Let's allow users who want to use `conn_id` to use them, but if there are standard `conn_id`s for amazon, google, and others - why shouldn't we use them? If you think it is a valid use case, we can find alternative ways to implement it!

Regarding your example, `BaseDatabase` is an abstract base class - so I don't think that use case is valid. Subclasses that do not have a default `conn_id` defined can raise an exception when trying to instantiate

src/astro/databases/__init__.py
tatiana: @kaxil at the moment, it is up to the BaseDatabase subclass (e.g. `SqliteDatabase,` `BigqueryDatabase`)  to declare and use the default connection. I could change the interface for the connection parameter to be mandatory in the BaseDatabase instantiation method and have this class retrieve the attribute `DEFAULT_CONNECTION` from the module is defined.

@dimberman, I detailed the motivation in this other comment:
https://github.com/astro-projects/astro/pull/320#discussion_r859571663
This type of implicit behavior is already used by Airflow users when they instantiate hooks, for instance.

src/astro/databases/sqlite.py
```diff
@@ -0,0 +1,116 @@
...
+from astro.constants import (
+    DEFAULT_CHUNK_SIZE,
+    AppendConflictStrategy,
+    LoadExistStrategy,
+)
+from astro.databases.base import BaseDatabase
+from astro.sql.tables import Table
+from astro.utils.load import load_file_into_dataframe
+
+DEFAULT_CONN_ID = "sqlite_default"
```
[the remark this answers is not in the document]
tatiana: I didn't realize that existed! :raised_hands: Thanks, @kaxil !

conftest.py
tatiana: Indeed, @dimberman , we don't need the suffix `fixture`!

So far, we have not been following a convention to name our fixtures. One of them is called `test_table`, which is confusing - since it is prefixed just like the tests. And we have other fixtures which don't have a prefix or suffix (e.g. `output_table` and `sql_server.`)

Having some naming convention would help users unfamiliar with the code base quickly identify if a particular test argument is a fixture or a normal parameter.

That said, I'm fine dropping that if the rest of the team prefers it :)

conftest.py
tatiana: Yep, unfortunately it does - I'll add a comment!

src/astro/databases/base.py
tatiana: @dimberman, I know exactly how you feel about it - I had the same feeling of disgust and pain :joy:  It would be great that table owned this property. In my original implementation, that's where it belonged:
https://github.com/astro-projects/astro/pull/320/commits/b306e8df50b55a49c479095a280f6ccdfcd96b56#diff-33f4b62fcc766d55fbfd03d22eb256a156aae2cdf81977b5ec501880b6a7729eR19-R45

One way we may be able to bring this method back to the `Table` class is if we try to create the qualified name based on the metadata that was set for the `Table` - ignoring database types. I'm not sure if this will cause any issues in the future, but it may work:
a) if there is no metadata, return `name`
b) if there is `metadata.schema`, without `metadata.database`, return `schema.name`
c) if both `metadata.schema` and `metadata.database` are declared, return `database.schema.name`

Would this implicit behavior be good enough for all the currently supported databases and Redshift?

The benefit of having this in the Database class is that we stick to having database-specific customizations explicit and declared in a single class/module.

src/astro/databases/base.py
tatiana: What do you think would be the benefit of this? The parts of the code which will use this method know `table` in advance, and pass it as an argument.

src/astro/databases/base.py
tatiana: This method is currently customizable per database. As an example, BQ does not support `IF EXISTS` in the drop table statement. Since we are aiming to have an agnostic Table class, how do you suggest we move the `drop_table` functionality there?

src/astro/databases/sqlite.py
tatiana: @dimberman in a conversation we had while you were off, @kaxil, @utkarsharma2, and I discussed having a single function that would merge both features, with a different parameter `if_conflicts`. I thought this could be polemic - so I'd love your thoughts on this before further implementing it.

src/astro/utils/path.py
tatiana: It converts `Path("/home/tati/Code/astro/src/astro/databases/sqlite.py")` to `"astro.databases.sqlite"`, and it is used by `get_database_from_conn_id` to automatically map connection types which have the same name as our modules to the respective file.

src/astro/databases/base.py
```diff
@@ -0,0 +1,259 @@
...
+        source_dataframe.to_sql(
+            target_table.name,
+            con=self.sqlalchemy_engine,
+            if_exists=if_exists,
+            chunksize=chunk_size,
+            method="multi",
+            index=False,
+        )
+
+    def append_table(
```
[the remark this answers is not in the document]
tatiana: So far, I feel that any operation with a table/database argument should belong to `Database` - since they can be customized per `Database` if needed.

Most operators will build and offer end-user functionalities by using `Database` and any other abstractions we may need. Operators are still responsible for templating, handling users' arguments, instantiating tables if necessary, and sending info to xcom.

Do you have any concerns with this approach, @utkarsharma2 ?

src/astro/databases/base.py
tatiana: My understanding was that any operation directly affecting databases - which may be customized depending on the database, should be implemented within the Database class. Do you see any benefit in extracting this feature outside of the Database class, @dimberman ?

src/astro/databases/__init__.py
tatiana: @dimberman @kaxil I roll back my previous position, I'll remove the handling of default connection here. In this particular case, we need `conn_id`  to identify `conn_type` - as explained by @kaxil . I'll fix this.

kaxil:
Since https://github.com/astro-projects/astro/pull/317 is merged, there are some merge conflicts

src/astro/databases/base.py
```diff
@@ -0,0 +1,267 @@
...
+        """
+        # Initially this method belonged to the Table class.
+        # However, in order to have an agnostic table class implementation,
+        # we are keeping all methods which vary depending on the database within the Database class.
+        raise NotImplementedError
+
+    # ---------------------------------------------------------
+    # Table creation & deletion methods
+    # ---------------------------------------------------------
+    def create_table(self, table: Table) -> None:
```
utkarsharma2, line 104: Can we rename it to create_db_table() to distinguish it from Table entity we have in our code.

src/astro/databases/base.py
```diff
@@ -0,0 +1,267 @@
...
+    def drop_table(self, table: Table) -> None:
+        """
+        Delete a SQL table, if it exists.
+
+        :param table: The table to be deleted.
+        """
+        statement = self._drop_table_statement.format(
+            self.get_table_qualified_name(table)
+        )
+        self.hook.run(statement)
```
utkarsharma2, line 141: can we use - self.run_sql() instead of self.hook.run()??

src/astro/databases/base.py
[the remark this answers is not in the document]
utkarsharma2: Not now, I think it would be cleaner to handle things per database if need be with this approach, compared to handling it in Operators.

tests/files/locations/test_base.py
```diff
@@ -28,15 +28,6 @@ def local_file():  # skipcq: PY-D0003
 
 @pytest.mark.parametrize("filepath", sample_filepaths, ids=sample_filepaths_ids)
 def test_validate_path_with_supported_filepaths(local_file, filepath):
-    """Test is_valid_path with supported paths"""
-    location = location_factory(filepath)
+    """Test is_valid_patfilesh with supported paths"""
```
utkarsharma2:
```suggestion
    """Test is_valid_path files with supported paths"""
```

tests/databases/test_base_database.py
```diff
@@ -0,0 +1,70 @@
...
+from astro.databases.base import BaseDatabase
+from astro.sql.tables import Table
+
+
+class DatabaseSubclass(BaseDatabase):
+    pass
+
+
+def test_subclass_missing_hook_raises_exception():
+    db = DatabaseSubclass(conn_id="fake_conn_id")
```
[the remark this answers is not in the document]
tatiana: I just renamed the test function to be aligned with what the test does:
`test_subclass_missing_not_implemented_methods_raise_exception`

I'd rather we have meaningful test names than have to add docstrings to all of them, when possible.

src/astro/databases/base.py
```diff
@@ -0,0 +1,267 @@
...
+        Copy the content of a table to a target file of supported type, in a supported location.
+
+        :param source_table: An existing table in the database
+        :param target_file: The path to the file to which we aim to dump the content of the database.
+        """
+        # TODO: we probably want to have a class to abstract File. The File object would contain any normalization
+        # and the filetype, in case the extension isn't representative of the file type.
+
+        # TODO: the check if the file exists should be specific per filesystem, this does not work for remote files
+        if if_exists == "exception" and os.path.isfile(target_file):
```
utkarsharma2, line 243: FYI - we can replace this code once Files is in - 
 

```
file = File(self.output_file_path, self.output_conn_id)
if self.overwrite or not file.exists():
    file.write(df)
else:
    raise FileExistsError(f"{self.output_file_path} file already exists.")

```

kaxil (approved):

src/astro/databases/base.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @property
+    def connection(self) -> sqlalchemy.engine.base.Connection:
+        """
+        Return a Sqlalchemy connection object for the given database.
+        """
+        return self.sqlalchemy_engine.connect()
+
+    @property
+    def sqlalchemy_engine(self) -> sqlalchemy.engine.base.Engine:
```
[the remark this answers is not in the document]
tatiana: That's a good question, this property is being used internally. We can prefix it with `_` as a warning for users to use it from operators.

PR #325, opened 2022-04-19
Handled multiple database tables in dataframe operator. 

utkarsharma2:
Closes: https://github.com/astro-projects/astro/issues/323

Within method - _set_variables_from_first_table
we try to fill in all DB params from the first table passes as a param to the dataframe operator, there was a corner case not handled there.

Check to see if all tables belong to same conn_id. Otherwise, we this can go wrong for cases

    When we have tables from different DBs.
    When we have tables from different conn_id, since they can be configured with different database/schema etc.


src/astro/utils/table_handler.py
```diff
@@ -16,28 +16,50 @@ def _set_variables_from_first_table(self):
...
             if self.parameters:
-                param_tables = [t for t in self.parameters.values() if type(t) == Table]
-                if param_tables:
+                param_tables = [
+                    t for t in self.parameters.values() if isinstance(t, Table)
+                ]
+                conn_id_set = {
+                    t.conn_id for t in self.parameters.values() if isinstance(t, Table)
+                }
+                if param_tables and len(conn_id_set) == 1:
```
tatiana, line 62: @utkarsharma2 as we spoke, since this was a bug found in an Astro release, we need a regression test to make sure the same bug doesn't appear again - particularly given the refactoring we're doing.

tatiana (commented):
@utkarsharma2 it seems there is something strange because there is a commit that is part of this PR that is different from the commit in the main related to the 0.8.1 release(a0d0ddfdc7038470d24657967bfb912a8916b870).
Also, I suggest the commit message have the word `Fix` since it fixes a live bug

src/astro/utils/table_handler.py
tatiana: It should be possible to unit test `_set_variables_from_first_table`

utkarsharma2: I have added a UnitTest case checking all possible conditions.

tests/miscellaneous/test_table_handler.py
```diff
@@ -0,0 +1,216 @@
...
+
+    assert handler.conn_id == "conn_1"
+    assert handler.database == "database_1"
+    assert handler.schema == "scheme_1"
+    assert handler.warehouse == "warehouse_1"
+    assert handler.role == "role_1"
+
+
+def test__set_variables_from_first_table_with_different_db_tables_in_parameters():
+    """Test _set_variables_from_first_table() when the tables passed are with same tables in parameters"""
```
tatiana, line 185: "with same tables" -> "have different databases"

tatiana (approved):
Looks great, @utkarsharma2! Thanks for solving this!

Issue #323, opened 2022-04-19
Astro Build's Integration Test breaking on 0.8.1


utkarsharma2:
broken on - 0.8.1 , but works with 0.7.0.

In this test dag, task_5 joins tables from task_3 (snowflake) and task_4 (postgres). The task’s print statement shows the joined table, suggesting successful ingestion and transformation. However, the error below suggests the the postgres output conn is expected to contain the database element of the snowflake connection.
In other words, the task fails becausepostgres_conn doesn’t have the database attribute associated with snowflake_conn.

```
import time
from datetime import datetime

import pandas as pd
from airflow.decorators import dag, task
from airflow.models import Variable
from airflow.utils import timezone
from airflow.utils.dates import days_ago
from astro import dataframe as df
from astro import sql as aql
from astro.sql.table import Table, TempTable

@df()
def task_1_func():
    return pd.DataFrame({'a':[1,2,3]})

@aql.transform(conn_id='postgres_conn')
def task_2_func(execution_date: Table):
    return """SELECT * FROM actor WHERE startdate < '{{ execution_date }}'"""

@aql.transform(conn_id='snowflake_conn')
def task_3_func():
    return """SELECT * FROM "ASTROBUILD"."BUILDSCHEMA"."MYTABLE" LIMIT 10;"""

@aql.transform(conn_id='postgres_conn')
def task_4_func():
    return """SELECT * FROM actor LIMIT 10;"""

@df(conn_id='postgres_conn')
def task_5_func(task_3: pd.DataFrame, task_4: pd.DataFrame):
    df=task_3.join(task_4)
    print(df)
    return df

@dag(schedule_interval='0 0 * * *', start_date=datetime(2022, 4, 15, 11, 28, 8), catchup=False, tags=['tag_1', 'tag_1'])
def dag_1():
    task_1 = task_1_func()
    task_2 = task_2_func(output_table=Table(conn_id='postgres_conn', schema='tmp_astro', table_name='tmp_astro_dag_1_task_2'), execution_date=Table(conn_id='postgres_conn', table_name='execution_date'))
    task_3 = task_3_func(output_table=Table(conn_id='snowflake_conn', schema='tmp_astro', table_name='tmp_astro_dag_1_task_3'))
    task_4 = task_4_func(output_table=Table(conn_id='postgres_conn', schema='tmp_astro', table_name='tmp_astro_dag_1_task_4'))
[...]

PR #326, opened 2022-04-19
Fix 323

utkarsharma2:
Closes: #323 

Within method - _set_variables_from_first_table
we try to fill in all DB params from the first table passes as a param to the dataframe operator, there was a corner case not handled there. 

Check to see if all tables belong to same conn_id. Otherwise, we this can go wrong for cases
1. When we have tables from different DBs.
2. When we have tables from different conn_id, since they can be configured with different database/schema etc.


src/astro/utils/table_handler.py
```diff
@@ -16,28 +16,46 @@ def _set_variables_from_first_table(self):
...
+            if table_index and len(conn_id_set) == 1:
                 first_table = self.op_args[table_index[0]]
         elif not first_table:
             table_kwargs = [
                 x
                 for x in inspect.signature(self.python_callable).parameters.values()
                 if (
                     x.annotation == Table
-                    and type(self.op_kwargs[x.name]) == Table
+                    and type(self.op_kwargs[x.name]) is Table
```
tatiana, line 33: @utkarsharma2 normally it is better to use `isinstance(self.op_kwargs[x.name], Table)`, because it's faster and also considers inheritance!

tests/operators/test_sql_decorator.py
```diff
@@ -0,0 +1,60 @@
...
+                "table_name": test_utils.get_table_name("test"),
+            },
+        }
+    ],
+    indirect=True,
+)
+def test_sql_decorator_basic_functionality(sample_dag, sql_server, test_table):
+    """Test basic sql execution of SqlDecoratedOperator."""
+
+    def handler_func(result):
```
tatiana, line 40: How could we move this to become an `assert` within the test itself?

src/astro/utils/table_handler.py
```diff
@@ -16,28 +16,46 @@ def _set_variables_from_first_table(self):
...
         if self.op_args:
-            table_index = [x for x, t in enumerate(self.op_args) if type(t) == Table]
-            if table_index:
+            table_index = [x for x, t in enumerate(self.op_args) if type(t) is Table]
+            conn_id_set = {x.conn_id for x in self.op_args if type(x) is Table}
+            # Check to see if all tables belong to same conn_id. Otherwise, we this can go wrong for cases
+            # 1. When we have tables from different DBs.
+            # 2. When we have tables from different conn_id, since they can be configured with different
+            # database/schema etc.
+            if table_index and len(conn_id_set) == 1:
```
tatiana, line 25: If this condition is not met, should we raise an exception? Should we have a test to cover if the right exception is raised?

src/astro/utils/table_handler.py
utkarsharma2: Sure, make sense.

src/astro/utils/table_handler.py
utkarsharma2: I think it's a valid case if you have tables from two different connections. We just can't select DB-related params from the first table in that case.

PR #328, opened 2022-04-26
Refactoring file types

utkarsharma2:
Closes: #336


src/astro/files/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+import pandas as pd
+import smart_open
+
+from astro.constants import LOAD_DATAFRAME_BYTES_LIMIT, FileType
+from astro.files.locations import location_factory
+from astro.files.type import type_factory
+from astro.utils.file import get_size
+
+
+class File:
```
dimberman, line 14: Is there a reason that this class exists in the __init__ file? Are we planning for this to part of the public API?

src/astro/files/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def get_size(self) -> int:
+        """
+        Return the size (bytes) of the given file.
+
+        :return: File size in bytes
+        :rtype: int
+        """
+        path = pathlib.Path(self.path)
+        return os.path.getsize(path)
```
dimberman, line 36: Wouldn't this only work for local files?

dimberman: Or is the idea that this is how we handle the file once it has been loaded?

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,107 @@
...
+        Check if the given path is either a valid URI or a local file
+
+        :param path: Either local filesystem path or remote URI
+        """
+        try:
+            Location.get_location_type(path)
+        except ValueError:
+            return False
+
+        result = urlparse(path)
```
dimberman, line 57: Would we need to wrap this in a try/catch? What happens with an invalid URL?

src/astro/files/locations/base.py
```diff
@@ -0,0 +1,107 @@
...
+                    f"Unsupported scheme '{file_scheme}' from path '{path}'"
+                )
+        return location
+
+    def exists(self) -> bool:
+        """Check if the file exists or not"""
+
+        def null_scheme(_: Any = None) -> dict:
+            """dummy function to get dummy creds"""
+            return {}
```
dimberman, line 99: Can you explain this a bit further?

src/astro/files/locations/gcs.py
```diff
@@ -0,0 +1,31 @@
...
+        """Resolve GS file paths with prefix"""
+        url = urlparse(self.path)
+        bucket_name = url.netloc
+        prefix = url.path[1:]
+        hook = gcs.GCSHook(gcp_conn_id=self.conn_id) if self.conn_id else gcs.GCSHook()
+        prefixes = hook.list(bucket_name=bucket_name, prefix=prefix)
+        paths = [
+            urlunparse((url.scheme, url.netloc, keys, "", "", "")) for keys in prefixes
+        ]
+        return paths
```
dimberman, line 31: Do we need all of these paths? Are we not able to simply use the '*' notation?

src/astro/files/locations/s3.py
```diff
@@ -0,0 +1,44 @@
+import os
+from typing import Dict, List, Tuple
+from urllib.parse import urlparse, urlunparse
+
+from astro.constants import FileLocation
+from astro.files.locations.base import Location
+from astro.utils.dependencies import BotoSession, s3
+
+
+class S3(Location):
```
dimberman, line 10: It's really nice how easy this will make adding future file locations as needed (especially if community members want to add a more esoteric one)

dimberman (changes requested):

src/astro/files/base.py
```diff
@@ -0,0 +1,80 @@
...
+
+import pandas as pd
+import smart_open
+
+from astro.constants import FileType
+from astro.files.locations import location_factory
+from astro.files.type import type_factory
+
+
+class File:
```
kaxil, line 13: Do we really need this class, I thought we just needed FileType and FlieLocation

kaxil:
Since https://github.com/astro-projects/astro/pull/317 is merged, there are some merge conflicts

kaxil:
Sry Utkarsh, conflicts again because we merged the other PR 😬

src/astro/files/type/csv.py
```diff
@@ -0,0 +1,28 @@
+import io
+
+import pandas as pd
+
+from astro.constants import FileType as FileTypeConstants
+from astro.files.type.base import FileType
+
+
+class CSV(FileType):
+    def read(self, stream, **kwargs) -> pd.DataFrame:
```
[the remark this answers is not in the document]
kaxil: +1

src/astro/files/type/csv.py
```diff
@@ -0,0 +1,28 @@
...
+        """read csv file from one of the supported locations and return dataframe
+
+        :param stream: file stream object
+        """
+        kwargs.pop("normalize_config")
+        return pd.read_csv(stream, **kwargs)
+
+    def write(self, df: pd.DataFrame, stream: io.TextIOWrapper) -> None:
```
kaxil, line 18: same as Daniel's comment above

src/astro/files/__init__.py
utkarsharma2: Moved it in base.py file

src/astro/files/__init__.py
utkarsharma2: We need an implementation for all file locations, for now, I'll just move this to Local.py with the base class having the size method. I'll create a separate PR for this.

src/astro/files/locations/base.py
utkarsharma2: @dimberman good catch, I'll add it.

src/astro/files/locations/base.py
utkarsharma2: Moved out of the code.

src/astro/files/locations/gcs.py
utkarsharma2: Can you explain this more? do you mean for packing variables while we pass it to urlunparse() ?

src/astro/files/type/csv.py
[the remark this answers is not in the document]
utkarsharma2: is type definition not enough for this? 

and in the future, we can overload this method to cater to different inputs?

src/astro/files/base.py
utkarsharma2: I think it provides nice abstraction, where the only responsibility of client code is to specify path of the file. and everything else is taken care of. 

Also, is there any case where we'll be using location and not types or types and not locations logic separately?

tatiana: @utkarsharma2 @kaxil I agree it would be great to have the `File` abstraction - or something similar to it, e.g.`Dataset`.
When implementing the `Base,Database` class, in parts like: https://github.com/astro-projects/astro/pull/320/files#diff-5280fa37ed21b32d1f80cf6e5768fe10eccf88b0ea0a33b16a3601ca41f82de1R223-R224, I really felt we could benefit by having a central `File` representation which would be used as a container to al file-related bits (location, connection, type, etc). It was not clear to me - for those methods - we benefited by splitting FileType from FileLocation.

src/astro/files/base.py
```diff
@@ -0,0 +1,80 @@
...
+
+import pandas as pd
+import smart_open
+
+from astro.constants import FileType
+from astro.files.locations import create_file_location
+from astro.files.type import create_type_factory
+
+
+class File:
```
tatiana, line 13: @utkarsharma2 I think it would be great to add a docstring explaining if we expect the `File` to be user-facing.

src/astro/files/base.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_size(self) -> int:
+        """
+        Return the size (bytes) of the given file.
+
+        :return: File size in bytes
+        :rtype: int
+        """
+        path = pathlib.Path(self.path)
+        return os.path.getsize(path)
```
tatiana, line 35: I strongly suggest this method is implemented per FileLocation, since the current implementation only works for local files.

src/astro/files/base.py
```diff
@@ -0,0 +1,80 @@
...
+
+        :return: True or False
+        :rtype: bool
+        """
+        result: bool = self.type.name == FileType.PARQUET
+        return result
+
+    def write(self, df: pd.DataFrame) -> None:
+        """Write dataframe to all supported files formats
+        :param df: pandas dataframe
```
tatiana, line 49: IIR from what @kaxil mentioned, Sphinx expects a blank line between the description and params.

src/astro/files/base.py
```diff
@@ -0,0 +1,80 @@
...
+            return self.type.read(stream, normalize_config=normalize_config, **kwargs)
+
+    def exists(self) -> bool:
+        """Check if the file exists or not"""
+        file_exists: bool = self.location.exists()
+        return file_exists
+
+
+def get_files(path_pattern: str, conn_id: Optional[str] = None) -> List[File]:
+    """get file objects by resolving path_pattern from local/object stores
```
tatiana, line 73: `s/get/Get`

src/astro/files/locations/__init__.py
```diff
@@ -16,14 +16,38 @@
...
-
     :param path: Path to a file in the filesystem/Object stores
     :param conn_id: Airflow connection ID
     """
     filetype: FileLocation = BaseFileLocation.get_location_type(path)
     module_path = DEFAULT_CONN_TYPE_TO_MODULE_PATH[filetype.value]
-    module_name = module_path.split(".")[-1]
-    class_name = module_name.title() + "Location"
     module_ref = importlib.import_module(module_path)
+    class_name = get_class_name(module_ref)
```
tatiana, line 25: It feels the other both `FileLocation` and `Database` could also benefit from this method, could you refactor them so they are consistent?

src/astro/files/locations/base.py
```diff
@@ -50,6 +55,29 @@ def is_valid_path(path: str) -> bool:
         except ValueError:
             return False
 
+        try:
```
tatiana, line 58: @utkarsharma2 There are a few cases in which `is_valid_path` is not being unit tested, could you cover those, please?

src/astro/files/locations/google/gcs.py
```diff
@@ -6,7 +6,7 @@
 from astro.utils.dependencies import gcs
 
 
-class GcsLocation(BaseFileLocation):
+class GCSLocation(BaseFileLocation):
```
tatiana, line 9: In the same way that `Http` is capitalized using Title, I suggest we stick to title capitalization for all classes. Otherwise, we can be consistent and rename all of them, which should be uppercase to uppercase.

src/astro/files/type/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+from astro.constants import FileType as FileTypeConstants
+from astro.files.type.base import FileType
+from astro.files.type.csv import CSV
+from astro.files.type.json import Json
+from astro.files.type.ndjson import NdJson
+from astro.files.type.parquet import Parquet
+
+
+def create_type_factory(path: str):
```
tatiana, line 12: It would be great if we rename this to `create_file_type`, so it is consistent with FileLocation and Database.

src/astro/files/type/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Return a FileType given the filepath. Uses a naive strategy, using the file extension.
+
+    :param filepath: URI or Path to a file
+    :type filepath: str or pathlib.PosixPath
+    :return: The filetype (e.g. csv, ndjson, json, parquet)
+    :rtype: astro.constants.FileType
+    """
+    if isinstance(filepath, pathlib.PosixPath):
+        extension = filepath.suffix[1:]
```
tatiana, line 32: It's worth adding a unittest to cover this!

src/astro/files/type/csv.py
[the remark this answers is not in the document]
tatiana: I also vote for explicit typing here. This way we are consistent with the `Database` implementation `read_from_dataframe` would be very clear and explicit. If in the future we need a generic function, we can add it.

src/astro/files/type/json.py
```diff
@@ -0,0 +1,28 @@
+import io
+
+import pandas as pd
+
+from astro.constants import FileType as FileTypeConstants
+from astro.files.type.base import FileType
+
+
+class Json(FileType):
```
tatiana, line 9: I'm fine with using title capitalization for everything. However, if we want to have GCSLocation, this should be renamed to JSON (JavaScript Object Notation).

My preference is to name everything as title (the first letter is capital, others are lowercase) - just to simplify things - but I'm happy either way, as long as we are consistent. 

Also, in the case of databases and file locations, we had suffixed the classes with what they are. Should we name this JsonFileType?

src/astro/files/type/ndjson.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Union
+
+import pandas as pd
+
+from astro.constants import DEFAULT_CHUNK_SIZE
+from astro.constants import FileType as FileTypeConstants
+from astro.files.type.base import FileType
+
+
+class NdJson(FileType):
```
tatiana, line 12: Either we name it `Ndjson` or `NDJSON` (New Delimited JavaScipt Object Notation),

tests/files/test_file.py
```diff
@@ -0,0 +1,191 @@
...
+@patch("astro.files.base.smart_open.open")
+def test_write(mocked_smart_open, filetype, locations):
+    data = {"id": [1, 2, 3], "name": ["First", "Second", "Third with unicode पांचाल"]}
+    df = pd.DataFrame(data=data)
+    filetype_to_class = {
+        FileType.JSON: "astro.files.type.Json.write",
+        FileType.CSV: "astro.files.type.CSV.write",
+        FileType.NDJSON: "astro.files.type.NdJson.write",
+        FileType.PARQUET: "astro.files.type.Parquet.write",
+    }
```
tatiana, line 108: Do we have an alternative to having to create and maintain this dictionary? I noticed this pattern in a few tests in this PR and it feels we can improve it. Perhaps a fixture..?

tests/files/type/test_type_base.py
```diff
@@ -0,0 +1,34 @@
...
+
+from astro.constants import FileType
+from astro.files.type import get_filetype
+
+sample_file = pathlib.Path(pathlib.Path(__file__).parent.parent, "data/sample.csv")
+sample_filepaths_per_filetype = [
+    (FileType.CSV, "sample.csv"),
+    (FileType.JSON, "sample.json"),
+    (FileType.NDJSON, "sample.ndjson"),
+    (FileType.PARQUET, "sample.parquet"),
```
tatiana, line 13: I'd love to see this type of thing unified in a fixture or using another strategy - so we don't need to redefine it per test/per test file.

src/astro/files/type/csv.py
utkarsharma2: Updated the method names to 
- read_to_dataframe()
- write_from_dataframe()

src/astro/files/base.py
utkarsharma2: @tatiana, 
1. I'll the docstring.
2. @dimberman had a different view on moving this in __init__.py. File was initially in __init__.py then moved it to base.py.

Also, the API I want for this module is to just expose the Class File, which then can be used by client code, and not end user. 

If we move File class to __init__.py, there we need to import FileType, create_file_location and create_type_factory, which will the be exposed in the namespace, which I don't want the client code to use directly as it break the intended API.

src/astro/files/base.py
utkarsharma2: done.

src/astro/files/base.py
utkarsharma2: Good catch! Done.

src/astro/files/locations/google/gcs.py
utkarsharma2: Done.

src/astro/files/type/__init__.py
utkarsharma2: Done.

src/astro/files/type/json.py
utkarsharma2: 1. Added suffix - FileType
2. Changed classes names to -
   CSVFileType, JSONFileType, NDJSONFileType, ParquetFileType

src/astro/files/type/ndjson.py
utkarsharma2: Went with - NDJSON

src/astro/files/base.py
tatiana: @utkarsharma2 by "make it available" I meant to allow users to import it using:
```
from astro.files import File
```

This can be achieved by simply

src/astro/files/base.py
```diff
@@ -0,0 +1,80 @@
...
+import pandas as pd
+import smart_open
+
+from astro.constants import FileType
+from astro.files.locations import create_file_location
+from astro.files.type import create_type_factory
+
+
+class File:
+    def __init__(self, path: str, conn_id: Optional[str] = None):
```
tatiana, line 14: Could we have an optional parameter: `type`, in case the user does not Astro to guess the filetype by the file extension.

src/astro/files/base.py
tatiana: Should we change this method to use the `FileLocation` implementation of `get_size`?

src/astro/files/base.py
tatiana: @utkarsharma2 I noticed you marked some comments as resolved, but they don't seem updated here - is there any chance the changes were not pushed yet?

src/astro/files/locations/__init__.py
```diff
@@ -16,14 +16,38 @@
...
     location: BaseFileLocation = getattr(module_ref, class_name)(path, conn_id)
     return location
+
+
+def get_class_name(module_ref: Any, suffix: str = "Location") -> str:
```
tatiana, line 30: Perhaps this function could be placed in `astro.utils.path` or somewhere else, so we can reuse it in other parts of the code.

tests/files/test_file.py
utkarsharma2: Added a fixture type_method_map_fixture

tests/files/type/test_type_base.py
utkarsharma2: added a fixture - type_method_map_fixture

src/astro/files/base.py
utkarsharma2: I have moved this method to respective location classes, with -1 returning for locations that are nonlocal.

src/astro/files/base.py
utkarsharma2: It's already added.

src/astro/files/locations/__init__.py
utkarsharma2: Moved and updated the database package to use it.

src/astro/files/base.py
utkarsharma2: Done.

src/astro/databases/__init__.py
```diff
@@ -29,7 +29,6 @@ def create_database(conn_id: str) -> BaseDatabase:
     conn_type = BaseHook.get_connection(conn_id).conn_type
     module_path = CONN_TYPE_TO_MODULE_PATH[conn_type]
     module = importlib.import_module(module_path)
-    module_name = module_path.split(".")[-1]
-    class_name = module_name.title() + "Database"
+    class_name = get_class_name(module_ref=module, suffix="Database")
```
tatiana, line 32: That's great, @utkarsharma2 . Thank you very much for making them all consistent! :tada:

src/astro/files/base.py
tatiana: @utkarsharma2 I had a conversation with @kaxil yesterday about this topic, and he agreed with it :raised_hands: 
I'm marking this conversation as resolved.

src/astro/files/base.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        Return a FileType given the filepath. Uses a naive strategy, using the file extension.
+
+        :return: True or False
+        :rtype: bool
+        """
+        result: bool = self.type.name == FileType.PARQUET
+        return result
+
+    def write_from_dataframe(self, df: pd.DataFrame) -> None:
```
tatiana, line 55: I was initially confused with the name. What do you think of `create_from_dataframe`?

I think the confusion is that usually when we have `write,` we'd expect a `to`. In this case, since the class is a file, it would be redundant to say `write_file_from_dataframe`, so I can fully understand why you dropped the "file".

src/astro/files/base.py
```diff
@@ -0,0 +1,97 @@
...
+        """Write dataframe to all supported files formats
```
tatiana, line 56: What do you think of:
```
Create a file in the desired location using the values of a dataframe.
```

src/astro/files/type/ndjson.py
```diff
@@ -0,0 +1,59 @@
...
+        :param df: pandas dataframe
+        :param stream: file stream object
+        """
+        df.to_json(stream, orient="records", lines=True)
+
+    @property
+    def name(self):
+        return FileTypeConstants.NDJSON
+
+    def flatten_ndjson(
```
tatiana, line 33: Since we're within NDJSON, perhaps this method could be called `normalize` or `flatten`

tatiana (approved):
@utkarsharma2 I have some minor questions/requests on naming - please, let me know if you'd like to talk about it!
I'm pre-approving it - so we can move forward.

dimberman (approved):
+1 once Tatiana's comments are addressed

src/astro/files/base.py
utkarsharma2: Done.

src/astro/files/base.py
utkarsharma2: Done.

src/astro/files/base.py
```diff
@@ -0,0 +1,97 @@
...
+
+        :param df: pandas dataframe
+        """
+        with smart_open.open(
+            self.path, mode="wb", transport_params=self.location.transport_params
+        ) as stream:
+            self.type.write_from_dataframe(stream=stream, df=df)
+
+    def read_to_dataframe(
```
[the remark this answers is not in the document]
utkarsharma2: Makes more sense. Adding it.

src/astro/files/type/ndjson.py
utkarsharma2: Done.

PR #349, opened 2022-05-04
[WIP] Create UniversalTransferOperator

sunank200:
- Add interface for `DataProvider` and `DataTransfer`. 
- Create a python module and folder structure.
- Add a class definition for the `Dataset` class.
- Implement `S3DataProvider` and `FTPDataProvider`.
- Add class definitions for `UniversalTransferOperator`.

### DIFF
--- a/src/universal_transfer_operator/data_providers/constants.py
+++ b/src/universal_transfer_operator/data_providers/constants.py
@@ -1,0 +1,71 @@
+from enum import Enum
+
+
+class SupportedDataProvider(Enum):
+    """
+    List of all the supported DataProviders.
+    """
+
+    LOCAL = "local"
+    S3 = "s3"
+    FTP = "ftp"
+    HDFS = "hdfs"
+    SFTP = "sftp"
+    GCS = "gcs"
+    MYSQL = "mysql"
+    POSTGRES = "postgres"
+    BIGQUERY = "bigquery"
+    REDSHIFT = "redshift"
+
+
+class Database(Enum):
+    """
+    List of all the supported Databases DataProviders
+    """
+
+    POSTGRES = "postgres"
+    POSTGRESQL = "postgres"
+    SQLITE = "sqlite"
+    BIGQUERY = "bigquery"
+    SNOWFLAKE = "snowflake"
+
+
+class FileSystem(Enum):
+    """
+    List of all the supported FileSystem DataProviders
+    """
+
+    LOCAL = "local"
+    GCS = "gcs"  # Google Cloud Storage
+    S3 = "s3"  # Amazon S3
+
+
+class API(Enum):
+    """
+    List of all the supported API DataProviders
+    """
+
+    HTTP = "http"
+    HTTPS = "https"
+
+
+class SupportedFileType(Enum):
+    CSV = "csv"
+    JSON = "json"
+    NDJSON = "ndjson"
+    PARQUET = "parquet"
+
+
+SUPPORTED_FILE_LOCATIONS = [const.value for const in FileSystem]
+SUPPORTED_FILE_TYPES = [const.value for const in SupportedFileType]
+SUPPORTED_DATABASES = list({const.value for const in Database})
+
+CONN_TYPE_TO_DATABASE = {
+    "postgres": Database.POSTGRES,
+    "postgresql": Database.POSTGRES,
+    "sqlite": Database.SQLITE,
+    "bigquery": Database.BIGQUERY,
+    "gcpbigquery": Database.BIGQUERY,
+    "google_cloud_platform": Database.BIGQUERY,
+    "snowflake": Database.SNOWFLAKE,
+}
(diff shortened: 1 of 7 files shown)


src/universal_transfer_operator/data_providers/constants.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    POSTGRES = "postgres"
+    POSTGRESQL = "postgres"
+    SQLITE = "sqlite"
+    BIGQUERY = "bigquery"
+    SNOWFLAKE = "snowflake"
+
+
+class FileSystem(Enum):
```
tatiana, line 33: What are your thoughts on naming this `FileLocation`, as we did in Astro?
The reason is that S3 and GCS are not file systems. They are object stores.

src/universal_transfer_operator/data_providers/constants.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    List of all the supported FileSystem DataProviders
+    """
+
+    LOCAL = "local"
+    GCS = "gcs"  # Google Cloud Storage
+    S3 = "s3"  # Amazon S3
```
tatiana, line 40: There are times when a user is simply downloading a file from an HTTP/HTTPSs endpoint - as opposed to accessing an API with credentials and paginating.  I believe we could benefit by having both HTTP & HTTPs options here.

src/universal_transfer_operator/data_providers/constants.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class API(Enum):
+    """
+    List of all the supported API DataProviders
+    """
+
```
tatiana, line 47: What type of API do we want to support? In the past I heard from the Data team they'd be interested in using Salesforce API, Stripe, Prometheus/Thanos, and Splunk, for instance. Would we like to offer more specific API interfaces or generic ones?

src/universal_transfer_operator/data_providers/ftp/filesystem/ftp.py
```diff
@@ -0,0 +1,82 @@
...
+    FTPDataProvider class is responsible to create FTP connection, read dataset and write dataset.
+    It inherits FTPHook and implements DataProviders.
+    """
+
+    def read_dataset(
+        self,
+        remote_full_path: str,
+        local_full_path_or_buffer: str,
+        callback: Any = None,
+    ) -> Any:
```
tatiana, line 19: It would be great if we could be more specific about what this method returns. Would `None` be a good return value?

src/universal_transfer_operator/universal_transfer_operator.py
```diff
@@ -0,0 +1,95 @@
...
+        }
+    """
+
+    def __init__(
+        self,
+        *,
+        source_dataset: Dataset,
+        destination_dataset: Dataset,
+        optimization_params: Optional[Dict],
+        **kwargs
```
tatiana, line 44: I really like how clean is the interface of the UniversalTransferOperator!

src/universal_transfer_operator/dataset.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def get_provider(self) -> DataProviders:
+        """Return DataProvider based on conn_type."""
+        (
+            hook_class_name,
+            conn_id_param,
+            package_name,
+            hook_name,
+            connection_type,
+        ) = ProvidersManager().hooks.get(self.conn_type, (None, None, None, None, None))
```
tatiana, line 292: I didn't know Airflow had this singleton class, that's very cool!

src/universal_transfer_operator/dataset.py
```diff
@@ -0,0 +1,306 @@
...
+
+    #### MySQL:
+    Dataset hierarchy:
+     * Host
+     * Port
+
+    Naming hierarchy:
+     * Database
+     * Table
+
```
tatiana, line 62: Do we want to have this level of detail in the base Dataset class docstring?

src/universal_transfer_operator/data_providers/constants.py
sunank200: But when a user downloads from an HTTP/HTTPS endpoint, isn't it a `GET` REST call? Or am I understanding it wrong?

src/universal_transfer_operator/data_providers/constants.py
sunank200: We can let users define the APIs in connection and we can support `GET`, `POST` and `PATCH` given source dataset and destination. Various authentication mechanisms like basic, OAuth, jwt token etc can be passed as part of the connection as well.

src/universal_transfer_operator/dataset.py
sunank200: I was thinking if we use sphinx, this doc-string would be useful. I could also add these details in a separate doc-string.

From the user's perspective, I felt it's important to know how URI has to be passed for different providers. What do you think?

sunank200:
Closing this for now and will open it in future once DataProviders are ready.

PR #353, opened 2022-05-04
[WIP] Sql decorator rebase

dimberman:
closes https://github.com/astro-projects/astro/issues/344

**Acceptance criteria**
* [ ] Identify current responsibilities of the SQLDecorator (e.g. execute SQL, templating, resolve queries, create tables, etc) and propose a minimum interface for the new version
* [ ] Implement the minimal interface necessary for the Base SQLDecorator class (e.g. templating?)


src/astro/dataframe/__init__.py
```diff
@@ -29,10 +30,18 @@ def dataframe(
...
     if task_id:
         param_map["task_id"] = task_id
+    df_class = SqlDataframeOperator
+    if kwargs.get("_experimental", False):
+        from astro.sql.operators.sql_dataframe_refactor import (
+            SqlDataframeOperator as DFNew,
+        )
+
+        df_class = DFNew
+        kwargs.pop("_experimental")
```
dimberman, line 40: @tatiana @kaxil should we have a "experimental" flag per-task, or would it make more sense to just have an environment variable that automatically turns on experimental across the board? I can see it either way

kaxil: I would say once we have the SqlDecoratedOperator ready, we should just remove and replace the old table with the new one.

kaxil (commented):
We will also need to update it to the refactor in https://docs.google.com/document/d/1coJwrqm3wbCcgop_9fQSiQDdZOMLZ9r7_ENDOyO8JrQ/edit?usp=sharing

src/astro/databases/google/bigquery.py
```diff
@@ -36,6 +36,9 @@ def sqlalchemy_engine(self) -> Engine:
     def default_metadata(self) -> Metadata:
         return Metadata(schema=settings.SCHEMA, database=self.hook.project_id)
 
+    def schema_exists(self, schema):
+        return False
```
tatiana, line 40: Why are we returning `False` here?
Should we check if the BQ namespace exists?

src/astro/databases/snowflake.py
```diff
@@ -75,3 +77,46 @@ def load_pandas_dataframe_to_table(
...
+        :return: Dictionary with values with Table type replaced with the table name
+        """
+        for k, v in parameters.items():
+            if isinstance(v, Table):
+                context[k] = "IDENTIFIER(:" + k + ")"
+            else:
+                context[k] = ":" + k
+        return context
+
+    def process_sql_parameters(self, parameters: Dict) -> Dict:
```
tatiana, line 106: Is there anything specific to the Database here - or could we keep this together with our base operator?

src/astro/databases/google/bigquery.py
dimberman: @tatiana in the old `schema_exists` we return False if the DB is not postgres or snowflake 

```def schema_exists(hook, schema, conn_type):
    if conn_type in ["postgresql", "postgres"]:
        created_schemas = hook.run(
            "SELECT schema_name FROM information_schema.schemata;",
            handler=lambda x: [y[0] for y in x.fetchall()],
        )
        return schema.upper() in [c.upper() for c in created_schemas]
    elif conn_type == "snowflake":
        created_schemas = [
            x["SCHEMA_NAME"]
            for x in hook.run("SELECT SCHEMA_NAME from information_schema.schemata;")
        ]
        return schema.upper() in [c.upper() for c in created_schemas]
    return False
```

I was following existing code for parity.

src/astro/databases/base.py
```diff
@@ -251,3 +251,64 @@ def export_table_to_file(
...
+            else:
+                context[k] = ":" + k
+        return context
+
+    def process_sql_parameters(self, parameters: Dict):
+        """
+        Used for DB-specific processing on parameters. It is used in Snowflake in conjunction with
+        add_templates_to_context to pass the name of the table
+        """
+        return parameters
```
tatiana, line 314: In which circumstanced would we not want to unfold a Table into a qualified name?

src/astro/databases/base.py
```diff
@@ -251,3 +251,64 @@ def export_table_to_file(
 
         df = self.export_table_to_pandas_dataframe(source_table)
         target_file.create_from_dataframe(df)
+
+    # ---------------------------------------------------------
+    # Schema Management
+    # ---------------------------------------------------------
+
+    def create_schema_if_needed(self, schema):
```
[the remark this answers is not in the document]
dimberman: The way this function is implemented it always checks if the schema exists and then does nothing if it does exist, I worry if we just call it "create_schema" someone might forget that fact when subclassing.

src/astro/databases/snowflake.py
```diff
@@ -75,3 +77,46 @@ def load_pandas_dataframe_to_table(
             chunk_size=chunk_size,
             quote_identifiers=False,
         )
+
+    def add_templates_to_context(self, parameters: Dict, context: Dict) -> Dict:
```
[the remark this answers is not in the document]
dimberman: @tatiana the alternative is that the operator now has to carry the complexity of each database's specific templating logic. This keeps to the very clean interface where someone can implement a new database only by creating a new database class (otherwise they would need to go into the operators to add each database and we're back where we started).

FWIW the database class is still agnostic to airflow stuff as in this case the context is just a function parameter, instead of something that the database class has to hold on to.

src/astro/databases/base.py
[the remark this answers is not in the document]
tatiana: In other operations (e.g. `create_table` and `drop_table`, we handle if a table exists or not implicitly). The docstring explains the behavior, so I wouldn't worry about people subclassing it.

src/astro/databases/google/bigquery.py
tatiana: Since the schema creation statement has `not exist`, do we need this function to check if a schema exists?
```
_create_schema_statement: str = "CREATE SCHEMA IF NOT EXISTS {}"
```
If we decide we need it, I think it is worth implementing it in all the databases which support schemas.

In the case of BQ, we could use the hook:
https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_api/airflow/providers/google/cloud/hooks/bigquery/index.html#airflow.providers.google.cloud.hooks.bigquery.BigQueryHook.get_dataset

src/astro/databases/postgres.py
```diff
@@ -24,4 +27,50 @@ def hook(self):
...
         schema = self.hook.schema
-        return Metadata(schema=schema)
+        return Metadata(database=schema)
+
+    def schema_exists(self, schema):
+        created_schemas = self.hook.run(
+            "SELECT schema_name FROM information_schema.schemata;",
+            handler=lambda x: [y[0] for y in x.fetchall()],
+        )
+        return schema.upper() in [c.upper() for c in created_schemas]
```
tatiana, line 37: Would it be worth to use SQL-only to check if this table exists? E.g.:
`SELECT schema_name FROM information_schema.schemata WHERE lower(schema_name) = lower({})`

src/astro/dataframe/__init__.py
tatiana: +1, the `TruncateOperator` was fully replaced

src/astro/sql/operators/sql_dataframe.py
```diff
@@ -61,7 +61,7 @@ def __init__(
             **kwargs,
         )
 
-    def handle_op_args(self):
+    def load_op_arg_table_into_dataframe(self):
```
tatiana, line 64: Why don't we replace `_get_dataframe` by the `database.export_table_to_pandas_dataframe` method?

src/astro/sql/operators/sql_dataframe_refactor.py
```diff
@@ -0,0 +1,113 @@
...
+from astro.sql.tables import Metadata
+from astro.sql.tables import Table as NewTable
+from astro.utils.dataframe_function_handler import DataframeFunctionHandler
+from astro.utils.table_handler_new import TableHandler
+
+
+class SqlDataframeOperator(DataframeFunctionHandler, DecoratedOperator, TableHandler):
+    def __init__(
+        self,
+        conn_id: str = "",
```
tatiana, line 18: I suggest we only support users passing Table objects to this operator without supporting users passing the following arguments: `conn_id, `database`, `schema`, `warehouse` and `role`.

src/astro/sql/operators/sql_dataframe_refactor.py
```diff
@@ -0,0 +1,113 @@
...
+        :param database: Database for input table
+        :param schema:  schema for input table
+        :param warehouse: (Snowflake) Which warehouse to use for the input table
+        :param kwargs:
+        """
+        self.conn_id = conn_id
+        self.database = database
+        self.schema = schema
+        self.warehouse = warehouse
+        self.role = role
```
tatiana, line 41: Why do we need these properties to be associated to the operator?

src/astro/sql/operators/sql_dataframe_refactor.py
```diff
@@ -0,0 +1,113 @@
...
+        self.parameters = {}
+        self.kwargs = kwargs or {}
+        self.write_to_table = False
```
tatiana, line 44: Why do we need this?

src/astro/databases/postgres.py
dimberman: Ooooo that would be great so we're not loading all of this stuff locally!

src/astro/databases/postgres.py
```diff
@@ -24,4 +27,51 @@ def hook(self):
...
+        """
+        Create a table with the dataframe's contents.
+        If the table already exists, append or replace the content, depending on the value of `if_exists`.
+
+        :param source_dataframe: Local or remote filepath
+        :param target_table: Table in which the file will be loaded
+        :param if_exists: Strategy to be used in case the target table already exists.
+        :param chunk_size: Specify the number of rows in each batch to be written at a time.
+        """
+        self.create_schema_if_needed(target_table.metadata.schema)
```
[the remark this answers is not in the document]
tatiana: It would break because we're passing schema to `source_dataframe.to_sql(`, right?

src/astro/sql/operators/sql_dataframe_refactor.py
tatiana: Yes, that's my thought, @dimberman ! IT would allow us to simplify our interfaces:
- `Table`
- `File`
- `pandas.Dataframe`

src/astro/sql/operators/sql_decorator_refactor.py
```diff
@@ -0,0 +1,167 @@
...
+        self.load_op_kwarg_dataframes_into_sql()
+
+        # Get SQL from function and render templates in the SQL String
+        self.read_sql_from_function()
+        self.move_function_params_into_sql_params(context)
+        self.template(context)
+        self.parameters = self.database_impl.process_sql_parameters(self.parameters)  # type: ignore
+        self.handle_schema()
+
+        if self.raw_sql:
```
tatiana, line 107: It feels like we are giving two very different behaviors to a single decorator - would it make sense splitting them?

**Behavior 1**

Input:
- table(s)
- SQL statement

Output:
- table

**Behavior 2**

Input:
- table(s)
- SQL Statement
- function

Output:
- the result of running SQL statement + applying function (this can potentially trash the xcom database depending on the backend used)

src/astro/databases/postgres.py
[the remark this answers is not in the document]
dimberman: @tatiana yes

src/astro/sql/operators/sql_decorator_refactor.py
dimberman: I think we could split in the future, but for now 90% of the functionality is the same (Everything leading up to actually running the query).

src/astro/sql/operators/sql_decorator_refactor.py
```diff
@@ -0,0 +1,167 @@
...
+from astro.databases import create_database
+from astro.sql.table import Table as OldTable
+from astro.sql.table import TempTable
+from astro.sql.tables import Metadata, Table
+from astro.utils.dataframe_function_handler import DataframeFunctionHandler
+from astro.utils.sql_refactor import SQLHandler
+from astro.utils.table_handler_new import TableHandler
+
+
+class SqlDecoratedOperator(
```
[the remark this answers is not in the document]
dimberman: renamed

src/astro/sql/operators/sql_dataframe_refactor.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+    @staticmethod
+    def convert_old_table_to_new(table):
+        """
+        This function is only temporary until other functions use the new table format.
+
+        Converts a TempTable or a Table object into the new Table format.
+        :param table:
+        :return:
```
kaxil, line 55: Can you take care of adding description to all the param and return statements please.

src/astro/sql/operators/sql_dataframe_refactor.py
```diff
@@ -0,0 +1,103 @@
...
+            self.write_to_table = True
+        self.op_args = self.kwargs.get("op_args")  # type: ignore
+        self.identifiers_as_lower = identifiers_as_lower
+
+        super().__init__(
+            **kwargs,
+        )
+
+    @staticmethod
+    def convert_old_table_to_new(table):
```
kaxil, line 49: While we are at it, let's add the return Types.

```suggestion
    def convert_old_table_to_new(table) -> Table:
```

src/astro/sql/__init__.py
```diff
@@ -38,6 +36,7 @@ def transform(
         database=database,
         schema=schema,
         warehouse=warehouse,
```
tatiana, line 38: My vote would be for table-related properties to be given as part of the table objects (conn_id, database, schema, warehouse, etc)

src/astro/sql/__init__.py
```diff
@@ -51,6 +50,7 @@ def run_raw_sql(
     schema: Optional[str] = None,
     warehouse: Optional[str] = None,
     handler: Optional[Callable] = None,
+    **kwargs,
 ):
     return transform_decorator(
```
tatiana, line 55: Why don't we have a dedicated operator to handle running raw SQL?

src/astro/sql/operators/sql_dataframe.py
```diff
@@ -1,35 +1,20 @@
...
-    SnowflakeHook,
-    postgres_sql,
-)
-from astro.utils.load import load_dataframe_into_sql_table
+from astro.utils.dataframe_function_handler import DataframeFunctionHandler
 from astro.utils.table_handler import TableHandler
 
 
-class SqlDataframeOperator(DecoratedOperator, TableHandler):
+class SqlDataframeOperator(DataframeFunctionHandler, DecoratedOperator, TableHandler):
```
tatiana, line 12: Since this already lives within the `sql` folder, perhaps we could call this `DataframeOperator` only, and the module could be named `dataframe.py`

src/astro/sql/operators/transform.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(
+        self,
+        conn_id: Optional[str] = None,
+        autocommit: bool = False,
+        parameters: Optional[dict] = None,
+        handler: Optional[Function] = None,
+        database: Optional[str] = None,
+        schema: Optional[str] = None,
+        warehouse: Optional[str] = None,
+        role: Optional[str] = None,
```
tatiana, line 29: We probably want to keep `conn_id``, and these other properties declared within the `Table` instances

src/astro/sql/operators/transform.py
```diff
@@ -0,0 +1,197 @@
...
+        raw_sql=False,
```
tatiana, line 30: Why does the `transform` operator need to support `raw_sql`?

src/astro/sql/operators/transform.py
```diff
@@ -0,0 +1,197 @@
...
+class TransformOperator(
+    SQLHandler, DataframeFunctionHandler, DecoratedOperator, TableHandler
+):
+    """Handles all decorator classes that can return a SQL function"""
+
+    # todo: Add docstrings
+    def __init__(
+        self,
+        conn_id: Optional[str] = None,
+        autocommit: bool = False,
```
tatiana, line 23: I can't see where this property is being used at the moment, should we remove it from all operators for now?

src/astro/sql/operators/transform.py
```diff
@@ -0,0 +1,197 @@
...
+        parameters: Optional[dict] = None,
```
tatiana, line 24: Would it make sense to rename this to `sql_parameters` or similar, so it is very clear these are parameters to be used in the `sql_statement`?

src/astro/databases/base.py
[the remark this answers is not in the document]
utkarsharma2: Can we have a param to decide if we have to perform a check or not?
Example - 
create_schema(if_exist=True)

That way the parameter justifies the logic of the check? and then we can change the name to - create_schema()

src/astro/databases/base.py
```diff
@@ -250,3 +250,66 @@ def export_table_to_file(
...
+        :param schema: DB Schema - a namespace that contains named objects like (tables, functions, etc)
+        """
+        raise NotImplementedError
+
+    # ---------------------------------------------------------
+    # Context & Template Rendering methods (Transformations)
+    # ---------------------------------------------------------
+
+    def add_templates_to_context(
+        self, parameters: Dict, context: Dict
```
utkarsharma2, line 284: @dimberman I think this function has very little to do with Database, should we keep it in SQLDecorator itself?

src/astro/databases/google/bigquery.py
utkarsharma2: @dimberman - do we ever use 'check schema' check without 'create schema'? If not we should merge it. 

And create_schema() can be a good place to raise a Custom exception if we fail to create the schema in the case where the user lack privilege.

src/astro/databases/postgres.py
```diff
@@ -24,5 +27,58 @@ def hook(self):
 
     @property
     def default_metadata(self) -> Metadata:
-        # TODO: comment why
-        return Metadata(database=self.hook.schema)
+        schema = self.hook.get_connection(self.conn_id).schema
+        return Metadata(database=schema)
```
utkarsharma2, line 31: Is the database getting used anywhere for Postgres? Also, why do we match schema to db here? is it because of the Postgres hook?

src/astro/utils/sql_handler.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def handle_schema(self):
+        """
+        In this function we ensure that both the temporary schema and the explicitly stated schema exist.
+        We also ensure that if there is a table conflict, the conflicting table is dropped so the new data
+        can be added.
+        :return:
+        """
+        # Create DEFAULT SCHEMA if not created.
+        self.database_impl.create_schema_if_needed(SCHEMA)
```
utkarsharma2, line 55: @dimberman - Low priority issue, shouldn't we just care about the schema of the tables involved? assuming default schema is populated to tables when it is missing? 

Use case: when all the operators have tables/output tables that have a named schema. We will be creating the SCHEMA for no reason.

src/astro/utils/sql_handler.py
```diff
@@ -0,0 +1,97 @@
...
+        if not self.raw_sql:
```
utkarsharma2, line 56: Should we be checking for raw_sql here? this function should just be concerned about the schemas, right?

src/astro/utils/sql_handler.py
```diff
@@ -0,0 +1,97 @@
...
+            # Create a table name for the temp table
+            self.database_impl.create_schema_if_needed(self.schema)
+            self.database_impl.drop_table(
+                Table(
```
utkarsharma2, line 60: I think this action is misplaced. should not be part of - handle_schema() function.

src/astro/utils/sql_handler.py
```diff
@@ -0,0 +1,97 @@
...
+from airflow.operators.dummy import BaseOperator
+from sqlalchemy.sql.functions import Function
+
+from astro.databases.base import BaseDatabase
+from astro.settings import SCHEMA
+from astro.sql.table import Metadata, Table
+
+
+class SQLHandler(BaseOperator):
+    """Contains helper functions for creating and running SQL queries"""
```
utkarsharma2, line 13: I think this class is doing many things, do you think it would be better if we have a class that
1. Resolves sql_queries and parameter templates
2. Then takes care of executing them with the Database package. 

and the 1. dropping of table and 2. Schema operation can be taken care of in Operator itself.

src/astro/utils/sql_handler.py
```diff
@@ -0,0 +1,97 @@
...
+                    ),
+                )
+            )
+        else:
+            # If there's no SQL to run we raise an error
+            if self.sql == "" or not self.sql:
+                raise AirflowException("There's no SQL to run")
+
+    @property
+    def output_table_name(self):
```
utkarsharma2, line 75: Why do have this property? we can always access the table object - self.table.name?

src/astro/utils/table_handler.py
```diff
@@ -1,13 +1,25 @@
...
-import pandas
+from astro.databases import create_database
+from astro.sql.table import Metadata, Table
 
-from astro.settings import SCHEMA
-from astro.sql.table import Table
 
+class TableHandler(ABC):
+    """This class contains any functions involving modifying or reading from a Table object"""
+
```
utkarsharma2, line 11: Do we use this class anywhere besides SqlDecorator? If we are not using it anywhere but with SqlDecorator, can we merge it?

src/astro/sql/operators/transform.py
```diff
@@ -0,0 +1,197 @@
...
+        self.load_op_kwarg_dataframes_into_sql()
+
+        # Get SQL from function and render templates in the SQL String
+        self.read_sql_from_function()
+        self.move_function_params_into_sql_params(context)
+        self.template(context)
+        self.parameters = self.database_impl.process_sql_parameters(self.parameters)  # type: ignore
+        self.handle_schema()
+
+        if self.raw_sql:
```
utkarsharma2, line 70: What is/are the case where we will be using raw_sql with transform op?

src/astro/sql/operators/transform.py
```diff
@@ -0,0 +1,197 @@
...
+
+        :param context: Airflow's Context dictionary used for rendering templates
+        """
+        if self.op_kwargs:
+            self.parameters.update(self.op_kwargs)  # type: ignore
+        if self.op_args:
+            params = list(inspect.signature(self.python_callable).parameters.keys())
+            for i, arg in enumerate(self.op_args):
+                self.parameters[params[i]] = arg  # type: ignore
+        if context:
```
utkarsharma2, line 131: can we move this to SQLHandler(assuming that class deals with all templating)?

src/astro/sql/operators/transform.py
utkarsharma2: This feels a little off, decorators shouldn't be returning SQL query results, right?

src/astro/utils/dataframe_function_handler.py
```diff
@@ -0,0 +1,87 @@
...
+    identifiers_as_lower: bool = False
+    conn_id: str = ""
+
+    def load_op_arg_dataframes_into_sql(self):
+        """Identifies dataframes in op_args and loads them to the table"""
+        final_args = []
+        for arg in self.op_args:
+            if isinstance(arg, pd.DataFrame):
+                self.database_impl.load_pandas_dataframe_to_table(
+                    source_dataframe=arg, target_table=self.output_table
```
utkarsharma2, line 29: Shouldn't we create a new table object here? otherwise, this will keep updating the same output table.

tatiana:
@dimberman Is it worth closing this PR since we are having smaller PRs, like https://github.com/astronomer/astro-sdk/pull/373?

Issue #331, opened 2022-05-03
Create BigqueryDatabase without file operations


tatiana:
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

Similar module: https://github.com/astro-projects/astro/blob/main/src/astro/databases/sqlite.py

**Acceptance criteria**
* [ ] Create `BigqueryDatabase`, implementing all the necessary methods except for `load_file_to_table` and `export_table_to_file`.
* [ ] All the quality checks passed
* [ ] All the functions have typehints
* [ ] All the new functions and modules have docstrings
* [ ] The introduced code has 100% of test coverage

kaxil:
The only thing missing is `append/merge`

kaxil:
closing this in favor of https://github.com/astronomer/astro-sdk/issues/343

PR #356, opened 2022-05-05
Added bigquery database support in database package

utkarsharma2:
Closes: #338 and #331

**Acceptance criteria**
* [x] Implement `BigqueryDatabase.load_file_to_table`,and `BigqueryDatabase.export_table_to_file`.
* [x] All the quality checks passed
* [x] All the functions have typehints
* [x] All the new functions and modules have docstrings
* [x] The introduced code has 100% of test coverage

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -14,7 +14,6 @@ from airflow.utils.session import create_session, provide_session
 
 from astro.constants import Database, FileLocation, FileType
 from astro.databases import create_database
-from astro.files import File
 from astro.settings import SCHEMA
 from astro.sql.table import Table, TempTable, create_unique_table_name
 from astro.sql.tables import Table as NewTable
@@ -208,7 +207,7 @@ def database_table_fixture(request):
     """
     params = request.param
     table = params.get("table", NewTable())
-    filepath = params.get("filepath", "")
+    file = params.get("file", None)
 
     database_name = params["database"]
     database_name_to_conn_id = {
@@ -219,8 +218,8 @@ def database_table_fixture(request):
     database = create_database(conn_id)
 
     database.drop_table(table)
-    if filepath:
-        database.load_file_to_table(File(str(filepath)), table)
+    if file:
+        database.load_file_to_table(file, table)
     yield database, table
 
     database.drop_table(table)
--- a/conftest.py
+++ b/conftest.py
@@ -210,7 +210,10 @@ def database_table_fixture(request):
     filepath = params.get("filepath", "")
 
     database_name = params["database"]
-    database_name_to_conn_id = {Database.SQLITE: "sqlite_default"}
+    database_name_to_conn_id = {
+        Database.SQLITE: "sqlite_default",
+        Database.BIGQUERY: "google_cloud_default",
+    }
     conn_id = database_name_to_conn_id[database_name]
     database = create_database(conn_id)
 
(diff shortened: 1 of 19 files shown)


conftest.py
```diff
@@ -210,13 +211,16 @@ def database_table_fixture(request):
...
+        Database.SQLITE: "sqlite_default",
+        Database.BIGQUERY: "google_cloud_default",
+    }
     conn_id = database_name_to_conn_id[database_name]
     database = create_database(conn_id)
 
     database.drop_table(table)
     if filepath:
-        database.load_file_to_table(filepath, table)
+        database.load_file_to_table(File(str(filepath)), table)
```
tatiana, line 223: @utkarsharma2, what if we changed the fixture to receive `File`? This way, the user can customize more parameters if they want to..!
Also, did you find any limitations with having a single fixture for both the database and table? This was something I was unsure when I introduced the fixture - if we have a need, we can always break it into two..!

src/astro/databases/base.py
```diff
@@ -139,10 +135,11 @@ def drop_table(self, table: Table) -> None:
     # ---------------------------------------------------------
     def load_file_to_table(
         self,
-        source_file: str,
+        source_file: File,
```
tatiana, line 138: :see_no_evil: I thought this was part of this ticket:
https://github.com/astro-projects/astro/issues/330

I'm checking if this PR covers all necessary changes - it may well close 330

src/astro/databases/base.py
```diff
@@ -139,10 +135,11 @@ def drop_table(self, table: Table) -> None:
...
         target_table: Table,
         if_exists: LoadExistStrategy = "replace",
         chunk_size: int = DEFAULT_CHUNK_SIZE,
+        normalize_config: Optional[dict] = None,
```
tatiana, line 142: I strongly recommend that `normalize_config` is an argument in `File` as opposed to the `load_file_to_table` operation. The reason is that this parameter is file-specific and doesn't make sense for many of the use-cases!

src/astro/databases/base.py
```diff
@@ -230,32 +232,8 @@ def export_table_to_file(
...
-            FileType.JSON: {"orient": "records"},
-            FileType.NDJSON: {"orient": "records", "lines": True},
-        }
-        # TODO: until here - the lines above should be simplified by the File object
-
-        with smart_open.open(
-            target_file, mode="wb", transport_params=file_location.transport_params
-        ) as stream:
-            serializer[filetype](stream, **serializer_params.get(filetype, {}))
+        target_file.create_from_dataframe(df)
```
tatiana, line 239: I love all this code deleted :heart_eyes: it looks super neat!

tatiana (approved):
@utkarsharma2, this is looking amazing! Great work!

I have one minor change request (on the DB interface for `load_file_to_table` and one question on the fixture side. Just to optimize time-zone differences, I'm already approving :rocket:

conftest.py
utkarsharma2: 1. Sounds good, I'll make the change. 
2. We were actually always using a table in the context of a certain DB and I think that's why I haven't seen any limitations as of now.

src/astro/databases/base.py
utkarsharma2: Good catch, making the change.

Issue #332, opened 2022-05-03
Create Postgres without file operations


tatiana:
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

**Acceptance criteria**
* [ ] Create `PostgresDatabase`, implementing all the necessary methods except for `load_file_to_table` and `export_table_to_file`.
* [ ] All the quality checks passed
* [ ] The introduced code has 100% of test coverage

Issue #337, opened 2022-05-03
 Add file operations to the PostgresDatabase 


tatiana:
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

Similar module: https://github.com/astro-projects/astro/blob/main/src/astro/databases/sqlite.py

**Dependencies**
* Depends on: #336
* Depends on: #332 

**Acceptance criteria**
* [ ] Implement `PostgresDatabase.load_file_to_table`,and `PostgresDatabase.export_table_to_file`.
* [ ] All the quality checks passed
* [ ] All the functions have typehints
* [ ] All the new functions and modules have docstrings
* [ ] The introduced code has 100% of test coverage

PR #358, opened 2022-05-10
Postgres database

utkarsharma2:
Closes: #332  and #337

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -213,6 +213,7 @@ def database_table_fixture(request):
     database_name_to_conn_id = {
         Database.SQLITE: "sqlite_default",
         Database.BIGQUERY: "google_cloud_default",
+        Database.POSTGRES: "postgres_conn",
     }
     conn_id = database_name_to_conn_id[database_name]
     database = create_database(conn_id)
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -62,7 +62,9 @@ class BaseDatabase(ABC):
         :param sql_statement: Contains SQL query to be run against database
         :param parameters: Optional parameters to be used to render the query
         """
-        return self.hook.run(sql_statement, parameters)
+        if parameters is None:
+            parameters = {}
+        return self.connection.execute(sql_statement, parameters)
 
     def table_exists(self, table: Table) -> bool:
         """
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -58,6 +58,7 @@ class BaseDatabase(ABC):
     def run_sql(self, sql_statement: str, parameters: Optional[dict] = None):
         """
         Return the results to running a SQL statement.
+        Return the results to running a SQL statement.
 
         Whenever possible, this method should be implemented using Airflow Hooks,
         since this will simplify the integration with Async operators.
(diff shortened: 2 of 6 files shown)


conftest.py
```diff
@@ -213,6 +213,7 @@ def database_table_fixture(request):
     database_name_to_conn_id = {
         Database.SQLITE: "sqlite_default",
         Database.BIGQUERY: "google_cloud_default",
+        Database.POSTGRES: "postgres_conn",
```
kaxil, line 216: The default is `postgres_default`

https://github.com/apache/airflow/blob/e36868ba1b8a26f838851d1d09cca9db64cbcdc8/airflow/providers/postgres/hooks/postgres.py#L62

```suggestion
        Database.POSTGRES: "postgres_default",
```

src/astro/databases/base.py
```diff
@@ -61,7 +61,9 @@ def run_sql(self, sql_statement: str, parameters: Optional[dict] = None):
         :param sql_statement: Contains SQL query to be run against database
         :param parameters: Optional parameters to be used to render the query
         """
-        return self.hook.run(sql_statement, parameters)
+        if parameters is None:
+            parameters = {}
+        return self.connection.execute(sql_statement, parameters)
```
kaxil, line 66: Why this change?

src/astro/databases/base.py
```diff
@@ -86,7 +88,9 @@ def get_table_qualified_name(self, table: Table) -> str:
         # Initially this method belonged to the Table class.
         # However, in order to have an agnostic table class implementation,
         # we are keeping all methods which vary depending on the database within the Database class.
-        raise NotImplementedError
+        schema = table.metadata.schema if table.metadata else None
+        qualified_name: str = schema + "." + table.name if schema else table.name
```
kaxil, line 92:
```suggestion
        qualified_name: str = f"{schema}.{table.name}" if schema else table.name
```

src/astro/databases/postgres.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def get_table_qualified_name(self, table: Table) -> str:
+        """
+        Return table qualified name. This is Database-specific.
+        For instance, in Sqlite this is the table name. In Snowflake, however, it is the database, schema and table
+
+        :param table: The table we want to retrieve the qualified name for.
+        """
+        name: str = table.name
+        return name
```
kaxil, line 31:
```suggestion
        return table.name
```

kaxil:
Tests are failing too - 

https://github.com/astro-projects/astro/runs/6369868357?check_suite_focus=true#step:10:564

kaxil:
> @kaxil - I think the default connection is not configured correctly. connection string - postgresql://postgres:airflow@postgres/airflow
> 
> 'postgres' is added in place of IP/localhost etc. Do we need to override the connection object with the correcnt location?

"postgres" for hostname will work if their is a postgres container, example: https://github.com/astronomer/astronomer-providers/blob/main/dev/docker-compose.yaml#L15

src/astro/databases/postgres.py
utkarsharma2: I did this because of MyPy, mypy thinks that the function returns the 'Any' if we don't explicitly give type here. Any better way to do this.

kaxil: The following should work too
```suggestion
        return str(table.name)
```

Issue #346, opened 2022-05-03
 Refactor the `save_file` operator


tatiana:
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

**Dependencies**
* Depends on: #336  
* Depends on: #330  
* Depends on: #344

**Acceptance criteria**
* [ ] Implement the `save_file` operator using the new `Database` class and using the `create_database(conn_id)` function. It should also use the `File` class and other abstractions.
* [ ] All the quality checks passed
* [ ] All the functions have type hints
* [ ] All the new functions and modules have docstrings
* [ ] The introduced code has 100% of test coverage

PR #360, opened 2022-05-12
Refactor load file and Save file op to use Files and Database packages

utkarsharma2:
Closes: #345 
Closes: #346


conftest.py
```diff
@@ -136,6 +136,81 @@ def test_table(request, sql_server):  # noqa: C901
             hook.run(f"DROP SCHEMA IF EXISTS {table.metadata.schema} CASCADE;")
 
 
+@pytest.fixture
+def table_fixture(request, sql_server):  # noqa: C901
+    tables = []
```
tatiana, line 141: @utkarsharma2 since we've renamed this, would it be possible for the `table_fixture` to be closer to the `database_table_fixture`, using the `Database` class? I put quite some effort to make sure tear down was working properly for all databases!

src/astro/databases/google/bigquery.py
```diff
@@ -58,3 +58,7 @@ def load_pandas_dataframe_to_table(
             chunksize=chunk_size,
             project_id=self.hook.project_id,
         )
+
+    def replace_illegal_columns_chars(self, char) -> str:
```
tatiana, line 62: What do you think if, within the Database, we only defined:
* `illegal_column_name_chars: List[chr]`
* `illegal_column_name_char_replacement: chr`

And we allowed the Operator to use those and replace them, as necessary? I understand that the replacement happens in the context of loading a JSON which has to be normalized - so perhaps the reserve itself should be kept to the load operation, whereas this bit - which is Database-specific, is held in the DB.

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -34,15 +33,15 @@ def __init__(
         path: str,
         output_table: Table,
         file_conn_id: Optional[str] = "",
```
tatiana, line 35: It would be great if the following arguments could be contained within `File`:
* `path`
* `file_conn_id`
* `ndjson_normalize_sep`

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -56,23 +55,18 @@ def execute(self, context: Any) -> Table:
...
-            database=create_database_from_conn_id(self.output_table.conn_id),
+            database=database,
         )
 
-        hook = get_hook(
-            conn_id=self.output_table.conn_id,
-            database=self.output_table.metadata.database,
-            schema=self.output_table.metadata.schema,
-            warehouse=self.output_table.metadata.warehouse,
+        return self.load_data(
```
tatiana, line 64: Perhaps we could call `database.load_pandas_dataframe_to_table` from here, without the need to have a `load_data` method1

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -82,32 +76,25 @@ def load_data(
         for file in get_files(
```
tatiana, line 76: This can probably be simplified by using the `File` instance

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -34,15 +33,15 @@ def __init__(
         path: str,
         output_table: Table,
         file_conn_id: Optional[str] = "",
-        chunksize: int = DEFAULT_CHUNK_SIZE,
-        if_exists: str = "replace",
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+        if_exists: LoadExistStrategy = "replace",
```
tatiana, line 37: This is great! :heart:

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -55,7 +51,8 @@ def execute(self, context: dict) -> None:
         """
```
tatiana, line 51: Github didn't allow me to comment on lines 33-36 :see_no_evil: So here it goes:

The properties:
*  `output_file_path`
* `output_conn_id`
* `output_file_format`

Can be wrapped into a single argument `output_file: File`

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -70,52 +67,6 @@ def execute(self, context: dict) -> None:
...
-                f"Support types: {list(hook_class.keys())}"
-            )
-
-        db = create_database(input_table.conn_id)
-        table_name = db.get_table_qualified_name(input_table)
-        return pd.read_sql(
-            f"SELECT * FROM {table_name}",
-            con=input_hook.get_sqlalchemy_engine(),
-        )
-
```
tatiana, line 118: :heart: great to see so many lines deleted :raised_hands:

src/astro/utils/postgres_transform.py
```diff
@@ -9,4 +9,5 @@ def add_templates_to_context(parameters, context):
             context[k] = database.get_table_qualified_name(v)
         else:
             context[k] = ":" + k
+            context[k] = ":" + k
```
tatiana, line 12: Do we need this line here?

tests/operators/test_agnostic_load_file.py
```diff
@@ -155,7 +154,9 @@ def test_unique_task_id_for_same_path(sample_dag):
             params = {
                 "path": str(CWD) + "/../data/homes.csv",
                 "file_conn_id": "",
-                "output_table": Table(name=OUTPUT_TABLE_NAME, conn_id="postgres_conn"),
+                "output_table": Table(
+                    conn_id="postgres_conn", metadata=Metadata(database="pagila")
```
tatiana, line 158: Shouldn't we be using `postgres_conn_pagila` here and not pass any metadata?

conftest.py
utkarsharma2: Sure, we can I'll make this change.

utkarsharma2: @tatiana I'll raise a separate PR for this.

src/astro/sql/operators/agnostic_load_file.py
utkarsharma2: We have to handle the patterns in path, that is done in load_data.

src/astro/sql/operators/agnostic_load_file.py
utkarsharma2: This is coming from files, we are using this method to resolve pattern in paths.

src/astro/utils/postgres_transform.py
utkarsharma2: Nope, removed it.

tests/operators/test_agnostic_load_file.py
utkarsharma2: So we only need that for prepopulated tables, here we are populating the table with new data.

src/astro/utils/postgres_transform.py
kaxil: I can still see it :)

tests/operators/test_agnostic_load_file.py
```diff
@@ -128,9 +128,9 @@ def test_aql_replace_existing_table(sample_dag, test_table, sql_server):
 )
 @pytest.mark.parametrize(
     "test_table",
-    [{"is_temp": True}, {"is_temp": False}],
+    [{}],
```
kaxil, line 131: What are we parameterizing over?

src/astro/databases/base.py
```diff
@@ -41,6 +41,10 @@ def hook(self) -> BaseHook:
         """Return an instance of the database-specific Airflow hook."""
         raise NotImplementedError
 
+    def replace_illegal_columns_chars(self, char: str) -> str:
```
tatiana, line 44: Perhaps, from a Database perspective, we could just define the following attributes:
* `illegal_characters: List[chr]`  
* `illegal_characters_replacement: chr: `
And allow the operator to use them as needed.

tests/operators/test_agnostic_load_file.py
utkarsharma2: So the table fixture returns the Table object(s) based on the config passed in the List and the table no. of config passed is the number of tables returned.

Maybe we can have a count instead of this weird-looking list if the tables are the default tables?

src/astro/databases/base.py
utkarsharma2: Done.

src/astro/files/type/__init__.py
```diff
@@ -44,7 +44,16 @@ def get_filetype(filepath: Union[str, pathlib.PosixPath]) -> FileTypeConstants:
     if isinstance(filepath, pathlib.PosixPath):
```
kaxil, line 44: Should `type` be renamed to `types` in module name like `files` and `databases`? That can happen in separate PR too

utkarsharma2: Ya, I was thinking the same. It collides with python's type() BIF.

src/astro/utils/postgres_transform.py
utkarsharma2: Changes were not pushed by that time.

src/astro/sql/operators/agnostic_load_file.py
utkarsharma2: besides with ndjson_normalize_sep, all params are moved to the File instance.

example_dags/example_google_bigquery_gcs_load_and_save.py
```diff
@@ -52,8 +55,9 @@ def extract_top_5_movies(input_df: pd.DataFrame):
     aql.save_file(
         task_id="save_to_gcs",
         input_data=t2,
-        output_file_path=f"{gcs_bucket}/{{ task_instance_key_str }}/top_5_movies.csv",
-        output_file_format="csv",
-        output_conn_id="gcp_conn",
-        overwrite=True,
+        output_file=File(
+            path=f"{gcs_bucket}/{{ task_instance_key_str }}/top_5_movies.csv",
```
kaxil, line 59: Did this work?

example_dags/example_sqlite_load_transform.py
```diff
@@ -27,7 +28,9 @@ def top_five_animations(input_table: Table):
 ) as dag:
 
     imdb_movies = aql.load_file(
-        path="https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb.csv",
+        input_file=File(
+            path="https://raw.githubusercontent.com/astro-projects/astro/main/tests/data/imdb.csv"
```
kaxil, line 32:
```suggestion
            path="https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb.csv"
```

src/astro/databases/base.py
```diff
@@ -33,6 +33,9 @@ class BaseDatabase(ABC):
     _drop_table_statement: str = "DROP TABLE IF EXISTS {}"
     _create_table_statement: str = "CREATE TABLE IF NOT EXISTS {} AS {}"
 
+    illegal_column_name_chars: List[str] = []
+    illegal_column_name_chars_replacement: List[str] = []
```
kaxil, line 37: Would be good to add a comment above these two on what are they used for and when we should override it

src/astro/databases/postgres.py
```diff
@@ -26,3 +33,66 @@ def hook(self):
...
+        return qualified_name
+
+    def table_exists(self, table: Table) -> bool:
+        """
+        Check if a table exists in the database.
+
+        :param table: Details of the table we want to check that exists
+        """
+        schema = getattr(table.metadata, "schema", None)
+        schema = schema.lower() if schema else None
```
kaxil, line 96: A comment might help on why we lowercase it here

src/astro/databases/snowflake.py
```diff
@@ -63,7 +63,7 @@ def load_pandas_dataframe_to_table(
 
         db.prep_table(
             source_dataframe,
-            target_table.name,
+            target_table.name.lower(),
```
kaxil, line 66: A comment might help on why we lowercase it here

src/astro/files/base.py
```diff
@@ -36,6 +38,10 @@ def __init__(
     def path(self):
         return self.location.path
 
+    @property
+    def conn_id(self):
```
kaxil, line 42: Needs return type - typehints

src/astro/files/base.py
```diff
@@ -36,6 +38,10 @@ def __init__(
     def path(self):
         return self.location.path
```
kaxil, line 39: Needs return type - typehints

src/astro/sql/operators/agnostic_load_file.py
```diff
@@ -53,68 +46,50 @@ def execute(self, context: Any) -> Table:
...
 
-    def load_data(
-        self, path: str, hook: BaseHook, file_conn_id: Optional[str] = None
-    ) -> Table:
+    def load_data(self, input_file: File, database: BaseDatabase) -> Table:
         """Loads csv/parquet table from local/S3/GCS with Pandas.
         Infers SQL database type based on connection then loads table to db.
         """
-        self.log.info(f"Loading {self.path} into {self.output_table}...")
+        self.log.info(f"Loading {self.input_file.path} into {self.output_table}...")
```
kaxil, line 64:
```suggestion
        self.log.info("Loading %s into %s ...", self.input_file.path, self.output_table)
```

kaxil: We should probably use https://github.com/globality-corp/flake8-logging-format in pre-commit Hooks so that we always use Lazy loading of args in logs.

I did it for Providers repo a while back: https://github.com/astronomer/astronomer-providers/pull/212

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -25,28 +21,21 @@ class SaveFile(BaseOperator):
...
-        "output_conn_id",
-        "output_file_format",
-    )
+    template_fields = ("input_data", "output_file")
 
     def __init__(
         self,
         input_data: Union[Table, pd.DataFrame],
-        output_file_path: str = "",
+        output_file: File,
```
kaxil, line 29: We should update docstring on line 17-21

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -56,7 +45,8 @@ def execute(self, context: dict) -> None:
         """
         # Infer db type from `input_conn_id`.
         if type(self.input_data) is Table:
```
kaxil, line 47:
```suggestion
        if isinstance(self.input_data, Table):
```

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -56,7 +45,8 @@ def execute(self, context: dict) -> None:
...
-            df = self.convert_sql_table_to_dataframe()
+            database = create_database(self.input_data.conn_id)
+            df = database.export_table_to_pandas_dataframe(self.input_data)
         elif type(self.input_data) is pd.DataFrame:
```
kaxil, line 50:
```suggestion
        elif isinstance(self.input_data, pd.DataFrame):
```

tests/databases/test_bigquery.py
```diff
@@ -222,7 +222,7 @@ def test_export_table_to_file_file_already_exists_raises_exception(
         database.export_table_to_file(source_table, File(str(filepath)))
     err_msg = exception_info.value.args[0]
     assert err_msg.startswith("The file")
-    assert err_msg.endswith("tests/data/sample.csv already exists.")
+    assert err_msg.endswith("already exists.")
```
kaxil, line 225: What does the full message show, it should still show the path name no?

tests/databases/test_postgres.py
```diff
@@ -189,7 +189,7 @@ def test_export_table_to_file_file_already_exists_raises_exception(
         database.export_table_to_file(source_table, File(str(filepath)))
     err_msg = exception_info.value.args[0]
     assert err_msg.startswith("The file")
-    assert err_msg.endswith("tests/data/sample.csv already exists.")
+    assert err_msg.endswith("already exists.")
```
kaxil, line 192: Same as above

tests/databases/test_snowflake.py
```diff
@@ -195,7 +195,7 @@ def test_export_table_to_file_file_already_exists_raises_exception(
         database.export_table_to_file(source_table, File(str(filepath)))
     err_msg = exception_info.value.args[0]
     assert err_msg.startswith("The file")
-    assert err_msg.endswith("tests/data/sample.csv already exists.")
+    assert err_msg.endswith("already exists.")
```
kaxil, line 198: Same as above

tests/operators/test_agnostic_load_file.py
kaxil: empty list is weird, anything more explicit than this will be better

src/astro/sql/operators/agnostic_save_file.py
```diff
@@ -25,28 +21,21 @@ class SaveFile(BaseOperator):
...
-        "output_file_format",
-    )
+    template_fields = ("input_data", "output_file")
 
     def __init__(
         self,
         input_data: Union[Table, pd.DataFrame],
-        output_file_path: str = "",
+        output_file: File,
         output_conn_id: Optional[str] = None,
```
kaxil, line 30: `self.output_conn_id` is not used anywhere, can be removed now that we have `output_file`?

src/astro/sql/operators/save_file.py
```diff
@@ -0,0 +1,86 @@
+from typing import Optional, Union
```
kaxil, line 1: Do we plan to rename `save_file` to  `export_file` (https://docs.google.com/document/d/1coJwrqm3wbCcgop_9fQSiQDdZOMLZ9r7_ENDOyO8JrQ/edit?usp=sharing) in a separate PR or should we keep this as `save_file`?

src/astro/utils/load.py
```diff
@@ -98,38 +99,43 @@ def flatten_ndjson(
 
 
 def populate_normalize_config(
```
kaxil, line 101: Should we move this to `sql/operators/load_file` ? And what about other functions and methods in this file, are they needed? Or can they be deleted

src/astro/sql/operators/agnostic_load_file.py
kaxil: Why we didn't move ndjson?

src/astro/databases/postgres.py
```diff
@@ -26,3 +33,66 @@ def hook(self):
...
+        """
+
+        # To do - replace with create_schema() method.
+        conn = self.hook.get_connection(target_table.conn_id)
+        user = conn.login
+        schema = getattr(target_table.metadata, "schema", None)
+        create_schema_query = (
+            f"CREATE SCHEMA IF NOT EXISTS {schema} AUTHORIZATION {user}"
+        )
+        self.run_sql(sql_statement=create_schema_query, parameters={})
```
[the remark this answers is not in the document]
utkarsharma2: I was waiting for Daniel PR to get merged, that PR has this method.

src/astro/sql/operators/save_file.py
utkarsharma2: Let's do it in a separate PR.

src/astro/utils/load.py
utkarsharma2: We can certainly move this to load_file. I'll check for other functions as I'm not sure about their usage.

src/astro/databases/postgres.py
utkarsharma2: I'll add it

src/astro/databases/snowflake.py
utkarsharma2: I'll add it

conftest.py
```diff
@@ -136,6 +136,81 @@ def test_table(request, sql_server):  # noqa: C901
...
+                    database="sqlite",
+                ),
+            }
+        elif database == Database.BIGQUERY:
+            default_table_options = {
+                "conn_id": hook.gcp_conn_id,
+                "metadata": Metadata(
+                    schema=SCHEMA,
+                ),
+            }
```
tatiana, line 184: I strongly believe we should be moving out of this old pattern and using the `Database` class, similar to the `database_table_fixture`

src/astro/sql/operators/load_file.py
```diff
@@ -0,0 +1,112 @@
...
+        self.if_exists = if_exists
+        self.ndjson_normalize_sep = ndjson_normalize_sep
+        self.normalize_config: Dict[str, str] = {}
+
+    def execute(self, context: Any) -> Table:
+        """
+        Load an existing dataset from a supported file into a SQL table.
+        """
+        if self.input_file.conn_id:
+            BaseHook.get_connection(self.input_file.conn_id)
```
tatiana, line 50: Why do we have these lines?

src/astro/sql/operators/load_file.py
```diff
@@ -0,0 +1,112 @@
...
+
+        database = create_database(self.output_table.conn_id)
+        self.normalize_config = populate_normalize_config(
+            ndjson_normalize_sep=self.ndjson_normalize_sep,
+            database=database,
+        )
+
```
tatiana, line 57: Before loading the data into the database, for consistency, we should use the `Database.populate_table_metadata`:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/base.py#L114
Similar to:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/sql/operators/transform.py#L267
So we retrieve metadata from the connection or environment variable, if needed.

src/astro/sql/operators/save_file.py
```diff
@@ -0,0 +1,86 @@
...
+        self.kwargs = kwargs
+
+    def execute(self, context: dict) -> None:
+        """Write SQL table to csv/parquet on local/S3/GCS.
+
+        Infers SQL database type based on connection.
+        """
+        # Infer db type from `input_conn_id`.
+        if isinstance(self.input_data, Table):
+            database = create_database(self.input_data.conn_id)
```
tatiana, line 44: Before fetching the data from the source table, for consistency, we should use the Database.populate_table_metadata:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/base.py#L114
Similar to:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/sql/operators/transform.py#L267

So we retrieve metadata from the connection or environment variable if needed.

src/astro/sql/operators/save_file.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(
+        self,
+        input_data: Union[Table, pd.DataFrame],
+        output_file: File,
+        if_exists: ExportExistsStrategy = "exception",
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+        self.output_file = output_file
+        self.input_data = input_data
```
tatiana, line 33: I am worried that our interfaces for `load` and `save` are not symmetric ATM:
* `load`: from file to table-only
* `save`: from table to either file or database

Normalizing this would be a breaking change to `load.`

@utkarsharma2 @kaxil @dimberman I'd love your thoughts on this:
1. Do we want to normalize?
2. If yes to (1), as part of this or in a separate PR?
3. If separate PR (2), when?

tests/operators/test_load_file.py
```diff
@@ -1,12 +1,12 @@
 """
-Unittest module to test Agnostic Load File function.
```
tatiana, line 2: Please, let's keep this test inside the same path as the operator:
`tests/sql/operators`

I just realized there is a missing s to the directory I created, it would be great if the refactored operators could follow this pattern:
https://github.com/astronomer/astro-sdk/tree/main/tests/sql/operator

tests/operators/test_load_file.py
```diff
@@ -40,16 +41,17 @@
 )
```
tatiana, line 41: Do we have a test case covering the scenario: https://github.com/astronomer/astro-sdk/issues/366?

tests/operators/test_save_file.py
```diff
@@ -1,5 +1,5 @@
 """
-Unittest module to test Agnostic Load File function.
+Unittest module to test Load File function.
```
tatiana, line 2: Same comment as for the load test file location

tests/operators/test_save_file.py
```diff
@@ -24,10 +24,11 @@
 import astro.dataframe as adf
 import astro.sql as aql
 from astro.constants import SUPPORTED_DATABASES, SUPPORTED_FILE_TYPES, Database
+from astro.files import File
 from astro.settings import SCHEMA
 
 # Import Operator
-from astro.sql.operators.agnostic_save_file import save_file
+from astro.sql.operators.save_file import save_file
```
tatiana, line 31: Do we have a test case covering the scenario: https://github.com/astronomer/astro-sdk/issues/365?

src/astro/sql/operators/save_file.py
tatiana: I just spoke to @dimberman , he gave a good suggestion: if the user does not specify an `output_table` in `load_file`, we load to a dataframe.

@mikeshwe , feel free to loop in this discussion!

The suggestion is:
```
dataframe = aql.loadfile(File("s3://some-bucket/some-prefix"))
```

mikeshwe: I do like the sleek, minimalist look of this function! The current load_file requires an output_table parameter, AFAIK, which is unfortunate if we are just loading to a table object and don't care what the table is in the database, e.g,:

`orders_data = aql.load_file(
        path=S3_FILE_PATH + '/orders_data_header.csv',
        file_conn_id=S3_CONN_ID,
        output_table=Table(database=SNOWFLAKE_DATABASE,conn_id=SNOWFLAKE_CONN_ID))`

The part where I'm currently confused is that does it matter whether this:

`aql.loadfile(File("s3://some-bucket/some-prefix"))` 

returns a dataframe or a Table? Because even if we want to return a Table object, it seems unnecessary to have to specify the output_table parameter, because the SDK should just take care of that in the background--at most we should only have to globally specify some default backend for the SDK to do its magic, like in a .env variable.

IOW, it would be great if the same minimalist call to `load_file` could return either a dataframe or a Table object--maybe we have a boolean or enumerated type to specify which one?

tatiana: @mikeshwe thanks for your contribution! If we make the `output_table` argument optional and introduce the boolean argument, how do you suggest we retrieve the connection to create the temporary table ?

example_dags/example_google_bigquery_gcs_load_and_save.py
utkarsharma2: Yes, it did! :)

tests/operators/test_agnostic_load_file.py
utkarsharma2: @kaxil, I'll take care of this issue in a separate PR @tatiana also wanted a change in this Fixture. I'll address them both there.

src/astro/utils/load.py
utkarsharma2: I looked up, other methods of load files, they are mainly used in sql_dataframe.py and sql_decorator.py, which I think Daniel PR should cover.

src/astro/sql/operators/load_file.py
utkarsharma2: to test if the conn_id exists or not.

conftest.py
utkarsharma2: This was reintroduced in rebase. Removed it.

src/astro/sql/operators/load_file.py
utkarsharma2: @tatiana, What would happen in case user wants to keep the default schema of DB and with that intent did not provide the schema in the table object?

utkarsharma2: Done.

src/astro/sql/operators/save_file.py
utkarsharma2: Done.

tests/operators/test_save_file.py
utkarsharma2: I'll address this in a different PR, it's already too big.

tests/operators/test_load_file.py
utkarsharma2: I'll address this in a different PR, it's already too big.

tests/operators/test_load_file.py
utkarsharma2: Do we need that? we have every load_file() test using that interface.

tests/operators/test_save_file.py
utkarsharma2: I'll address this in a different PR, it's already too big.

src/astro/sql/operators/save_file.py
utkarsharma2: @tatiana 
1. Do they need to be symmetric? Also, if we didn't have the need so far, why do we need it now?
2. I think there are easier alternatives to getting a data frame from a file right, pd.readCSV() etc. It offers little value.

tests/operators/test_save_file.py
utkarsharma2: I'll add this.

src/astro/sql/operators/agnostic_load_file.py
utkarsharma2: @kaxil Well, one thought was that property is not consumed by any function within File Class. The function(_populate_normalize_config) which uses it lives within the LoadFile Op file, and it should live at this location because it has knowledge about both file class and database class.

src/astro/sql/operators/save_file.py
dimberman: @utkarsharma2 I disagree. We are still handling the complexity and keeping stuff within our ecosystem. This will also offer value when we later offer lineage and data validation checks. Telling people "well just write a python function and use pandas.read_csv" would be asking them to leave our ecosystem and miss out on later benefits.

dimberman: Also if they later decide they want to move that data to a database they would only need to add an output_table vs. they would need to rewrite everything if they use pandas.read_csv. We are also handling the complexity around pulling from a variety of datastores.

kaxil: I like @dimberman 's suggestion of an optional `output_table` param and do agree with the argument of one interface instead of recommend pd.read_csv

src/astro/sql/operators/load_file.py
kaxil: We should add this as a comment, also, it feels doesn't feel like a good way to check! We probably should add a method in OSS Airflow too to check if a Connection exists or not.

Until then though, we should do the following:

```suggestion
        if self.input_file.file_conn_id:
            try:
                BaseHook.get_connection(self.file_conn_id)
            except AirflowNotFoundException:
                # TODO: We should use our custom exception here
                raise AirflowNotFoundException(
                    f"Connection ('{self.input_file.conn_id}') does not exist in Airflow. "
                    "Please pass a valid connection ID to File()"
                )
```

We can even have a separate method/function:

```python
        def connection_exists(conn_id: str) -> bool
            try:
                BaseHook.get_connection(self.file_conn_id)
                return True
            except AirflowNotFoundException:
                return False
```

src/astro/databases/snowflake.py
```diff
@@ -61,9 +61,12 @@ def load_pandas_dataframe_to_table(
         if target_table.metadata:
             schema = getattr(target_table.metadata, "schema", None)
 
+        # within prep_table() we use pandas drop() function which is used when we pass 'if_exists=replace'.
+        # There is an issue where has_table() works with uppercase table names but the function meta.reflect() don't.
+        # To prevent the issue we are passing table name in lowercase.
```
kaxil, line 66: Is there a GitHub issue on the SqlAlchemy repo that you know of for this issue? If so, we can add it in the comments, and replace it once that issue is fixed

kaxil: (Not a PR blocking comment)

src/astro/sql/operators/load_file.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
-        self.output_table: Table = output_table
```
kaxil: This isn't needed because, we already have typehint for `output_table` in constructor

src/astro/files/type/__init__.py
```diff
@@ -44,7 +44,16 @@ def get_filetype(filepath: Union[str, pathlib.PosixPath]) -> FileTypeConstants:
...
-        extension = filepath.split(".")[-1]
+        extension = ""
+        tokenized_path = filepath.split(".")
+        if len(tokenized_path) > 1:
+            extension = tokenized_path[-1]
+
+    if extension == "":
+        raise ValueError(
+            f"Missing file extension, cannot determine filetype from path '{filepath}'."
+            f" If this path contains pattern pass 'filetype' param."
```
kaxil, line 55: what do you mean by "pattern" here?

dimberman: @kaxil I think this means that we can't autoparse the type so the user needs to explicitly state it. I changed the comment to reflect this.

src/astro/sql/operators/load_file.py
```diff
@@ -0,0 +1,174 @@
...
+        super().__init__(**kwargs)
+        self.output_table = output_table
+        self.input_file = input_file
+        self.chunk_size = chunk_size
+        self.kwargs = kwargs
+        self.if_exists = if_exists
+        self.ndjson_normalize_sep = ndjson_normalize_sep
+        self.normalize_config: Dict[str, str] = {}
+
+    def execute(self, context: Any) -> Table:
```
kaxil, line 45:
```suggestion
    def execute(self, context: Any) -> Union[Table, pd.DataFrame]::
```

kaxil: Would be good to add this in docstrings too with:

```
:return: If ``output_table`` is passed this operator returns a Table object. If not passed, returns a dataframe.
```

src/astro/utils/load.py
```diff
@@ -97,43 +97,6 @@ def flatten_ndjson(
...
-            normalize_config["record_prefix"] = replacement
-
-        sep = normalize_config.get("sep")
-        if sep is None or sep == illegal_char:
-            normalize_config["sep"] = replacement
-
-    return normalize_config
-
-
 def load_file_rows_into_dataframe(
```
kaxil, line 100: Did you check whether we need other functions in these file or can they be deleted

tests/operators/test_load_file.py
```diff
@@ -40,16 +43,17 @@
...
 @pytest.mark.parametrize(
     "test_table",
-    [{"is_temp": True}, {"is_temp": False}],
+    [{}],
```
kaxil, line 46: We still need to fix this empty list thing

src/astro/sql/operators/load_file.py
```diff
@@ -0,0 +1,179 @@
...
+
+    template_fields = ("output_table", "input_file")
+
+    def __init__(
+        self,
+        input_file: File,
+        output_table: Optional[Table] = None,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+        if_exists: LoadExistStrategy = "replace",
+        ndjson_normalize_sep: str = "_",
```
kaxil, line 36: We need to think about how we expose `ndjson_normalize_sep`. It is too specific to one `file_type`.

kaxil (approved):
I am going to merge this but there are still some unaddressed comments that we should fix @utkarsharma2 in the follow-up PR, please take a look at those comments.

PR #363, opened 2022-05-16
Refactor truncate operator

### LINKED ISSUE #342: Refactor the `truncate` operator
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

**Dependencies**
* Depends on: #340 
* Depends on: #344

**Acceptance criteria**
* [ ] Implement the `truncate` operator using the new `Database` class and using the `create_database(conn_id)` function

tatiana:
Resolve  #342

There are some decisions taken during the development of this ticket that will affect other parts of the project:

1. Retrieve the schema and other metadata from the connection if the table doesn't contain metadata. The only exception is BigQuery - since it seems there is no standard way of defining the schema for this type of connection. In the case of BigQuery, we're retrieving the schema from the environment variable if not defined in the table creation.

2. Use SQLAlchemy to run SQL queries. The initial intent to use Airflow hooks was to simplify the future adoption of AsyncOperators. However, each of the currently supported DBs implements the hooks differently (for BQ, for instance, `run_sql` returns a job ID). Additionally, each hook produces the output in a different format. Since we are focused on the refactoring, for now, and not on AsyncOperators - I decided to stick to the standard and straightforward interface for performing SQL queries. We can revisit this in the future.

### DIFF
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -1,15 +1,14 @@
+"""Google BigQuery table implementation."""
+
 import pandas as pd
 from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
 from sqlalchemy import create_engine
 from sqlalchemy.engine.base import Engine
 
-from astro.constants import (
-    DEFAULT_CHUNK_SIZE,
-    AppendConflictStrategy,
-    LoadExistStrategy,
-)
+from astro import settings
+from astro.constants import DEFAULT_CHUNK_SIZE, LoadExistStrategy
 from astro.databases.base import BaseDatabase
-from astro.sql.tables import Table
+from astro.sql.tables import Metadata, Table
 
 DEFAULT_CONN_ID = BigQueryHook.default_conn_name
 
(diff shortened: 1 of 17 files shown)


src/astro/databases/google/bigquery.py
```diff
@@ -36,39 +35,19 @@ def sqlalchemy_engine(self) -> Engine:
...
     def get_table_qualified_name(self, table: Table) -> str:
         """
-        Return the table qualified name.
+        Return table qualified name for BigQuery.
 
         :param table: The table we want to retrieve the qualified name for.
         """
-        schema = table.metadata.schema if table.metadata else None
-        qualified_name: str = schema + "." + table.name if schema else table.name
+        if table.metadata is not None and table.metadata.schema is not None:
```
kaxil, line 42: nit:

```suggestion
        if table.metadata and table.metadata.schema:
```

src/astro/databases/snowflake.py
```diff
@@ -24,6 +25,31 @@ def hook(self):
...
+        """
+        Return table qualified name for Snowflake.
+
+        :param table: The table we want to retrieve the qualified name for.
+        """
+        if table.metadata is not None and table.metadata.schema is not None:
+            qualified_name = f"{table.metadata.schema}.{table.name}"
+        else:
+            qualified_name = table.name
+        return qualified_name
```
kaxil, line 38: Hmm do you think we can move this to BaseDB since it is same for most of at least 3 of them?

tatiana: That was the idea, @kaxil, but I can rollback! What are your thoughts?

kaxil: Yeah I would be in favor of moving this logic to `BaseDatabase.get_table_qualified_name`

kaxil (approved):

PR #370, opened 2022-05-19
Run tests against multiple airflow versions

kaxil:
We want to make sure that Astro SDK continues working with Airflow 2.2.5 and Airflow 2.3 - so this commit adds this capability in nox.

A problem I foresee is some tests failing intermittently as some of our "integration tests" don't have unique ids for tables, dataset which means there might be times when an integration test is running in 2.2.5 which deletes a resource that is being used by same test running for 2.3.

One "quick" solution is to run them sequentially but that will increase total test time. A better solution will be to fix the tests but that will take time.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,7 +101,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        group: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 ]
+        group: [ 1, 2, 3, 4, 5, 6 ]
     runs-on: ubuntu-latest
     services:
       postgres:
@@ -133,7 +133,7 @@ jobs:
       - run: python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
       - run: sqlite3 /tmp/sqlite_default.db "VACUUM;"
       - run: pip3 install nox
-      - run: nox -s test -- --splits 12 --group ${{ matrix.group }} --cov=src --cov-report=xml --cov-branch
+      - run: nox -s unit_test -- --splits 6 --group ${{ matrix.group }} --cov=src --cov-report=xml --cov-branch
       - name: Upload coverage
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -133,7 +133,7 @@ jobs:
       - run: python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
       - run: sqlite3 /tmp/sqlite_default.db "VACUUM;"
       - run: pip3 install nox
-      - run: nox -s test -- --splits 12 --group ${{ matrix.group }} --cov=src --cov-report=xml --cov-branch
+      - run: nox -s unit_test -- --splits 6 --group ${{ matrix.group }} --cov=src --cov-report=xml --cov-branch
```
tatiana, line 136: Did this change increase the time it takes for the CI to run tests? If yes, in how much?

noxfile.py
```diff
@@ -21,12 +21,29 @@ def dev(session: nox.Session) -> None:
...
+    # Log all the installed dependencies
+    session.log("Installed Dependencies:")
+    session.run("pip3", "freeze")
     session.run("airflow", "db", "init")
-    session.run("pytest", *session.posargs)
+    session.run("pytest", "-k", "not integration", *session.posargs)
+
+
+@nox.session(python=["3.7", "3.8", "3.9"])
+@nox.parametrize("airflow", ["2.2.5", "2.3"])
```
tatiana, line 36: This is great, @kaxil !

noxfile.py
```diff
@@ -21,12 +21,29 @@ def dev(session: nox.Session) -> None:
 
 
 @nox.session(python=["3.7", "3.8", "3.9"])
-def test(session: nox.Session) -> None:
+def unit_test(session: nox.Session) -> None:
```
tatiana, line 24: I like the split between unit and integration tests!

tatiana:
@kaxil, once the tests are passing, let's merge this! It is probably worth double-checking that coverage is working as expected after the tests are fixed.

tatiana (approved):

tatiana:
@kaxil, how is this going? Do we have any risks that Astro may stop working with previous versions of Airflow if this is not merged sooner than later?

kaxil:
> @kaxil, how is this going? Do we have any risks that Astro may stop working with previous versions of Airflow if this is not merged sooner than later?

No risk, until we release a new version and update any libraries in `setup.py`. Will get to it in a day or two

kaxil:
We want to make sure that Astro SDK continues working with Airflow 2.2.5 and Airflow 2.3 - so this commit adds this capability in nox.

A problem I foresee is some tests failing intermittently as some of our "integration tests" don't have unique ids for tables, dataset which means there might be times when an integration test is running in 2.2.5 which deletes a resource that is being used by same test running for 2.3.

One "quick" solution is to run them sequentially but that will increase total test time. A better solution will be to fix the tests but that will take time.

PR #371, opened 2022-05-19
CI: Get list of the slowest 30 test durations

kaxil:
Uses https://docs.pytest.org/en/latest/how-to/usage.html#profiling-test-execution-duration to get a list of slowest 30 tests for which the durations is over 1.0s long.

This is a small change but will help identify if a test takes too long. Found a need to do that while working on https://github.com/astro-projects/astro/pull/367 and https://github.com/astro-projects/astro/pull/370

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = ["--durations=30 --durations-min=1.0"]
+addopts = ["--durations=30", "--durations-min=1.0"]
 env_files = [".env"]
 testpaths = ["tests"]
 markers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = ["--durations=30", "--durations-min=1.0"]
+addopts = "--durations=30 --durations-min=1.0"
 env_files = [".env"]
 testpaths = ["tests"]
 markers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,7 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
+addopts = ["--durations=30 --durations-min=1.0"]
 env_files = [".env"]
 testpaths = ["tests"]
 markers = [


tatiana:
@kaxil, that's very interesting. I didn't know about this feature!

Even though we're using `"pytest>=6.0"`, which should support this feature, the CI didn't seem particularly happy with it; any thoughts about it? It may relate to how `pytest` parses the config on the `pyproject.toml`.

```
nox > pytest tests/test_example_dags.py -k sqlite --cov=src --cov-report=xml --cov-branch
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: argument --durations: invalid int value: '30 --durations-min=1.0' (via addopts config)

nox > Command pytest tests/test_example_dags.py -k sqlite --cov=src --cov-report=xml --cov-branch failed with exit code 4
```

pyproject.toml
```diff
@@ -88,6 +88,7 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
+addopts = ["--durations=30 --durations-min=1.0"]
```
kaxil, line 91:
```suggestion
addopts = ["--durations=30", "--durations-min=1.0"]
```

pyproject.toml
```diff
@@ -88,6 +88,7 @@ provider_info = "astro.__init__:get_provider_info"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
+addopts = ["--durations=30", "--durations-min=1.0"]
```
kaxil, line 91:
```suggestion
addopts = "--durations=30 --durations-min=1.0"
```

kaxil:
@tatiana fixed :)

tatiana:
That's great, @kaxil , thank you! :tada:
What is causing the current tests to fail? It seems they are passing in main, it would be great to address the root cause of the issue before merging this PR.

kaxil:
> That's great, @kaxil , thank you! 🎉 What is causing the current tests to fail? It seems they are passing in main, it would be great to address the root cause of the issue before merging this PR.

It's the one I posted on Slack about: 
```
E       sqlalchemy.exc.ProgrammingError: (snowflake.connector.errors.ProgrammingError) 090153 (22000): 01a46079-0604-a49f-0000-682108eb3a66: The result set size exceeded the max number of rows(10000) supported for SHOW statements. Use LIMIT option to limit result set to a smaller number.
E       [SQL: SHOW /* sqlalchemy:get_table_names */ TABLES IN astroflow_ci]
E       (Background on this error at: http://sqlalche.me/e/14/f405)
```

But I don't think we should hold this PR for that failure, it is completely unrelated, this PR barely does not impact the library just gives us the slowest test.

tatiana (approved):

utkarsharma2 (approved):

PR #380, opened 2022-05-20
Support table.name correct sizing after setting table.schema

tatiana:
@dimberman identified a scenario where the table name was being created incorrectly.

There are situations in the code when we want to associate metadata (e.g., schema) to a temporary table after it was created. When this happens, the previous implementation of `table.name` would auto-generate the name during initialization without considering the schema (which had not been set yet).

This PR introduces test cases and adjusts the implementation to generate only the `table.name` once it is needed (when trying to access the attribute).

### DIFF
--- a/src/astro/sql/table.py
+++ b/src/astro/sql/table.py
@@ -1,14 +1,14 @@
 import random
 import string
 from dataclasses import dataclass, field, fields
-from typing import List, Optional, Union
+from typing import List, Union
 
 from airflow.models import DagRun, TaskInstance
 from sqlalchemy import Column, MetaData
 
 from astro.constants import UNIQUE_TABLE_NAME_LENGTH
 
-MAX_TABLE_NAME_LENGTH = 45
+MAX_TABLE_NAME_LENGTH = 62
 
 
 @dataclass
@@ -45,17 +45,14 @@ class Table:
     # SQL table itself
     # Some ideas: TableRef, TableMetadata, TableData, TableDataset
     conn_id: str = ""
-    name: str = ""
+    name: Union[str, property] = ""
+    _name: str = field(init=False, repr=False, default="")
     metadata: Metadata = field(default_factory=Metadata)
-    columns: Optional[List[Column]] = None
+    columns: List[Column] = field(default_factory=list)
     temp: bool = False
 
     def __post_init__(self):
-        if self.columns is None:
-            self.columns = []
-
-        if not self.name:
-            self.name = self._create_unique_table_name()
+        if not self._name:
             self.temp = True
 
     def _create_unique_table_name(self) -> str:
@@ -78,6 +75,18 @@ class Table:
         else:
             alchemy_metadata = MetaData()
         return alchemy_metadata
+
+    @property  # type: ignore
+    def name(self) -> str:
+        if self.temp and not self._name:
+            self._name = self._create_unique_table_name()
+        return self._name
+
+    @name.setter
+    def name(self, value: Union[str, property]) -> None:
+        if not isinstance(value, property):
+            self._name = value
+            self.temp = False
 
 
 # TODO: deprecate by the end of the refactoring
--- a/tests/sql/test_table.py
+++ b/tests/sql/test_table.py
@@ -1,0 +1,39 @@
+from astro.sql.table import Table
+
+
+def test_table_with_explicit_name():
+    """Check that we respect the name of table when it's named an we set it to non-temp"""
+    table = Table(conn_id="some_connection", name="some_name")
+    assert table.conn_id == "some_connection"
+    assert table.name == "some_name"
+    assert not table.temp
+
+
+def test_table_without_name():
+    """Check that we create a name, that it is always the same name, and that we set temp name to True"""
+    table = Table(conn_id="some_connection")
+    assert table.conn_id == "some_connection"
+    name1 = table.name
+    name2 = table.name
+    assert name1
+    assert isinstance(name1, str)
+    assert len(table.name) == 62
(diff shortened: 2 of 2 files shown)


src/astro/sql/table.py
```diff
@@ -79,6 +76,42 @@ def sqlalchemy_metadata(self) -> MetaData:
             alchemy_metadata = MetaData()
         return alchemy_metadata
 
+    def __getattribute__(self, attribute_name):
```
kaxil, line 79: Can we not create a `name` property it as follows, similar to the example in https://docs.python.org/3/library/functions.html#property:

```python
@property
def name(self):
     # Custom logic you have in getter 
     if self.temp and not self._name:
         return self._create_unique_table_name()
     return self._name
```

tatiana: @kaxil, my first attempts used properties, and they were not successful. Some attempts I tried:

**Attempt 1**

```
class Table:
    _name: str=""

    @property
    def name(self) -> str:
        if self.temp and not self._name:
            return self._create_unique_table_name()

    @name.setter
    def name(self, value: str) -> None:
        self._name = value
```

Which resulted in all tests failing:
```
    def test_table_with_explicit_name():
        """Check that we respect the name of table when it's named an we set it to non-temp"""
        table = Table(conn_id="some_connection", name="some_name")
        assert table.conn_id == "some_connection"
>       assert table.name == "some_name"
E       AssertionError: assert None == 'some_name'
E        +  where None = Table(conn_id='some_connection', name=None, metadata=Metadata(schema=None, account=None, database=None, host=None, region=None, role=None, warehouse=None), columns=[], temp=False).name

tests/sql/test_table.py:8: AssertionError
------------------------------------------------------------------------------- Captured stdout setup --------------------------------------------------------------------------------
[2022-05-20 14:16:23,654] {crypto.py:82} WARNING - empty cryptography key - values will not be stored encrypted.
--------------------------------------------------------------------------------- Captured log setup ---------------------------------------------------------------------------------
WARNING  airflow.models.crypto:crypto.py:82 empty cryptography key - values will not be stored encrypted.
______________________________________________________________________________ test_table_without_name _______________________________________________________________________________

    def test_table_without_name():
        """Check that we create a name, that it is always the same name, and that we set temp name to True"""
        table = Table(conn_id="some_connection")
        assert table.conn_id == "some_connection"
        name1 = table.name
        name2 = table.name
>       assert name1
E       assert None

tests/sql/test_table.py:18: AssertionError
_________________________________________________________________________ test_table_without_name_and_schema _________________________________________________________________________

    def test_table_without_name_and_schema():
        """Check that the table name is smaller when there is metadata associated to the table."""
        table = Table(conn_id="some_connection")
        table.metadata.schema = "abc"
>       assert isinstance(table.name, str)
E       AssertionError: assert False
E        +  where False = isinstance(None, str)
E        +    where None = Table(conn_id='some_connection', name=None, metadata=Metadata(schema='abc', account=None, database=None, host=None, region=None, role=None, warehouse=None), columns=[], temp=False).name

```

**Attempt 2**

```
    @property
    def name(self) -> str:
        if self.temp and not self._name:
            return self._create_unique_table_name()

    @name.setter
    def name(self, value: str) -> None:
        if type(value) is not property:
            self._name = value

```

Which resulted in:
```
    def test_table_with_explicit_name():
        """Check that we respect the name of table when it's named an we set it to non-temp"""
        table = Table(conn_id="some_connection", name="some_name")
        assert table.conn_id == "some_connection"
>       assert table.name == "some_name"
E       AssertionError: assert None == 'some_name'
E        +  where None = Table(conn_id='some_connection', name=None, metadata=Metadata(schema=None, account=None, database=None, host=None, region=None, role=None, warehouse=None), columns=[], temp=False).name

tests/sql/test_table.py:8: AssertionError
------------------------------------------------------------------------------- Captured stdout setup --------------------------------------------------------------------------------
[2022-05-20 14:18:58,617] {crypto.py:82} WARNING - empty cryptography key - values will not be stored encrypted.
--------------------------------------------------------------------------------- Captured log setup ---------------------------------------------------------------------------------
WARNING  airflow.models.crypto:crypto.py:82 empty cryptography key - values will not be stored encrypted.
______________________________________________________________________________ test_table_without_name _______________________________________________________________________________

    def test_table_without_name():
        """Check that we create a name, that it is always the same name, and that we set temp name to True"""
        table = Table(conn_id="some_connection")
        assert table.conn_id == "some_connection"
        name1 = table.name
        name2 = table.name
        assert name1
        assert isinstance(name1, str)
        assert len(table.name) == 62
>       assert name1 == name2
E       AssertionError: assert 'o7dkhq82gf0a...z3lt5lieag5to' == 'y91dzxtv8s7h...y2ilklpivkum6'
E         - y91dzxtv8s7hm231hnobldu8wquc6z8cgruknp2o45gap3q0gy2ilklpivkum6
E         + o7dkhq82gf0am1k2f1mt9gpg3iu7xmnvkwsjkc10xvd0w9jniz3lt5lieag5to

tests/sql/test_table.py:21: AssertionError

```

I made several attempts of using properties changing the field types and the data class definition itself. Since we have other priorities, I thought it was better to have a working ugly version and move to the other more critical PRs - we can replace the current implementation or an alternative working one!

dimberman: @tatiana in the property code you wrote, it looks like you don't return a name if temp=false. Could that be why tests were failing?

tatiana: Great catch, @dimberman ! Thank you :heart:  I just updated the code and pushed a new version :tada:

kaxil (approved):

PR #383, opened 2022-05-21
Proposal - merge and append feature for postgres.

utkarsharma2:
Proposal for Postgres merge/append operation.


tests/databases/test_postgres.py
```diff
@@ -320,3 +320,137 @@ def test_create_table_from_select_statement(database_table_fixture):
...
+            "target_tables_cols": ["sell", "list"],
+            "conflict_strategy": "ignore",
+            "expected_df": pd.DataFrame(
+                [
+                    {"sell": 142, "list": 160},
+                    {"sell": 162, "list": 330},
+                    {"sell": 175, "list": 540},
+                ]
+            ),
+        },
```
dimberman, line 353: Why does this function require so much set up? This doesn't feel scalable

src/astro/constants.py
```diff
@@ -62,3 +62,5 @@ class Database(Enum):
 
 # TODO: check how snowflake names these
 AppendConflictStrategy = Literal["append", "replace", "exception"]
+
+DBMergeConflictStrategy = Literal["ignore", "update"]
```
tatiana, line 66: None of the other Strategies has DB in the name. Why aren't we using `AppendConflictStrategy` (which was meant for the hybrid append-merge)?

src/astro/databases/base.py
```diff
@@ -1,13 +1,13 @@
...
+from typing import List, Optional
 
 import pandas as pd
 import sqlalchemy
 from airflow.hooks.base import BaseHook
 
 from astro.constants import (
     DEFAULT_CHUNK_SIZE,
-    AppendConflictStrategy,
+    DBMergeConflictStrategy,
```
tatiana, line 10: Why was the previous Strategy replaced? It contained all the cases for append and merge.

src/astro/databases/base.py
```diff
@@ -201,15 +201,44 @@ def append_table(
         self,
         source_table: Table,
         target_table: Table,
-        if_conflicts: AppendConflictStrategy = "exception",
+        source_tables_cols: Optional[List[str]] = None,
+        target_tables_cols: Optional[List[str]] = None,
```
tatiana, line 205: The table definition has a property `columns`:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/sql/table.py#L45

Why aren't we using it in these two lines?

src/astro/databases/base.py
```diff
@@ -201,15 +201,44 @@ def append_table(
         self,
         source_table: Table,
         target_table: Table,
-        if_conflicts: AppendConflictStrategy = "exception",
```
tatiana, line 204: Why was this parameter removed?

src/astro/databases/base.py
```diff
@@ -201,15 +201,44 @@ def append_table(
...
+        :param target_tables_cols: List of columns name in target table that will be used in appending
+        """
+        return self.merge_table(
+            source_table=source_table,
+            target_table=target_table,
+            source_tables_cols=source_tables_cols,
+            target_tables_cols=target_tables_cols,
+        )
+
+    def merge_table(
```
tatiana, line 224: The end goal was to have a single method for append and merge, with a single `Strategy` (previously defined). why have we created a new method?

src/astro/databases/base.py
```diff
@@ -201,15 +201,44 @@ def append_table(
...
+        self,
+        source_table: Table,
+        target_table: Table,
+        conflict_strategy: DBMergeConflictStrategy = "ignore",
+        merge_cols: Optional[List[str]] = None,
+        source_tables_cols: Optional[List[str]] = None,
+        target_tables_cols: Optional[List[str]] = None,
```
tatiana, line 231: The table definition has a property `columns`:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/sql/table.py#L45

Why aren't we using it in these two lines? Ideally, one of the refactoring goals was to avoid having multiple function arguments that refer to the table, file, etc. Those properties should be contained in the objects of `Table`, and `File`, where possible. The exception is arguments of the transformation itself.

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
+
+        statement = "INSERT INTO {main_table} ({target_columns}) SELECT {append_columns} FROM {append_table}"
+
+        if len(merge_cols) > 0:
+            if conflict_strategy == "ignore":
+                statement += " ON CONFLICT ({merge_keys}) DO NOTHING"
+            elif conflict_strategy == "update":
+                statement += (
+                    " ON CONFLICT ({merge_keys}) DO UPDATE SET {update_statements}"
+                )
```
tatiana, line 73: so far, we've used constants like: `_create_schema_statement` to define this type of query in the `Database`. Is there any reason for not using them here?

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
+        merge_cols = merge_cols or []
+
+        if len(source_tables_cols) != len(target_tables_cols):
+            raise ValueError(
+                "Params 'source_tables_cols' and  'target_tables_cols' should contain same number of cols."
+            )
+
+        def identifier_args(table):
+            schema = table.metadata.schema
+            return (schema, table.name) if schema else (table.name,)
```
tatiana, line 63: We could have a method available only in the `PostgresDatabase` class (it doesn't have to be declared inside the append/merge function), named `_psycopg_identifier`, which returns an instance of `postgres_sql.Identifier` with the desired arguments.

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
+
+        statement = "INSERT INTO {main_table} ({target_columns}) SELECT {append_columns} FROM {append_table}"
+
+        if len(merge_cols) > 0:
```
tatiana, line 67: `if merge_cols`?

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
         return Metadata(database=self.hook.schema)
+
+    def merge_table(
+        self,
+        source_table: Table,
+        target_table: Table,
+        conflict_strategy: DBMergeConflictStrategy = "ignore",
+        merge_cols: Optional[List[str]] = None,
+        source_tables_cols: Optional[List[str]] = None,
+        target_tables_cols: Optional[List[str]] = None,
```
tatiana, line 40: Commented previously:
https://github.com/astronomer/astro-sdk/pull/383/files#r878776864

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
+
+        if len(merge_cols) > 0:
+            if conflict_strategy == "ignore":
+                statement += " ON CONFLICT ({merge_keys}) DO NOTHING"
+            elif conflict_strategy == "update":
+                statement += (
+                    " ON CONFLICT ({merge_keys}) DO UPDATE SET {update_statements}"
+                )
+
+        source_tables_cols = [postgres_sql.Identifier(c) for c in source_tables_cols]
```
tatiana, line 75: We could have a dedicated method to return table columns in the Psyco format: `_psycopg_table_columns`.
Smaller functions can be easily unit-tested.

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
+        merge_cols: Optional[List[str]] = None,
+        source_tables_cols: Optional[List[str]] = None,
+        target_tables_cols: Optional[List[str]] = None,
+    ) -> None:
+        """Merge two tables based on merge keys
+
+        :param source_table: Contains the rows to be appended to the target_table
+        :param target_table: Contains the destination table in which the rows will be appended
+        :param conflict_strategy: Action that needs to be taken in case there is a conflict
+        :param merge_cols: List of cols that are checked for uniqueness while merging,
```
tatiana, line 47: `conflict_columns` may be a more suitable name. I'd also advise using a list of Sqlalchemy `Column` instances - as we used in the `Table` definition, so we are consistent thought our interface about how we represent these columns.

src/astro/databases/postgres.py
```diff
@@ -26,3 +29,68 @@ def hook(self):
...
+            main_table=postgres_sql.Identifier(*identifier_args(target_table)),
+            append_columns=postgres_sql.SQL(",").join(source_tables_cols),
+            append_table=postgres_sql.Identifier(*identifier_args(source_table)),
+            update_statements=postgres_sql.SQL(",").join(update_statements),
+            merge_keys=postgres_sql.SQL(",").join(
+                [postgres_sql.Identifier(x) for x in merge_cols]
+            ),
+        )
+
+        hook = PostgresHook(postgres_conn_id=target_table.conn_id)
```
tatiana, line 94: why aren't we using `self.hook`?

tests/databases/test_postgres.py
```diff
@@ -320,3 +320,137 @@ def test_create_table_from_select_statement(database_table_fixture):
...
+        source_dataframe=target_df, target_table=target_table
+    )
+    db.load_pandas_dataframe_to_table(
+        source_dataframe=source_df, target_table=source_table
+    )
+
+    db.run_sql(
+        sql_statement=f"ALTER TABLE {db.get_table_qualified_name(target_table)} "
+        f"ADD CONSTRAINT con_{target_table.name} UNIQUE ({','.join(merge_cols)})"
+    )
```
tatiana, line 387: How are we tearing down this constraint?

tests/databases/test_postgres.py
```diff
@@ -320,3 +320,137 @@ def test_create_table_from_select_statement(database_table_fixture):
...
+
+    merge_cols = parameters["merge_cols"]
+
+    db = create_database(conn_id=target_table.conn_id)
+    db.load_pandas_dataframe_to_table(
+        source_dataframe=target_df, target_table=target_table
+    )
+    db.load_pandas_dataframe_to_table(
+        source_dataframe=source_df, target_table=source_table
+    )
```
tatiana, line 382: how are we tearing down these tables?

tests/databases/test_postgres.py
```diff
@@ -320,3 +320,137 @@ def test_create_table_from_select_statement(database_table_fixture):
...
+    target_table = Table(conn_id="postgres_conn")
+    source_table = Table(conn_id="postgres_conn")
+
+    db = create_database(conn_id=target_table.conn_id)
+    db.load_pandas_dataframe_to_table(
+        source_dataframe=target_df, target_table=target_table
+    )
+    db.load_pandas_dataframe_to_table(
+        source_dataframe=source_df, target_table=source_table
+    )
```
tatiana, line 439: We should tear down these tables, regardless if the test succeeds or not.

src/astro/databases/base.py
dimberman: @tatiana merge and append can still be wildly different in implementation even if it looks like one "endpoint" to the user. This is why I have not been a fan of merging these functions.

src/astro/databases/base.py
kaxil: https://en.wikipedia.org/wiki/Merge_(SQL) - Yea Merge for most DB MySQL, Postres, SQLite this is Insert and do some actions on conflicts

kaxil: - https://www.postgresql.org/docs/current/sql-insert.html
- https://dev.mysql.com/doc/refman/8.0/en/insert.html

whereas Databricks & Snowflake do have a Megre function

- https://docs.databricks.com/delta/delta-update.html#upsert-into-a-table-using-merge
- https://docs.snowflake.com/en/sql-reference/sql/merge.html

src/astro/databases/base.py
dimberman: @tatiana the reason for these two lists is that the source table and the target table might have different names for these columns, so we would want to make sure that we merge the correct columns.

src/astro/databases/postgres.py
utkarsharma2: Done.

src/astro/databases/postgres.py
utkarsharma2: We can, but that would require us to invert conditions and a whole bunch of code would be under if condition, I think it is cleaner this way.

tests/databases/test_postgres.py
utkarsharma2: @dimberman 
1. I think it is simpler to understand what is happening within this test this way as opposed to moving this data in fixtures/functions.

2. I don't see how we can reduce the setup code here, since that is all the parameters that the operator requires.

3. I'm getting your point regarding the scaling of this.

src/astro/databases/base.py
utkarsharma2: we need a separate set of cols since the user can choose a subset of cols as opposed to all the cols specified in the Table object, which is a user-defined construct and altering is not something we should do.

src/astro/databases/base.py
utkarsharma2: We agreed on new Conflict Strategy values and started using them.

src/astro/databases/base.py
utkarsharma2: We have started to use the new function combine_tables() function for all purposes.

tests/databases/test_postgres.py
utkarsharma2: Used fixture to teardown the table created.

tests/databases/test_postgres.py
utkarsharma2: Used fixture to teardown the table created.

src/astro/databases/postgres.py
```diff
@@ -110,3 +126,62 @@ def table_exists(self, table: Table) -> bool:
 
         inspector = sqlalchemy.inspect(self.sqlalchemy_engine)
         return bool(inspector.dialect.has_table(self.connection, table.name, schema))
+
+    def combine_tables(
+        self,
+        source_table: Table,
+        target_table: Table,
+        source_to_target_columns_map: Dict[str, str],
```
tatiana, line 134: Could this parameter be optional?
A widespread use case here would be to combine tables with the same columns.

tatiana: @kaxil @mikeshwe any thoughts?

utkarsharma2:
Since these features are picked and added by other PRs - https://github.com/astronomer/astro-sdk/pull/439 and https://github.com/astronomer/astro-sdk/pull/422, closing this one.

Issue #364, opened 2022-05-17
Add `_tmp_` prefix to Temporary Table


kaxil:
This will allow users to identify and delete all the temp tables until we have a proper fix in https://github.com/astro-projects/astro/issues/187

PR #384, opened 2022-05-22
Add prefix to unnamed temporary table name

tatiana:
Resolve: #364

### DIFF
--- a/src/astro/sql/table.py
+++ b/src/astro/sql/table.py
@@ -1,8 +1,7 @@
-from __future__ import annotations
-
 import random
 import string
 from dataclasses import dataclass, field, fields
+from typing import List, Optional
 
 from airflow.models import DagRun, TaskInstance
 from sqlalchemy import Column, MetaData
@@ -21,8 +20,8 @@ class Metadata:
     """
 
     # This property is used by several databases, including: Postgres, Snowflake and BigQuery ("namespace")
-    schema: str | None = None
-    database: str | None = None
+    schema: Optional[str] = None
+    database: Optional[str] = None
 
     def is_empty(self):
         """Check if all the fields are None."""
@@ -44,10 +43,10 @@ class Table:
     # SQL table itself
     # Some ideas: TableRef, TableMetadata, TableData, TableDataset
     conn_id: str = ""
-    name: str | property = ""
+    name: str = ""
     _name: str = field(init=False, repr=False, default="")
     metadata: Metadata = field(default_factory=Metadata)
-    columns: list[Column] = field(default_factory=list)
+    columns: List[Column] = field(default_factory=list)
     temp: bool = False
 
     def __post_init__(self):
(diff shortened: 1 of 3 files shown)


tatiana:
@kaxil @dimberman @utkarsharma2 do we want to add to the scope of this ticket to have an Airflow task to clean up temporary tables, using the logic @dimberman  had proposed previously?

Example:
```
def cleanup():
   hook = get_snowflake_hook()
   eng = hook.get_sqlalchemy_engine()
   table_names = [t for t in eng.table_names() if t.startswith("test_dag")]
   queries = [f"DROP TABLE IF EXISTS {t}" for t in table_names]
   hook.run(queries)
```

The implementation is quite simple. The tricky bit may only be to make sure when we test this. We will need to test this feature in a dedicated unique schema - so one test doesn't affect the others. This would work for Postgres and BQ, but probably not for Snowflake, since we don't have enough superpowers there.

src/astro/sql/table.py
```diff
@@ -19,8 +21,8 @@ class Metadata:
     """
 
     # This property is used by several databases, including: Postgres, Snowflake and BigQuery ("namespace")
-    schema: Union[str, None] = None
-    database: Union[str, None] = None
+    schema: str | None = None
```
kaxil, line 24: We should use `Optional[str]` instead

kaxil: I will follow it up with a PR

src/astro/sql/table.py
```diff
@@ -49,22 +53,26 @@ def __post_init__(self):
...
             random.choice(string.ascii_lowercase + string.digits)
-            for _ in range(MAX_TABLE_NAME_LENGTH - schema_length)
+            for _ in range(MAX_TABLE_NAME_LENGTH - schema_length - prefix_length)
         )
+        if prefix:
+            unique_id = f"{prefix}{unique_id}"
+
         return unique_id
 
-    # Due to some weird issue in python typing, I can not actually record that this function returns a table
```
kaxil, line 64: Fixed it in https://github.com/astronomer/astro-sdk/pull/384/commits/13b68de9ccbe00ffebd3ed64b256c03bce48741a , it is because the return type is actually the part of the Class definition itself.

>When a type hint contains names that have not been defined yet, that definition may be expressed as a string literal, to be resolved later.

Reference: https://peps.python.org/pep-0484/#forward-references

kaxil (approved):

src/astro/sql/table.py
tatiana: @kaxi I understand the problem, and I solved it by importing `from __future__ import annotations`, which will be the default behavior starting in Python 3.11. The downside with this approach was that Black automatically converted several types of annotations.

src/astro/sql/table.py
tatiana: This was changed by `Black` as part of our pre-commits, as explained here:
https://github.com/astronomer/astro-sdk/pull/384#discussion_r878956136

src/astro/sql/table.py
tatiana: @kaxil I think I know what happened. When I added:
```
 from __future__ import annotations
```
One of the pre-commit checks said that the imports from typing were no longer necessary. 

When I removed those imports, black changed the syntax for type checking.

PR #391, opened 2022-05-22
Refactor dataframe operator

dimberman:
Piggybacking on the recent fixes to the sql decorator, we can now simplify the SQL dataframe decorator to only inherit from DecoratedOperator and take advantage of functions created in the Database class.

This refactor will further decouple our classes, as well as separate out core functionality into functions that are easier to test with unit tests.


src/astro/sql/operators/dataframe.py
```diff
@@ -39,35 +35,37 @@ def load_op_arg_table_into_dataframe(
...
-        k: _get_dataframe(v)
+        k: _get_dataframe(v, identifiers_as_lower=identifiers_as_lower)
         if param_types.get(k).annotation is pd.DataFrame and isinstance(v, Table)  # type: ignore
         else v
         for k, v in op_kwargs.items()
     }
 
 
-class SqlDataframeOperator(DecoratedOperator, TableHandler):
+class SqlDataframeOperator(DecoratedOperator):
```
kaxil, line 62: Should we call this `DataframeOperator` similar to `TransformOperator`

src/astro/sql/operators/dataframe.py
```diff
@@ -39,35 +35,37 @@ def load_op_arg_table_into_dataframe(
...
     def __init__(
         self,
         conn_id: Optional[str] = None,
         database: Optional[str] = None,
         schema: Optional[str] = None,
-        identifiers_as_lower: Optional[bool] = True,
+        identifiers_as_lower: bool = True,
```
kaxil, line 68: Can we add a docstring for `identifiers_as_lower` after  L80

src/astro/sql/operators/dataframe.py
```diff
@@ -81,7 +79,7 @@ def __init__(
         :param warehouse: (Snowflake) Which warehouse to use for the input table
         :param kwargs:
         """
-        self.conn_id = conn_id
+        self.conn_id: str = conn_id or ""
```
kaxil, line 82: Why do we need an empty string?

src/astro/sql/operators/dataframe.py
```diff
@@ -91,70 +89,41 @@ def __init__(
...
+            self.op_kwargs, self.python_callable, self.identifiers_as_lower
+        )
 
         pandas_dataframe = self.python_callable(*self.op_args, **self.op_kwargs)
         if self.output_table:
-            self.populate_output_table()
-            self.output_table.metadata.schema = (  # type: ignore
-                self.output_table.metadata.schema or SCHEMA
+            self.output_table.conn_id = self.output_table.conn_id or self.conn_id
+            db = create_database(self.output_table.conn_id or self.conn_id)
```
kaxil, line 120:
```suggestion
            db = create_database(self.output_table.conn_id)
```

since we already handle the "or" condition on L119

src/astro/sql/operators/dataframe.py
```diff
@@ -81,7 +79,7 @@ def __init__(
         :param warehouse: (Snowflake) Which warehouse to use for the input table
         :param kwargs:
         """
```
kaxil, line 81: We should also add what this operator returns (df or table) so users know what to expect. And this will be helpful when the Sphinx docs are published

src/astro/sql/operators/dataframe.py
```diff
@@ -91,70 +89,41 @@ def __init__(
...
-            self.populate_output_table()
-            self.output_table.metadata.schema = (  # type: ignore
-                self.output_table.metadata.schema or SCHEMA
+            self.output_table.conn_id = self.output_table.conn_id or self.conn_id
+            db = create_database(self.output_table.conn_id or self.conn_id)
+            self.output_table = db.populate_table_metadata(self.output_table)
+            db.load_pandas_dataframe_to_table(
+                source_dataframe=pandas_dataframe,
+                target_table=self.output_table,
+                if_exists="replace",
```
kaxil, line 125: Should we expose both the allowed strategies: `["replace", "append"]` on the Operator itself?

src/astro/sql/operators/dataframe.py
```diff
@@ -91,70 +89,41 @@ def __init__(
...
-    def handle_op_kwargs(self):
-        param_types = inspect.signature(self.python_callable).parameters
-        self.op_kwargs = {
-            k: _get_dataframe(v, identifiers_as_lower=self.identifiers_as_lower)
-            if param_types.get(k).annotation == pd.DataFrame and type(v) == Table
-            else v
-            for k, v in self.op_kwargs.items()
-        }
-
     def execute(self, context: Dict):  # skipcq: PYL-W0613
```
kaxil, line 99:
```suggestion
    def execute(self, context: Dict) -> Union[Table, pd.DataFrame]:
```

src/astro/sql/operators/dataframe.py
```diff
@@ -91,70 +89,41 @@ def __init__(
...
-                conn_id=self.output_table.conn_id,
-                database=self.output_table.metadata.database,
-                schema=self.output_table.metadata.schema,
-            )
-            load_dataframe_into_sql_table(pandas_dataframe, self.output_table, hook)
             return self.output_table
         else:
             return pandas_dataframe
-
-    def get_snow_hook(self, table: Table) -> SnowflakeHook:
```
kaxil, line 148: Love all the deleted code ❤️

src/astro/utils/first_table.py
```diff
@@ -0,0 +1,96 @@
+import inspect
```
kaxil, line 1: Can we rename the module name to `src/astro/utils/first_table.py` -> `src/astro/utils/table.py`.

That way we have all the utils that are shared between dataframe and transform in a single file. Ideally would love to not have utils 😆 but looks like this needs to be there for now, atleast.

src/astro/sql/operators/dataframe.py
dimberman: Let's see if that's a user request.

kaxil (approved):

kaxil:
@dimberman When you merge the PR can you clear up the description, otherwise the commit description for the merge commit has lot of "comments" from all the commits on the PR.

Example, check the one that was merged: https://github.com/astronomer/astro-sdk/commit/25616bbc88c99921a3f1c8ef1595e9edc01a9078 

image

vs This one: https://github.com/apache/airflow/commit/73b9163a8f55ce3d5bf6aec0a558952c27dd1b55

Issue #413, opened 2022-05-26
Table unable to have Airflow templated names


tatiana:
**Describe the bug**

@guohui-gao reported this issue. 

Given a DAG which
* uses one of the SQL decorators (`transform`, `run_raw_sql` or `dataframe`)
* one of the arguments of this SQL decorator is a `Table` that has a name using an Airflow template (e.g. `{{ ds }}`)

The user will get errors related to that table having an invalid identifier.

Example:

```
    tickets_daily_table = Table(name="TICKETS_{{ts_nodash}}", metadata=transform_meta)
    create_daily_table = create_daily_table(tickets_daily_table)
```

**Version**
* Astro: 0.9.0
* Databases: all

**To Reproduce**

1. Checkout the branch `fix-templated-table-name` (f02a0693c512d8ba6168b7bf131dae99541097b5) with a test that shows this issue
```
git checkout fix-templated-table-name
```

2. Run the test
```
$ PYTHONPATH=`pwd`/src pytest -k test_transform_with_templated_table_name
```

3. See the error
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "-": syntax error
[SQL: DROP TABLE IF EXISTS test-is-{{ test_mode }}]

```

**Expected behavior**

Users should be able to use Airflow templates in table names in `astro-sdk-python` SQL decorators.

PR #417, opened 2022-05-28
Enable table to have Airflow templated names

sunank200:
- Add template field - `name` to `Table` class.
- Render template for all the Tables in op_args and op_kwargs.

closes: #413


tatiana (approved):
That's super neat, @sunank200 ! Thanks a lot :tada:

PR #420, opened 2022-05-31
add repr to all classes #283

### LINKED ISSUE #283: Add `__repr__` to all our classes
**Reference**:
- https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

Currently, without `__repr__`, if you create an instance of a Database class and print it out, it will give the memory address as shown below.
```python
>>> from astro.databases.snowflake import SnowflakeDatabase
>>> SnowflakeDatabase("d")
<astro.databases.snowflake.SnowflakeDatabase object at 0x7ff3c80afe80>
```

As part of this GitHub issue, we should evaluate all of custom classes (`FileType`, `FileLocation` etc) and add `__repr__` methods wherever they are missing.

Example:

https://github.com/astronomer/astro-sdk/blob/5a511d265c6aed10248a1e36fd986518eaec0875/src/astro/files/base.py#L87-L93

dimberman:
**Reference**:
- https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

Currently, without `__repr__`, if you create an instance of a Database class and print it out, it will give the memory address as shown below.
```python
>>> from astro.databases.snowflake import SnowflakeDatabase
>>> SnowflakeDatabase("d")
<astro.databases.snowflake.SnowflakeDatabase object at 0x7ff3c80afe80>
```

As part of this GitHub issue, we should evaluate all of custom classes (`FileType`, `FileLocation` etc) and add `__repr__` methods wherever they are missing.

Example:

https://github.com/astronomer/astro-sdk/blob/5a511d265c6aed10248a1e36fd986518eaec0875/src/astro/files/base.py#L87-L93

closes #283

### DIFF
--- a/src/astro/files/types/base.py
+++ b/src/astro/files/types/base.py
@@ -38,3 +38,6 @@ class FileType(ABC):
     def __str__(self):
         """String representation of type"""
         return self.name.value
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}(path="{self.path}",conn_id="{self.normalize_config}")'
--- a/src/astro/files/types/base.py
+++ b/src/astro/files/types/base.py
@@ -40,4 +40,4 @@ class FileType(ABC):
         return self.name.value
 
     def __repr__(self):
-        return f'{self.__class__.__name__}(path="{self.path}",conn_id="{self.normalize_config}")'
+        return f'{self.__class__.__name__}(path="{self.path}")'
--- a/src/astro/sql/table.py
+++ b/src/astro/sql/table.py
@@ -23,9 +23,6 @@ class Metadata:
     schema: Optional[str] = None
     database: Optional[str] = None
 
-    def __repr__(self):
-        return f'{self.__class__.__name__}(schema="{self.schema}", database="{self.database}")'
-
     def is_empty(self):
         """Check if all the fields are None."""
         values = [getattr(self, field.name) for field in fields(self)]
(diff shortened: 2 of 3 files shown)


src/astro/sql/table.py
```diff
@@ -23,6 +23,9 @@ class Metadata:
     schema: Optional[str] = None
     database: Optional[str] = None
 
+    def __repr__(self):
```
kaxil, line 26: Since `Metadata` and `Table` are using `@dataclass` they don't need `__repr__`:

https://docs.python.org/3/library/dataclasses.html

kaxil (commented):
Can you link the Github issue in the PR description so that the issue is closed when the PR is merged?

Check https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

kaxil (commented):
Does `FileType`, `FileLocation` not need `__repr__` as the PR description says that you'll evaluate those custom classes?

src/astro/files/types/base.py
```diff
@@ -38,3 +38,6 @@ def name(self):
     def __str__(self):
         """String representation of type"""
         return self.name.value
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}(path="{self.path}",conn_id="{self.normalize_config}")'
```
kaxil, line 43: `normalize_config` is optional so just path would be good

```suggestion
        return f'{self.__class__.__name__}(path="{self.path}")'
```

kaxil (approved):

Issue #297, opened 2022-04-07
 Refactor tests so they tear down all the tables they create in Snowflake and BQ #176 


tatiana:
**Context**
Recently we realized that the CI Snowflake schema had over 3k tables created by tests that were not automatically deleted.

We identified and deleted those by running:
```
select 'drop table ' || table_schema || '.' || table_name || ';' QUERY
from information_schema.tables  where table_schema = 'ASTROFLOW_CI' and not table_name like 'ADOPTION%' and not table_name = 'homes_reporting';
```

We also manually cleaned the BQ namespace.

Many of these tables may be the result of legacy tests, which were not correctly dropping (there was a recent refactor to most integration tests)

**To Reproduce**

1. Check how many tables are in the Snowflake `AIRFLOW_CI` namespace by running:
```
select count(*) from information_schema.tables;
```

2. Run the tests in the CI

3. Check how many tables are left after the tests finish (repeating step 1)

At the moment, this difference is bigger than 0.

**Expected behavior**

* The tests should not create tables without tearing them down.

PR #421, opened 2022-05-31
Refactor Integration tests to use a single endpoint (precursor to #297)

dimberman:
We recently needed to clean up 3k tables on our snowflake instance and want to make sure that we don't need to do this again. To fix this we will make sure that all tests use a single table creation mechanism and that this mechanism deletes any tables created. 

While this issue is linked to #297 , I would not say that it closes it. Rather it's a necessary first step to simplify our testing architecture to make it easier to find all points that create tables.


tatiana (approved):
Wow, that's massive, @dimberman ! Thank you very much for unifying the fixtures! :clap: :clap: :clap: 
Please, feel free to merge when the checks are passing!

Issue #415, opened 2022-05-26
Fix or remove "tests/integration_test_dag.py"


kaxil:
We have https://github.com/astronomer/astro-sdk/blob/main/tests/integration_test_dag.py file which is not being used in our test suite. I identified it when I passed it explicitly to pytest for https://github.com/astronomer/astro-sdk/pull/414/commits/42bcaefaece5ea1ce56b5766c501fe2d077fab26

If the tests in this file are already covered by other tests we have we should remove it.

The test is failing with the following error:

```

Location: US
Job ID: 269b39af-58b6-4ff0-afa5-ecaf6b672ef6

[SQL: ALTER TABLE astroflow_ci._tmp_ti2fyeniqgk86m8alla51zvx6oiqe2q6nsquwkkmu6vlg ADD CONSTRAINT airflow UNIQUE (list,sell)]
(Background on this error at: http://sqlalche.me/e/14/4xp6).
Traceback (most recent call last):
  File "/home/runner/work/astro-sdk/astro-sdk/.nox/test-3-8-airflow-2-2-5/lib/python3.8/site-packages/google/cloud/bigquery/dbapi/cursor.py", line 203, in _execute
    self._query_job.result()
  File "/home/runner/work/astro-sdk/astro-sdk/.nox/test-3-8-airflow-2-2-5/lib/python3.8/site-packages/google/cloud/bigquery/job/query.py", line 1499, in result
    do_get_result()
  File "/home/runner/work/astro-sdk/astro-sdk/.nox/test-3-8-airflow-2-2-5/lib/python3.8/site-packages/google/api_core/retry.py", line 283, in retry_wrapped_func
    return retry_target(
  File "/home/runner/work/astro-sdk/astro-sdk/.nox/test-3-8-airflow-2-2-5/lib/python3.8/site-packages/google/api_core/retry.py", line 190, in retry_target
    return target()
  File "/home/runner/work/astro-sdk/astro-sdk/.nox/test-3-8-airflow-2-2-5/lib/python3.8/site-packages/google/cloud/bigquery/job/query.py", line 1489, in do_get_result
    super(QueryJob, self).result(retry=retry, timeout=timeout)
  File "/home/runner/work/astro-sdk/astro-sdk/.nox/test-3-8-airflow-2-2-5/lib/python3.8/site-packages/google/cloud/bigquery/job/base.py", line 728, in result
    return super(_AsyncJob, self).result(timeout=timeout, **kwargs)
[...]

PR #423, opened 2022-06-02
Fix the test for BigQuery with constraint

sunank200:
- `RETURN` in case of BigQuery without any constraints. The only constraint that BigQuery supports is `NOT NULL`.

closes: #415

### DIFF
--- a/tests/integration_test_dag.py
+++ b/tests/integration_test_dag.py
@@ -7,6 +7,7 @@ from airflow.utils import timezone
 
 from astro import sql as aql
 from astro.databases import create_database
+from astro.databases.google.bigquery import BigqueryDatabase
 from astro.databases.sqlite import SqliteDatabase
 from astro.files import File
 from astro.sql.table import Table
@@ -74,6 +75,8 @@ def add_constraint(table: Table):
     db = create_database(table.conn_id)
     if isinstance(db, SqliteDatabase):
         return "CREATE UNIQUE INDEX unique_index ON {{table}}(list,sell)"
+    if isinstance(db, BigqueryDatabase):
+        return "RETURN"
     return "ALTER TABLE {{table}} ADD CONSTRAINT airflow UNIQUE (list,sell)"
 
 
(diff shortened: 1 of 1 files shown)


src/astro/sql/operators/merge.py
```diff
@@ -37,7 +36,7 @@ def __init__(
         self.target_conflict_columns = target_conflict_columns
         self.source_to_target_columns_map = source_to_target_columns_map
         self.if_conflicts = if_conflicts
-        task_id = task_id or get_unique_task_id(self.target_table.name + "_merge")
+        task_id = task_id or "merge"
```
tatiana, line 39: What would happen if this task is run multiple times without defining the `task_id`? Does `task_id` need to be unique?

sunank200: I have added `get_unique_task_id("_merge")` to the `task_id` in the [commit](https://github.com/astronomer/astro-sdk/pull/423/commits/605f7e9a8703ec179d296a2e07a4f8cfc4d61cb7).

tests/integration_test_dag.py
```diff
@@ -60,18 +63,22 @@ def add_constraint(table: Table):
     db = create_database(table.conn_id)
     if isinstance(db, SqliteDatabase):
         return "CREATE UNIQUE INDEX unique_index ON {{table}}(list,sell)"
+    # The only constraint that BigQuery supports is NOT NULL
+    # Hence, skipping the constraint for Bigquery.
+    if isinstance(db, BigqueryDatabase):
+        return "RETURN"
```
tatiana, line 69: @sunank200 what do you think if the `Database` classes exposed a method `setup_append`, which would:
* PostgresDatabase & Snowflake: create a constraint
* BQ: Do nothing
* SQLite: Create the index

This would allow us to not have to handle database-specific logic in this method, keeping it agnostic to database

tatiana: Sorry, @sunank200 , I meant `setup_merge` - they are needed for the merge operation, not append

sunank200: sure. I have named it to merge only.

tatiana (commented):
Looks great, thanks, @sunank200 ! Thank you very much!

tatiana (approved):

PR #424, opened 2022-06-02
fixes to make benchmark tests work

dimberman:
adjusts benchmark tests to the new API so we can start optimizing load_file and export_file


tatiana (approved):
It looks great, @dimberman ! Do we need any other adjustments or this change is enough to run the benchmark?

Issue #400, opened 2022-05-24
Split Transform Operator into two


kaxil:
Currently, the `TransformOperator` does two things:

1) Runs raw SQL query (and handles the result of the output) and returns the result.
2) Runs SQL and handles schema and table (drops table if it exists and create schema if needed) and returns a Table.

Tatiana summed up well:

**Behavior 1**

Input:
- table(s)
- SQL statement

Output:
- table

**Behavior 2**

Input:
- table(s)
- SQL Statement
- function

Output:
- the result of running SQL statement + applying function (this can potentially trash the xcom database depending on the backend used)

_Originally posted by @tatiana in https://github.com/astronomer/astro-sdk/pull/353#discussion_r876095469_

PR #425, opened 2022-06-06
Split Transform and RawSQL decorators

tatiana:
Closes: #400


src/astro/sql/__init__.py
```diff
@@ -40,16 +47,19 @@ def run_raw_sql(
...
-):
-    return transform_decorator(
+    **kwargs,
+) -> TaskDecorator:
+
+    kwargs["conn_id"] = conn_id
+    kwargs["parameters"] = parameters
+    kwargs["database"] = database
+    kwargs["schema"] = schema
+    kwargs["handler"] = handler
```
kaxil, line 57: wdyt about the following?
```suggestion
    kwargs.update({
        "conn_id": conn_id,
        "parameters": parameters,
        "database": database,
        "schema": schema,
        "handler": handler,
    })
```

kaxil, line 57: wdyt about the following?
```suggestion
    kwargs.update({
        "conn_id": conn_id,
        "parameters": parameters,
        "database": database,
        "schema": schema,
        "handler": handler,
    })
```

src/astro/sql/operators/base.py
```diff
@@ -0,0 +1,222 @@
...
+        super().__init__(
+            **kwargs,
+        )
+
+    def execute(self, context: Dict):
+        first_table = find_first_table(
+            op_args=self.op_args,  # type: ignore
+            op_kwargs=self.op_kwargs,
+            python_callable=self.python_callable,
+            parameters=self.parameters or {},  # type: ignore
```
kaxil, line 49: is the `or {}` needed? since we already do that on L36

kaxil, line 49: is the `or {}` needed? since we already do that on L36

src/astro/sql/operators/base.py
```diff
@@ -0,0 +1,222 @@
...
+        self.sql = sql
+        self.parameters = parameters or {}
+        self.database = (self.op_kwargs.pop("database", database),)
+        self.schema = (self.op_kwargs.pop("schema", schema),)
+        self.op_args: Dict[str, Union[Table, pd.DataFrame]] = {}
+        super().__init__(
+            **kwargs,
+        )
+
+    def execute(self, context: Dict):
```
kaxil, line 44:
```suggestion
    def execute(self, context: Dict) -> None:
```

src/astro/sql/operators/raw_sql.py
```diff
@@ -0,0 +1,24 @@
...
+class RawSQLOperator(BaseSQLOperator):
+    """
+    Given a SQL statement, (optional) tables and a (optional) function, execute the SQL statement
+    and apply the function to the results, returning the result of the function.
+
+    Disclaimer: this could potentially trash the XCom Database, depending on the XCom backend used
+    and on the SQL statement/function declared by the user.
+    """
+
+    def execute(self, context: Dict):
```
kaxil, line 15: Should we add the return type ?

kaxil (approved):

src/astro/sql/operators/raw_sql.py
tatiana: @kaxil I found this one a bit tricky since it can potentially return anything - depending on the function defined by the user. What do you suggest?

Also, I had mixed feelings if we should by default return None or the results. Any feedback is very much appreciated

tatiana: @kaxil I'm merging this, and we can improve it in a next PR

kaxil: How about just adding `-> Any`

src/astro/sql/operators/base.py
```diff
@@ -0,0 +1,222 @@
...
+    ):
+        self.kwargs = kwargs or {}
+        self.op_kwargs: Dict = self.kwargs.get("op_kwargs") or {}
+        self.output_table: Table = self.op_kwargs.pop("output_table", Table())
+        self.handler = self.op_kwargs.pop("handler", handler)
+        self.conn_id = self.op_kwargs.pop("conn_id", conn_id)
+        self.sql = sql
+        self.parameters = parameters or {}
+        self.database = (self.op_kwargs.pop("database", database),)
+        self.schema = (self.op_kwargs.pop("schema", schema),)
```
kaxil, line 38:
```python
        self.database = self.op_kwargs.pop("database", database)
        self.schema = self.op_kwargs.pop("schema", schema)
```

src/astro/sql/operators/raw_sql.py
tatiana: What if the handler is not provided, @kaxil ? Do you think it would be more intuitive to return None or the result of running the query? I feel the second would be more intuitive for an operator called `run_raw_sql` - IMO, we should not enforce users to define the handler to retrieve results.

The tricky part is that there may be too many results, and that could thrash the database - but it feels more intuitive

kaxil: cc @mikeshwe Something to discuss on tomorrow's call

src/astro/sql/operators/base.py
tatiana: Addressing this in #426 426

src/astro/sql/operators/raw_sql.py
tatiana: BTW: addressing the `Any` comment in #426

Issue #335, opened 2022-05-03
Implement append_table in PostgresDatabase


tatiana:
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

Placeholder: https://github.com/astro-projects/astro/blob/main/src/astro/databases/base.py#L183-L197

**Acceptance criteria**
* [ ] Implement `SqliteDatabase.append` (potentially as part of `BaseDatabase.append`
* [ ] All the quality checks passed
* [ ] All the functions have typehints
* [ ] All the new functions and modules have docstrings
* [ ] The introduced code has 100% of test coverage

PR #439, opened 2022-06-07
Refactor `AppendOperator`

### LINKED ISSUE #343: Refactor the `append/merge` operator
**Describe the context**
This task is part of the [Astro Python SDK refactor]().

**Dependencies**
* Depends on: #340 
* Depends on: #335 
* Depends on: #344 

**Acceptance criteria**
* [ ] Implement the `append` operator using the new `Database` class and using the `create_database(conn_id)` function. It should contain both the behavior of the previously existing `merge` and `append`.

kaxil:
This PR refactors the Append operator to use the new interfaces.

Breaking Changes that I would love to get some feedback on:
- Removed casting functionality
- changed the interface from which uses `source_to_target_columns_map` instead of `columns`:

```python
    def append(
        self,
        main_table: Table,
        columns: List[str],
        casted_columns: dict,
        append_table: Table,
    ):
```

to

```python
def append(
    source_table: Table,
    target_table: Table,
    source_to_target_columns_map: Optional[Dict[str, str]] = None,
    **kwargs,
):
```

This PR intends to keep the same logic as now and just use new interfaces so we can delete all the old files from old interfaces. The scope to whether or not we should have a single function or not is kept for https://github.com/astronomer/astro-sdk/pull/383

closes #343
closes https://github.com/astronomer/astro-sdk/issues/335

Co-Authored-By: Utkarsh Sharma <utkarsharma2@gmail.com>

### DIFF
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -4,10 +4,11 @@ from typing import Dict, List, Optional, Tuple
 import pandas as pd
 import sqlalchemy
 from airflow.hooks.base import BaseHook
+from sqlalchemy import column, insert, select
+from sqlalchemy.sql.schema import Table as SqlaTable
 
 from astro.constants import (
     DEFAULT_CHUNK_SIZE,
-    AppendConflictStrategy,
     ExportExistsStrategy,
     LoadExistStrategy,
     MergeConflictStrategy,
@@ -248,7 +249,6 @@ class BaseDatabase(ABC):
         source_table: Table,
         target_table: Table,
         source_to_target_columns_map: Dict[str, str],
-        if_conflicts: AppendConflictStrategy = "exception",
     ) -> None:
         """
         Append the source table rows into a destination table.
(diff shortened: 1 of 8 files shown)


src/astro/databases/base.py
```diff
@@ -248,18 +250,35 @@ def append_table(
...
-        """
-        raise NotImplementedError
+        :param source_to_target_columns_map: Dict of source_table columns names to target_table columns names
+        """
+        target_table_sqla = self.get_sqla_table(target_table)
+        source_table_sqla = self.get_sqla_table(source_table)
+        if (
+            source_to_target_columns_map is None
+            or len(source_to_target_columns_map) == 0
+        ):
```
tatiana, line 267:
```suggestion
        if not source_to_target_columns:
```

tatiana (commented):
@kaxil I just noticed coverage doesn't seem to be running for this PR - this issue was not introduced in this PR, but we probably want to restore that before merging further PRs.

tests/integration_test_dag.py
```diff
@@ -75,9 +75,9 @@ def run_append(output_specs: List):
     )
 
     aql.append(
-        columns=["sell", "living"],
-        main_table=load_main,
-        append_table=load_append,
+        source_to_target_columns_map={"sell": "sell", "living": "living"},
```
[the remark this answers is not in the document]
kaxil: I agree, it feels tedious and probably most of the times the col names would be the same. Thoughts on the name of the parameter? `source_to_target_columns_map` doesn't seem appropriate , how about just `columns` and we allow passing either a) list or tuples containing col names b) dict containing source to target mapping.

If you agree, should we do the same for merge too?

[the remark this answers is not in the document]
kaxil: Yes I will tackle them separately, thanks.

src/astro/sql/operators/append.py
```diff
@@ -1,110 +1,46 @@
...
     def __init__(
         self,
-        append_table: Table,
-        main_table: Table,
-        columns: Optional[List[str]] = None,
-        casted_columns: Optional[dict] = None,
+        source_table: Table,
+        target_table: Table,
+        source_to_target_columns_map: Optional[Dict[str, str]] = None,
+        task_id: str = "",
```
sunank200, line 26: Is `task_id` required here as this operator is inherited from `BaseOperator`?

sunank200 (commented):
Overall LGTM. But certain tests seem to be failing.

src/astro/databases/base.py
```diff
@@ -248,18 +250,32 @@ def append_table(
...
         :param target_table: Contains the destination table in which the rows will be appended
-        :param source_to_target_columns_map: Dict of target_table columns names to source_table columns names
-        :param if_conflicts: The strategy to be applied if there are conflicts.
-        """
-        raise NotImplementedError
+        :param source_to_target_columns_map: Dict of source_table columns names to target_table columns names
+        """
+        target_table_sqla = self.get_sqla_table(target_table)
+        source_table_sqla = self.get_sqla_table(source_table)
+        if not source_to_target_columns:
```
kaxil, line 264:
```suggestion
        if not source_to_target_columns_map:
```

tatiana (approved):

PR #440, opened 2022-06-07
timebox: cleanup operator

dimberman:


src/astro/sql/__init__.py
```diff
@@ -1,10 +1,16 @@
 from typing import Callable, Dict, Iterable, List, Mapping, Optional, Union
 
 import pandas as pd
-from airflow.decorators.base import TaskDecorator, task_decorator_factory
+
+try:
+    from airflow.decorators.base import TaskDecorator, task_decorator_factory
+except Exception:
```
tatiana, line 7: @dimberman it would be great to use a specific exception here (e.g. `ImportError`)? To use `except Exception` could hide other errors and make it tricky to debug.

tatiana (commented):
Hey @dimberman, I think it is an excellent start on the cleanup journey! Great work!

Giving my two pennies:

(1) It would be great to have the documentation updated to show this feature as part of this PR

(2) My feeling is that if the user does not specify `tables_to_cleanup`, we should clean up all temporary tables by default when using `cleanup`. They can name tables if they don't want them to be deleted.

dimberman:
@tatiana how would we go about setting the inheritance where this could clean up all temp tables even if they're not specified? We would need to wait until those tasks are done to do so. 

With that in mind these are the only ways I can think of doing it:

1. create an `on_success` and `on_failure` callback that can pull in all tables and check if they're temps and then delete them (this is what I proposed earlier).

This solution would look like this

```
from astro.sql import astro_table_cleanup_callback

dag = DAG(on_success_callback=astro_table_cleanup_callback, on_failure_callback=astro_table_cleanup_callback)

with dag:
   ....
```

2. Create a custom DAG object that adds those callbacks automatically (this would feel the cleanest from the user perspective but would involve the most work on our end).

```from astro.dag import DAG```

3. have every function manually set dependencies in the old-school manner, which basically breaks the entire new astro feel

```
with dag:
   homes_data1 = load_file(.... )

    homes_data2 = load_file(... )

    extracted_data = extract_data(...)

    transformed_data = transform_data( ... )

    filtered_data = filter_data(...)

    create_results_table = create_table(conn_id=SNOWFLAKE_CONN_ID)


    # result of append will not be deleted
    record_results = append()
    record_results.set_upstream(create_results_table)

    end = cleanup()
    [homes_data2, homes_data1, record_results, extracted_data, transformed_data] >> end
```
I'm also open to any other ideas but that's my understanding of the problem as it stands.

cc @kaxil

**Review by tatiana — APPROVED** [review]
@dimberman Since we had a timebox for this task, I reckon the current approach is better than none.

After aligning with you and @kaxil, I understand that we should avoid (1) because callback logs are not readily viewable to users (they are in the DAG parsing component).

Regarding suggestion (2), would there be a way of accomplishing this without relying on callbacks?

Lastly, if (2) isn't an option, we can have a future ticket to support (3) to complement the implementation of this PR.

Unfortunately, Airflow listeners don't have DAG-level events: https://airflow.apache.org/docs/apache-airflow/stable/listeners.html. That would be an excellent use case for them.

Since we had a timebox for this task, I'm happy for this PR to be merged once the checks pass.

File: tests/integration_test_dag.py [hunk 2]
```diff
@@ -130,4 +130,5 @@ def test_full_dag(database_table_fixture, sample_dag):
             output_file=File(path="/tmp/out_agg.csv"),
             if_exists="replace",
         )
+        aql.cleanup()
```
tatiana, line 133: @dimberman what does this look like in the DAG UI? Do we see the task & all the tasks on which it depends on?

dimberman: @tatiana it looks like this. The task will run in parallel and had no "airflow based" dependencies.
Screen Shot 2022-06-16 at 10 25 18 AM

src/astro/sql/operators/cleanup.py
```diff
@@ -0,0 +1,136 @@
...
+def resolve_tables_from_tasks(tasks, context):
+    """
+    For the moment, these are the only two classes that create temporary tables.
+    This function allows us to only resolve xcom for those objects (to reduce how much data is brought into the worker).
+
+    We also process these values one at a time so the system can garbage collect non-table objects (otherwise we might
+    run into a situation where we pull in a bunch of dataframes and overwhelm the worker).
+    :param tasks:
+    :param context:
+    :return:
```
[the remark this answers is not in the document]
kaxil: yes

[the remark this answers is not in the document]
kaxil: `:param tasks:` says nothing about `tasks` param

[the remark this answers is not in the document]
kaxil: yes

src/astro/sql/operators/cleanup.py
```diff
@@ -0,0 +1,164 @@
...
+
+    template_fields = ("tables_to_cleanup",)
+
+    def __init__(
+        self,
+        *,
+        tables_to_cleanup: Optional[List[Table]] = None,
+        task_id: str = "",
+        run_sync_mode: bool = False,
+        single_worker_mode: bool = False,
```
kaxil, line 63: Why do we need a param for single worker mode?

kaxil (commented):
PR title needs updating too -- as the PR title will become the commit message subject

tests/sql/operators/test_cleanup.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "database_table_fixture",
+    SUPPORTED_DATABASES_WITH_FILE,
+    indirect=True,
+    ids=["sqlite", "postgres", "bigquery", "snowflake"],
+)
+def test_cleanup_non_table(database_table_fixture):
```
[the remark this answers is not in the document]
dimberman: @kaxil that's fair. I've change the test to only run against sqlite (As that's essentially the same as mocking since it's all local/runs super quickly).

src/astro/sql/operators/cleanup.py
```diff
@@ -0,0 +1,164 @@
...
+            if not self.run_immediately:
+                self.wait_for_dag_to_finish(context)
+            self.tables_to_cleanup = self.get_all_task_outputs(context=context)
+        self.log.info(
+            "Tables found for cleanup: %s",
+            ",".join([t.name for t in self.tables_to_cleanup]),
+        )
+        temp_tables = filter_for_temp_tables(self.tables_to_cleanup)
+        for table in temp_tables:
+            self.drop_table(table)
```
utkarsharma2, line 87: @dimberman  I think we should make this fault-tolerant if there is an error due to network fault, the subsequent tables deletion will be skipped.

dimberman: @utkarsharma2 I don't think we need to worry about that as there are two incoming fixes that will make that irrelevant.

1. @sunank200 is going to make this operator asynchronous
2. In airflow 2.4 we will have DAG-level teardown operations so this won't even be necessary anymore.

Issue #284, opened 2022-04-04
Check that `enable_xcom_pickling` is set to `True`


kaxil:
The Python SDK relies on XCom pickling (`export AIRFLOW__CORE__ENABLE_XCOM_PICKLING=True`). However, we don't check for it explicitly in the library which causes a failure as by default it is not enabled. For new users, the stacktrace isn't that helpful, so we should check for this config to be enabled before running any tasks

PR #446, opened 2022-06-08
Check that enable_xcom_pickling is set to True

sunank200:
- Check if `AIRFLOW__CORE__ENABLE_XCOM_PICKLING` is set to `True` in environment variable or  `enable_xcom_pickling=true` in airflow.cfg else raise exception in `__init__()`.
- Test for checking enable_xcom_pickling

closes: #284


tests/test_constants.py
```diff
@@ -18,3 +24,19 @@ def test_supported_file_types():
 def test_supported_databases():
     expected = ["bigquery", "postgres", "snowflake", "sqlite"]
     assert sorted(SUPPORTED_DATABASES) == expected
+
+
+@mock.patch.dict(
+    os.environ, {"AIRFLOW__CORE__ENABLE_XCOM_PICKLING": "False"}, clear=True
+)
+def test_enable_xcom_pickling_set_false_in_env():
+    import astro
```
tatiana, line 33: Perhaps this could be on top of the file, WDYT?

src/astro/__init__.py
```diff
@@ -30,3 +34,11 @@ def get_provider_info() -> dict:
         "hook-class-names": [],
         "extra-links": [],
     }
+
+
+if not conf.getboolean(section="core", key="enable_xcom_pickling"):
+    raise OSError(
+    raise OSError(
+        "AIRFLOW__CORE__ENABLE_XCOM_PICKLING environment variable needs to be set to True or enable_xcom_pickling=true "
+        "in airflow.cfg before importing astro-sdk-python."
```
tatiana, line 43: It may be worth confirming with @mikeshwe this message..! Due to timezone, I think we can merge the PR as it is, and improve it after his feedback

tests/test_constants.py
sunank200: The reason we need that import was just for the tests. That was the reason I kept it inside tests

src/astro/__init__.py
kaxil: There are also multiple ways of setting configuration including secrets backend as documented in https://airflow.apache.org/docs/apache-airflow/stable/howto/set-config.html

So probably just worth simplifying that set `enable_xcom_pickling` airflow config to true. And if you'd like to be more helpful to user, then say " One of the simplest ways is by adding the environment variable AIRFLOW__CORE__ENABLE_XCOM_PICKLING=True`

PR #447, opened 2022-06-09
POC of creating a table cleanup task

dimberman:
This POC proves that we can extend the Airflow DAG object to create a
task that lives downstream of all other tasks, and that said task can
access the output of all previous tasks. We can later further optimize
this to ensure that it only ever pulls table objects (based on the task
type).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -68,6 +68,19 @@ def sample_dag():
 
 
 @pytest.fixture
+def sample_astro_dag():
+    from astro.dag import DAG as AstroDAG
+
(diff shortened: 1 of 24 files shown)


kaxil (commented):
I am in favor of https://github.com/astronomer/astro-sdk/pull/440 instead of this PR.

Some of my reasons:
- The Graph View becomes difficult to view for larger DAGs since all the tasks will have a dependency line to the last task
- Users don't get the flexibility of when to delete a table. If a DAG has lots of branches or even have a lot of parallel tasks, they wouldn't want to wait until the end of the task and would rather want to clean up the table after a set of Task (probably within a TaskGroup)

Issue #286, opened 2022-04-04
Refactor methods and functions to decrease Code Complexity


kaxil:
- Split bigger functions and methods into smaller chunks. This improves readability as well as makes it easy to add unit tests for them.
- Mark internal functions/emthods by adding `_` as the prefix

This can easily be verified by Flake8's McCabe complexity score. Currently is it **12** for the repo:
https://github.com/astro-projects/astro/blob/a8982c15fce7f1d4e460c8045b91f33ab7a5eed6/.flake8#L3

This feature is quite useful to detect over-complex code. According to McCabe, anything that goes beyond 10 is too complex. See https://en.wikipedia.org/wiki/Cyclomatic_complexity.

**Acceptance Criteria**:
- A Complexity score of <10 for Flake8.
- Single responsibility principle

PR #448, opened 2022-06-09
Enforce flake8 complexity of 6

kaxil:
Before the refactors, our Code Complexity was 12. After the refactor and after changes in this PR our Code Complexity is 6 :tada:

This PR enforces the code complexity via flake8 so we can ensure it doesn't increases in future.

closes https://github.com/astronomer/astro-sdk/issues/286

### DIFF
--- a/tests/benchmark/run.py
+++ b/tests/benchmark/run.py
@@ -25,7 +25,7 @@ def subtract(after_dict, before_dict):
     return {key: after_dict[key] - before_dict.get(key, 0) for key in after_dict}
 
 
-def profile(func, *args, **kwargs):
+def profile(func, *args, **kwargs):  # noqa: C901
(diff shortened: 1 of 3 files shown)


tests/benchmark/run.py
```diff
@@ -25,7 +25,7 @@ def subtract(after_dict, before_dict):
     return {key: after_dict[key] - before_dict.get(key, 0) for key in after_dict}
 
 
-def profile(func, *args, **kwargs):
+def profile(func, *args, **kwargs):  # noqa: C901
```
kaxil, line 28: Added an ignore comment here as the Code Complexity here is better for readability, logical flow and its in our benchmark script.

tatiana (approved):
Exciting times! Thanks for setting this up, @kaxil - great to see the improvement

Issue #433, opened 2022-06-06
Define datasets to be used in the benchmark


tatiana:
**Context**

We already have datasets of sizes:
* 45 KB
* 472 KB
* 20 MB
* 279 MB

They are available in GCP GCS.

The goal of this ticket is to have additional datasets, so we can benchmark with larger amounts of data.

**Acceptance criteria**

Using real datasets, create additional datasets for benchmarking, of sizes:
* [ ] 1 GB
* [ ] 5 GB
* [ ] 10 GB

Also:
* [ ] Make them available in GCS (gs://astro-sdk/benchmark/datasets/)
* [ ] Document them in `tests/benchmark`

PR #449, opened 2022-06-10
Document (new) load_file benchmark datasets

tatiana:
Closes: #433

Some comments:
* In the past, we used to download the data to where we were running the benchmarks
* Since we're now using up to 10GB of data, I thought it would be a better idea to have the datasets in cloud storage (GS/S3), and have Astro SDK querying them.
* From a performance perspective, this means we will have to take into account the network latency when analyzing the results. As long as those are consistent across the tests, I believe the analysis won't be compromised.

I created a new bucket for the datasets, in case we want to make it public. Any thoughts?

### DIFF
--- a/tests/benchmark/datasets.md
+++ b/tests/benchmark/datasets.md
@@ -1,0 +1,106 @@
+# Datasets
+
+## Overview
+
+The following datasets are used for running performance benchmarking on Astro SDK `load_file` feature.
+
+All of them are based on real external datasets. For reproducibility of the benchmark tests, they were copied to:
+`gs://astro-sdk/benchmark/`
+
+Within `gs://astro-sdk/benchmark/`, there are two paths:
+* `original`: files with original size
+* `trimmed`: files trimmed to meet the desired sizes (for example, 10 GB)
+
+|            | size   | format | rows      | columns | path to trimmed file(s)                     | description                 |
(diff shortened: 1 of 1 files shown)


tests/benchmark/datasets.md
```diff
@@ -0,0 +1,106 @@
...
+
+Within `gs://astro-sdk/benchmark/`, there are two paths:
+* `original`: files with original size
+* `trimmed`: files trimmed to meet the desired sizes (for example, 10 GB)
+
+|            | size   | format | rows      | columns | path to trimmed file(s)                     | description                 |
+|------------|--------|--------|-----------|---------|---------------------------------------------|-----------------------------|
+| ten_kb     | 10 KB  | csv    | 158       | 8       | covid_overview/covid_overview_10kb.csv      | UK Covid overview sample    |
+| hundred_kb | 100 KB | csv    | 748       | 9       | tate_britain/artist_data_100kb.csv          | Tate Gallery artist sample  |
+| ten_mb     | 10 MB  | csv    | 600,000   | 3       | imdb/title_ratings_10mb.csv                 | IMDB title ratings sample   |
```
kaxil, line 18: Should we include parquet for one of the datasets?

tatiana: Great idea, @kaxil! I'll do it

tatiana: Addressed, @kaxil !

PR #450, opened 2022-06-10
Allow list and tuples as columns names in Append & Merge

kaxil:
addresses https://github.com/astronomer/astro-sdk/pull/439#discussion_r892181788

Tt feels tedious and probably most of the times the col names would be the same. This commit renames `source_to_target_columns_map` param in Append and Merge operator to `columns` and allows passing either a) list or tuples containing col names b) dict containing source to target mapping.

**Before**:

```python
append(
    source_table=filtered_data,
    target_table=Table(name="homes_reporting", conn_id=SNOWFLAKE_CONN_ID),
    source_to_target_columns_map={
        "sell": "sell",
        "list": "list",
        "variable": "variable",
        "value": "value",
    },
)
```

**After**:

```python
append(
    source_table=filtered_data,
    target_table=Table(name="homes_reporting", conn_id=SNOWFLAKE_CONN_ID),
    columns=["sell", "list", "variable", "value"]
)
```

**ToDo**:
- [x] Add unit tests for passing different types and verify that it is converted to dict
- [ ] Probably a separate task to update docs and add a task in example DAGs with one of list and dict as col names before we release 0.10.0 - https://github.com/astronomer/astro-sdk/issues/466

### DIFF
--- a/src/astro/sql/operators/append.py
+++ b/src/astro/sql/operators/append.py
@@ -1,10 +1,12 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from airflow.models.baseoperator import BaseOperator
 
 from astro.databases import create_database
 from astro.sql.table import Table
 from astro.utils.task_id_helper import get_unique_task_id
+
+APPEND_COLUMN_TYPE = Optional[Union[List[str], Tuple[str], Dict[str, str]]]
 
 
 class AppendOperator(BaseOperator):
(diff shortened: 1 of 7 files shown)


src/astro/sql/operators/append.py
```diff
@@ -22,14 +27,19 @@ def __init__(
...
-        source_to_target_columns_map: Optional[Dict[str, str]] = None,
+        columns: APPEND_COLUMN_TYPE = None,
         task_id: str = "",
         **kwargs: Any,
     ) -> None:
         self.source_table = source_table
         self.target_table = target_table
-        self.source_to_target_columns_map = source_to_target_columns_map or {}
-
+        if isinstance(columns, (list, tuple)):
```
utkarsharma2, line 36: Hey @kaxil, nothing major, but can we move lines 36 - 41 to a function and use that function in the append and merge operators just to avoid some code duplication?

utkarsharma2 (approved):

sunank200 (approved):
LGTM

src/astro/sql/operators/append.py
kaxil: In this case, [DAMP](https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests) (“Descriptive and Meaningful Phrases”) approach is better than DRY (“Don’t Repeat Yourself”) as they are hardly 4-5 lines :)

PR #451, opened 2022-06-11
Enable mypy for all files and some more cleanup

kaxil:
- This PR enables mypy for all the files
- Deletes redundant code
- Adds and corrects typehints in some of the missing places


src/astro/databases/postgres.py
```diff
@@ -27,11 +27,11 @@ def __init__(self, conn_id: str = DEFAULT_CONN_ID):
         super().__init__(conn_id)
 
     @property
-    def sql_type(self):
+    def sql_type(self) -> str:
         return "postgresql"
```
utkarsharma2, line 31: @kaxil Shouldn't we use constants here?

utkarsharma2 (approved):

sunank200 (approved):
Overall LGTM

src/astro/databases/postgres.py
sunank200: Enum of types from constants.py?

Issue #434, opened 2022-06-06
Export the metrics to GCS when running the benchmark on the cloud


tatiana:
**Dependencies**
* Depends on: #432 

**Acceptance criteria**
* [ ] Export the following metrics when running the benchmark tool on the cloud to GCS:
  * `total_time`
  * `memory_rss`
  * `memory_shared`
  * `memory_pss`
  * `cpu_time_user`
  * `cpu_time_system`
  * `dataset size`
  * `code github hash`
  
TO DEFINE: dataset & table names

Issue #432, opened 2022-06-06
Change benchmark tool to run in the cloud (GCP)


tatiana:
**Acceptance criteria**
* [ ] Be able to run the benchmark, with different configurations, in the cloud (GCP and GKE)

**Reference**
* https://github.com/astronomer/qa-scenario-dags

PR #457, opened 2022-06-15
Added feature to publish benchmark data to bigquery table.

utkarsharma2:
Closes: #434 

Earlier we were storing benchmark data in `/tmp/<some-file>.ndjson` file locally, but now we need to perform the benchmarks in the CI pipeline and hence we need the data to be persisted. With this PR we are publishing the benchmark data in the Bigquery table - `astronomer-dag-authoring.Benchmark.load_files_to_database`

Publishing is not enabled by default, because developers can run benchmarks locally but this data may not be intended to be published. In order to allow the publishing of data, an environment variable needs to be created `ASTRO_PUBLISH_BENCHMARK_DATA`.

Note - We assume the existence of a table in the Bigquery, and the schema of the table is tightly coupled with profiling data to be stored. If we decide to change the profiling data in the future, we may need to alter the table schema.

### DIFF
--- a/tests/benchmark/run.py
+++ b/tests/benchmark/run.py
@@ -6,9 +6,12 @@ import sys
 import time
 
 import airflow
+import pandas as pd
 import psutil
 from airflow.executors.debug_executor import DebugExecutor
 from airflow.utils import timezone
+
+from astro.databases import create_database
 
 
 def elapsed_since(start):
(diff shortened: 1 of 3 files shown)


tests/benchmark/run.py
```diff
@@ -25,6 +28,80 @@ def subtract(after_dict, before_dict):
     return {key: after_dict[key] - before_dict.get(key, 0) for key in after_dict}
 
 
+def prep_insert_statement(
+    df: pd.DataFrame, table_name="load_files_to_database", schema="Benchmark"
```
sunank200, line 32: What would happen when the user runs the same benchmarking script multiple times? `dag_id` would be same in that case as the row inserted before

sunank200 (commented):
Overall LGTM.

I had a question - Reason to put the benchmark metrics to Big Query is to make it queryable?

utkarsharma2:
@sunank200 - yes, the metrics are more useful to us if we make them queriable, we can later decide to do a weekly comparison on them and alert if they are less than a certain threshold.

tests/benchmark/run.py
utkarsharma2: @sunank200 - I have logged an [issue](https://github.com/astronomer/astro-sdk/issues/460) for this bug and a few other bugs with the script.

tatiana: @sunank200 @utkarsharma2 We're adapting the benchmark to run as K8s jobs, as part of #432  so I believe the `dag_id` won't be an issue (each job will be run in a separate pod) - WDYT?

tatiana (approved):
LGTM, thanks, @utkarsharma2 !

Issue #460, opened 2022-06-16
Issues with benchmarking script


utkarsharma2:
When running benchmarking script there are a few things missing from the script which can lead to unexpected behavior

1. Table should be deleted post dag run.
2. Table names should be unique to support concurrent runs.
     the table names are derived from the path stem right now, which will produce the same table name every time the script runs and may hinder the concurrent runs. 
     `table_name = Path(dataset_path).stem`
3. With the concurrent run we also need unique dag_id, when is currently derived from - 
    `dag_name = f"load_file_{dataset_name}_into_{database_name}"` 
4. The data populated in the table is not the same as a dataset. 
5. Script doesn't work with airflow 2.2.5 and 2.3
   error - Cannot use relative path: `sqlite:///./unittests.db` to connect to sqlite. Please use absolute path such as `sqlite:////tmp/airflow.db`.

**Version**
* Astro: 0.9.2
* OS: Mac Monterey

tatiana:
Hey @utkarsharma2 , as we spoke, I believe this is a very high priority and it justifies bringing it to the current sprint.

The only point which I think won't be an issue is:
```
3. With the concurrent run we also need unique dag_id, when is currently derived from -
dag_name = f"load_file_{dataset_name}_into_{database_name}"
```
Because we'll run the benchmark as K8s jobs, and the dag_id shouldn't conflict. That's an improvement we could make in future.

PR #465, opened 2022-06-16
Fixed issues - 1. Table should be deleted post dag run, 2. Generate u…

utkarsharma2:
Closes: #460 
Fixes 1, 2, and 4th issue with benchmarking script:

When running benchmarking script there are a few things missing from the script which can lead to unexpected behavior

Issues List : 

1. Table should be deleted post dag run. - Fixed
2. Table names should be unique to support concurrent runs.
     the table names are derived from the path stem right now, which will produce the same table name every time the script runs and may hinder the concurrent runs.  -  Fixed
     `table_name = Path(dataset_path).stem`
3. ~~With the concurrent run we also need unique dag_id, when is currently derived from - 
    `dag_name = f"load_file_{dataset_name}_into_{database_name}"`~~ (not valid issue)
4. The data populated in the table is not the same as a dataset.  --  Fixed
5. Script doesn't work with airflow 2.2.5 and 2.3
   error - Cannot use relative path: `sqlite:///./unittests.db` to connect to sqlite. Please use absolute path such as `sqlite:////tmp/airflow.db`.


tests/benchmark/config.json
```diff
@@ -59,6 +59,12 @@
       "rows": 385817,
       "conn_id": "bigquery",
       "file_type": "ndjson"
+    },
+    {
+        "name": "few_kb",
+        "size": "45K",
+        "path": "../data/covid_overview.csv",
+        "rows": 732
```
kaxil, line 67: Does this not need `conn_id` key like other records.

nit: formatting/tabs seems off compared to other records

utkarsharma2: @kaxil - the conn_id is optional for local files.

kaxil (approved):
❤️ PR description

PR #469, opened 2022-06-17
Create an AEP for table cleanup

dimberman:
Create proposal for cleaning up temporary tables using a table function instead of relying on temporary schemas. This AEP should also provide a good template for future users who want to propose features to the astro-sdk.


docs/aep/AEP-2-table-cleanup.md
```diff
@@ -0,0 +1,186 @@
...
+| <!-- -->       |  <!-- -->                                                 |
+| -------------- | --------------------------------------------------------- |
+| **State**      | Accepted  |
+| **Discussion** | https://github.com/astronomer/astro-sdk/issues/436                                      |
+| **Created**    | 2022-06-05                                                |
+
+
+## Motivation
+
+When running `aql.transform` or `aql.load_file` jobs, users will develop temporary tables in their databases every time they run a DAG. These extra tables are currently handled by placing all tables into a “temporary schema”. The thought behind these schemas is that a DBA can easily delete the schema if it starts to get full. This answer is insufficient so a better solution for table cleanup should be enacted.
```
tatiana, line 19:
```suggestion
When running `aql.transform` or `aql.load_file` jobs, the Astro SDK implicitly creates temporary tables in the database every time the DAG runs. These extra tables are currently handled by placing all tables into a “temporary schema”. The thought behind these schemas is that a DBA can easily delete the schema if it starts to get full. This answer is insufficient so a better solution for table cleanup should be enacted.
```

tatiana: @dimberman Could you confirm in the code if we are always storing the tables in the temporary schema, or if there are times when we are actually storing them in the schema set in the connection?
This is the code I believe may be leading to this behaviour: https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/base.py#L148

docs/aep/AEP-2-table-cleanup.md
```diff
@@ -0,0 +1,186 @@
...
+    aql.cleanup(tables_to_cleanup=[my_homes_table, sample_table, df_res])
+```
+Pros:
+* Requires no airflow knowledge
+* Users can control when the temp tables are deleted. They can delete it after the end of a task(s) or at the end of DAG.
+* Easy to debug, as tasks logs are accessible from the webserver as opposed to callbacks
+* Allows using tasks trigger rules which gives control of running cleanup even when certain task fails
+
+Cons:
+* If users add new tasks to the DAG, they need to remember to add it to the list
```
tatiana, line 133: to the cleanup list

docs/aep/AEP-2-table-cleanup.md
```diff
@@ -0,0 +1,186 @@
...
+* Allows using tasks trigger rules which gives control of running cleanup even when certain task fails
+
+Cons:
+* This ruins the feeling of the astro python SDK seeming “new”. We are exiting the functional paradigm and are now back in the land of bitshifting tasks.
+* If users add new tasks to the DAG, they need to remember to add it to the list
+* It might find all tables and delete them before they necessarily SHOULD be deleted if the downstream tasks are not added to the dependencies list
+
+### Option 5: A listener
+
+We looked into using the new airflow listener feature for this feature, but unfortunately found it was not compatible. Even if it was, it would require
```
tatiana, line 163: It may be worth adding a link and explaining why it is not compatible (ATM, in Airflow 2.3, listeners are at a task-level, and not at a DAG level)

tatiana (approved):
Fantastic analysis, @dimberman , thank you very much for documenting it!
There are some minor comments inline.

docs/aep/AEP-2-table-cleanup.md
```diff
@@ -0,0 +1,186 @@
...
+
+Ideally we would like to have the transparency of a task, but without the extra logic or DAG structure complexity.
+
+We believe that we can accomplish this using an asynchronous operator.
+
+Using an asynchronous operator, we can create a task that does not NEED to run downstream of every other task. Instead this task can run in parallel, and only actually perform cleanups once every other task in the DAG has been accounted for (either passing, failing, or being skipped).
+
+![](img/AEP-2/example-with-async-task.png)
+
+This structure will allow the DAG structure to remain relatively unchanged, keep additional code to a minimum for the DAG author AND allow users the ability to control if they only want certain tables deleted (e.g. if you give it a list of tasks it will only delete for those tasks, otherwise it will wait for all tasks to complete).
```
utkarsharma2, line 186: @dimberman How will the _cleanup task know what tables to delete and keep? will it be based on naming convention(_tmp<random_string>) or will it delete all tables?

kaxil (approved):
Love it, very well done, feel free to merge it once Tatiana's comments are addressed

docs/aep/AEP-2-table-cleanup.md
dimberman: @tatiana if the user explicitly sets a schema in the connection then yes that will take precedence. This is by design (As we can't assume the same temp schema name across different databases). This also works perfectly with the cleanup step as once we set this up (and soon have a teardown step in the DAG), we won't need temporary schemas anyways.

PR #471, opened 2022-06-20
Add support to running the benchmark in the cloud 

tatiana:
Fixes: #432

What is missing:
* Fix `make run_job` (running the job in GKE). ATM it outputs:
```
+ python3 -W ignore /opt/app/run.py --dataset=few_kb --database=bigquery --revision 1644c4c --chunk-size=1000000
  File "/opt/app/run.py", line 138, in <module>
    run_dag(
  File "/opt/app/run.py", line 57, in wrapper
    result = func(*args, **kwargs)
  File "/opt/app/run.py", line 96, in run_dag
    dag.run(
  File "/usr/local/lib/python3.9/dist-packages/airflow/models/dag.py", line 2247, in run
    job.run()
  File "/usr/local/lib/python3.9/dist-packages/airflow/jobs/base_job.py", line 244, in run
    self._execute()
  File "/usr/local/lib/python3.9/dist-packages/airflow/utils/session.py", line 71, in wrapper
    return func(*args, session=session, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/airflow/jobs/backfill_job.py", line 860, in _execute
    raise BackfillUnfinished(err, ti_status)
airflow.exceptions.BackfillUnfinished: Some task instances failed:
```

The logs can be seen by running:
```
 kubectl logs benchmark-66v77-8sbp8 -n benchmark
```


tests/benchmark/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM debian:bullseye
```
tatiana, line 1: We can consider using Alpine since it's a more lightweight image - less prone to security issues.

tests/benchmark/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Assert the Astro SDK is available in the PYTHONPATH
+RUN mkdir -p $AIRFLOW_HOME
+WORKDIR $AIRFLOW_HOME
+COPY run.sh $AIRFLOW_HOME/
+COPY run.py $AIRFLOW_HOME/
+COPY config.json $AIRFLOW_HOME/
+COPY dags $AIRFLOW_HOME/
+COPY astro-sdk/src/astro $AIRFLOW_HOME/
+
+CMD ./run.sh
```
tatiana, line 41: ATM the script is running as a super user and we're able to write to the folder of the application itself (`AIRFLOW_HOME`) because it creates `unittests.db`. These tend to not be good practices.

Since we're planning to soon migrate to use Astro cloud - I don't think they are critical, but we can address them as part of the current PR if others feel strongly about it.

tests/benchmark/infrastructure/kubernetes/job.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: batch/v1
+kind: Job
```
kaxil, line 2: Does this container use all the available resources? Worth confirming

tests/benchmark/Dockerfile
dimberman: I think using our AP-airflow image will make it MUCH faster as we won't need to build any of this from scratch.

tests/benchmark/Dockerfile
dimberman: @tatiana I don't feel strongly about this for a benchmark. I would be concerned if this was an image people would use for production, but for testing I think it's fine to keep things simple.

tests/benchmark/Dockerfile
tatiana: Is this an improvement we could do in a separate PR?

PR #476, opened 2022-06-20
Fail `LoadFile` operator when `input_file` does not exist

utkarsharma2:
Closes: #476 
 
**Describe the bug**
When we use the load_file operator and pass a local file and the file doesn't exist, the operator does nothing and fails silently. 

**Version**
* Astro: [e.g. 0.9.2]

**To Reproduce**
1. Use the load_file operator:
    ```
     aql.load_file(  # noqa: F841
            input_file=File(
                path='/tmp/non_existing_file.csv',
            ),
            task_id="load_csv",
            output_table=table_metadata,
            chunk_size=chunk_size,
     )
    ```
2. Now the operator will run without any warning or failure making the user think that the operator ran successfully.

**Expected behavior**
**Case 1 (default)**

> The operator should fail with an error if param `if_file_doesnt_exist`  is set to `exception` saying that the file doesn't exist.

**Case 2**

> if the `if_file_doesnt_exist` is set to `suppress` the operator should not raise an exception.

To accommodate this behaviour added a parameter `if_file_doesnt_exist`  to `load_file()` operator and `get_file()` function.

### DIFF
--- a/src/astro/files/base.py
+++ b/src/astro/files/base.py
@@ -1,9 +1,9 @@
-from typing import Any, List, Optional, Union, cast
+from typing import List, Optional, Union
 
 import pandas as pd
 import smart_open
 
-from astro.constants import FileNotFoundStrategy, FileType
+from astro.constants import FileType
 from astro.files.locations import create_file_location
 from astro.files.types import create_file_type
 
@@ -35,12 +35,12 @@ class File:
         )
 
     @property
-    def path(self) -> Union[List[str], str]:
-        return cast(list, self.location.path)
+    def path(self) -> str:
+        return self.location.path
 
     @property
     def conn_id(self) -> Optional[str]:
-        return cast(str, self.location.conn_id)
+        return self.location.conn_id
 
     @property
     def size(self) -> int:
@@ -89,9 +89,7 @@ class File:
             f'{self.__class__.__name__}(location="{self.location}",type="{self.type}")'
         )
 
-    # To Do: path property can resolve into lists of path(list[str]) or single path(str).
-    # Right now code is assuming value of string which is wrong need to make changes all across code.
-    def __str__(self) -> Any:
+    def __str__(self) -> str:
         return self.location.path
 
 
@@ -100,7 +98,6 @@ def get_files(
     conn_id: Optional[str] = None,
     filetype: Union[FileType, None] = None,
     normalize_config: Optional[dict] = None,
-    if_file_doesnt_exist: FileNotFoundStrategy = "exception",
 ) -> List[File]:
     """get file objects by resolving path_pattern from local/object stores
     path_pattern can be
@@ -112,7 +109,6 @@ def get_files(
     :param conn_id: Airflow connection ID
     :param filetype: constant to provide an explicit file type
     :param normalize_config: parameters in dict format of pandas json_normalize() function
-    :param if_file_doesnt_exist: determines the strategy in case the file path/pattern doesn't result in existing files
     """
     location = create_file_location(path_pattern, conn_id)
     files = [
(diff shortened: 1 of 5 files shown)


src/astro/files/base.py
```diff
@@ -34,13 +34,15 @@ def __init__(
             path=path, filetype=filetype, normalize_config=normalize_config
         )
 
+    # To Do: path property can resolve into lists of path(list[str]) or single path(str).
+    # Right now code is assuming value of string which is wrong need to make changes all across code.
     @property
-    def path(self) -> str:
+    def path(self) -> Any:
```
tatiana, line 40: It may be worth being more specific about this, using `Union` of list of strings and strings

src/astro/sql/operators/load_file.py
```diff
@@ -33,6 +34,7 @@ def __init__(
         chunk_size: int = DEFAULT_CHUNK_SIZE,
         if_exists: LoadExistStrategy = "replace",
         ndjson_normalize_sep: str = "_",
+        if_file_doesnt_exist: FileNotFoundStrategy = "exception",
```
tatiana, line 37: An alternative to `if_file_doesnt_exist` could be `if_inexistent`

src/astro/files/base.py
```diff
@@ -34,13 +34,15 @@ def __init__(
             path=path, filetype=filetype, normalize_config=normalize_config
         )
 
+    # To Do: path property can resolve into lists of path(list[str]) or single path(str).
```
kaxil, line 37:
```suggestion
    # ToDo: path property can resolve into lists of path(list[str]) or single path(str).
```

kaxil: "to do" wouldn't appear as a TODO in an IDE with spaces

src/astro/files/base.py
kaxil: +1

src/astro/files/base.py
```diff
@@ -98,6 +102,7 @@ def get_files(
     conn_id: Optional[str] = None,
     filetype: Union[FileType, None] = None,
     normalize_config: Optional[dict] = None,
+    if_file_doesnt_exist: FileNotFoundStrategy = "exception",
```
kaxil, line 105: Can you think of a use-case where someone would want to suppress the error? It most certainly should be error if the file doesn't exists, as the operator is no-op

Let's simplify this an just error when the file doesn't exist

src/astro/files/base.py
```diff
@@ -34,13 +34,15 @@ def __init__(
...
+    # Right now code is assuming value of string which is wrong need to make changes all across code.
     @property
-    def path(self) -> str:
+    def path(self) -> Any:
         return self.location.path
 
     @property
     def conn_id(self) -> Optional[str]:
-        return self.location.conn_id
+        return cast(str, self.location.conn_id)
```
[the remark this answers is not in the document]
utkarsharma2: Due to mypy, I need to be explicit about the type.

src/astro/files/base.py
utkarsharma2: @kaxil - The initial intent was to cater to cases where we have dags that monitor a folder for the presence of files and process them if they exist. But I think we can achieve that with the file_sensor operator as well.

kaxil: Yeah, that should not be the use case of this operator. We have an async version of FileSensor for that too.

src/astro/files/base.py
utkarsharma2: @tatiana  We need to use typing.cast() for this. Updated the code.

kaxil (approved):
I have simplified the PRin https://github.com/astronomer/astro-sdk/pull/476/commits/9cc0be52ddd71ce7bc6e11c1a02628c0d40dbecb, feel free to create a separate PR if there were other things you wanted to add.

Did it so that we can at least close the loop on this

PR #484, opened 2022-06-22
Create sphinx documentation for readthedocs

dimberman:
# Description
Currently there is no readthedocs site for the Astro SDK. This means that users need to read through our github (which is unindexed) or our codebase to know wh

## What is the new behavior?
Now we can use sphinx-autoapi and sphinx to automatically generate documentation for us based on our function and class comments. While this is not NEARLY sufficient as total documentation, it is still a huge step forward.

No

### Checklist


docs/conf.py
```diff
@@ -10,30 +10,33 @@
...
 
-project = "Astro Projects"
-author = "Astronomer Inc."
+project = "astro-sdk"
+copyright = "2022, Astronomer inc."  # noqa
+author = "Astronomer inc."
 
 # The full version, including alpha/beta/rc tags
-release = "0.6.0"
+release = "0.10.0"
```
tatiana, line 24: Is it possible we could retrieve this from `astro.__init__.__versions__`?

tatiana (approved):
@dimberman it looks great, thanks for setting up things on ReadTheDocs!

As we spoke, my concern was how we're rendering `Literals` ATM - ideally we'd show users a list of the possible values for them, and this is not happening.

I don't know the best solution, but I believe this is important to be done before the 1.0.0 release. I'm happy to have that logged as a separate ticket since this didn't have acceptance criteria - and we achieved the goal: have the docs in readthedocs.

dimberman:
@tatiana let's make that its own ticket as I think that warrants a discussion of a few options.

src/astro/sql/__init__.py
```diff
@@ -31,6 +31,31 @@ def transform(
...
+    you do not want the parameter converted.
+
+    :param python_callable:
+    :param multiple_outputs:
+    :param conn_id:
+    :param parameters:
+    :param database:
+    :param schema:
+    :param handler:
+    :param kwargs:
```
kaxil, line 55: This PR should not have been merged with these changes!

This is how it looks like:

image

kaxil: https://astro-sdk.readthedocs.io/en/create-docs-ci/autoapi/astro/sql/index.html

kaxil: There needs to be a description for it

src/astro/sql/__init__.py
```diff
@@ -59,6 +84,34 @@ def run_raw_sql(
...
+    :return:
```
kaxil, line 113: Same as https://github.com/astronomer/astro-sdk/pull/484/files#r905309852

src/astro/sql/__init__.py
```diff
@@ -59,6 +84,34 @@ def run_raw_sql(
...
 ) -> TaskDecorator:
+    """
+    Given a python function that returns a SQL statement and (optional) tables, execute the SQL statement and output
+    the result into a SQL table.
+
+    Use this function as a decorator like so:
+
+      @transform
+      def my_sql_statement(table1: Table) -> Table:
+          return "DROP TABLE {{table1}}"
```
kaxil, line 95: We need to use rst code blocks for this to render as code:

image

Example:

https://github.com/apache/airflow/blob/main/airflow/providers/google/cloud/operators/gcs.py#L90-L99

src/astro/sql/__init__.py
```diff
@@ -59,6 +84,34 @@ def run_raw_sql(
...
+
+    In this example, by identifying parameters as `Table` objects, astro knows to automatically convert those
+    objects into tables (if they are, for example, a dataframe). Any type besides table will lead astro to assume
+    you do not want the parameter converted.
+
+    Please note that the `run_raw_sql function will not create a temporary table. It will either return the result of a
```
kaxil, line 101: It is rst so need using double quotes not single for inline code formatting- ``run_raw_sql``

src/astro/sql/__init__.py
```diff
@@ -78,6 +131,18 @@ def run_raw_sql(
...
+    Clean up temporary tables once either the DAG or upstream tasks are done
+
+    The cleanup operator allows for two possible scenarios: Either a user wants to clean up a specific set of tables
+    during the DAG run, or the user wants to ensure that all temporary tables are deleted once the DAG run is finished.
+    The idea here is to ensure that even if a user doesn't have access to a "temp" schema, that astro does not leave
+    hanging tables once execution is done.
+
+    :param tables_to_cleanup: A list of tables to cleanup, defaults to waiting for all upstream tasks to finish
+    :param kwargs:
+    :return:
```
kaxil, line 144: Same as the other comments, missing description!

kaxil:
Would have been better to list the GitHub issue this relates to

src/astro/sql/__init__.py
dimberman: @kaxil based on discussion with tatiana I was given the sense that this PR would be more about creating the infrastructure for a readthedocs site. I think that actually fixing the docs can be handled in subsequent tickets (As there are a lot of places we should be fixing the docs now that we can see them)

PR #485, opened 2022-06-23
Remove need of `single_worker_mode` env var for Cleanup

kaxil:
We can identify `single_worker_mode` by checking if it is using `SequentialExecutor` or `DebugExecutor`.

This PR also does some minor fixes to add missing docstrings


kaxil:
Added a unit test too in https://github.com/astronomer/astro-sdk/pull/485/commits/9d23cf4067eee35738c074a5aafff0591b0a540f -- if you'll want to re-review @dimberman @tatiana @utkarsharma2 @sunank200

sunank200 (approved):
Overall LGTM.

tests/sql/operators/test_cleanup.py
```diff
@@ -204,3 +209,43 @@ def test_is_dag_running():
...
+    dag = DAG(
+        "test_error_raised_with_blocking_op_executors", start_date=datetime(2022, 1, 1)
+    )
+    cleanup_task = CleanupOperator(dag=dag)
+    dr = DagRun(dag_id=dag.dag_id)
+    with mock.patch(
+        "astro.sql.operators.cleanup.CleanupOperator._is_dag_running"
+    ) as mock_is_dag_running, mock.patch.dict(
+        os.environ,
+        {"AIRFLOW__CORE__EXECUTOR": executor},
```
sunank200, line 244: Is this trying to set `env` while running?

tests/sql/operators/test_cleanup.py
```diff
@@ -204,3 +209,43 @@ def test_is_dag_running():
...
+    ):
+        mock_is_dag_running.side_effect = [True, False]
+        if raise_error:
+            with pytest.raises(AirflowException):
+                cleanup_task.execute({"dag_run": dr})
```
sunank200, line 249: It also might be good to assert the exception message.

sunank200 (approved):

dimberman:
@kaxil Are you sure this would work? It seems like if the conf says "LocalExecutor" but you run `airflow test` with a DebugExecutor, then the system might accidentally think you are running a localexecutor

tests/sql/operators/test_cleanup.py
kaxil: Yes

kaxil:
> @kaxil Are you sure this would work? It seems like if the conf says "LocalExecutor" but you run `airflow test` with a DebugExecutor, then the system might accidentally think you are running a localexecutor

Good point, how about now?

https://github.com/astronomer/astro-sdk/blob/372c0409305d3d3dc657b48f4d082bce652f0ae1/src/astro/sql/operators/cleanup.py#L132-L147

Test:

https://github.com/astronomer/astro-sdk/blob/372c0409305d3d3dc657b48f4d082bce652f0ae1/src/astro/sql/operators/cleanup.py#L132-L147

kaxil:
> @kaxil but what about a case where someone does a backfilljob and assigns a LocalExecutor or a CeleryExecutor manually with the `executor` parameter?

The "airflow dags backfill" CLI command doesn't support passing an executor though. I doubt it works with Celery even if you want it via `dag.run()` 🤷 but let me make it more robust

kaxil:
How about now with the changes in https://github.com/astronomer/astro-sdk/pull/485/commits/34048199e0e3c9b3c94588673d36c3669adc3c7f ?

tatiana:
Looks good, @kaxil !

tatiana (approved):

dimberman (approved):

Issue #481, opened 2022-06-22
Add native support for data transfers optimisation


sunank200:
**Please describe the feature you'd like to see**
Currently, we do not support the native transfers between different file locations and databases in `LoadFile` operator even though the external provider library supports the optimised approach to transfer the data. For example, to transfer between S3 to Snowflake we could use [COPY INTO <table>](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html) command to load data from a Cloud Storage bucket into tables that would make the transfer way faster. 

**Describe the solution you'd like**
Me and @utkarsharma2  were discussing various approaches. One of the most optimised approaches looks like using existing airflow hooks to achieve that. For example, [S3ToSnowflakeOperator](https://github.com/apache/airflow/blob/main/airflow/providers/snowflake/transfers/s3_to_snowflake.py#L27) has the whole logic already implemented. I propose using existing transfer operators from airflow providers where ever possible. If that provider doesn't exist we could implement that logic on our end on the respective location classes in astro-sdk-python.

For each database class like [SnowflakeDatabase](https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/snowflake.py#L16), [BigqueryDatabase](https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/google/bigquery.py#L18) we could maintain a list of source storage types from which the optimised transfer is possible. In the `LoadFile` `execute()` method we can do the following:

- Check if the Source file storage(`input_file`) to Destination database(`output_table`) optimised transfer exists or not. (This mapping can be maintained in each database and location classes). For each database class like `SnowflakeDatabase`, `BigqueryDatabase` we could maintain a list of source storage types from which the optimised transfer is possible.

- If optimised transfer exists, fetch the relevant hooks from airflow providers and do the transfer. Another way would be to use the Transfer Operators implemented on airflow directly if it exists. ( Example - [S3ToSnowflakeOperator](https://github.com/apache/airflow/blob/main/airflow/providers/snowflake/transfers/s3_to_snowflake.py#L27))

**Are there any alternatives to this feature?**
Is there another way we could solve this problem or enable this use-case?
Next step would be to use chunk or split the file and do parallel async uploads. But using optimised transfers using airflow hooks would improve the performance drastically.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
[...]

Issue #430, opened 2022-06-06
Support loading datasets of 10GB in Snowflake in less than 5 min


tatiana:
**Dependencies**
* Depends on: #438 
* Depends on: #492
* Depends on: #493
* Depends on: #494

**Acceptance criteria**
* [ ] Make changes to how we load data into Snowflake.
* [ ] Re-run the benchmark and identify the performance improvements (?)
* [ ] The changes must work for all file types supported (CSV, JSON, NDJSON, Parquet)
* [ ] The changes made would also work on S3 and GCS

tatiana:
I believe this was closed by mistake, we haven't merged the changes into master yet..!

PR #487, opened 2022-06-23
Optimize loading datasets for snowflake [WIP]

sunank200:
# Description
## What is the current behaviour?

Currently, we do not support the native transfers between different file locations and databases in the LoadFile operator even though the external provider library supports the optimised approach to transfer the data. For example, to transfer between GCP to Snowflake we could use [COPY INTO](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html)

closes: #430 
closes: #481
closes: #492
closes: #493 
closes: #494
 

## What is the new behaviour?

- Added the interface for check_optimised_path_and_transfer in database.py
- Check if the optimised path exists and then transfer in LoadFile
- Add optimisation for GCS to snowflake for the following cases:
    - If the Table and schemas are specified by the user, run `COPY INTO` to copy data from GCS to snowflake.
    - If the table and schema exist, then create a snowflake stage based on the snowflake integration provided by the user. Generate the query to COPY INTO and run.
    - If the Schema and table don't exist and the file type is PARQUET, use `INFER_SCHEMA` of the snowflake to create the table first. Then run the `COPY INTO` query
    -  If the table and schema do not exist, create an empty table for `CSV` and `NDJSON`. Detect schema using pandas and fetch dataset for chunked size and create table from the file and generate the query to COPY INTO and run.
- Add `optional_args` argument to the `Table` data class to pass additional params like `integration` for snowflake.

Yes

### Checklist

### DIFF
--- a/src/astro/constants.py
+++ b/src/astro/constants.py
@@ -11,6 +11,8 @@ else:
 DEFAULT_CHUNK_SIZE = 1000000
 PYPI_PROJECT_NAME = "astro-sdk-python"
 
+LOAD_COLUMN_AUTO_DETECT_ROWS = 1
+
 
 class FileLocation(Enum):
     LOCAL = "local"
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -242,7 +242,7 @@ class BaseDatabase(ABC):
         Checks if optimised path for transfer between File location to database exists
         and if it does, it transfers it and returns true else false.
         """
-        raise NotImplementedError()
+        return False
 
     def load_pandas_dataframe_to_table(
         self,
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -242,7 +242,7 @@ class BaseDatabase(ABC):
         Checks if optimised path for transfer between File location to database exists
         and if it does, it transfers it and returns true else false.
         """
-        pass
+        raise NotImplementedError()
 
     def load_pandas_dataframe_to_table(
         self,
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -254,6 +254,7 @@ class BaseDatabase(ABC):
         :param if_exists: Update strategy for file
         :param kwargs:
         """
+        # This method can be overridden by each class separately.
         return
 
     def load_pandas_dataframe_to_table(
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -230,19 +230,31 @@ class BaseDatabase(ABC):
             chunk_size,
         )
 
-    def check_optimised_path_and_transfer(
+    def check_optimised_path(self, source_file: File, target_table: Table) -> bool:
+        """
+        Check if there is an optimised path for source to destination.
+        :param source_file: File from which we need to transfer data
+        :param target_table: Table that needs to be populated with file data
+        """
+        return False
+
+    def optimised_transfer(
         self,
         source_file: File,
         target_table: Table,
-        chunk_size: int = DEFAULT_CHUNK_SIZE,
         if_exists: LoadExistStrategy = "replace",
         **kwargs,
-    ) -> bool:
+    ) -> None:
         """
         Checks if optimised path for transfer between File location to database exists
-        and if it does, it transfers it and returns true else false.
+        and if it does, it transfers it.
+
+        :param source_file: Source file
+        :param target_table: Output target table on snowflake
+        :param if_exists: Update strategy for file
+        :param kwargs:
         """
-        return False
+        return
 
     def load_pandas_dataframe_to_table(
         self,
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -231,7 +231,12 @@ class BaseDatabase(ABC):
         )
 
     def check_optimised_path_and_transfer(
-        self, source_file: File, target_table: Table
+        self,
+        source_file: File,
+        target_table: Table,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
     ) -> bool:
         """
         Checks if optimised path for transfer between File location to database exists
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -230,6 +230,15 @@ class BaseDatabase(ABC):
             chunk_size,
         )
 
+    def check_optimised_path_and_transfer(
+        self, source_file: File, target_table: Table
+    ) -> bool:
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        pass
+
     def load_pandas_dataframe_to_table(
         self,
         source_dataframe: pd.DataFrame,
--- a/src/astro/databases/snowflake.py
+++ b/src/astro/databases/snowflake.py
@@ -6,8 +6,14 @@ from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
 from pandas.io.sql import SQLDatabase
 from snowflake.connector import pandas_tools
 
-from astro.constants import DEFAULT_CHUNK_SIZE, LoadExistStrategy, MergeConflictStrategy
+from astro.constants import (
+    DEFAULT_CHUNK_SIZE,
+    FileLocation,
+    LoadExistStrategy,
+    MergeConflictStrategy,
+)
 from astro.databases.base import BaseDatabase
+from astro.files import File
 from astro.sql.table import Metadata, Table
 
 DEFAULT_CONN_ID = SnowflakeHook.default_conn_name
@@ -20,6 +26,7 @@ class SnowflakeDatabase(BaseDatabase):
     """
 
     def __init__(self, conn_id: str = DEFAULT_CONN_ID):
+        self.native_support: Tuple = (FileLocation.GS, FileLocation.S3)
         super().__init__(conn_id)
 
     @property
@@ -56,6 +63,23 @@ class SnowflakeDatabase(BaseDatabase):
         ]
         qualified_name = ".".join(name for name in qualified_name_lists if name)
         return qualified_name
+
+    def check_optimised_path_and_transfer(
+        self, source_file: File, target_table: Table
+    ) -> bool:
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        if FileLocation(source_file) in self.native_support:
+            # TODO:
+            """
+            1. snowflake_hook = self.hook()
+            2. create schema and table if not exists using pandas
+            3. read file and use COPY INTO
+            4. return True
+            """
+        return False
 
     def load_pandas_dataframe_to_table(
         self,
--- a/src/astro/databases/snowflake.py
+++ b/src/astro/databases/snowflake.py
@@ -292,13 +292,15 @@ class SnowflakeDatabase(BaseDatabase):
         )
 
         create_table_from_infer_schema_query = (
-            f"CREATE OR REPLACE TABLE "
-            f"{self.get_table_qualified_name(target_table)} USING TEMPLATE "
-            f"(SELECT ARRAY_AGG(OBJECT_CONSTRUCT(*)) FROM TABLE "
-            f"(INFER_SCHEMA "
-            f"(LOCATION=>'@{stage_name}/{os.path.basename(source_file.path) or ''}',"
-            f" FILE_FORMAT=>'parquet_format')));"
+            "CREATE OR REPLACE TABLE %s USING TEMPLATE (SELECT ARRAY_AGG(OBJECT_CONSTRUCT(*)) FROM TABLE ("
+            "INFER_SCHEMA (LOCATION=>'@%s/%s', FILE_FORMAT=>'parquet_format')));"
+            % (
+                self.get_table_qualified_name(target_table),
+                stage_name,
+                os.path.basename(source_file.path) or "",
+            )
         )
+
         execution_info = snowflake_hook.run(
             create_table_from_infer_schema_query, autocommit=True
         )
--- a/src/astro/databases/snowflake.py
+++ b/src/astro/databases/snowflake.py
@@ -76,17 +76,30 @@ class SnowflakeDatabase(BaseDatabase):
         qualified_name = ".".join(name for name in qualified_name_lists if name)
         return qualified_name
 
-    def check_optimised_path_and_transfer(
+    def check_optimised_path(self, source_file: File, target_table: Table) -> bool:
+        """
+        Check if there is an optimised path for source to destination.
+
+        :param source_file: File from which we need to transfer data
+        :param target_table: Table that needs to be populated with file data
+        """
+        return bool(self.native_support.get(source_file.location.location_type))
+
+    def optimised_transfer(
         self,
         source_file: File,
         target_table: Table,
-        chunk_size: int = DEFAULT_CHUNK_SIZE,
         if_exists: LoadExistStrategy = "replace",
         **kwargs,
-    ) -> bool:
+    ) -> None:
         """
         Checks if optimised path for transfer between File location to database exists
         and if it does, it transfers it and returns true else false.
+
+        :param source_file: Source file
+        :param target_table: Output target table on snowflake
+        :param if_exists: Update strategy for file
+        :param kwargs:
         """
         method_name = self.native_support.get(source_file.location.location_type)
         if method_name:
@@ -95,23 +108,23 @@ class SnowflakeDatabase(BaseDatabase):
                 if transfer_method(
                     source_file=source_file,
                     target_table=target_table,
-                    chunk_size=chunk_size,
                     if_exists=if_exists,
                     **kwargs,
                 ):
-                    return True
+                    return
             except Exception as exec_err:
                 raise exec_err
-        return False
+        return
 
     @staticmethod
     def get_gcs_project_id_from_conn(conn: connection) -> str:
         """
         Get GCS project id from conn
+
         :param conn: Airflow's connection
         """
-        if conn.extra and conn.extra.get("project"):
-            return str(conn.extra.get("project"))
+        if conn.extra and conn.extra_dejson.get("project"):
+            return str(conn.extra_dejson["project"])
         elif conn.host:
             return str(conn.host)
         raise ValueError(f"conn_id {conn.conn_id} has no project id.")
@@ -127,6 +140,7 @@ class SnowflakeDatabase(BaseDatabase):
         Checks if the table and schema exists on Snowflake.
         If table and schema exists, create stage based on integration passed by user.
         Generate query and run COPY INTO command and return True else returns False
+
         @param source_file: Source file
         @param target_table: Output target table on snowflake
         @param chunk_size: Chunk size for the file
@@ -181,6 +195,7 @@ class SnowflakeDatabase(BaseDatabase):
     def create_stage(self, source_file: File, snowflake_hook: SnowflakeHook) -> str:
         """
         Creates a stage name based on source_file path and storage integration provided.
+
         @param source_file: Source file
         @param snowflake_hook: Snowflake hook
         @return: Stage name created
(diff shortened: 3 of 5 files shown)


src/astro/sql/operators/load_file.py
```diff
@@ -80,13 +80,19 @@ def load_data_to_table(self, input_file: File) -> Table:
         for file in get_files(
             input_file.path, input_file.conn_id, normalize_config=self.normalize_config
         ):
-            database.load_pandas_dataframe_to_table(
-                source_dataframe=file.export_to_dataframe(),
+            if not database.check_optimised_path_and_transfer(
```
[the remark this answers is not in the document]
sunank200: Added the change in latest commit

src/astro/databases/base.py
```diff
@@ -230,6 +230,20 @@ def load_file_to_table(
...
+        target_table: Table,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ) -> bool:
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        raise NotImplementedError()
```
utkarsharma2, line 245: Should we raise an exception here? we might have databases that don't have optimized paths to them like SQLite and Postgres.

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        """
+        Checks if the table and schema exists on Snowflake.
+        If table and schema exists, create stage based on integration passed by user.
+        Generate query and run COPY INTO command and return True else returns False
+        @param source_file: Source file
+        @param target_table: Output target table on snowflake
+        @param chunk_size: Chunk size for the file
+        @param if_exists: Update strategy for file
+        @param kwargs:
+        @return:bool
```
kaxil, line 135:
```suggestion
        Generate query and run COPY INTO command and return True else returns False

        :param source_file: Source file
        :param target_table: Output target table on snowflake
        :param chunk_size: Chunk size for the file
        :param if_exists: Update strategy for file
        :param kwargs:
        :return: bool
```

`:return:` should says what is being returned, `:rtype` shows the return type

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ) -> bool:
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        method_name = self.native_support.get(source_file.location.location_type)
+        if method_name:
+            transfer_method = self.__getattribute__(method_name)
```
kaxil, line 93: why use internal method

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            self.drop_stage(stage_name=stage_name, snowflake_hook=snowflake_hook)
+            return True
+        return False
+
+    def create_stage(self, source_file: File, snowflake_hook: SnowflakeHook) -> str:
+        """
+        Creates a stage name based on source_file path and storage integration provided.
+        @param source_file: Source file
+        @param snowflake_hook: Snowflake hook
+        @return: Stage name created
```
kaxil, line 186:
```suggestion

        :param source_file: Source file
        :param snowflake_hook: Snowflake hook
        :return: Stage name created
```

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            )
+        execution_info = snowflake_hook.run(create_stage_sql, autocommit=True)
+        logging.info(execution_info)
+        return stage_name
+
+    @staticmethod
+    def drop_stage(stage_name: str, snowflake_hook: SnowflakeHook) -> None:
+        """
+        Runs the snowflake query to drop stage if exists
+        @param stage_name: Stage name
```
kaxil, line 252: same as dot notation

sunank200: Done

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        return False
+
+    @staticmethod
+    def get_gcs_project_id_from_conn(conn: connection) -> str:
+        """
+        Get GCS project id from conn
+        :param conn: Airflow's connection
+        """
+        if conn.extra and conn.extra.get("project"):
+            return str(conn.extra.get("project"))
```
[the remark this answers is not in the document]
sunank200: Done

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
         qualified_name = ".".join(name for name in qualified_name_lists if name)
         return qualified_name
 
+    def check_optimised_path_and_transfer(
```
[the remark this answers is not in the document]
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/487/commits/2191c036f3f91da354ef410f875bc26eb858ca4a)

src/astro/databases/base.py
sunank200: done

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,238 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+                    if_exists=if_exists,
+                    **kwargs,
+                ):
+                    return True
+            except Exception as exec_err:
+                raise exec_err
+        return False
+
+    @staticmethod
+    def get_gcs_project_id_from_conn(conn: connection) -> str:
```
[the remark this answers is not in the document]
sunank200: This can be a static method as its only used by gcp connection

src/astro/databases/snowflake.py
```diff
@@ -20,6 +34,11 @@ class SnowflakeDatabase(BaseDatabase):
     """
 
     def __init__(self, conn_id: str = DEFAULT_CONN_ID):
+        self.native_support: Dict[Any, str] = {
+            FileLocation.GS: "gs_to_snowflake",
+            FileLocation.S3: "s3_to_snowflake",
+        }
+        self.integration: Optional[str] = None
```
tatiana, line 41: It may be worth naming it `storage_integration` and adding a link to some Snowflake documentation about it, e.g.:
https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration.html

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,252 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        # FILE_FORMAT=(TYPE=CSV, TRIM_SPACE=TRUE) COPY_OPTIONS=(MATCH_BY_COLUMN_NAME=CASE_INSENSITIVE)
+        # storage_integration = <storage_integration_name>;
+
+        # COPY INTO query would be:
+        # COPY INTO sample_table from '@WORKSPACE_STAGE/sample.csv';
+        FILE_FORMAT_MAPPING = {
+            FileType.CSV: "CSV",
+            FileType.PARQUET: "PARQUET",
+            FileType.NDJSON: "JSON",
+        }
```
tatiana, line 218: What happens if the user was attempting to transfer something of `FileType.JSON`?

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,252 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            "stage_"
+            + random.choice(string.ascii_lowercase)
+            + "".join(
+                random.choice(string.ascii_lowercase + string.digits) for _ in range(7)
+            )
+        )
+
+        parsed_url = urlparse(source_file.path, allow_fragments=False)
+        bucket_type = snowflake_query_file_storage_mapping.get(
+            parsed_url.scheme.lower(), "gcs"
```
tatiana, line 230: Do we want to default the bucket_type to `gcs`? Perhaps we should raise an exception if it is not one of the pre-defined supported ones.

src/astro/sql/operators/load_file.py
```diff
@@ -80,13 +83,23 @@ def load_data_to_table(self, input_file: File) -> Table:
...
+                    **self.kwargs,
+                )
+            else:
+                database.load_pandas_dataframe_to_table(
+                    source_dataframe=file.export_to_dataframe(),
+                    target_table=self.output_table,
+                    if_exists=if_exists,
+                    chunk_size=self.chunk_size,
+                )
+                if_exists = "append"
```
tatiana, line 102: It would be great if we kept the operator as simple as possible and handled optimised paths from within the database class, as part of the `load_file_to_table` method

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,252 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            create_stage_sql = (
+                "CREATE OR REPLACE STAGE {} URL='{}' FILE_FORMAT=(TYPE={}, TRIM_SPACE=TRUE) "
+                "COPY_OPTIONS=(MATCH_BY_COLUMN_NAME=CASE_INSENSITIVE) storage_integration = {};".format(
+                    stage_name,
+                    location_url,
+                    FILE_FORMAT_MAPPING.get(get_filetype(source_file.path)),
+                    self.integration,
+                )
+            )
+        execution_info = snowflake_hook.run(create_stage_sql, autocommit=True)
```
tatiana, line 258: It is probably worth it for us to capture any exception here related to the user not having permissions to create a stage in snowflake, explaining which set of permissions is necessary and explaining how they can disable this behaviour (creating stage) if they don't have permissions for this.

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,252 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+                    if_exists=if_exists,
+                    **kwargs,
+                ):
+                    return
+            except Exception as exec_err:
+                raise exec_err
+        return
+
+    @staticmethod
+    def get_gcs_project_id_from_conn(conn: connection) -> str:
```
tatiana, line 120: Where are we using this method? Does it receive a `file_conn` or a `db_conn`?

sunank200: I was planing to get the chunk of file in using GCP connection and fetch first two to detect schema for CSV. I am implementing that part.

src/astro/databases/snowflake.py
sunank200: Makes sense. Adding the exception handling.

src/astro/databases/snowflake.py
sunank200: Removed this. Using pandas to do that.

src/astro/databases/snowflake.py
sunank200: [Done]([Add exception handling and change return types](https://github.com/astronomer/astro-sdk/pull/487/commits/863c4e9c8160f91412565a52aff93939f680638d))

src/astro/databases/snowflake.py
sunank200: [Done]([Add exception handling and change return types](https://github.com/astronomer/astro-sdk/pull/487/commits/863c4e9c8160f91412565a52aff93939f680638d))

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,252 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            FileType.NDJSON: "JSON",
+        }
+        snowflake_query_file_storage_mapping = {"gs": "gcs", "s3": "s3"}
+        stage_name = (
+            "stage_"
+            + random.choice(string.ascii_lowercase)
+            + "".join(
+                random.choice(string.ascii_lowercase + string.digits) for _ in range(7)
+            )
+        )
```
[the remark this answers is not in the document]
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/487/commits/863c4e9c8160f91412565a52aff93939f680638d)

src/astro/databases/snowflake.py
sunank200: Add check in `check_optimised_path()`
[Done](https://github.com/astronomer/astro-sdk/pull/487/commits/863c4e9c8160f91412565a52aff93939f680638d)

src/astro/databases/snowflake.py
```diff
@@ -57,6 +76,252 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            transfer_method = self.__getattribute__(method_name)
+            try:
+                if transfer_method(
+                    source_file=source_file,
+                    target_table=target_table,
+                    if_exists=if_exists,
+                    **kwargs,
+                ):
+                    return
+            except Exception as exec_err:
```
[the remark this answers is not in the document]
sunank200: I was trying to send the control to without native support in case it was failing but changed it now.

src/astro/databases/snowflake.py
sunank200: Done

src/astro/databases/snowflake.py
sunank200: This is trying to fetch the method based on file location type

src/astro/databases/snowflake.py
```diff
@@ -20,6 +35,15 @@ class SnowflakeDatabase(BaseDatabase):
     """
 
     def __init__(self, conn_id: str = DEFAULT_CONN_ID):
+        self.native_support: Dict[Any, str] = {FileLocation.GS: "gs_to_snowflake"}
```
utkarsharma2, line 38: @sunank200  We should define native_support and  optimized_path_supported_file_types as static variables. Also, can we be consistent with [PR](https://github.com/astronomer/astro-sdk/pull/489) for the names of the variables?

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,321 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        :param source_file: Source file
+        :param target_table: Output target table on snowflake
+        :param if_exists: Update strategy for file
+        :param kwargs:
+        """
+        # Fetch snowflake integration passed by the user as part of output table.
+        self.storage_integration = target_table.optional_args.get(
+            "storage_integration", None
+        )
+        if self.storage_integration is None:
```
utkarsharma2, line 115: @sunank200  This checked should be moved to - `check_optimised_path().`

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,321 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
         return qualified_name
 
+    def check_optimised_path(self, source_file: File, target_table: Table) -> bool:
+        """
+        Check if there is an optimised path for source to destination.
+
+        :param source_file: File from which we need to transfer data
+        :param target_table: Table that needs to be populated with file data
+        """
+        file_type = self.optimized_path_supported_file_types.get(source_file.type.name)
```
utkarsharma2, line 91: @sunank200 - Maybe not right now, can these checks can partially or completely be dependent on different sources? Like for S3. Like we may require 'storage_integration' for one and not for another or some different way to authenticate.

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,321 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+
+        # Create new table from file if snowflake table does not exists
+        if if_exists == "replace" and not self.table_exists(table=target_table):
+            self.drop_table(target_table)
+            self.create_emtpy_table(
+                source_file=source_file, target_table=target_table, if_exists=if_exists
+            )
+
+        # check if snowflake table exists
+        if self.table_exists(table=target_table):
```
utkarsharma2, line 160: @sunank200 Why do we have this check here? aren't we making sure that a exists at this point?

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,321 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+                if_exists=if_exists,
+                **kwargs,
+            )
+
+    def gs_to_snowflake(
+        self,
+        source_file: File,
+        target_table: Table,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
```
utkarsharma2, line 138: @sunank200  we need to make sure the users can pass other options related to the `copy into` query it might be useful for a lot of common cases.

src/astro/constants.py
```diff
@@ -11,6 +11,8 @@
 DEFAULT_CHUNK_SIZE = 1000000
 PYPI_PROJECT_NAME = "astro-sdk-python"
 
+LOAD_COLUMN_AUTO_DETECT_ROWS = 1
```
dimberman, line 14: Can you leave a comment explaining what this variable does?

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,322 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+            )
+        method_name = self.native_support.get(source_file.location.location_type)
+        if method_name:
+            transfer_method = self.__getattribute__(method_name)
+            transfer_method(
+                source_file=source_file,
+                target_table=target_table,
+                if_exists=if_exists,
+                **kwargs,
+            )
```
dimberman, line 131: @tatiana @kaxil how do you feel about this? This feels a bit... magic to me.

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,322 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        # where stage is WORKSPACE_STAGE, following steps has to be completed.
+
+        # The path of stage should be to the folder the file is in. Command to create STAGE would look
+        # like following:
+        # CREATE OR REPLACE STAGE <warehouse_name>.WORKSPACE_STAGE URL='gcs://sample-bucket/workspace'
+        # FILE_FORMAT=(TYPE=CSV, TRIM_SPACE=TRUE) COPY_OPTIONS=(MATCH_BY_COLUMN_NAME=CASE_INSENSITIVE)
+        # storage_integration = <storage_integration_name>;
+
+        # COPY INTO query would be:
+        # COPY INTO sample_table from '@WORKSPACE_STAGE/sample.csv';
```
dimberman, line 195: Why is this all in comments instead of in the doctring? This seems like info that might be useful to users.

dimberman (changes requested):

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,322 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+        FILE_FORMAT_MAPPING = {
+            FileType.CSV: "CSV",
+            FileType.PARQUET: "PARQUET",
+            FileType.NDJSON: "JSON",
+        }
+        snowflake_query_file_storage_mapping = {"gs": "gcs", "s3": "s3"}
+        stage_name = generate_stage_name()
```
tatiana, line 202: The `stage_name` needs to be the qualified name, otherwise, Snowflake raises an exception.

src/astro/databases/snowflake.py
```diff
@@ -57,6 +81,322 @@ def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
...
+
+            logging.info("Executing COPY command...")
+            execution_info = self.hook.run(
+                copy_query, autocommit=kwargs.get("autocommit", True)
+            )
+            logging.info("COPY command completed")
+            logging.info(execution_info)
+            self.drop_stage(stage_name=stage_name)
+
+    def create_stage(self, source_file: File) -> str:
```
tatiana, line 177: I separated this part in a PR of its own, with respective tests, and auxiliary methods for checking if a stage exists / to delete a stage:
https://github.com/astronomer/astro-sdk/pull/523

Issue #429, opened 2022-06-06
Support loading datasets of 10GB in BQ in less than 5 min


tatiana:
**Dependencies**
* Depends on: #437 

**Acceptance criteria**
* [X] Run the bq load command to see the "quickest" way we could load the file from GCS -> BQ (depending on this, adjust how long it is acceptable for us to do the same operation in Astro SDK)
* [ ] Make changes to how we load data into BigQuery.
* [ ] Re-run the benchmark and identify the performance improvements (?)
* [ ] The changes must work for all file types supported (CSV, JSON, NDJSON, Parquet)
* [ ] The changes made would also work on S3 and GCS

utkarsharma2:
Below are the results of the profiling on the load_file operator with the `gs://astro-sdk/benchmark/trimmed/imdb/title_ratings_10mb.csv` dataset.

```
454728 function calls (452486 primitive calls) in 69.968 seconds

   Ordered by: cumulative time
   
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     29/1    0.000    0.000   69.969   69.969 {built-in method builtins.exec}
        1    0.000    0.000   69.969   69.969 load_file.py:63(load_data)
        1    0.000    0.000   69.969   69.969 load_file.py:71(load_data_to_table)
     8273    0.021    0.000   59.489    0.007 socket.py:655(readinto)
     8273    0.020    0.000   59.451    0.007 ssl.py:1230(recv_into)
     8273    0.014    0.000   59.428    0.007 ssl.py:1090(read)
     8273   59.412    0.007   59.412    0.007 {method 'read' of '_ssl._SSLSocket' objects}
       41    0.000    0.000   54.168    1.321 _request_helpers.py:109(wait_and_retry)
        1    0.000    0.000   52.174   52.174 base.py:74(export_to_dataframe)
       54    0.001    0.000   51.561    0.955 sessions.py:457(request)
       51    0.002    0.000   51.237    1.005 requests.py:421(request)
       54    0.003    0.000   51.036    0.945 sessions.py:613(send)
       54    0.003    0.000   51.016    0.945 adapters.py:395(send)
       54    0.005    0.000   50.953    0.944 connectionpool.py:522(urlopen)
       54    0.004    0.000   50.900    0.943 connectionpool.py:361(_make_request)
        4    0.000    0.000   49.615   12.404 _decorators.py:302(wrapper)
[...]

PR #489, opened 2022-06-24
 Load files from GCS to Bigquery using BigqueryHook

utkarsharma2:
# Description
When we run the load_file operator for a table in Bigquery and a file in GCS, the data is first loaded to local from the GCS and then uploaded in Bigquery. Because of this, we require both memory and network bandwidth on the worker node. 

partially closes: #429 

## What is the new behavior?
We have used an optimized path that can directly ingest data from GCS to Bigquery, therefore removing both memory and bandwidth requirements on a worker node. Because of this, the speed and amount of data transfer increase. 

No

### Checklist
- [X] Created tests cases
- [X] Add a variable to control the selection of optimized path

### Results
Baseline using `GCSToBigQueryOperator` using [benchmark_gcs_to_bigquery.py](tests/benchmark/dags/benchmark_gcs_to_big_query.py)

|Dataset                                    |Size | Duration(seconds)  |
|-------------------------------------------|-----|--------------------|
|covid_overview/covid_overview_10kb.csv     |10 KB| 5.129522           |
|tate_britain/artist_data_100kb.csv         |100KB| 3.319834           |
|imdb/title_ratings_10mb.csv                |10MB | 5.558414           |
|stackoverflow/stackoverflow_posts_1g.ndjson|1GB  | 85.409014          |
|trimmed/pypi/*                             |5GB  | 48.973093          |

Results post optimization
|Dataset                                    |Size | Duration(seconds)  |
|-------------------------------------------|-----|--------------------|
|covid_overview/covid_overview_10kb.csv     |10 KB| 13.57           |
|tate_britain/artist_data_100kb.csv         |100KB| 16.10          |
|imdb/title_ratings_10mb.csv                |10MB | 19.40         |
|stackoverflow/stackoverflow_posts_1g.ndjson|1GB  | 30.26          |
|trimmed/pypi/*                             |5GB  | 59.90          |


src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+                target_table=target_table,
+                chunk_size=chunk_size,
+                if_exists=if_exists,
+                **kwargs,
+            )
+            return True
+
+        return False
+
+    def get_project_id_from_conn(self, conn: connection) -> str:
```
sunank200, line 176: This can be a static method.

utkarsharma2: Done.

utkarsharma2:
@sunank200 Added new becnhmaking results in PR.

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ) -> bool:
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        method_name = self.OPTIMIZED_PATHS.get(source_file.location.location_type)
+        if method_name:
```
kaxil, line 163: ok look like you do it here, but I think it would be good to have those as two methods.

as if optimized path is not available, we don't actually do the transfer

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+        return False
+
+    @staticmethod
+    def get_project_id_from_conn(conn: connection) -> str:
+        """
+        Get project id from conn
+        :param conn: Airflow's connection
+        """
+        if conn.extra and conn.extra.get("project"):
+            return str(conn.extra.get("project"))
```
kaxil, line 183:
```suggestion
            return conn.extra["project"]
```

This might fix it and not need `str()`

kaxil: We already check on L182 that "project" key is present in extra.

Also you might actually want `extra_dejson` not `extra`

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+        elif conn.host:
+            return str(conn.host)
+        raise ValueError(f"conn_id {conn.conn_id} has no project id.")
```
kaxil, line 186: Host as Project id?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+
+    def gs_to_bigquery(
+        self,
+        source_file: File,
+        target_table: Table,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,  # skipcq: PYL-W0613
```
kaxil, line 192: Would we ever use chunk_size in this function?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ) -> None:
+        """
+        Transfer data from gcs to bigquery
+        :param source_file: Source file that is used as source of data
+        :param target_table: Table that will be created on the bigquery
+        :param chunk_size: Specify the number of records in each batch to be written at a time
```
kaxil, line 200: Not sure chunk size matters here

tests/sql/operators/test_load_file.py
```diff
@@ -715,3 +717,98 @@ def test_load_file_should_fail_loudly(sample_dag, invalid_path, caplog):
...
+            "gs",
+            "bigquery",
+        ): "astro.databases.google.bigquery.BigqueryDatabase.gs_to_bigquery"
+    }
+
+    source = file_uri.split(":")[0]
+    destination = db.sql_type
+    mock_path = optimised_path_to_method[(source, destination)]
+
+    with mock.patch(mock_path) as gs_to_bigquery:
```
kaxil, line 760: If you want to mock the path, why not call the task directly instead of trying to run the DAG?

src/astro/databases/base.py
```diff
@@ -419,3 +419,17 @@ def get_sqlalchemy_template_table_identifier_and_parameter(
             self.get_table_qualified_name(table),
             self.get_table_qualified_name(table),
         )
+
+    def check_optimised_path_and_transfer(
```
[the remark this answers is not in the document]
utkarsharma2: Made the change.

src/astro/databases/google/bigquery.py
utkarsharma2: Made the change.

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+            )
+            return True
+
+        return False
+
+    @staticmethod
+    def get_project_id_from_conn(conn: connection) -> str:
+        """
+        Get project id from conn
+        :param conn: Airflow's connection
```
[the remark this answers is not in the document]
utkarsharma2: Done.

src/astro/databases/google/bigquery.py
utkarsharma2: Resolved.

src/astro/databases/google/bigquery.py
utkarsharma2: I hope not, :) Removing.

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,96 @@ def merge_table(
...
+        self,
+        source_file: File,
+        target_table: Table,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,  # skipcq: PYL-W0613
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ) -> None:
+        """
+        Transfer data from gcs to bigquery
+        :param source_file: Source file that is used as source of data
```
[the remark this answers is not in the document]
utkarsharma2: Done.

src/astro/databases/google/bigquery.py
utkarsharma2: Removed.

src/astro/sql/operators/load_file.py
```diff
@@ -19,6 +19,7 @@ class LoadFile(BaseOperator):
     :param ndjson_normalize_sep: separator used to normalize nested ndjson.
     :param chunk_size: Specify the number of records in each batch to be written at a time.
     :param if_exists: Overwrite file if exists. Default False.
+    :param optimise_load: variable to control use of optimised path for datatransfer.
```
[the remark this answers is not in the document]
utkarsharma2: Done.

src/astro/sql/operators/load_file.py
```diff
@@ -158,6 +170,7 @@ def load_file(
     task_id: Optional[str] = None,
     if_exists: LoadExistStrategy = "replace",
     ndjson_normalize_sep: str = "_",
+    optimise_load: bool = True,
```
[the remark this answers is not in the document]
utkarsharma2: **use_native_support** sounds good to me

tests/sql/operators/test_load_file.py
```diff
@@ -715,3 +717,98 @@ def test_load_file_should_fail_loudly(sample_dag, invalid_path, caplog):
...
```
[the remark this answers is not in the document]
utkarsharma2: Updated test case to check the method call and verify the args and kwags passed.

tests/sql/operators/test_load_file.py
utkarsharma2: Done.

src/astro/sql/operators/load_file.py
```diff
@@ -82,13 +85,23 @@ def load_data_to_table(self, input_file: File) -> Table:
...
+                    **self.kwargs,
+                )
+            else:
+                database.load_pandas_dataframe_to_table(
+                    source_dataframe=file.export_to_dataframe(),
+                    target_table=self.output_table,
+                    if_exists=if_exists,
+                    chunk_size=self.chunk_size,
+                )
+                if_exists = "append"
```
tatiana, line 104: Why are we setting `if_exists` to append here?

tests/sql/operators/test_load_file.py
```diff
@@ -715,3 +717,129 @@ def test_load_file_should_fail_loudly(sample_dag, invalid_path, caplog):
...
+                "configuration": {
+                    "load": {
+                        "sourceUris": [file_uri],
+                        "destinationTable": {"tableId": test_table.name},
+                    }
+                }
+            },
+            "expected_args": (),
+        }
+    }
```
tatiana, line 780: What is the benefit of having many different properties defined in a dictionary, as opposed to having separate variables containing the desired values?
```
source_file_location = "gs"
destination_database = "bigquery"
method_path = "airflow.providers.google.cloud.hooks.bigquery.BigQueryHook.insert_job"
expected_kwargs = {
                "configuration": {
                    "load": {
                        "sourceUris": [file_uri],
                        "destinationTable": {"tableId": test_table.name},
                    }
                }
expected_args = ()
```

tests/sql/operators/test_load_file.py
```diff
@@ -715,3 +717,129 @@ def test_load_file_should_fail_loudly(sample_dag, invalid_path, caplog):
...
+    expected_kwargs = optimised_path_to_method[(source, destination)]["expected_kwargs"]
+    expected_args = optimised_path_to_method[(source, destination)]["expected_args"]
+
+    with mock.patch(mock_path) as method:
+        load_file(
+            input_file=File(file_uri),
+            output_table=test_table,
+        ).operator.execute({})
+        assert method.called
+        assert is_dict_subset(superset=method.call_args.kwargs, subset=expected_kwargs)
```
tatiana, line 794: Why are we checking for a subset and not an exact value?

src/astro/sql/operators/load_file.py
tatiana, line 104: Why are we setting `if_exists` to append here?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,100 @@ def merge_table(
...
+        """
+        Transfer data from gcs to bigquery
+
+        :param source_file: Source file that is used as source of data
+        :param target_table: Table that will be created on the bigquery
+        :param if_exists: Overwrite table if exists. Default 'replace'
+        """
+        bq_hook = BigQueryHook(
+            gcp_conn_id=target_table.conn_id,
+        )
```
tatiana, line 209: any reason why we're not using `self.hook`?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,100 @@ def merge_table(
...
+
+        if if_exists == "replace":
+            self.drop_table(target_table)
```
tatiana, line 212: We may want to do this regardless if it is an optimised implementation or not. Would there be a higher level function where we could place this?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,100 @@ def merge_table(
...
+
+        file_type_to_bq_source_format = {
+            FileType.CSV: "CSV",
+            FileType.NDJSON: "NEWLINE_DELIMITED_JSON",
+            FileType.PARQUET: "PARQUET",
```
tatiana, line 217: What happens if the user had a `FileType.JSON`? Should we raise an exception?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,100 @@ def merge_table(
...
+        }
+
+        conn = BaseHook.get_connection(target_table.conn_id)
```
tatiana, line 220: do we need `BaseHook`, or could we be using `self.hook`?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,100 @@ def merge_table(
...
+
+        load_job_config = {
+            "sourceUris": [source_file.path],
+            "destinationTable": {
+                "projectId": self.get_project_id_from_conn(conn),
+                "datasetId": target_table.metadata.schema,
+                "tableId": target_table.name,
+            },
+            "createDisposition": "CREATE_IF_NEEDED",
+            "writeDisposition": "WRITE_APPEND",
```
tatiana, line 230:
```
if_exists == "replace" -> WRITE_TRUNCATE
if_exists == "append" -> WRITE_APPEND
```
https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.WriteDisposition.html

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +146,100 @@ def merge_table(
...
+            "sourceFormat": file_type_to_bq_source_format[source_file.type.name],
+            "autodetect": True,
+        }
+
+        # Since bigquery has other options besides used here, we need to expose them to end user.
```
tatiana, line 235: Could you give examples?

src/astro/sql/operators/load_file.py
```diff
@@ -82,13 +85,23 @@ def load_data_to_table(self, input_file: File) -> Table:
...
+                    if_exists=self.if_exists,
+                    **self.kwargs,
+                )
+            else:
+                database.load_pandas_dataframe_to_table(
+                    source_dataframe=file.export_to_dataframe(),
+                    target_table=self.output_table,
+                    if_exists=if_exists,
+                    chunk_size=self.chunk_size,
+                )
```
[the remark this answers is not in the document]
utkarsharma2: @tatiana we can definitely use load_file_to_table. I'll make the change.

Regarding the first point, I think it's lean enough right now. But If you feel strongly about it we can move it to the database.

src/astro/sql/operators/load_file.py
utkarsharma2: @tatiana we are iterating over files and the data needs to be appended in the same table.

tests/sql/operators/test_load_file.py
utkarsharma2: @tatiana I think variables are useful when we are reusing them, here this config is only consumed at one location.

tests/sql/operators/test_load_file.py
utkarsharma2: @because Matching complete dicts also entail we create the exact same dicts in the test cases that might have a certain logic to them and implementing them might be overkill for test case and as this test case is meant to be generic for all source-destination paths it becomes crucial for it to not have that logic. Instead of that, we can check for certain important args and kwargs and that's why we are using subset logic.

src/astro/sql/operators/load_file.py
utkarsharma2: Probably duplicate comment. Addressed above.

src/astro/databases/google/bigquery.py
utkarsharma2: @tatiana @kaxil @sunank200  Nice catch! Yes, we should, and I'll update the code reflecting the same. 

Also, there are two things here
1. I think we should drop the support for the JSON files as they are not used for large data transfer the preferred file type is NDJSON and most DBs actually mean NDJSON when they say JSON.
2. We should also include fileTypes in figuring out optimized paths. Right now we are using only (source, destination) it should be (source, destination, fileType) since there might be a support to ingest some fileType and not others by DBs and we should take that into account.

src/astro/databases/google/bigquery.py
utkarsharma2: Done.

src/astro/databases/google/bigquery.py
utkarsharma2: @tatiana my bad I can't find the doc but I had the impression that 'WRITE_TRUNCATE' retains the schema of an existing table.  Have made the required changes to code.

src/astro/databases/google/bigquery.py
utkarsharma2: Added google link for options.

src/astro/databases/google/bigquery.py
utkarsharma2: As per our discussion, removing it since this feature is no longer needed because the google provider >= 6.4.2 take care of this.

src/astro/databases/google/bigquery.py
utkarsharma2: Done

tests/sql/operators/test_load_file.py
tatiana: @utkarsharma2 seems to make sense, could you give an example of a parameter we don't care about in the test?

src/astro/sql/operators/load_file.py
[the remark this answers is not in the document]
tatiana: Hi @utkarsharma2 , yes, please, I'd be very grateful if we could use the original method in the `BaseDatabase` class

src/astro/sql/operators/load_file.py
tatiana: The operator should not have to worry about iterating through files - this should be an internal implementation. from within the Database. Also, it is probably worth adding a comment saying this line is used when we have multiple files.

tests/sql/operators/test_load_file.py
tatiana: I just realised this is a generic method which will be reused across multiple optimised paths, @utkarsharma2 - I understand now :)

src/astro/databases/google/bigquery.py
tatiana: @utkarsharma2:
1. I'm fine dropping support to JSON that is not new-line delimited.
2. I have mixed feelings about this. It would be great if the Astro SDK converts between file types, if necessary, to support the optimised path regardless of the file type.

src/astro/sql/operators/load_file.py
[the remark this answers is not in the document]
kaxil: +1

src/astro/databases/google/bigquery.py
kaxil: 1. I think we discussed on Friday that we will fall-back to the non-optimized path if the filetype isn't supported or raise an error for now and drop the support in future major releases
2. I agree we should include filetype in finding optimized path as doing without it means a) we will either need to fallback to pandas or b) convert the file on the source. So happy for you to include filetype in optimization check. And CSV, NDJSON, PARQUET should be supported by most of the DBs anyways

kaxil (commented):
Can you also update the PR title :) -- "Optimise loading files from GCS to Bigquery" -- or something like that

src/astro/databases/google/bigquery.py
utkarsharma2: Noted. Made the required code changes.

src/astro/sql/operators/load_file.py
utkarsharma2: Done.

src/astro/sql/operators/load_file.py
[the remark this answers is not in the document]
utkarsharma2: Done.

tests/sql/operators/test_load_file.py
utkarsharma2: @tatiana - {'source_file': File(location="gs://dag-authoring/test/8e14a473-3f61-42d6-b903-f9daa22b75c50.csv",type="csv"), 'target_table': Table(conn_id='bigquery', name='_tmp_nomxb3e1i1m5tlxtt3rra55y6tjlap28gibfwgzed587wdo2', metadata=Metadata(schema='tmp_astro', database='astronomer-dag-authoring'), columns=[], temp=True), 'if_exists': 'replace'}

.github/ci-test-connections.yaml
```diff
@@ -31,8 +31,9 @@ connections:
   - conn_id: bigquery
     conn_type: bigquery
     description: null
-    extra: null
-    host: "astronomer-dag-authoring"
+    extra:
+      project: "astronomer-dag-authoring"
```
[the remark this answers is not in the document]
kaxil: The following is the source code to get the project id
https://github.com/apache/airflow/blob/d9694733cafd9a3d637eb37d5154f0e1e92aadd4/airflow/providers/google/common/hooks/base_google.py#L327-L336

https://github.com/apache/airflow/blob/d9694733cafd9a3d637eb37d5154f0e1e92aadd4/airflow/providers/google/common/hooks/base_google.py#L314-L325

src/astro/databases/base.py
```diff
@@ -220,15 +222,56 @@ def load_file_to_table(
...
+        else:
+            dataframe = source_file.export_to_dataframe()
+            self.load_pandas_dataframe_to_table(
+                dataframe,
+                target_table,
+                if_exists,
+                chunk_size,
+            )
+
+    def load_multiple_file_to_table(
```
kaxil, line 246:
```suggestion
    def load_multiple_files_to_table(
```

src/astro/databases/base.py
```diff
@@ -419,3 +462,25 @@ def get_sqlalchemy_template_table_identifier_and_parameter(
...
+        source_file: File,
+        target_table: Table,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ):
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        pass
```
kaxil, line 486: Would be good to raise an Exception/NotImplemented error instead here

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +150,96 @@ def merge_table(
...
+        target_table: Table,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ):
+        """
+        Checks if optimised path for transfer between File location to database exists
+        and if it does, it transfers it and returns true else false.
+        """
+        method_name = self.OPTIMIZED_PATHS.get(source_file.location.location_type)
+        if method_name:
```
kaxil, line 177: If a file type isn't supported, it will silently pass here. Should we raise an error, what are your thoughts?

src/astro/databases/google/bigquery.py
kaxil: We don't need this do we?

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +150,96 @@ def merge_table(
...
+        # Since bigquery has other options besides used here, we need to expose them to end user.
+        # https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationLoad
+        load_job_config.update(kwargs)
+
+        job_config = {
+            "jobType": "LOAD",
+            "load": load_job_config,
+            "labels": {"target_table": target_table.name},
+        }
+        bq_hook.insert_job(
```
kaxil, line 243: Can we not change this to `self.hook.insert_job` and remove `bq_hook`?

src/astro/files/base.py
```diff
@@ -36,11 +36,11 @@ def __init__(
 
     @property
     def path(self) -> str:
-        return self.location.path
+        return self.location.path  # type: ignore
```
kaxil, line 39: We should solve this instead of ignoring it -- can be in a separate PR

utkarsharma2: @kaxil I'll raise a separate PR for this.

src/astro/databases/google/bigquery.py
utkarsharma2: Done.

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +150,96 @@ def merge_table(
...
+        )
+
+        write_disposition_val = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
+
+        conn = self.hook.get_connection(target_table.conn_id)
+
+        load_job_config = {
+            "sourceUris": [source_file.path],
+            "destinationTable": {
+                "projectId": self.get_project_id_from_conn(conn),
```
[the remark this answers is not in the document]
utkarsharma2: Done.

src/astro/databases/google/bigquery.py
utkarsharma2: Make sense, added code change.

src/astro/databases/base.py
utkarsharma2: Done.

src/astro/databases/google/bigquery.py
```diff
@@ -135,3 +150,101 @@ def merge_table(
...
+            )
+
+    @staticmethod
+    def get_project_id_from_conn(conn: connection) -> str:
+        """
+        Get project id from conn
+
+        :param conn: Airflow's connection
+        """
+        if conn.extra_dejson and conn.extra_dejson.get("project"):
```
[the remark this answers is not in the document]
tatiana: Where are we using this method?

src/astro/sql/operators/load_file.py
```diff
@@ -76,19 +79,18 @@ def load_data_to_table(self, input_file: File) -> Table:
             ndjson_normalize_sep=self.ndjson_normalize_sep,
             database=database,
         )
-        if_exists = self.if_exists
-        for file in get_files(
+        files = resolve_file_path_pattern(
```
tatiana, line 82: @utkarsharma2 , let's use in the operator `database.load_file_to_table`, and let's allow the method   `database.load_file_to_table` to decide if and how to iterate through files if needed. From the Astro Python SDK perspective, `load_file` means wildcard.

As an example, the BigQuery load job supports receiving the wildcard for the files to be matched, so we do not need to iterate through files in this particular path.

utkarsharma2: Moved the logic into database's base class.

src/astro/databases/google/bigquery.py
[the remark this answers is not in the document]
utkarsharma2: We are not using it now. Removed it.

tatiana (approved):
Thanks, @utkarsharma2 ! I'll use this as a base to extend Ankit's work on Snowflake, so we have consistent interfaces! 🙌 
Feel free to merge once the tests pass!

PR #490, opened 2022-06-27
Remove old code in `astro/utils/`

kaxil:
This code is no longer needed as we now import a Database class (Bigquery, Snowflake) or a File (GCS) via [create_database](https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/__init__.py#L23)  or [create_file_location](https://github.com/astronomer/astro-sdk/blob/main/src/astro/files/locations/__init__.py#L16-L27):

- Removed `MissingPackage` class and its import usage 

# Description

Just redundant code is being deleted. The code that is deleted was used to import packages using try..catch

## What is the new behavior?

- Same behaviour

No

### Checklist


src/astro/files/locations/google/gcs.py
```diff
@@ -1,9 +1,10 @@
 from typing import Dict, List
 from urllib.parse import urlparse, urlunparse
 
+from airflow.providers.google.cloud.hooks.gcs import GCSHook
```
dimberman, line 4: @kaxil would this not cause import errors if the provider isn't installed?

dimberman (changes requested):

src/astro/files/locations/google/gcs.py
kaxil: Since this module isn't imported directly from anywhere we should go good in my opinion.

This module is only imported from https://github.com/astronomer/astro-sdk/blob/main/src/astro/files/locations/__init__.py#L16-L27

dimberman: ohhh interesting ok.

dimberman (approved):

Issue #497, opened 2022-06-30
Benchmarking: Read task duration from airflows database.


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we are calculating the efficiency of load_oprator by running a dag and then collecting profiling data by measuring the start and end time of the dag. The time calculated in this way is not accurate and also includes tasks like scheduling of tasks and deletion of the table created and other overheads by airflow. 

**total time** = dag end time - dag start time

**Describe the solution you'd like**
We can directly collect task run time by airflow's database and use it in our benchmarking results

**Are there any alternatives to this feature?**
Not that I'm aware of, but I'm open to suggestions

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

dimberman:
cc: @kaxil  I think we can probably do something like this to get the duration for the task

``` 
from airflow.utils.session import provide_session
from airflow.models import TaskInstance

@provide_session
def get_load_task_duration(dag, execution_date, session=None):
    ti: TaskInstance = session.query(TaskInstance).filter(TaskInstance.dag_id == dag.dag_id,
        TaskInstance.run_id == execution_date, TaskInstance.task_id == "load").first()   
    return ti.duration
```

We'll just need to figure out the task_id.

PR #500, opened 2022-07-04
Benchmarking: read task duration from taskInstance model of airflow instead of profiling dag.

utkarsharma2:
# Description
Currently, we are calculating the efficiency of load_oprator by running a dag and then collecting profiling data by measuring the start and end times of the dag. The time calculated in this way is not accurate and also includes tasks like scheduling of tasks and deletion of the table created and other overheads by airflow.

total time = dag end time - dag start time

closes: #497 

## What is the new behavior?
We can directly collect task run time by airflow's database and use it in our benchmarking results

No


tatiana (approved):
Good improvement, @utkarsharma2 !
Could you share some numbers on the difference between these times before and after?

PR #506, opened 2022-07-04
Improve cleanup retries

tatiana:
Change the default cleanup retry configuration, so if users are running the steps in the tutorial with cleanup, the example will gracefully work.


src/astro/sql/operators/base.py
```diff
@@ -59,7 +59,7 @@ def execute(self, context: Dict) -> None:
 
         # Find and load dataframes from op_arg and op_kwarg into Table
         self.create_output_table_if_needed()
-        self.op_args = load_op_arg_dataframes_into_sql(
+        self.op_args = load_op_arg_dataframes_into_sql(  # type: ignore
```
kaxil, line 62: Why was this needed? lgtm otherwise

PR #510, opened 2022-07-05
Benchmark postgres

### LINKED ISSUE #431: Evaluate the performance of loading datasets with Astro Python SDK 0.9.0 into Postgres
**Dependencies**
* Depends on: #432
* Depends on: #434
* Depends on: #433 
* Depends on: #435

**Acceptance criteria**
* Trigger the benchmark to load all the supported (CSV-format, available in GCS) datasets into Postgres (K8s-hosted), exporting the resulting metrics. Use a large worker node. Set timeout to 1h.
* Document the resulting metrics in the Astro SDK repo, within: `tests/benchmark`

dimberman:
# Description
Currently we use old benchmarking numbers for postgres that rely on a local docker image of postgres. These numbers are unreliable because they don't take into account that a remote postgres might have more resources and can take advantage of faster internet. 

closes #431

## What is the new behavior?
We have now added postgres to the new benchmarking script s.t. it can run in a remote kubernetes cluster.

### Checklist


tests/benchmark/analyse.py
```diff
@@ -46,7 +47,10 @@ def analyse_results(results_filepath):
     data = []
     with open(results_filepath) as fp:
         for line in fp.readlines():
-            data.append(json.loads(line.strip()))
+            try:
+                data.append(json.loads(line.strip()))
+            except JSONDecodeError as j:
+                print(j)
```
pankajastro, line 53: what about using a logging library instead of print here?

tests/benchmark/analyse.py
```diff
@@ -46,7 +47,10 @@ def analyse_results(results_filepath):
     data = []
     with open(results_filepath) as fp:
         for line in fp.readlines():
-            data.append(json.loads(line.strip()))
+            try:
+                data.append(json.loads(line.strip()))
```
tatiana, line 51: In which scenario this happened, @dimberman ? Is this expected?

PR #518, opened 2022-07-07
Rename `BaseSQLOperator` class

dimberman:
# Description
Currently the base class for all SQL Decorators is called "BaseSQLOperator". This is a misnomer, as this class makes a lot of assumptions around being a function decorator. Future contributors might be confused by this name and try to use it on non-decorator classes.

## What is the new behavior?
Change the name to the more accurate "BaseSQLDecoratorOperator" to ensure that users would only extend this class for creating decorators.

No


src/astro/sql/operators/base_decorator.py
```diff
@@ -12,7 +12,7 @@
 from astro.utils.table import find_first_table
 
 
-class BaseSQLOperator(DecoratedOperator):
+class BaseSQLDecoratorOperator(DecoratedOperator):
```
kaxil, line 15: Should it be `BaseSQLDecoratedOperator` instead of "Decorator? Mainly coz it is inheriting `DecoratedOperator`, wdyt?

```suggestion
class BaseSQLDecoratedOperator(DecoratedOperator):
```

kaxil (approved):

Issue #512, opened 2022-07-06
Optimise S3 to Bigquery data transfer


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we are transferring the data between S3 to Bigquery vi GCS to local and then local to Bigquery.

**Describe the solution you'd like**
We can request Bigquery via API to submit the job at bigquery to do the direct transfer, which can lot quicker.

**Additional context**
https://cloud.google.com/bigquery-transfer/docs/s3-transfer#bq

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #519, opened 2022-07-08
Use native path for S3 to Bigquery in load_file operator

utkarsharma2:
# Description
Currently, we are transferring the data between S3 to Bigquery vi GCS to local and then local to Bigquery.

related: #512 

## What is the new behavior?
We can request Bigquery via API to submit the job at bigquery to do the direct transfer, which can lot quicker.
https://cloud.google.com/bigquery-transfer/docs/s3-transfer#bq

No

### Checklist

### DIFF
--- a/.github/ci-test-connections.yaml
+++ b/.github/ci-test-connections.yaml
@@ -52,3 +52,5 @@ connections:
     conn_type: aws
     description: null
     extra: null
+    login: $AWS_ACCESS_KEY_ID
+    password: $AWS_SECRET_ACCESS_KEY
--- a/.github/ci-test-connections.yaml
+++ b/.github/ci-test-connections.yaml
@@ -52,5 +52,3 @@ connections:
     conn_type: aws
     description: null
     extra: null
-    login: $AWS_ACCESS_KEY_ID
-    password: $AWS_SECRET_ACCESS_KEY
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -207,7 +207,7 @@ class BaseDatabase(ABC):
         """
         Remove all rows from a given table.
 
-        :param table: The table to be deleted.
+        :param table: Table from which all rows will be deleted.
         """
         statement = self._truncate_table_statement.format(
             self.get_table_qualified_name(table)
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -479,3 +479,36 @@ class BaseDatabase(ABC):
         and if it does, it transfers it and returns true else false.
         """
         raise NotImplementedError
+
+    def create_empty_table(
+        self, source_file: File, target_table: Table, chunksize: int = 1000
+    ):
+        """
+        Infer schema from source file and create and empty table in database
+
+        :param source_file: File from which we need to transfer data
+        :param target_table: Table that needs to be populated with file data
+        :param chunksize: No. of rows to use to infer schema
+        """
+        df = source_file.export_to_dataframe(chunksize=chunksize)
+        # When we pass chunksize we get reference to data and not the data.
+        # We need to use read() to get the 1st chunk.
+        df = df.read()
+        self.load_pandas_dataframe_to_table(
+            source_dataframe=df,
+            target_table=target_table,
+            if_exists="replace",
+            chunk_size=chunksize,
+        )
+        self.truncate_table(target_table)
+
+    def get_project_id(self, target_table) -> str:
+        """
+        Get project id from the hook.
+
+        :param target_table: table object that the hook is derived from.
+        """
+        try:
+            return str(self.hook.project_id)
+        except AttributeError:
+            raise ValueError(f"conn_id {target_table.conn_id} has no project id")
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -1,7 +1,6 @@
 from abc import ABC
 from typing import Dict, List, Optional, Tuple, Union
 
-import pandas
 import pandas as pd
 import sqlalchemy
 from airflow.hooks.dbapi import DbApiHook
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -4,6 +4,7 @@ from typing import Dict, List, Optional, Tuple, Union
 import pandas as pd
 import sqlalchemy
 from airflow.hooks.dbapi import DbApiHook
+from pandas.io.sql import get_schema
 from sqlalchemy import column, insert, select
 from sqlalchemy.sql import ClauseElement
 from sqlalchemy.sql.elements import ColumnClause
@@ -37,7 +38,6 @@ class BaseDatabase(ABC):
     _create_schema_statement: str = "CREATE SCHEMA IF NOT EXISTS {}"
     _drop_table_statement: str = "DROP TABLE IF EXISTS {}"
     _create_table_statement: str = "CREATE TABLE IF NOT EXISTS {} AS {}"
-    _truncate_table_statement: str = "TRUNCATE TABLE {}"
     # Used to normalize the ndjson when appending fields in nested fields.
     # Example -
     #   ndjson - {'a': {'b': 'val'}}
@@ -199,17 +199,6 @@ class BaseDatabase(ABC):
         :param table: The table to be deleted.
         """
         statement = self._drop_table_statement.format(
-            self.get_table_qualified_name(table)
-        )
-        self.run_sql(statement)
-
-    def truncate_table(self, table: Table) -> None:
-        """
-        Remove all rows from a given table.
-
-        :param table: Table from which all rows will be deleted.
-        """
-        statement = self._truncate_table_statement.format(
             self.get_table_qualified_name(table)
         )
         self.run_sql(statement)
@@ -503,22 +492,7 @@ class BaseDatabase(ABC):
         # When we pass chunksize we get reference to data and not the data.
         # We need to use read() to get the 1st chunk.
         df = df.read()
-
-        self.load_pandas_dataframe_to_table(
-            source_dataframe=df,
-            target_table=target_table,
-            if_exists="replace",
-            chunk_size=chunksize,
+        schema_statement = get_schema(
+            df, self.get_table_qualified_name(target_table), con=self.sqlalchemy_engine
         )
-        self.truncate_table(target_table)
-
-    def get_project_id(self, target_table) -> str:
-        """
-        Get project id from the hook.
-
-        :param target_table: table object that the hook is derived from.
-        """
-        try:
-            return str(self.hook.project_id)
-        except AttributeError:
-            raise ValueError(f"conn_id {target_table.conn_id} has no project id")
+        self.run_sql(schema_statement)
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -205,7 +205,7 @@ class BaseDatabase(ABC):
 
     def truncate_table(self, table: Table) -> None:
         """
-        Delete a SQL table, if it exists.
+        Remove all rows from a given table.
 
         :param table: The table to be deleted.
         """
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -37,7 +37,7 @@ class BaseDatabase(ABC):
     _create_schema_statement: str = "CREATE SCHEMA IF NOT EXISTS {}"
     _drop_table_statement: str = "DROP TABLE IF EXISTS {}"
     _create_table_statement: str = "CREATE TABLE IF NOT EXISTS {} AS {}"
-
+    _truncate_table_statement: str = "TRUNCATE TABLE {}"
     # Used to normalize the ndjson when appending fields in nested fields.
     # Example -
     #   ndjson - {'a': {'b': 'val'}}
@@ -199,6 +199,17 @@ class BaseDatabase(ABC):
         :param table: The table to be deleted.
         """
         statement = self._drop_table_statement.format(
+            self.get_table_qualified_name(table)
+        )
+        self.run_sql(statement)
+
+    def truncate_table(self, table: Table) -> None:
+        """
+        Delete a SQL table, if it exists.
+
+        :param table: The table to be deleted.
+        """
+        statement = self._truncate_table_statement.format(
             self.get_table_qualified_name(table)
         )
         self.run_sql(statement)
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -485,7 +485,7 @@ class BaseDatabase(ABC):
         :param source_file: File from which we need to transfer data
         :param target_table: Table that needs to be populated with file data
         :param if_exists: Overwrite file if exists. Default False
-        :param native_support_kwargs: kwargs to be used by method involved in native support flow
+        :param native_support_kwargs: kwargs to be used by native loading command
         """
         raise NotImplementedError
 
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -1,6 +1,7 @@
 from abc import ABC
 from typing import Dict, List, Optional, Tuple, Union
 
+import pandas
 import pandas as pd
 import sqlalchemy
 from airflow.hooks.dbapi import DbApiHook
@@ -479,19 +480,16 @@ class BaseDatabase(ABC):
         raise NotImplementedError
 
     def create_empty_table(
-        self, source_file: File, target_table: Table, chunksize: int = 1000
+        self, source_file: File, target_table: Table, nrows: int = 1000
     ):
         """
         Infer schema from source file and create and empty table in database
 
         :param source_file: File from which we need to transfer data
         :param target_table: Table that needs to be populated with file data
-        :param chunksize: No. of rows to use to infer schema
+        :param nrows: No. of rows to use to infer schema
         """
-        df = source_file.export_to_dataframe(chunksize=chunksize)
-        # When we pass chunksize we get reference to data and not the data.
-        # We need to use read() to get the 1st chunk.
-        df = df.read()
+        df = source_file.export_to_dataframe(nrows=nrows)
         schema_statement = get_schema(
             df, self.get_table_qualified_name(target_table), con=self.sqlalchemy_engine
         )
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -226,6 +226,7 @@ class BaseDatabase(ABC):
         if_exists: LoadExistStrategy = "replace",
         chunk_size: int = DEFAULT_CHUNK_SIZE,
         use_native_support: bool = True,
+        native_support_kwargs: Optional[Dict] = None,
         **kwargs,
     ):
         """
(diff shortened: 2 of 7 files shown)


tatiana:
> ## What is the new behavior?
> 
> We can request Bigquery via API to submit the job at bigquery to do the direct transfer, which can lot quicker. https://cloud.google.com/bigquery-transfer/docs/s3-transfer#bq


@utkarsharma2 could you give numbers on the performance improvement?

.github/ci-test-connections.yaml
```diff
@@ -52,3 +52,5 @@ connections:
     conn_type: aws
     description: null
     extra: null
+    login: $AWS_ACCESS_KEY_ID
+    password: $AWS_SECRET_ACCESS_KEY
```
tatiana, line 56: What changed for us to need to explicitly give these values now?

src/astro/databases/base.py
```diff
@@ -203,6 +203,17 @@ def drop_table(self, table: Table) -> None:
         )
         self.run_sql(statement)
 
+    def truncate_table(self, table: Table) -> None:
+        """
+        Delete a SQL table, if it exists.
```
tatiana, line 208:
```suggestion
        Remove all rows from a given table.
```

src/astro/databases/base.py
```diff
@@ -203,6 +203,17 @@ def drop_table(self, table: Table) -> None:
...
+
+        :param table: The table to be deleted.
```
tatiana, line 210:
```suggestion
        :param table: Table from which all rows will be deleted.
```

src/astro/databases/base.py
```diff
@@ -461,10 +475,50 @@ def load_file_to_table_natively(
...
     ):
         """
         Checks if optimised path for transfer between File location to database exists
-        and if it does, it transfers it and returns true else false.
+        and if it does, it transfers it and returns true else false
+
+        :param source_file: File from which we need to transfer data
+        :param target_table: Table that needs to be populated with file data
+        :param if_exists: Overwrite file if exists. Default False
+        :param native_support_kwargs: kwargs to be used by method involved in native support flow
```
tatiana, line 488:
```suggestion
        :param native_support_kwargs: kwargs to be used by native loading command
```

src/astro/databases/base.py
```diff
@@ -461,10 +475,50 @@ def load_file_to_table_natively(
...
+        # We need to use read() to get the 1st chunk.
+        df = df.read()
+
+        self.load_pandas_dataframe_to_table(
+            source_dataframe=df,
+            target_table=target_table,
+            if_exists="replace",
+            chunk_size=chunksize,
+        )
+        self.truncate_table(target_table)
```
tatiana, line 513: It may be cleaner to use something like the following to create an empty table:
```
from pandas.io.sql import get_schema

schema_statement = get_schema(df, self.get_table_qualified_name(target_table), con=self.sqlalchemy_engine)
self.run_sql(schema_statement)
```
As opposed to loading stuff and having to delete them afterwards.

.github/ci-test-connections.yaml
utkarsharma2: @tatiana  We need to pass the S3 credentials for bigquery to access the remote file.

src/astro/databases/base.py
```diff
@@ -461,10 +475,50 @@ def load_file_to_table_natively(
...
+
+    def get_project_id(self, target_table) -> str:
```
tatiana, line 515: Why is this in the `BaseDatabase` class? Shouldn't it be in the `BigqueryDatabase`?

.github/ci-test-connections.yaml
tatiana: We can use a similar strategy to this:
https://github.com/astronomer/astro-sdk/pull/523/files#diff-b8140a95c509699b028734bb59473e42133480d0d99ae375e7ada290d8fb03bfR204-R205

Without the need to explicitly define them in the connection.

utkarsharma2: @tatiana I'll rebase when PR - [https://github.com/astronomer/astro-sdk/pull/523/](https://github.com/astronomer/astro-sdk/pull/523/files#diff-b8140a95c509699b028734bb59473e42133480d0d99ae375e7ada290d8fb03bfR204-R205) is merged to main.

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
         self.hook.insert_job(
             configuration=job_config,
         )
+
+    def s3_to_bigquery(
```
tatiana, line 261: This way we'd be consistent with the current naming conventions:
```suggestion
    def load_s3_file_to_table(
```

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        self,
+        source_file: File,
+        target_table: Table,
+        if_exists: LoadExistStrategy = "replace",
+        native_support_kwargs: Optional[Dict] = None,
+        **kwargs,
+    ):
+        """
+        Transfer data from S3 to Bigquery via invoking datatransfer job
```
tatiana, line 270:
```suggestion
        Load content of multiple files in S3 to output_table in Bigquery by using a datatransfer job.
```

We need to add a link to the data transfer job service! Also, we should say that there are costs associated to this transaction.

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+            self.create_empty_table(source_file, target_table)
+
+        transfer = S3ToBigqueryDataTransfer(
+            target_table=target_table,
+            source_file=source_file,
+            project_id=project_id,
+            native_support_kwargs=native_support_kwargs,
+            **kwargs,
+        )
+        transfer.run()
```
tatiana, line 293: What happens if the user does not have this API enabled in GCP? Should we capture this exception / give a friendly message in that case?

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        self.poll_duration = poll_duration
+        self.native_support_kwargs = native_support_kwargs
+        self.kwargs = kwargs
+
+    def run(self):
+        """
+        Algo to run S3 to Bigquery datatransfer
+        """
+        transfer_config_id = self.create_transfer_config()
+        try:
```
tatiana, line 336: Why don't we have an `except` clause? Are we at the risk of silencing possible issues?

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        :param target_table: Table that will be created on the bigquery
+        :param project_id: Bigquery project id
+        :param poll_duration: sleep duration between two consecutive job status checks. Unit - seconds. Default 1 sec.
+        :param native_support_kwargs: kwargs to be used by method involved in native support flow
+        """
+        self.client = BiqQueryDataTransferServiceHook(gcp_conn_id=target_table.conn_id)
+        self.target_table = target_table
+        self.source_file = source_file
+
+        conn = BaseHook.get_connection(source_file.conn_id)
```
tatiana, line 321: I exposed the hook in the FileLocation class - it may help here:
https://github.com/astronomer/astro-sdk/pull/523/files#diff-aaa4e821af24c22f598e2399b6d9afc54358185f4810740bc387b966f87da14eR18

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        s3_params = {
+            "destination_table_name_template": self.target_table.name,
+            "data_path": self.source_file.path,
+            "access_key_id": self.s3_login,
+            "secret_access_key": self.s3_password,
+            "file_format": self.s3_file_type,
+        }
+        s3_params.update(self.native_support_kwargs)
+
+        s3_params.update(self.kwargs)
```
tatiana, line 395: Why do we have both `kwargs` and `native_support_kwargs`?

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        Extract transfer_config_id from TransferConfig object
+        """
+        # ToDo: Look for a native way to extract 'transfer_config_id'
+        # name - "projects/103191871648/locations/us/transferConfigs/6302bf19-0000-26cf-a568-94eb2c0a61ee'.
+        # We need extract transferConfigs which is at the end of string.
+        tokens = config.name.split("transferConfigs/")
+        return str(tokens[-1])
+
+    @staticmethod
+    def get_run_id(config: TransferRun) -> str:
```
tatiana, line 373: We can have a unit test for this!

tests/sql/operators/test_load_file.py
```diff
@@ -836,8 +836,43 @@ def test_aql_load_file_optimized_path_method_is_not_called(
...
+    "database_table_fixture",
+    [
+        {
+            "database": Database.BIGQUERY,
+        }
+    ],
+    indirect=True,
+    ids=["Bigquery"],
+)
+def test_aql_load_file_optimized_path_method_is_not_called_1(
```
tatiana, line 855: @utkarsharma2 I'm slightly worried about the imbalance between the huge amount of code introduced by this PR and the little number of tests.

Should we try to cover some simple functions with unit test, where relevant, and perhaps handle some error scenarios?

src/astro/databases/google/bigquery.py
```diff
@@ -28,11 +47,9 @@ class BigqueryDatabase(BaseDatabase):
     logic in other parts of our code-base.
     """
 
-    NATIVE_PATHS = {FileLocation.GS: "gs_to_bigquery"}
-    NATIVE_PATHS_SUPPORTED_FILE_TYPES = {
-        FileType.CSV: "CSV",
-        FileType.NDJSON: "NEWLINE_DELIMITED_JSON",
-        FileType.PARQUET: "PARQUET",
+    NATIVE_PATHS = {
+        FileLocation.GS: "gs_to_bigquery",
```
kaxil, line 51: What's the value (`gs_to_bigquery`) in this dict used for?

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        native_support_kwargs: Optional[Dict] = None,
+        **kwargs,
+    ):
+        """
+        :param source_file: Source file that is used as source of data
+        :param target_table: Table that will be created on the bigquery
+        :param project_id: Bigquery project id
+        :param poll_duration: sleep duration between two consecutive job status checks. Unit - seconds. Default 1 sec.
+        :param native_support_kwargs: kwargs to be used by method involved in native support flow
+        """
```
kaxil, line 316: Move this to class docstring instead of __init__ docstrings

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +257,177 @@ def gs_to_bigquery(
...
+        # ToDo: Look for a native way to extract 'run_id'
+        # config.runs[0].name - "projects/103191871648/locations/us/
+        # transferConfigs/62d38894-0000-239c-a4d8-089e08325b54/runs/62d6a4df-0000-2fad-8752-d4f547e68ef4'.
+        # We need extract transferConfigs which is at the end of string.
+        tokens = config.runs[0].name.split("runs/")
+        return str(tokens[-1])
+
+    def create_transfer_config(self):
+        """Create bigquery transfer config on cloud"""
```
kaxil, line 385: What was the outcome of your research of whether we could use bigquery_dts available in Airflow or not -- based on our discussion y'day?

tests/sql/operators/test_load_file.py
kaxil: One option is to move the `kwargs_for_native_path` in a separate PR

src/astro/databases/base.py
```diff
@@ -203,6 +203,17 @@ def drop_table(self, table: Table) -> None:
         )
         self.run_sql(statement)
 
+    def truncate_table(self, table: Table) -> None:
```
kaxil, line 206: Need a unit test for this

src/astro/databases/base.py
kaxil: We need a unit test for this too

.github/ci-test-connections.yaml
utkarsharma2: Will make the change.

src/astro/databases/google/bigquery.py
utkarsharma2: We are not silencing possible issues, they will be raised for sure. But the intent of the code is to make sure we delete the transfer config once it's created, regardless of any issues.

Screenshot 2022-07-13 at 12 38 39 AM

src/astro/databases/google/bigquery.py
utkarsharma2: Duplicate.

src/astro/databases/google/bigquery.py
utkarsharma2: was a mistake will remove it.

src/astro/databases/google/bigquery.py
utkarsharma2: @kaxil they are the function names that are supposed to be called for a certain path(source, destination). So now the mapping is - 
```
NATIVE_PATHS = {
        FileLocation.GS: "load_gs_file_to_bigquery",
        FileLocation.S3: "load_s3_file_to_bigquery",
    }
```
and below code is used to run a specific function.
https://github.com/astronomer/astro-sdk/blob/32b6f157a0cf7a64a150cc246a5ac70d4fc39410/src/astro/databases/google/bigquery.py#L164

tests/sql/operators/test_load_file.py
utkarsharma2: @kaxil the integration test won't without this change, as we need to pass certain parameters that are specific to this dataset.

src/astro/databases/base.py
utkarsharma2: Added unit test.

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +256,189 @@ def gs_to_bigquery(
...
+        transfer.run()
+
+    def get_project_id(self, target_table) -> str:
+        """
+        Get project id from the hook.
+
+        :param target_table: table object that the hook is derived from.
+        """
+        try:
+            return str(self.hook.project_id)
```
tatiana, line 305: It would be great to use Airflow's implementation of this:
https://github.com/astronomer/astro-sdk/pull/489#discussion_r913838157

src/astro/databases/google/bigquery.py
```diff
@@ -236,3 +256,189 @@ def gs_to_bigquery(
...
+        :param native_support_kwargs: kwargs to be used by method involved in native support flow
+        """
+        native_support_kwargs = native_support_kwargs or {}
+
+        project_id = self.get_project_id(target_table)
+
+        if if_exists == "replace":
+            # We need to create an empty table as datatransfer job can only append to an existing table,
+            # so we need to create a table.
+            self.create_empty_table(source_file, target_table)
```
tatiana, line 287: We should also support the case where the user gives the Table schema using `Table(columns=[...])`. I'm okay with this being addressed in a separate PR if we think it is out of the scope. This will give us feature parity with Snowflake.

src/astro/databases/google/bigquery.py
utkarsharma2: I think google handles it at their end, and we should raise this exception since it will guarantee to have working links to the correct document.
Screenshot 2022-07-13 at 3 36 14 PM

src/astro/databases/google/bigquery.py
utkarsharma2: @tatiana I'll address this in a separate PR. Created a ticket for the same - https://github.com/astronomer/astro-sdk/issues/532

src/astro/databases/base.py
```diff
@@ -461,10 +464,32 @@ def load_file_to_table_natively(
...
+        and if it does, it transfers it and returns true else false
+
+        :param source_file: File from which we need to transfer data
+        :param target_table: Table that needs to be populated with file data
+        :param if_exists: Overwrite file if exists. Default False
+        :param native_support_kwargs: kwargs to be used by native loading command
         """
         raise NotImplementedError
+
+    def create_empty_table(
```
tatiana, line 481: Since we already have `create_table` - which creates an empty table using `table.columns`, perhaps we should name this `create_table_with_schema_autodetection`, WDYT?

tatiana (approved):
Thanks, @utkarsharma2 , one last minor comment!

utkarsharma2:
@tatiana - as suspected Native path is overkill for a smaller file. However, I was not able to pull data for Larger files as it was taking too long, I'll run the script on my local and we should have data by morning. 

Also, we should include file size for making the decision to go with the s3-to-bigquery native path or not. 

**Native**

{"duration": 174.000502, "memory_full_info": {"rss": 55508992, "vms": 299528192, "pfaults": 58038, "pageins": 447, "uss": 12341248}, "cpu_time": {"user": 0.040738088, "system": 0.034938644000000005, "children_user": 0.0, "children_system": 0.0}, "disk_usage": 5554176, "dag_id": "load_file_ten_kb_into_bigquery", "execution_date": "2022-07-13 20:19:24.638457+00:00", "revision": "dddab4c", "chunk_size": 1000000}

{"duration": 175.380213, "memory_full_info": {"rss": 56573952, "vms": 432734208, "pfaults": 57560, "pageins": 4, "uss": 12738560}, "cpu_time": {"user": 0.04194025800000001, "system": 0.035832556, "children_user": 0.0, "children_system": 0.0}, "disk_usage": 13266944, "dag_id": "load_file_hundred_kb_into_bigquery", "execution_date": "2022-07-13 20:24:48.855388+00:00", "revision": "dddab4c", "chunk_size": 1000000}

{"duration": 250.084764, "memory_full_info": {"rss": 57597952, "vms": 595263488, "pfaults": 58045, "pageins": 0, "uss": 13012992}, "cpu_time": {"user": 0.046639617999999994, "system": 0.046830241999999994, "children_user": 0.0, "children_system": 0.0}, "disk_usage": -7954432, "dag_id": "load_file_ten_mb_into_bigquery", "execution_date": "2022-07-13 20:27:56.481097+00:00", "revision": "dddab4c", "chunk_size": 1000000}



**Non Native -**

{"duration": 19.023265, "memory_full_info": {"rss": 53841920, "vms": 584056832, "pfaults": 23617, "pageins": 314, "uss": 12009472}, "cpu_time": {"user": 0.017896796000000003, "system": 0.005202350000000001, "children_user": 0.0, "children_system": 0.0}, "disk_usage": 434176, "dag_id": "load_file_ten_kb_into_bigquery", "execution_date": "2022-07-13 21:12:23.805734+00:00", "revision": "dddab4c", "chunk_size": 1000000}


{"duration": 18.644579, "memory_full_info": {"rss": 54509568, "vms": 411955200, "pfaults": 24069, "pageins": 1, "uss": 12079104}, "cpu_time": {"user": 0.020449904, "system": 0.005729686999999999, "children_user": 0.0, "children_system": 0.0}, "disk_usage": -10100736, "dag_id": "load_file_hundred_kb_into_bigquery", "execution_date": "2022-07-13 21:12:55.073895+00:00", "revision": "dddab4c", "chunk_size": 1000000}


{"duration": 47.02245, "memory_full_info": {"rss": 201797632, "vms": 883830784, "pfaults": 77794, "pageins": 0, "uss": 47960064}, "cpu_time": {"user": 0.03278598, "system": 0.011085567999999999, "children_user": 0.0, "children_system": 0.0}, "disk_usage": 5947392, "dag_id": "load_file_ten_mb_into_bigquery", "execution_date": "2022-07-13 21:13:25.891789+00:00", "revision": "dddab4c", "chunk_size": 1000000}

kaxil:
We should store this benchmark data in https://github.com/astronomer/astro-sdk/blob/main/tests/benchmark/results.md

Issue #503, opened 2022-07-04
Read configs from airflow's config


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we are reading the config info. directly from the env variables using - os.get(). Better way to read the values is from airflow configs. 

**Describe the solution you'd like**
We can use the below code to read from the airflow's config. By assuming a section in the airflow's config named - `astro`

```
from airflow.configuration import conf
conf.get('astro', 'config_name')
```

PR #520, opened 2022-07-08
Read default config if airflow's isn't defined

utkarsharma2:
# Description
Currently, the code for reading default `sql_schema` in setting.py errors out if the config is not defined in airflow's config.

related: #503 

## What is the new behavior?
Added a default value of not `None` so that the default value gets picked up.

Nope.

### Checklist

### DIFF
--- a/src/astro/settings.py
+++ b/src/astro/settings.py
@@ -1,4 +1,4 @@
 from airflow.configuration import conf
 
 DEFAULT_SCHEMA = "tmp_astro"
-SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=None) or DEFAULT_SCHEMA
+SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=DEFAULT_SCHEMA)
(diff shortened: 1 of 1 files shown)


kaxil (approved):

src/astro/settings.py
```diff
@@ -1,4 +1,4 @@
 from airflow.configuration import conf
 
 DEFAULT_SCHEMA = "tmp_astro"
-SCHEMA = conf.get("astro_sdk", "sql_schema") or DEFAULT_SCHEMA
+SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=None) or DEFAULT_SCHEMA
```
kaxil, line 4:
```suggestion
SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=DEFAULT_SCHEMA)
```

utkarsharma2: @kaxil my bad, changing it.

Issue #444, opened 2022-06-08
Plan a strategy to gracefully handle dataframes so users don't trash Airflow metadata database


tatiana:
**Context**

At the moment the `dataframe` decorator adds the entire content of a dataframe to Airflow XCom. Depending on how the user set XCom, this can potentially trash Airflow's metadata database (if the user is using the default XCom backend).

There are a few approaches we could use to deal with this:
* Give warnings to users if they use the dataframe decorator with the default backend
* Not allow users to use the dataframe decorator with the default backend if the dataframe is larger than a specific size
* Others

The goal with this ticket is propose a solution for the problem.

**Acceptance criteria**

* [ ] Log a ticket with the agreed approach

dimberman:
@tatiana I think one solution here would be to force users to explicitly set an environment variable of "yes I know that I'm putting dataframes in the airlfow metadata and yes I know that's dangerous". Something like "ASTRO__DATAFRAME__ALLOW_UNSAFE_STORAGE" cc: @kaxil

PR #522, opened 2022-07-08
create a possible solution to users passing large dataframes between …

dimberman:
…tasks

# Description
Currently, users can put full dataframes into the Airflow metadata DB by default (or without even knowing it) through our load and dataframe functions. this can be extremely dangerous as it will fill up the airflow metadata DB and cause all kinds of issues. We don't want to entirely block this ability (especially since we use it for our testing), but we also don't want to allow it by default.

closes https://github.com/astronomer/astro-sdk/issues/444

## What is the new behavior?
This new behavior basically forces users to either provide a safe custom backend that can store full dataframes, or to at minimum give an environment variable that they understand the risks involved with storing data directly to the metadata DB.

Yes, I would say this is SOMEWHAT breaking. It forces people to create a custom backend or set an environment variable. That said I would also call this a critical bugfix.

### Checklist


.env-template
```diff
@@ -11,3 +11,4 @@ SNOWFLAKE_WAREHOUSE=<add snowflake warehouse here>
 AWS_ACCESS_KEY_ID=<add aws api key here>
 AWS_SECRET_ACCESS_KEY=<add aws secret key here>
 AIRFLOW_VAR_FOO=templated_file_name
+ASTRO__DATAFRAME__ALLOW_UNSAFE_STORAGE=True
```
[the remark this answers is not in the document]
utkarsharma2: @dimberman We should use prefix - `AIRFLOW__ASTRO_SDK__` for env variables since we have standardized the config retrieval from airflow config manager. PR - https://github.com/astronomer/astro-sdk/issues/503

.github/workflows/ci.yaml
```diff
@@ -113,6 +113,7 @@ jobs:
       AIRFLOW__CORE__LOAD_DEFAULT_CONNECTIONS: True
       AIRFLOW__CORE__ENABLE_XCOM_PICKLING: True
       AIRFLOW_VAR_FOO: templated_file_name
+      ASTRO__DATAFRAME__ALLOW_UNSAFE_STORAGE: True
```
utkarsharma2, line 116:
```suggestion
      AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE=True: True
```

.github/workflows/ci.yaml
```diff
@@ -178,6 +179,8 @@ jobs:
...
```
utkarsharma2, line 182:
```suggestion
      AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE: True
```

.github/workflows/ci.yaml
```diff
@@ -236,6 +239,7 @@ jobs:
...
```
utkarsharma2, line 242:
```suggestion
      AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE: True
```

src/astro/settings.py
```diff
@@ -1,4 +1,4 @@
 from airflow.configuration import conf
 
 DEFAULT_SCHEMA = "tmp_astro"
-SCHEMA = conf.get("astro_sdk", "sql_schema") or DEFAULT_SCHEMA
+SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=None) or DEFAULT_SCHEMA
```
utkarsharma2, line 4: @dimberman you can rebase from main, this is fixed.

src/astro/sql/operators/dataframe.py
```diff
@@ -130,4 +134,10 @@ def execute(self, context: Dict) -> Union[Table, pd.DataFrame]:
             )
             return self.output_table
         else:
+            if (
+                conf.get("core", "xcom_backend") == "airflow.models.xcom.BaseXCom"
+                and os.getenv("ASTRO__DATAFRAME__ALLOW_UNSAFE_STORAGE", "False").lower()
```
utkarsharma2, line 139:
```suggestion
                and os.getenv("AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE", "False").lower()
```

src/astro/sql/operators/load_file.py
```diff
@@ -62,6 +69,12 @@ def load_data(self, input_file: File) -> Union[Table, pd.DataFrame]:
         if self.output_table:
             return self.load_data_to_table(input_file)
         else:
+            if (
+                conf.get("core", "xcom_backend") == "airflow.models.xcom.BaseXCom"
+                and os.getenv("ASTRO__DATAFRAME__ALLOW_UNSAFE_STORAGE", "False").lower()
```
utkarsharma2, line 74:
```suggestion
                and os.getenv("AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE", "False").lower()
```

src/astro/constants.py
```diff
@@ -45,3 +44,12 @@ class Database(Enum):
 
 # TODO: check how snowflake names these
 MergeConflictStrategy = Literal["ignore", "update", "exception"]
+
+LOAD_DATAFRAME_ERROR_MESSAGE = (
```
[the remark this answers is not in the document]
tatiana: I like the approach of a threshold, I also commented about this in:
https://github.com/astronomer/astro-sdk/pull/522/files#r917858466 :)

tests/sql/operators/test_load_file.py
```diff
@@ -66,6 +67,19 @@ def test_load_file_with_http_path_file(sample_dag, database_table_fixture):
...
+@mock.patch.dict(os.environ, {"ASTRO__DATAFRAME__ALLOW_UNSAFE_STORAGE": "False"})
+def test_unsafe_loading_of_dataframe(sample_dag):
+    with pytest.raises(AirflowException):
+        with sample_dag:
+            load_file(
+                input_file=File(
+                    "https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/homes_main.csv"
+                ),
+            )
+        test_utils.run_dag(sample_dag)
```
tatiana, line 80: Should we also have some assert to confirm that we logged the desired message to the end-user? Even a subset of it would be great!

dimberman: @tatiana we can make a custom exception and ensure that the custom exception is thrown. This PR was more to get sign off that this is a good solution.

src/astro/constants.py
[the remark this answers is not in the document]
dimberman: I'd like @kaxil 's thought on this too, but IMO we really shouldn't put ANY dataframes into the metadata database. the DB really isn't meant for this in the first place.

.env-template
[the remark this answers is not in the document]
dimberman: @tatiana i think the size limit idea would lead to very weird user interactions (e.g. someone is using the library and their DAGs suddenly start failing because their dataframe is too large).

Storing data in the metadata DB is an anti pattern we shouldn't promote

src/astro/sql/operators/dataframe.py
kaxil: use `conf.get` as this setting should come from there too like - https://github.com/astronomer/astro-sdk/pull/522#discussion_r917727385

src/astro/constants.py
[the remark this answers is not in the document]
kaxil: I agree, for now -- just relying on Custom XCom backend and "DATAFRAME_ALLOW_UNSAFE_STORAGE" (or similar config) should be good.

tests/sql/operators/test_load_file.py
kaxil: You can pass a "match" kwarg to `pytest.raises`. Example: https://stackoverflow.com/a/55982417 or https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions

Definition: https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises

utkarsharma2 (approved):
@dimberman  LGTM!

kaxil:
we missed a better PR title :) -- It is long! so it ends abruptly at "..." -- smaller PR description and commit subjects as listed in https://cbea.ms/git-commit/ is definitely desirable please

Issue #428, opened 2022-06-06
Support loading datasets of 10GB in Postgres in less than 5 min


tatiana:
**Context**

We started improving the performance of loading data into Postgres before starting refactoring. We used the `COPY` command. The work can be found at the branch: https://github.com/astronomer/astro-sdk/tree/issue-98-part7.

What was missing, at the time, was to address edge-cases of escaping quotes and dealing with null values.

**Dependencies**
* Depends on: #431 
* Depends on: #509 

**Acceptance criteria**
* [ ] Make changes to how we load data into Postgres.
* [ ] Re-run the benchmark and identify the performance improvements (?)
* [ ] The changes must work for all file types supported (CSV, JSON, NDJSON, Parquet)
* [ ] The changes made would also work on S3 and GCS

Issue #515, opened 2022-07-07
Issue when truncating after load_file with output_table


tatiana:
**Describe the bug**

Truncate after loading to a file failed:
```
    with DAG(dag_name, schedule_interval=None, start_date=START_DATE) as dag:
        chunk_size = int(os.getenv("ASTRO_CHUNK_SIZE", DEFAULT_CHUNK_SIZE))
        table_metadata = table_args.pop("metadata", {})
        if table_metadata:
            table = Table(metadata=Metadata(**table_metadata), **table_args)
        else:
            table = Table(**table_args)

        table_xcom = aql.load_file(  # noqa: F841
            input_file=File(
                path=dataset_path,
                conn_id=dataset_conn_id,
                filetype=FileType(dataset_filetype),
            ),
            task_id="load",
            output_table=table,
            chunk_size=chunk_size,
        )
        aql.truncate(table_xcom)
```

With the error message:

```
  File "/Users/al-cht01/Code/astro-sdk/src/astro/sql/operators/truncate.py", line 20, in __init__
    task_id = task_id or get_unique_task_id(table.name + "_truncate")
AttributeError: 'XComArg' object has no attribute 'name'
```
This was discovered during:
https://github.com/astronomer/astro-sdk/pull/514

**Version**
* Astro: 0.11.0

**Expected behaviour**
The DAG should succeed without errors

PR #531, opened 2022-07-12
Optimize postgres performance

dimberman:
# Description
Currently the postgres load numbers are far too slow for any production use-case. We need to find optimizations to bring the 10GB loading time to under 5 minutes for customers to reasonably use this feature.

Previous numbers:
| database                | dataset    | total_time   | memory_rss   | cpu_time_user   | cpu_time_system   |
|:------------------------|:-----------|:-------------|:-------------|:----------------|:------------------|
| postgres_conn_benchmark | few_kb     | 494.42ms     | 36.06MB      | 570.0ms         | 50.0ms            |
| postgres_conn_benchmark | ten_kb     | 689.05ms     | 42.96MB      | 540.0ms         | 40.0ms            |
| postgres_conn_benchmark | hundred_kb | 580.7ms      | 36.43MB      | 570.0ms         | 50.0ms            |
| postgres_conn_benchmark | ten_mb     | 44.67s       | 1.38GB       | 31.03s          | 4.03s             |
| postgres_conn_benchmark | one_gb     | 5.56min      | 62.5MB       | 14.07s          | 1.14s             |
| postgres_conn_benchmark | five_gb    | 24.44min     | 78.15MB      | 1.34min         | 5.73s             |
| postgres_conn_benchmark | ten_gb     | 45.64min     | 61.71MB      | 2.37min         | 11.48s            |

closes: [#428](https://github.com/astronomer/astro-sdk/issues/428)

## What is the new behavior?
In this PR we found two primary optimizations that have led to significant speed up

The first optimization is to not directly place a smart_open into the `read` function of pandas. There have been multiple reports of pandas and smart_open working very slow when used together. Instead, we read the `smart_open` file into an IO buffer (either StringIO or BytesIO). This also has the benefit that we do not need to write to disc if there is enough memory to hold the object.

The second optimization is postgres specific. The `pandas.to_sql` function uses the HIGHLY suboptimal `INSERT` statement for postgres. This is 10X slower than the `COPY` command. By saving objects as CSV buffers, we can use COPY to load the data into postgres as a stream instead of one value at a time.

Here are the new numbers with the optimizations:

| database   | dataset    | total_time   | memory_rss   | cpu_time_user   | cpu_time_system   |
|:-----------|:-----------|:-------------|:-------------|:----------------|:------------------|
| postgres   | ten_kb     | 432.93ms     | 43.19MB      | 540.0ms         | 60.0ms            |
| postgres   | hundred_kb | 417.2ms      | 31.34MB      | 560.0ms         | 40.0ms            |
| postgres   | ten_mb     | 2.21s        | 79.91MB      | 1.76s           | 170.0ms           |
| postgres   | one_gb     | 13.9s        | 35.43MB      | 7.97s           | 5.76s             |
| postgres   | five_gb    | 50.23s       | 43.27MB      | 26.64s          | 18.7s             |
| postgres   | ten_gb     | 3.11min      | 43.29MB      | 1.36min         | 1.48min           |

No

### Checklist

### DIFF
--- a/src/astro/databases/postgres.py
+++ b/src/astro/databases/postgres.py
@@ -1,9 +1,12 @@
 """Postgres database implementation."""
+import io
+from contextlib import closing
 from typing import Dict, List
 
 import pandas as pd
 import sqlalchemy
 from airflow.providers.postgres.hooks.postgres import PostgresHook
+from pandas.io.sql import SQLDatabase
 from psycopg2 import sql as postgres_sql
 
 from astro.constants import DEFAULT_CHUNK_SIZE, LoadExistStrategy, MergeConflictStrategy
@@ -74,15 +77,29 @@ class PostgresDatabase(BaseDatabase):
         if target_table.metadata and target_table.metadata.schema:
             self.create_schema_if_needed(target_table.metadata.schema)
             schema = target_table.metadata.schema.lower()
-        source_dataframe.to_sql(
-            target_table.name,
-            schema=schema,  # type: ignore
-            con=self.sqlalchemy_engine,
+        db = SQLDatabase(engine=self.sqlalchemy_engine)
+        # within prep_table() we use pandas drop() function which is used when we pass 'if_exists=replace'.
+        # There is an issue where has_table() works with uppercase table names but the function meta.reflect() don't.
+        # To prevent the issue we are passing table name in lowercase.
+        db.prep_table(
+            source_dataframe,
+            target_table.name.lower(),
+            schema=schema,
             if_exists=if_exists,
-            chunksize=chunk_size,
-            method="multi",
             index=False,
         )
+
+        output = io.StringIO()
+        source_dataframe.to_csv(output, sep=",", header=True, index=False)
+        output.seek(0)
+        table = self.get_table_qualified_name(target_table)
+        con = self.hook.get_conn()
+        with closing(con) as conn:
+            with closing(conn.cursor()) as cur:
+                cur.copy_expert(
+                    f"COPY {table} FROM STDIN DELIMITER ',' CSV HEADER;", output
+                )
+                conn.commit()
 
     @staticmethod
     def get_table_qualified_name(table: Table) -> str:  # skipcq: PYL-R0201
--- a/src/astro/databases/postgres.py
+++ b/src/astro/databases/postgres.py
@@ -6,7 +6,6 @@ from typing import Dict, List
 import pandas as pd
 import sqlalchemy
 from airflow.providers.postgres.hooks.postgres import PostgresHook
-from pandas.io.sql import SQLDatabase
 from psycopg2 import sql as postgres_sql
 
 from astro.constants import DEFAULT_CHUNK_SIZE, LoadExistStrategy, MergeConflictStrategy
@@ -73,18 +72,6 @@ class PostgresDatabase(BaseDatabase):
         :param if_exists: Strategy to be used in case the target table already exists.
         :param chunk_size: Specify the number of rows in each batch to be written at a time.
         """
-        if target_table.metadata and target_table.metadata.schema:
-            self.create_schema_if_needed(target_table.metadata.schema)
-
-        # create an empty table
-        db = SQLDatabase(engine=self.sqlalchemy_engine)
-        db.prep_table(
-            source_dataframe,
-            target_table.name.lower(),
-            schema=target_table.metadata.schema,
-            if_exists=if_exists,
-            index=False,
-        )
 
         output_buffer = io.StringIO()
         source_dataframe.to_csv(output_buffer, sep=",", header=True, index=False)
--- a/src/astro/databases/postgres.py
+++ b/src/astro/databases/postgres.py
@@ -83,15 +83,16 @@ class PostgresDatabase(BaseDatabase):
             if_exists=if_exists,
             index=False,
         )
-        output = io.StringIO()
-        source_dataframe.to_csv(output, sep=",", header=True, index=False)
-        output.seek(0)
-        table = self.get_table_qualified_name(target_table)
-        con = self.hook.get_conn()
-        with closing(con) as conn:
+        output_buffer = io.StringIO()
+        source_dataframe.to_csv(output_buffer, sep=",", header=True, index=False)
+        output_buffer.seek(0)
+        table_name = self.get_table_qualified_name(target_table)
+        postgres_conn = self.hook.get_conn()
+        with closing(postgres_conn) as conn:
             with closing(conn.cursor()) as cur:
                 cur.copy_expert(
-                    f"COPY {table} FROM STDIN DELIMITER ',' CSV HEADER;", output
+                    f"COPY {table_name} FROM STDIN DELIMITER ',' CSV HEADER;",
+                    output_buffer,
                 )
                 conn.commit()
 
--- a/src/astro/databases/postgres.py
+++ b/src/astro/databases/postgres.py
@@ -6,7 +6,6 @@ from typing import Dict, List
 import pandas as pd
 import sqlalchemy
 from airflow.providers.postgres.hooks.postgres import PostgresHook
-from pandas.io.sql import SQLDatabase
 from psycopg2 import sql as postgres_sql
 
 from astro.constants import DEFAULT_CHUNK_SIZE, LoadExistStrategy, MergeConflictStrategy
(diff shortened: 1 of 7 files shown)


src/astro/databases/postgres.py
```diff
@@ -70,20 +73,31 @@ def load_pandas_dataframe_to_table(
...
         )
 
+        output_buffer = io.StringIO()
+        source_dataframe.to_csv(output_buffer, sep=",", header=True, index=False)
+        output_buffer.seek(0)
+        table_name = self.get_table_qualified_name(target_table)
+        postgres_conn = self.hook.get_conn()
+        with closing(postgres_conn) as conn, closing(conn.cursor()) as cur:
+            cur.copy_expert(
+                f"COPY {table_name} FROM STDIN DELIMITER ',' CSV HEADER;",
```
tatiana, line 96: We'll probably want to allow users to customize the CSV delimiter and if the CSV has a header or not. That could be part of the `File` object.

tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -63,7 +63,7 @@ def create_dag(database_name, table_args, dataset):
             output_table=table,
             chunk_size=chunk_size,
         )
-        aql.truncate(my_table)
+        aql.cleanup([my_table])
```
tatiana, line 66: Why have we introduced this change?

tatiana:
Amazing results, @dimberman , some minor comments!

tests/files/test_file.py
```diff
@@ -137,12 +137,15 @@ def test_create_from_dataframe(
...
+def test_export_to_dataframe(filetype, locations, type_method_map_fixture):
     """Test export_to_dataframe() for all locations and filetypes"""
-    with patch(type_method_map_fixture[FileType(filetype)]) as mocked_read:
+    if filetype == "parquet":
+        data = str.encode("data")
+    else:
+        data = "data"
+    with patch(type_method_map_fixture[FileType(filetype)]) as mocked_read, patch(
+        "astro.files.base.smart_open.open", mock_open(read_data=data)
+    ) as mocked_smart_open:
```
tatiana, line 148: What is the benefit of declaring the patches within the test instead of using the decorator? It would be great if we were consistent with other tests.

dimberman: @tatiana from what I can tell `mock_open` must be used as part of the context statement. When I tried adding `mock_open` to the decorator it didn't work. I can try again.

tests/benchmark/dags/evaluate_load_file.py
dimberman: @tatiana two reasons

1) there are issues in truncate that I'm gonna make tickets for today
2) Truncate empties but doesn't delete the table. We want to delete the table here.

tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -46,7 +46,7 @@ def create_dag(database_name, table_args, dataset):
     dag_name = f"load_file_{dataset_name}_into_{database_name}"
 
     with DAG(dag_name, schedule_interval=None, start_date=START_DATE) as dag:
-        chunk_size = int(os.getenv("ASTRO_CHUNK_SIZE", DEFAULT_CHUNK_SIZE))
+        chunk_size = int(os.getenv("ASTRO_CHUNK_SIZE", str(DEFAULT_CHUNK_SIZE)))
```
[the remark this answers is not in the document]
dimberman: @tatiana deepsource doesn't like the fact that we're defaulting to an integer with `os.getenv`.

tests/sql/operators/test_load_file.py
```diff
@@ -507,26 +507,21 @@ def test_load_file_with_named_schema(sample_dag, database_table_fixture, file_ty
...
         },
-        {
-            "database": Database.POSTGRES,
-        },
     ],
     indirect=True,
-    ids=["snowflake", "bigquery", "postgresql"],
+    ids=["snowflake", "bigquery"],
 )
 def test_load_file_chunks(sample_dag, database_table_fixture):
```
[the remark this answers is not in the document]
dimberman: @tatiana AFAICT it seems that since we're now using StringIO that the buffering is handled for us, I can investigate further though.

dimberman:
@tatiana can you point me to where this work is being done (re: allowing users to define their own schema)?

tests/sql/operators/test_load_file.py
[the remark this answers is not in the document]
tatiana: Yes, please, @dimberman , I'd love to understand this.

tests/benchmark/dags/evaluate_load_file.py
tatiana: BTW: are there more bugs besides  https://github.com/astronomer/astro-sdk/issues/515?

tests/sql/operators/test_load_file.py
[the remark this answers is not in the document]
dimberman: @tatiana if you look [here](https://github.com/astronomer/astro-sdk/pull/531/files#diff-7160b78503b73f988abfbe41d63a4ff12e4fbeaead1ab29642aba0bf43ae0e1eL73-L89) you'll see that we no longer use to_sql, so there actually isn't anywhere to put a chunksize! Instead we're just using an IOBuffer which is far more performant and simply pulls raw data as needed.

tests/files/test_file.py
dimberman: @tatiana is that ok? :)

src/astro/databases/postgres.py
```diff
@@ -70,20 +73,31 @@ def load_pandas_dataframe_to_table(
...
-        schema = None
         if target_table.metadata and target_table.metadata.schema:
             self.create_schema_if_needed(target_table.metadata.schema)
-            schema = target_table.metadata.schema.lower()
-        source_dataframe.to_sql(
-            target_table.name,
-            schema=schema,  # type: ignore
-            con=self.sqlalchemy_engine,
+
+        # create an empty table
```
utkarsharma2, line 79: @dimberman - We have a standard function to create tables - https://github.com/astronomer/astro-sdk/pull/538. I think it would be a good idea to standardize the table creation.

src/astro/databases/postgres.py
```diff
@@ -70,20 +73,31 @@ def load_pandas_dataframe_to_table(
...
+            target_table.name.lower(),
+            schema=target_table.metadata.schema,
             if_exists=if_exists,
-            chunksize=chunk_size,
-            method="multi",
             index=False,
         )
 
+        output_buffer = io.StringIO()
+        source_dataframe.to_csv(output_buffer, sep=",", header=True, index=False)
```
utkarsharma2, line 90: @dimberman - how do we know the sep="," and header=True are always applied to all CSV? are we in control of how this CSV file is generated?

src/astro/files/base.py
```diff
@@ -72,12 +73,21 @@ def create_from_dataframe(self, df: pd.DataFrame) -> None:
             self.type.create_from_dataframe(stream=stream, df=df)
 
     def export_to_dataframe(self, **kwargs) -> pd.DataFrame:
-        """Read file from all supported location and convert them into dataframes"""
+        """Read file from all supported location and convert them into dataframes.
+
+        Due to noted issues with using smart_open with pandas (like
```
utkarsharma2, line 78: Thank for the comment was helpful! 👍

src/astro/files/base.py
```diff
@@ -72,12 +73,21 @@ def create_from_dataframe(self, df: pd.DataFrame) -> None:
...
+        https://github.com/RaRe-Technologies/smart_open/issues/524), we create a BytesIO or StringIO buffer
+        before exporting to a dataframe. We've found a sizable speed improvement with this optimization.
+        """
+
         mode = "rb" if self.is_binary() else "r"
```
utkarsharma2, line 83: @dimberman - maybe we can move this to a function of it's own, might be useful at other locations.

tests/benchmark/dags/evaluate_load_file.py
utkarsharma2, line 49:
```suggestion
        chunk_size = int(os.getenv("AIRFLOW__ASTRO_SDK__CHUNK_SIZE", str(DEFAULT_CHUNK_SIZE)))
```

src/astro/databases/postgres.py
dimberman: @utkarsharma2 yes we are. The CSV is being generated by pandas (because we first put the data into a dataframe). This makes the user interface much easier because otherwise we'd need to expose all the options.

src/astro/databases/postgres.py
dimberman: @tatiana this should also address your question about offering support for custom columns as well

src/astro/files/base.py
dimberman: @utkarsharma2 good call. For now I've moved it into a private function, so if later someone has a different use for it they can make it public.

tests/benchmark/dags/evaluate_load_file.py
dimberman: @utkarsharma2 this isn't really an astro config. It's just an env variable we set for the benchmarking.

tests/files/test_file.py
tatiana: Hey @dimberman if the value of read_data didn't change, we could do something like this:
`@patch( "astro.files.base.smart_open.open", new_callable=mock_open, read_data="data")`
But since it changes, I believe it would add much complexity to have this fixture dependent on the filetype! So it's okay, thank you!

tatiana (approved):

tatiana:
@dimberman Sorry for the delay!
> @tatiana can you point me to where this work is being done (re: allowing users to define their schema)?
I recently added it to the base class, so it should work for Postgres as well:
https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/base.py#L279-L282
https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/base.py#L204
https://github.com/astronomer/astro-sdk/blob/main/src/astro/databases/base.py#L167

Issue #534, opened 2022-07-13
Use native path to load local file to Bigquery


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we are transferring the data from local to Bigquery by first creating the dataframe and then using pandas.to_sql() to ingest data into bigquery. 

**Describe the solution you'd like**
Bigquery offers a python SDK to directly ingest data and we can save time that is wasted on converting files into data frames.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #535, opened 2022-07-13
Add native path from local to bigquery

utkarsharma2:
# Description
Currently, we are transferring the data from local to Bigquery by first creating the dataframe and then using pandas.to_sql() to ingest data into bigquery.

closes: #534 
related: #429 

## What is the new behavior?
We intend to use google's a python SDK to directly ingest data and we can save time that is wasted on converting files into data frames.

Nope

### DIFF
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -51,11 +51,6 @@ class BigqueryDatabase(BaseDatabase):
         FileLocation.S3: "load_s3_file_to_bigquery",
         FileLocation.LOCAL: "load_local_file_to_bigquery",
     }
-    NATIVE_PATHS_SUPPORTED_FILE_TYPES = {
-        FileType.CSV: "CSV",
-        FileType.NDJSON: "NEWLINE_DELIMITED_JSON",
-        FileType.PARQUET: "PARQUET",
-    }
 
     illegal_column_name_chars: List[str] = ["."]
     illegal_column_name_chars_replacement: List[str] = ["_"]
@@ -322,13 +317,18 @@ class BigqueryDatabase(BaseDatabase):
     ) -> None:
         """Transfer data from local to bigquery"""
         native_support_kwargs = native_support_kwargs or {}
+        # We need to maintain file_type to biqquery_format and not use NATIVE_PATHS_SUPPORTED_FILE_TYPES
+        # because the load_table_from_file expects 'JSON' value for ndjson file.
+        file_types_to_bigquery_format = {
+            FileType.CSV: "CSV",
+            FileType.NDJSON: "JSON",
+            FileType.PARQUET: "PARQUET",
+        }
 
         write_disposition_val = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
         client = self.hook.get_client()
         config = {
-            "source_format": self.NATIVE_PATHS_SUPPORTED_FILE_TYPES[
-                source_file.type.name
-            ],
+            "source_format": file_types_to_bigquery_format[source_file.type.name],
             "create_disposition": "CREATE_IF_NEEDED",
             "write_disposition": write_disposition_val[if_exists],
             "autodetect": True,
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -336,9 +336,9 @@ class BigqueryDatabase(BaseDatabase):
         config.update(native_support_kwargs)
         job_config = bigquery.LoadJobConfig(config)
 
-        with open(source_file.path, "rb") as source_file:
+        with open(source_file.path, "rb") as file:
             job = client.load_table_from_file(
-                source_file,
+                file,
                 job_config=job_config,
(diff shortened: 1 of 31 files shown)


src/astro/databases/google/bigquery.py
```diff
@@ -49,6 +49,12 @@ class BigqueryDatabase(BaseDatabase):
     NATIVE_PATHS = {
         FileLocation.GS: "load_gs_file_to_bigquery",
         FileLocation.S3: "load_s3_file_to_bigquery",
+        FileLocation.LOCAL: "load_local_file_to_bigquery",
+    }
+    NATIVE_PATHS_SUPPORTED_FILE_TYPES = {
+        FileType.CSV: "CSV",
+        FileType.NDJSON: "NEWLINE_DELIMITED_JSON",
```
dimberman, line 56: Maybe we should make these members of the parent class? It feels weird to have to redefine them for each class

src/astro/databases/google/bigquery.py
```diff
@@ -306,6 +312,38 @@ def get_project_id(self, target_table) -> str:
...
+                source_file.type.name
+            ],
+            "create_disposition": "CREATE_IF_NEEDED",
+            "write_disposition": write_disposition_val[if_exists],
+            "autodetect": True,
+        }
+        config.update(native_support_kwargs)
+        job_config = bigquery.LoadJobConfig(config)
+
+        with open(source_file.path, "rb") as file:
```
dimberman, line 339: How do we know the file is always binary?

dimberman (dismissed):

src/astro/databases/google/bigquery.py
utkarsharma2: @dimberman  Thanks, we don't. So I have added a property in FileTypes to convey the open modes - r/rb and used it in the open() function.

src/astro/databases/google/bigquery.py
utkarsharma2: @dimberman  It's a little weird that they require to pass a different value for NDJSON file types. That's why I maintained two different dicts. I cleared this a bit, hope you like it. :)

src/astro/files/types/parquet.py
```diff
@@ -9,6 +9,8 @@
 class ParquetFileType(FileType):
     """Concrete implementation to handle Parquet file type"""
 
+    open_mode = "rb"
```
tatiana, line 12: It's great to see us extending the classes created with the big refactor!

src/astro/databases/google/bigquery.py
```diff
@@ -306,6 +307,43 @@ def get_project_id(self, target_table) -> str:
...
+        target_table: Table,
+        if_exists: LoadExistStrategy = "replace",
+        native_support_kwargs: Optional[Dict] = None,
+        **kwargs,
+    ) -> None:
+        """Transfer data from local to bigquery"""
+        native_support_kwargs = native_support_kwargs or {}
+        # We need to maintain file_type to biqquery_format and not use NATIVE_PATHS_SUPPORTED_FILE_TYPES
+        # because the load_table_from_file expects 'JSON' value for ndjson file.
+        file_types_to_bigquery_format = {
```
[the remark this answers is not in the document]
utkarsharma2: @tatiana I think we should keep it in this place only. Since they are both used for the same thing, the google upstream API requires us to pass different values for NDJSON - NEWLINE_DELIMITED_JSON/JSON. So keeping mapping here will avoid some confusion.

src/astro/databases/google/bigquery.py
```diff
@@ -306,6 +307,43 @@ def get_project_id(self, target_table) -> str:
...
+            FileType.CSV: "CSV",
+            FileType.NDJSON: "JSON",
+            FileType.PARQUET: "PARQUET",
+        }
+
+        write_disposition_val = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
```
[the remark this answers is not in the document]
utkarsharma2: Yes, should move! I'll do it.

src/astro/databases/google/bigquery.py
dimberman: Ah yeah so I guess if it's text we need to encode it and pass it as binary.

utkarsharma2: Yes, for us it will always be binary read mode, irrespective of the file type.

tatiana (approved):

src/astro/databases/google/bigquery.py
```diff
@@ -302,6 +302,49 @@ def get_project_id(self, target_table) -> str:
         except AttributeError:
             raise ValueError(f"conn_id {target_table.conn_id} has no project id")
 
+    def load_local_file_to_bigquery(
```
tatiana, line 305: Since we are within the `BigqueryDatabase`, perhaps  this should be named `load_local_file_to_table`

utkarsharma2: Done.

tatiana:
@utkarsharma2 could you add to our benchmark metrics the numbers improvement achieved by this PR?

PR #537, opened 2022-07-14
Move config to store DF in XCom to settings file

kaxil:
This way if we want to ever rename the config, we only need to do it at a single place and it is easy to find which configs users can control.

At some point we should have a doc page that shows all the available configs

No

### Checklist

### DIFF
--- a/src/astro/sql/operators/dataframe.py
+++ b/src/astro/sql/operators/dataframe.py
@@ -2,12 +2,12 @@ import inspect
 from typing import Callable, Dict, Optional, Tuple, Union
 
 import pandas as pd
-from airflow.configuration import conf
 from airflow.decorators.base import DecoratedOperator
 
 from astro.constants import ColumnCapitalization
 from astro.databases import create_database
 from astro.exceptions import IllegalLoadToDatabaseException
+from astro.settings import ALLOW_UNSAFE_DF_STORAGE, IS_CUSTOM_XCOM_BACKEND
 from astro.sql.table import Table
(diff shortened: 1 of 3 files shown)


kaxil:
hm test is failing, I will re-open the PR when I am back to take a look

utkarsharma2 (approved):

pankajkoti (approved):

kaxil:
hmm test is still failing so will take a look and fix it

dimberman (approved):

dimberman:
```
sample_dag = <DAG: tv0gd8qx8v932rxh>

    @mock.patch.dict(
        os.environ, ***"AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE": "False"***
    )
    def test_unsafe_loading_of_dataframe(sample_dag):
        with pytest.raises(IllegalLoadToDatabaseException):
>           load_file(
                input_file=File(
                    "https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/homes_main.csv"
                ),
            ).operator.execute(***)
E           Failed: DID NOT RAISE <class 'astro.exceptions.IllegalLoadToDatabaseException'>
```

Looks like the boolean statement isn't being tripped off for some reason. I'd recommend running the test to print out the value of ALLOW_UNSAFE_DF_STORAGE, XCOM_BACKEND, and IS_CUSTOM_XCOM_BACKEND. It's possible one of them is getting an unexpected value.

Issue #525, opened 2022-07-12
aql.cleanup() will hang if there are failed tasks.


guohui-gao:
**Describe the bug**
I added aql.cleanup() to a dag to clear temp tables.
it will run after all task success. however, if there are failures in the tasks. it will hang, with the 
following errors:

[2022-07-12, 09:19:35 UTC] {cleanup.py:103} INFO - waiting on the following tasks to complete before cleaning up: [('complete', 'upstream_failed'), ('tear_down_tasks', 'removed'), ('create_staging_comments', 'upstream_failed'), ('merge_comments', 'upstream_failed'), ('upload_comments_to_gcs', 'upstream_failed'), ('extract_finish', 'upstream_failed'), ('create_comments', 'upstream_failed'), ('load_comments_to_snowflake_staging_db', 'upstream_failed'), ('cleanup_staging_comments', 'upstream_failed')]
[2022-07-12, 09:19:35 UTC] {cleanup.py:137} WARNING - You are currently running the 'waiting mode', where the task will wait for all other tasks to complete. Please note that for asynchronous executors (e.g. SequentialExecutor and DebugExecutor, this mode will cause locks.

**Version**
* Astro: [e.g. 0.11.0]
* OS: [eg. Debian]

**Expected behavior**
Cleanup would run whether all tasks success or if there are any task fail.

PR #540, opened 2022-07-15
Add missing task terminal states to CleanUp Operator

utkarsharma2:
# Description
When adding aql.cleanup() to a dag to clear temp tables. It will run after all task success. however, if there are failures in the tasks. it will hang up because some of the terminal states are not checked.

closes: #525

## What is the new behavior?
Added terminal states to check logic

No

## Alternative Solution
https://github.com/astronomer/astro-sdk/pull/541

### Checklist

### DIFF
--- a/src/astro/sql/operators/cleanup.py
+++ b/src/astro/sql/operators/cleanup.py
@@ -97,7 +97,15 @@ class CleanupOperator(BaseOperator):
             (ti.task_id, ti.state)
             for ti in task_instances
             if ti.task_id != self.task_id
-            and ti.state not in [State.SUCCESS, State.FAILED, State.SKIPPED]
+            and ti.state
+            not in [
+                State.SUCCESS,
+                State.FAILED,
+                State.SKIPPED,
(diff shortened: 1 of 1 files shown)


src/astro/sql/operators/cleanup.py
```diff
@@ -97,7 +97,15 @@ def _is_dag_running(self, task_instances: List[TaskInstance]) -> bool:
             (ti.task_id, ti.state)
             for ti in task_instances
             if ti.task_id != self.task_id
-            and ti.state not in [State.SUCCESS, State.FAILED, State.SKIPPED]
+            and ti.state
+            not in [
```
utkarsharma2, line 101: @kaxil @dimberman - am I missing any terminal states here?

dimberman (approved):

src/astro/sql/operators/cleanup.py
dimberman: @utkarsharma2 as far as I can tell this is all of them. We can make another PR if we find any more edge cases.

PR #541, opened 2022-07-15
Add all done trigger rule to cleanup op

utkarsharma2:
# Description
Current Implementation:
When adding aql.cleanup() to a dag to clear temp tables. It will run after all task success. however, if there are failures in the tasks.

closes: #525

## What is the new behavior?
Added trigger rule 'all_done' to the cleanup operator, so it executes at the end of the dag run and clear all the temp tables. 

One advantage of this is we don't have to keep iterating over all operators and checking the states. cleanup operator will only execute when all task are done. 

Nope.

### Checklist

### DIFF
--- a/src/astro/sql/operators/cleanup.py
+++ b/src/astro/sql/operators/cleanup.py
@@ -62,7 +62,11 @@ class CleanupOperator(BaseOperator):
         task_id = task_id or get_unique_task_id("_cleanup")
 
         super().__init__(
-            task_id=task_id, retries=retries, retry_delay=retry_delay, **kwargs
+            task_id=task_id,
+            retries=retries,
+            retry_delay=retry_delay,
+            trigger_rule="all_done",
+            **kwargs,
         )
 
     def execute(self, context: Context) -> None:


src/astro/sql/operators/cleanup.py
```diff
@@ -62,7 +62,11 @@ def __init__(
         task_id = task_id or get_unique_task_id("_cleanup")
 
         super().__init__(
-            task_id=task_id, retries=retries, retry_delay=retry_delay, **kwargs
+            task_id=task_id,
+            retries=retries,
+            retry_delay=retry_delay,
+            trigger_rule="all_done",
```
utkarsharma2, line 68: @kaxil @dimberman @tatiana 
1. Can we have two tasks with the 'all_done' trigger rule? 
2. If the answer to the previous question is yes, how is the order of execution decided?

tatiana: @utkarsharma2, this looks great. Could you confirm what the DAG looks like in the Airflow interface? If it has the same behaviour as now (that cleanup is not connected to all previous tasks, it feels we should go ahead with this implementation).

utkarsharma2:
Closing this since adding a task with an `all-done` trigger rule requires adding upstream dependencies to the task and will not be feasible for the end user to do. Plus it will make the dag view look messy.

Issue #304, opened 2022-04-11
Users should be able to set `tmp_schema` for each connection


pgzmnk:
**Please describe the feature you'd like to see**
Currently `tmp_schema` must exist on the database to which a SQL tasks will write output to. Feedback sessions with Astro Build surfaced that requiring a specific schema with write access to exist might be an unreasonable ask for a db admin. 

There's a feature to set the temp schema with the `AIRFLOW__ASTRO__SQL_SCHEMA` environment variable. However, forcing the same temp schema name for all conns is untenable. For example, conn1 might use the `tmp_schema` while conn2 might use a user defined `default_schema`. 

**Describe the solution you'd like**
A possible solution could be to let users specify the temp schema for each conn in a new field in 'extras' of the conn URI. 

This means that SQL Tasks write temp tables to the schema defined by the e.g. `tmp_schema` attribute in the connection's 'extras' section, if that attribute exists. Otherwise write to the default temp schema.

**Additional context**
This issue was encountered within the Astro Build UI.

**Acceptance Criteria**

- [] Users are able to define a schema

tatiana:
That's excellent feedback, @pgzmnk!

A PR in progress (#140) addresses the same issue but uses a different approach. It allows users to define default values per DAG instead of relying on Astro 0.7.x temporary schema environment variable. What are your thoughts about this alternative approach?

I like your proposal because it is focused on the database, as opposed to DAGs - it is more flexible and would allow a single DAG powered by Astro to interact with >1 database, which can have different temporary schemas.

pgzmnk:
This makes sense. Let's see what user feedback we get based on your proposed implementation.

Thank you for being on top of this.

tatiana:
@pgzmnk I am looking into implementing your proposal.
One of the challenges, for instance, is that Postgres expects the connection only to contain pre-defined fields:

tatiana:
For those who may be interested, the issues that happened with Postgres when trying to use this approach are documented in the branch `temp-schema-from-conn`: https://github.com/astronomer/astro-sdk/commit/8b0dfecb41a036bd849c185ec49377c7938cb5db

kaxil:
@feluelle @sunank200 Any updates on this issue.

This was brought up again internally at https://astronomer.slack.com/archives/C02B8SPT93K/p1663278078864369

PR #542, opened 2022-07-15
Add schema settings on database level

utkarsharma2:
# Description
When a user is using Astro-SDK for multiple databases. We impose a requirement of having the same-named SCHEMA in all the databases. Which is unrealistic.

related: #304

## What is the new behavior?
Now we have defined config variables to define SCHEMA per-database level.

Users can either configure env var:
AIRFLOW_ASTRO_SDK__POSTGRES_SQL_SCHEMA
AIRFLOW_ASTRO_SDK__BIGQUERY_SQL_SCHEMA
AIRFLOW_ASTRO_SDK__SNOWFLAKE_SQL_SCHEMA
or define them in airflow's config.

Nope

### Checklist

### DIFF
--- a/docs/getting-started/GETTING_STARTED.md
+++ b/docs/getting-started/GETTING_STARTED.md
@@ -392,3 +392,458 @@ or
 
 In all scenarios, even if the user gives a non-temporary table, only temporary
 tables will actually be deleted.
+
+## Configuring default schema
+
+We can configure the default schema that will be used for all operation involving database.
+environment variable :
+```shell
+AIRFLOW__ASTRO_SDK__SCHEMA="tmp"
+```
+or by updating Airflow's configuration
+```editorconfig
+[astro_sdk]
+schema = "tmp"
+```
+We can fur<h1 align="center">
+  Astro Python SDK
+</h1>
+  <h3 align="center">
+  Tutorial<br><br>
+</h3>
+
+# Introduction
+
+This tutorial demonstrates how to use the Astro Python SDK through a simple ETL example that you can run on your local machine:
+* **Extract** a file from S3 into a Snowflake relational table
+* **Transform** that table in Snowflake
+* **Load** that transformed table into a reporting table in Snowflake
+
+You'll need to use existing Amazon S3 and Snowflake accounts, or create some trial versions.
+
+***
+# Let's do some setup
+
+## Install Airflow
+
+* Install Python and `virtualenv` (ignore if already installed).
+
+  For macOS, you can follow these steps:
+    * Install homebrew:
+
+      ```shell
+      /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+      ```
+    * `brew install python`
+    * `brew install pip`
+    * `pip install virtualenv `
+
+* Create a new virtual environment by running:
+
+  ```shell
+  python3 -m virtualenv env
+  ```
+
+* Activate environment ``source env/bin/activate``
+* [Install Airflow](https://airflow.apache.org/docs/apache-airflow/stable/start/local.html)
+
+## Set up your data stores
+
+### Set up S3
+
+[Setup](https://docs.aws.amazon.com/AmazonS3/latest/userguide/GetStartedWithS3.html) an S3 account and storage bucket.
+
+Upload this datafile from your local machine to your newly created aws bucket
+
+```
+order_id,customer_id,purchase_date,amount
+ORDER1,CUST1,1/1/2021,100
+ORDER2,CUST2,2/2/2022,200
+ORDER3,CUST3,3/3/2023,300
+```
+
+### Set up Snowflake
+
+Sign up for a [free Snowflake trial](https://signup.snowflake.com/), selecting the Enterprise Edition.
+
+Once your workspace is ready, navigate to a worksheet and enter these commands:
+
+```sql
+create warehouse ASTRO_SDK_DW;
+create database ASTRO_SDK_DB;
+create schema ASTRO_SDK_SCHEMA;
+```
+Then, in the classic Snowflake console in the upper right, select the role, warehouse, database and schema
+to look like this:
+
+![classic-console](../images/snowflake-classic-console.png)
+
+Note that you can set the warehouse, database, and schema names to something else, but you'll just
+need to be consistent with whatever names you choose throughout the remainder of this tutorial.
+
+## Install Astro-SDK on your local machine
+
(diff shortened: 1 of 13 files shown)


utkarsharma2:
@tatiana @kaxil @dimberman Where should we document this feature?

pankajastro:
> @tatiana @kaxil @dimberman Where should we document this feature?
> 
> My assumption is that README.md and GETTING_STARTED.md should just focus on getting people familiar with the project and this feature is relatively advanced usage, as it handles cases where users are using lib to interact with multiple databases. Please let me know your thoughts.

This is more of a look-like feature so in my opinion, if we have an example then should add this as described there.

pankajastro:
Question: let's say I have configured `postgres_sql_schema` in airflow config and the value is `airflow_pg`  then what should be the response of `default_metadata` method  `tmp_astro` or `airflow_pg` after this change look like it would return `airflow_pg`? I think the `default_metadata` should always return the default value that we provide WDYT?

src/astro/settings.py
```diff
@@ -1,7 +1,12 @@
 from airflow.configuration import conf
 
-DEFAULT_SCHEMA = "tmp_astro"
+from astro.constants import DEFAULT_SCHEMA
+
 SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=DEFAULT_SCHEMA)
+POSTGRES_SCHEMA = conf.get("astro_sdk", "postgres_sql_schema", fallback=SCHEMA)
+BIGQUERY_SCHEMA = conf.get("astro_sdk", "bigquery_sql_schema", fallback=SCHEMA)
+SNOWFLAKE_SCHEMA = conf.get("astro_sdk", "snowflake_sql_schema", fallback=SCHEMA)
```
tatiana, line 8: WDYT of the following names?
```
bigquery_default_schema
postgres_sql_schema
snowflake_sql_schema
```

dimberman: +1

dimberman:
@kaxil since we now have a cleanup operator (and soon a cleanup DAG teardown command) should we maybe start moving away from tmp schemas? Or should we stay the course for now until 2.4?

tatiana:
> @tatiana @kaxil @dimberman Where should we document this feature?
> 
> My assumption is that README.md and GETTING_STARTED.md should just focus on familiarising people with the project. This feature is relatively advanced, as it handles cases where users use lib to interact with multiple databases. Please let me know your thoughts.

@utkarsharma2 I believe in any place the referenced the previous environment variable: `AIRFLOW__ASTRO_SDK__SQL_SCHEMA`. IIR, `docs/getting-started/GETTING_STARTED.md`. So this is not a complete breaking change, should we still continue supporting `AIRFLOW__ASTRO_SDK__SQL_SCHEMA` - as the default temp schema, in case the user does not define the specific ones?

src/astro/settings.py
kaxil: +1

kaxil:
> > @tatiana @kaxil @dimberman Where should we document this feature?
> > My assumption is that README.md and GETTING_STARTED.md should just focus on familiarising people with the project. This feature is relatively advanced, as it handles cases where users use lib to interact with multiple databases. Please let me know your thoughts.
> 
> @utkarsharma2 I believe in any place the referenced the previous environment variable: `AIRFLOW__ASTRO_SDK__SQL_SCHEMA`. IIR, `docs/getting-started/GETTING_STARTED.md`. So this is not a complete breaking change, should we still continue supporting `AIRFLOW__ASTRO_SDK__SQL_SCHEMA` - as the default temp schema, in case the user does not define the specific ones?

Ideally, we would have a page like https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html where we expose all the configurations and some lines on when to use which. But that can be it's own separate PR

src/astro/databases/base.py
```diff
@@ -158,7 +159,7 @@ def populate_table_metadata(self, table: Table) -> Table:
         if table.metadata and table.metadata.is_empty() and self.default_metadata:
```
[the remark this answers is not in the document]
utkarsharma2:
```
if table.metadata and table.metadata.is_empty() and self.default_metadata:  ## ---- 1
    table.metadata = self.default_metadata
if not table.metadata.schema:    ## ---- 2
    table.metadata.schema = self.DEFAULT_SCHEMA
```
@tatiana I think this scenario should be covered by the existing code. 

When table don't have metadata, will be assigned by 1.
Then we check if table has Schema or not if not we assign it from self.DEFAULT_SCHEMA -- 2

Also, self.DEFAULT_SCHEMA will always have value, because we use code(setting.py)
`SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=DEFAULT_SCHEMA)`
Where `DEFAULT_SCHEMA` is a constant.

src/astro/settings.py
utkarsharma2: Done.

src/astro/databases/base.py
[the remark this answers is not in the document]
tatiana: Makes sense, @utkarsharma2 !

docs/getting-started/GETTING_STARTED.md
```diff
@@ -392,3 +392,458 @@ or
...
+### bigquery
+environment variable
+```shell
+AIRFLOW__ASTRO_SDK__BIGQUERY_DEFAULT_SCHEMA="tmp"
+```
+or Airflow's configuration
+```
+[astro_sdk]
+bigquery_default_schema = "tmp"
+```
```
tatiana, line 849: It's a bit strange that all these lines are being introduced as new in this PR. They should be part of the existing tutorial.

docs/getting-started/GETTING_STARTED.md
```diff
@@ -392,3 +392,458 @@ or
...
+```editorconfig
+[astro_sdk]
+schema = "tmp"
+```
+We can fur<h1 align="center">
+  Astro Python SDK
+</h1>
+  <h3 align="center">
+  Tutorial<br><br>
+</h3>
```
tatiana, line 413: This should probably live in a dedicated page, within our Sphinx documents, as proposed by @kaxil

tatiana (commented):
@utkarsharma2, we're super close to merging this PR, just two last bits documentation-wise!

docs/getting-started/GETTING_STARTED.md
utkarsharma2: Make sense, I'll move it.

utkarsharma2:
Raising a separate PR since this one has lots of conflicts - https://github.com/astronomer/astro-sdk/pull/617

PR #550, opened 2022-07-18
Support running benchmark on load_file using native support

tatiana:
Based on the environment variable `ASTRO_ENABLE_NATIVE_LOAD`.
Enabled by default.

### DIFF
--- a/tests/benchmark/dags/evaluate_load_file.py
+++ b/tests/benchmark/dags/evaluate_load_file.py
@@ -47,6 +47,7 @@ def create_dag(database_name, table_args, dataset):
 
     with DAG(dag_name, schedule_interval=None, start_date=START_DATE) as dag:
         chunk_size = int(os.getenv("ASTRO_CHUNK_SIZE", str(DEFAULT_CHUNK_SIZE)))
+        use_native_support = bool(int(os.getenv("ASTRO_ENABLE_NATIVE_LOAD", "1")))
         table_metadata = table_args.pop("metadata", {})
         if table_metadata:
             table = Table(metadata=Metadata(**table_metadata), **table_args)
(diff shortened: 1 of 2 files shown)


tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -47,6 +47,7 @@ def create_dag(database_name, table_args, dataset):
 
     with DAG(dag_name, schedule_interval=None, start_date=START_DATE) as dag:
         chunk_size = int(os.getenv("ASTRO_CHUNK_SIZE", str(DEFAULT_CHUNK_SIZE)))
+        use_native_support = bool(int(os.getenv("ASTRO_ENABLE_NATIVE_LOAD", "1")))
```
utkarsharma2, line 50:
```suggestion
        use_native_support = conf.getboolean("astro_sdk", "enable_native_load", fallback=True)
```

utkarsharma2, line 50: @tatiana do you think we can move this feature to `config.json` `datasets` section? and that way we can run even the comparisons with native/non native paths?

tests/benchmark/run.sh
```diff
@@ -76,7 +76,7 @@ echo - Output: $(get_abs_filename $results_file)
           cat $results_file
 
           if [[ -z "${GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
```
utkarsharma2, line 78: @tatiana question: is there a recommended way to get credentials? since we are getting it from the env variable only?

dimberman: Yeah I had a lot of trouble with this, google doesn't make this easy so it would be great to have some basic docs on how to set it all up.

Issue #555, opened 2022-07-18
Create a TruncateOperator


dimberman:
Since the current TruncateOperator is essentially a DropTableOperator, we will need to build a TruncateOperator from scratch. 

A truncate operator should *not* delete a table. Rather, it should clear out all rows and leave the table empty. 

Acceptance criteria:
- [ ] A TruncateOperator class that clears all rows without deleting the table
- [ ] A corresponding `aql.truncate` function the enacts the TruncateOperator
- [ ] A new set of tests for the described functionality

Example test:

```python


@pytest.mark.integration
@pytest.mark.parametrize(
    "database_table_fixture",
    [
        {
            "database": Database.SQLITE,
            "file": File(DEFAULT_FILEPATH),
        },
        {
            "database": Database.POSTGRES,
            "file": File(DEFAULT_FILEPATH),
        },
        {
            "database": Database.BIGQUERY,
            "file": File(DEFAULT_FILEPATH),
        },
        {
            "database": Database.SNOWFLAKE,
            "file": File(DEFAULT_FILEPATH),
        },
    ],
    indirect=True,
    ids=["sqlite", "postgres", "bigquery", "snowflake"],
)
def test_truncate_with_table_metadata(database_table_fixture, sample_dag):
    """Test truncate operator for all databases."""
    database, test_table = database_table_fixture
    assert database.table_exists(test_table)

    @aql.dataframe
    def validate(df: pandas.DataFrame):
        assert len(df) == 0
    with sample_dag:
        aql.truncate(
            table=test_table,
        )
    test_utils.run_dag(sample_dag)

    assert database.table_exists(test_table)
```

PR #556, opened 2022-07-18
Rename aql.truncate to aql.drop_table and fix dependency issue

### LINKED ISSUE #515: Issue when truncating after load_file with output_table

### LINKED ISSUE #554: Convert current TruncateOperator into a DropTableOperator
The current TruncateOperator is implemented to drop tables instead of truncating them. This could cause critical issues with users who expect tables to exist after truncation. While not the correct action, there IS use for a DropTableOperator. 

Acceptance criteria:

- [ ] Create an operator that given a table will drop said table.
- [ ] If the user does not have proper permissions to drop the table, give proper warning in the form of an error
- [ ] create a corresponding AQL function

Example:
```
@pytest.mark.integration
@pytest.mark.parametrize(
    "database_table_fixture",
    [
        {
            "database": Database.SQLITE,
            "file": File(DEFAULT_FILEPATH),
        },
        {
            "database": Database.POSTGRES,
            "file": File(DEFAULT_FILEPATH),
        },
        {
            "database": Database.BIGQUERY,
            "file": File(DEFAULT_FILEPATH),
        },
        {
            "database": Database.SNOWFLAKE,
            "file": File(DEFAULT_FILEPATH),
        },
    ],
    indirect=True,
    ids=["sqlite", "postgres", "bigquery", "snowflake"],
)
def test_truncate_with_table_metadata(database_table_fixture, sample_dag):
    """Test truncate operator for all databases."""
    database, test_table = database_table_fixture
    assert database.table_exists(test_table)

    with sample_dag:
        aql.drop_table(
            table=test_table,
        )

    assert not database.table_exists(test_table)
```

dimberman:
# Description
Currently the functionality of the TruncateOperator actually acts like a DropTable Operator. We should therefore move this functionality into an `aql.drop_table` function and create a separate `truncate` function in a different PR.

Closes #554
Fixes #515

## What is the new behavior?
Now the function name properly reflects the actions performed. This is a critical bug as people can accidentally drop tables. 

Yes, but we can mitigate breaking changes by implementing #555 . It's worth noting that the behavior we're breaking is itself broken behavior.

### Checklist

### DIFF
--- a/src/astro/sql/operators/drop.py
+++ b/src/astro/sql/operators/drop.py
@@ -7,8 +7,10 @@ from astro.databases import create_database
 from astro.sql.table import Table
 
 
-class TruncateOperator(BaseOperator):
-    """Airflow Operator for truncating SQL tables."""
+class DropTableOperator(BaseOperator):
+    """Airflow Operator for dropping SQL tables."""
+
+    template_fields = ("table",)
 
     def __init__(
         self,
@@ -17,14 +19,15 @@ class TruncateOperator(BaseOperator):
         **kwargs,
     ):
         self.table = table
-        task_id = task_id or get_unique_task_id(table.name + "_truncate")
+        task_id = task_id or get_unique_task_id("_drop")
         super().__init__(
             task_id=task_id,
             **kwargs,
         )
 
-    def execute(self, context: Dict) -> None:  # skipcq: PYL-W0613
+    def execute(self, context: Dict) -> Table:  # skipcq: PYL-W0613
         """Method run when the Airflow runner calls the operator."""
         database = create_database(self.table.conn_id)
         self.table = database.populate_table_metadata(self.table)
         database.drop_table(self.table)
+        return self.table
--- a/example_dags/example_snowflake_partial_table_with_append.py
+++ b/example_dags/example_snowflake_partial_table_with_append.py
@@ -9,10 +9,10 @@ from astro.sql import (
     append,
     cleanup,
     dataframe,
+    drop_table,
     load_file,
     run_raw_sql,
     transform,
-    truncate,
 )
 from astro.sql.table import Metadata, Table
 
@@ -136,7 +136,7 @@ def example_snowflake_partial_table_with_append():
     # We truncate this table only to avoid wasting Snowflake resources
     # Why? Between 2022-03-25 and 2022-04-11 it accumulated 301G (89 million rows) because
     # this example DAG used to append rows without deleting them
-    truncate_results = truncate(
+    truncate_results = drop_table(
         table=Table(name="homes_reporting", conn_id=SNOWFLAKE_CONN_ID)
     )
     truncate_results.set_upstream(record_results)
--- a/src/astro/sql/__init__.py
+++ b/src/astro/sql/__init__.py
@@ -12,12 +12,12 @@ from astro.constants import MergeConflictStrategy
 from astro.sql.operators.append import APPEND_COLUMN_TYPE, AppendOperator
 from astro.sql.operators.cleanup import CleanupOperator
 from astro.sql.operators.dataframe import DataframeOperator
+from astro.sql.operators.drop import DropTableOperator
 from astro.sql.operators.export_file import export_file  # noqa: F401
 from astro.sql.operators.load_file import load_file  # noqa: F401
 from astro.sql.operators.merge import MERGE_COLUMN_TYPE, MergeOperator
 from astro.sql.operators.raw_sql import RawSQLOperator
 from astro.sql.operators.transform import TransformOperator  # noqa: F401
-from astro.sql.operators.truncate import TruncateOperator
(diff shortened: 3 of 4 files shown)


src/astro/sql/operators/drop.py
```diff
@@ -17,14 +19,15 @@ def __init__(
         **kwargs,
     ):
         self.table = table
-        task_id = task_id or get_unique_task_id(table.name + "_truncate")
+        task_id = task_id or get_unique_task_id("_drop")
```
tatiana, line 22: Don't you think having `table.name` as a prefix is more meaningful to users?

src/astro/sql/operators/drop.py
```diff
@@ -17,14 +19,15 @@ def __init__(
...
             **kwargs,
         )
 
-    def execute(self, context: Dict) -> None:  # skipcq: PYL-W0613
+    def execute(self, context: Dict) -> Table:  # skipcq: PYL-W0613
         """Method run when the Airflow runner calls the operator."""
         database = create_database(self.table.conn_id)
         self.table = database.populate_table_metadata(self.table)
         database.drop_table(self.table)
+        return self.table
```
tatiana, line 33: Why do we need to return `Table` here?

tatiana (commented):
@dimberman Thank you very much for the super quick PR on this!

There are some comments in-line.

I believe we also need to update the TUTORIAL with this change - could you confirm if there are any other docs which we may have to update as well?

src/astro/sql/operators/drop.py
utkarsharma2: +1 table objects will mostly be useless since the table is deleted in DB. 

@dimberman is there any case where you think this will be useful?

kaxil (commented):
Can you rename the PR title please - "drop table operator" doesn't say much. Would recommend checking https://cbea.ms/git-commit/ out

tatiana (approved):
Hey @dimberman , as we agreed, I'm approving this PR relying that you'll be doing the following:
- change the PR title
- use table title as task name prefix
- not return table
- rebase

src/astro/sql/operators/drop.py
dimberman: @tatiana maybe this would be a good use for the same type of operator we discussed for the raw_sql? Like an EmptyTable or something?

dimberman:
@tatiana there is no way to use the table name as a prefix for the task id. That's determined at runtime so we would get errors as it would be attempting to process an XcomArg.

src/astro/sql/operators/drop.py
dimberman: https://github.com/astronomer/astro-sdk/pull/556#issuecomment-1194869657

This was an error that made the TruncateOperator not work btw

src/astro/sql/operators/drop.py
dimberman: @tatiana I changed it to a "DropTableResult". There needs to be SOMETHING returned for Xcom to work (@kaxil not sure if this a bug in Airflow we should consider).

src/astro/sql/operators/drop.py
tatiana: @dimberman would there be an alternative way of giving a more friendly default user name, without the user having to specify it?

Would we be able to use templating or something else to give users a more friendly task name in the Airflow UI?

From a user perspective, displaying "drop_table_name" is a better experience than displaying "drop_10924"

src/astro/sql/operators/drop.py
tatiana: @dimberman I chatted to @ashb about this, since @kaxil is off this week.

I have a better understanding of this issue now. It feels, for now, the best option is indeed to return something - so we can extend the existing Task Flow API.

Issue #558, opened 2022-07-19
Fix 'nrows' export_to_dataframe() for NDJSON


utkarsharma2:
**Describe the bug**
Currently export_to_dataframe() is reading entire file and is not respecting `nrows` param.
nrows - https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

**Version**
* Astro: 0.11.0

**To Reproduce**
Read a file using file.export_to_dataframe(nrows=10) the output will be the entire file and not just 10 rows

**Expected behavior**
ile.export_to_dataframe(nrows=10) should only return 10 rows

PR #559, opened 2022-07-19
Handle nrows for export_to_dataframe()

utkarsharma2:
# Description
Currently, export_to_dataframe() is reading the entire file and is not respecting nrows param.
nrows - https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

Read a file using `file.export_to_dataframe(nrows=10)` the output will be the entire file and not just 10 rows

closes: #558

## What is the new behavior?
Read a file using file.export_to_dataframe(nrows=10) the output should  return dataframe with 10 rows

No

### DIFF
--- a/src/astro/files/types/ndjson.py
+++ b/src/astro/files/types/ndjson.py
@@ -32,10 +32,6 @@ class NDJSONFileType(FileType):
         return FileTypeConstants.NDJSON
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


tatiana (approved):

sunank200 (approved):
Looks good! But Deepsource needs some fix

kaxil (approved):

Issue #557, opened 2022-07-19
Implement fallback mechanism incase native support fails to default option


sunank200:
**Please describe the feature you'd like to see**
Currently, if the native support fails, the whole LoadFile fails with relevant errors. Ideally, it should fall back to the default option.

**Describe the solution you'd like**
Implement fallback mechanism incase native support fails to default option

**Acceptance Criteria**
- [ ] Add test for the added code.
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #560, opened 2022-07-19
Implement fallback mechanism incase native support fails to default option

sunank200:
# Description

Currently, if the native support fails, the whole LoadFile fails with relevant errors. Ideally, it should fall back to the default option when specific relevant exceptions are raised.

closes: #557 

## What is the new behavior?

Fall back to the default option when native support fails with specific relevant exceptions are raised.

### Checklist

### DIFF
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -196,7 +196,6 @@ class BigqueryDatabase(BaseDatabase):
         :param target_table: Table that needs to be populated with file data
         :param if_exists: Overwrite file if exists. Default False
         :param native_support_kwargs: kwargs to be used by method involved in native support flow
-        @return: returns True when successful else False
         """
         method_name = self.NATIVE_PATHS.get(source_file.location.location_type)
         if method_name:
@@ -231,7 +230,6 @@ class BigqueryDatabase(BaseDatabase):
         :param target_table: Table that will be created on the bigquery
         :param if_exists: Overwrite table if exists. Default 'replace'
         :param native_support_kwargs: kwargs to be used by method involved in native support flow
-        @return: returns True when successful else False
         """
         native_support_kwargs = native_support_kwargs or {}
 
@@ -287,7 +285,6 @@ class BigqueryDatabase(BaseDatabase):
         :param target_table: Table that will be created on the bigquery
         :param if_exists: Overwrite table if exists. Default 'replace'
         :param native_support_kwargs: kwargs to be used by method involved in native support flow
-        @return: returns True when successful else False
         """
         native_support_kwargs = native_support_kwargs or {}
 
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -1,11 +1,24 @@
 """Google BigQuery table implementation."""
-from typing import Dict, List, Tuple
+import time
+from typing import Dict, List, Optional, Tuple
 
 import pandas as pd
 from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
+from airflow.providers.google.cloud.hooks.bigquery_dts import (
+    BiqQueryDataTransferServiceHook,
+)
 from google.api_core.exceptions import NotFound as GoogleNotFound
+from google.cloud import bigquery_datatransfer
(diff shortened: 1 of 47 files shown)


src/astro/databases/google/bigquery.py
```diff
@@ -255,17 +257,23 @@ def load_gs_file_to_bigquery(
...
         }
-        self.hook.insert_job(
-            configuration=job_config,
-        )
+        try:
+            self.hook.insert_job(
+                configuration=job_config,
+            )
+        # Ignoring deepsource error as it needs to catch every other exception
+        except Exception as exe:  # skipcq: PYL-W0703
```
utkarsharma2, line 265: @sunank200 as per our discussion the other day, we need to handle errors in a way that if there is a config issue we retry with the default path, and if there is a network issue we can fail directly and not try the default path.

We can take this as a separate task since we need to explore what exception an underline lib can raise and then figure out where it is suitable to fail or retry with the default path.

utkarsharma2: Thanks, @bharanidharan14 for pointing it out.

utkarsharma2:
@sunank200 We can also introduce a boolean param to control if we want to fall back to the default path or not.

src/astro/databases/google/bigquery.py
sunank200: Sure. I will add a separate issue and PR for this.

sunank200: Added the change in this [commit](https://github.com/astronomer/astro-sdk/pull/560/commits/48fa4e8145fb54db5ac3f295e3af70d1742c71ff)

dimberman (approved):

src/astro/databases/snowflake.py
```diff
@@ -146,6 +157,22 @@ class SnowflakeDatabase(BaseDatabase):
     logic in other parts of our code-base.
     """
 
+    NATIVE_LOAD_EXCEPTIONS: Any = (
+        ValueError,
```
utkarsharma2, line 161: @sunank200 Why we have ValueError and AttributeError in this list?

src/astro/databases/google/bigquery.py
```diff
@@ -156,15 +192,14 @@ def merge_table(
             ON {' AND '.join(['T.' + col + '= S.' + col for col in target_conflict_columns])}\
             WHEN NOT MATCHED BY TARGET THEN INSERT ({','.join(target_columns)}) VALUES ({','.join(source_columns)})"
 
+        update_statement_map = ", ".join(
```
utkarsharma2, line 195: @sunank200 Why do we have this change in this PR?

tests/databases/test_bigquery.py
```diff
@@ -203,6 +204,54 @@ def test_load_file_to_table(database_table_fixture):
...
+        {
+            "database": Database.BIGQUERY,
+            "table": Table(metadata=Metadata(schema=SCHEMA)),
+        },
+    ],
+    indirect=True,
+    ids=["bigquery"],
+)
+def test_load_file_to_table_natively_for_not_optimised_path(database_table_fixture):
+    """Test loading on files to bigquery natively for non optimized path."""
```
utkarsharma2, line 220: @sunank200  Why is this test needed?

src/astro/databases/google/bigquery.py
sunank200: Deepsource was failing. This solves this deepsource

tests/databases/test_bigquery.py
sunank200: This is unit test for `load_file_to_table_natively`

src/astro/databases/snowflake.py
sunank200: This is to catch ValueError and ArributeError for `run_sql()` method of sqlalchemy. This can be improved by raising these two errors only while calling `run_sql()`

utkarsharma2: As per discussion, we should be handling the generic exception at the source and raise a custom exception that can be handled in via `NATIVE_LOAD_EXCEPTIONS`.

sunank200: Adding separate issues for this and tracking this in separate PR. But this should not be a blocker the release

Issue #564, opened 2022-07-25
Support customization of column titles capitalization when using dataframe & load_file


tatiana:
**Context**

To achieve a first end-to-end delivery of the Snowflake optimization:
https://github.com/astronomer/astro-sdk/pull/487
https://github.com/astronomer/astro-sdk/pull/544

There was an assumption that, to avoid issues with Snowflake mixed-capitalized column names, it was acceptable to convert all characters to uppercase. However, there are two issues with this approach:
* it is not consistent with the dataframe decorator parameter: `identifiers_as_lower`
* it does not support the use case where users may prefer to convert everything to uppercase

**Proposal**

The goal with this ticket is to add the following argument to both `aql.dataframe` and `aql.load_file` (snowflake-only), removing the previous `identifiers_as_lower` parameter:
```
columns_names_capitalization=["upper", "lower", "original"]
```

**Acceptance criteria**
* [ ] Replace the `aql.dataframe` argument `identifiers_as_lower` by `columns_names_capitalization`
* [ ] Add support for  `columns_names_capitalization` when using `load_file` to Snowflake
* [ ] Have tests covering these scenarios
* [ ] Update release notes

PR #544, opened 2022-07-15
Optimize Snowflake load_file using native COPY INTO

tatiana:
Fix: #430 

Reduced the time to load to Snowflake by 20% for 5GB datasets (from 24.46 min to 5.49 min). Further details in the PR results file.

Co-authored-by: Ankit Chaurasia <ankit.chaurasia@astronomer.io>

PR #565, opened 2022-07-25
Column capitalization

utkarsharma2:
# Description
To achieve a first end-to-end delivery of the Snowflake optimization:
https://github.com/astronomer/astro-sdk/pull/487
https://github.com/astronomer/astro-sdk/pull/544

There was an assumption that to avoid issues with Snowflake mixed-capitalized column names, it was acceptable to convert all characters to uppercase.

1.     it is not consistent with the dataframe decorator parameter: identifiers_as_lower
2.     it does not support the use case where users may prefer to convert everything to uppercase

closes: #564

## What is the new behavior?

The goal with this ticket is to add the following argument to both aql.dataframe and aql.load_file (snowflake-only), removing the previous identifiers_as_lower parameter:

columns_names_capitalization=["upper", "lower", "original"]

### Checklist
- [X] Replace the aql.dataframe argument identifiers_as_lower by columns_names_capitalization
- [ ] Add support for columns_names_capitalization when using load_file to Snowflake
- [X] Have tests covering these scenarios
- [ ] Update release notes

### DIFF
--- a/src/astro/constants.py
+++ b/src/astro/constants.py
@@ -44,3 +44,5 @@ ExportExistsStrategy = Literal["replace", "exception"]
 
 # TODO: check how snowflake names these
 MergeConflictStrategy = Literal["ignore", "update", "exception"]
+
+ColumnCapitalization = Literal["upper", "lower", "original"]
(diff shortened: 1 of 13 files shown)


utkarsharma2:
Duplicate PR.

PR #567, opened 2022-07-25
Add 'columns_names_capitalization' param to control the cols case 

utkarsharma2:
# Description
To achieve a first end-to-end delivery of the Snowflake optimization:
https://github.com/astronomer/astro-sdk/pull/487
https://github.com/astronomer/astro-sdk/pull/544

There was an assumption that to avoid issues with Snowflake mixed-capitalized column names, it was acceptable to convert all characters to uppercase.

1.     it is not consistent with the dataframe decorator parameter: identifiers_as_lower
2.     it does not support the use case where users may prefer to convert everything to uppercase

closes: #564

## What is the new behavior?

The goal with this ticket is to add the following argument to both aql.dataframe and aql.load_file (snowflake-only), removing the previous identifiers_as_lower parameter:

columns_names_capitalization=["upper", "lower", "original"]

### Checklist
- [X] Replace the aql.dataframe argument identifiers_as_lower by columns_names_capitalization
- [X] Add support for columns_names_capitalization when using load_file to Snowflake
- [X] Have tests covering these scenarios
- [ ] Update release notes

### DIFF
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -12,6 +12,7 @@ from sqlalchemy.sql.schema import Table as SqlaTable
 
 from astro.constants import (
     DEFAULT_CHUNK_SIZE,
+    ColumnCapitalization,
     ExportExistsStrategy,
     LoadExistStrategy,
     MergeConflictStrategy,
@@ -20,6 +21,7 @@ from astro.exceptions import NonExistentTableException
 from astro.files import File, resolve_file_path_pattern
 from astro.settings import LOAD_TABLE_AUTODETECT_ROWS_COUNT, SCHEMA
 from astro.sql.table import Metadata, Table
+from astro.utils.dataframe import convert_dataframe_col_case
 
 
 class BaseDatabase(ABC):
@@ -181,6 +183,7 @@ class BaseDatabase(ABC):
         table: Table,
         file: Optional[File] = None,
         dataframe: Optional[pd.DataFrame] = None,
+        columns_names_capitalization: ColumnCapitalization = "lower",
     ) -> None:
         """
         Create a SQL table, automatically inferring the schema using the given file.
@@ -188,6 +191,8 @@ class BaseDatabase(ABC):
         :param table: The table to be created.
         :param file: File used to infer the new table columns.
         :param dataframe: Dataframe used to infer the new table columns if there is no file
+        :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
+            in the resulting dataframe
         """
         if file is None:
             if dataframe is None:
@@ -199,6 +204,11 @@ class BaseDatabase(ABC):
             source_dataframe = file.export_to_dataframe(
                 nrows=LOAD_TABLE_AUTODETECT_ROWS_COUNT
             )
+        source_dataframe = convert_dataframe_col_case(
+            df=source_dataframe,
+            columns_names_capitalization=columns_names_capitalization,
+        )
+
         db = SQLDatabase(engine=self.sqlalchemy_engine)
         db.prep_table(
             source_dataframe,
@@ -213,6 +223,7 @@ class BaseDatabase(ABC):
         table: Table,
         file: Optional[File] = None,
         dataframe: Optional[pd.DataFrame] = None,
+        columns_names_capitalization: ColumnCapitalization = "lower",
     ) -> None:
         """
         Create a table either using its explicitly defined columns or inferring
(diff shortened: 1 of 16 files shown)


src/astro/sql/operators/load_file.py
```diff
@@ -116,6 +122,11 @@ def load_data_to_dataframe(self, input_file: File) -> Optional[pd.DataFrame]:
                 df = pd.concat([df, file.export_to_dataframe()])
             else:
                 df = file.export_to_dataframe()
+
+            df = convert_dataframe_col_case(
+                df=df, columns_names_capitalization=self.columns_names_capitalization
+            )
```
tatiana, line 128: I believe we would benefit from keeping this logic inside the database class itself, within `export_to_dataframe` - since our motivation to mess up with capitalization is because of issues with a specific database

src/astro/sql/operators/dataframe.py
```diff
@@ -114,13 +127,17 @@ def execute(self, context: Dict) -> Union[Table, pd.DataFrame]:
...
-            self.op_args, self.python_callable, self.identifiers_as_lower
+            self.op_args, self.python_callable, self.columns_names_capitalization
         )
         self.op_kwargs = load_op_kwarg_table_into_dataframe(
-            self.op_kwargs, self.python_callable, self.identifiers_as_lower
+            self.op_kwargs, self.python_callable, self.columns_names_capitalization
         )
 
         pandas_dataframe = self.python_callable(*self.op_args, **self.op_kwargs)
+        pandas_dataframe = convert_dataframe_col_case(
```
tatiana, line 137: I'd expect the column capitalization logic to be handled within the database class (`load_pandas_dataframe_to_table`) since the motivation to introduce this complexity was the limitations of databases.

src/astro/utils/dataframe.py
```diff
@@ -0,0 +1,20 @@
+import pandas as pd
```
tatiana, line 1: @utkarsharma2, we've been trying to move out of the utils package 🙈 
if we handle the change of capitalization from within the database class, this method could live there.

src/astro/utils/dataframe.py
```diff
@@ -0,0 +1,20 @@
...
+
+from astro.constants import ColumnCapitalization
+
+
+def convert_dataframe_col_case(
```
tatiana, line 6: Since the function name is already quite long, it may be better to name it `convert_columns_names_capitalization` - this way, we are consistent with the argument name "columns_names_capitalization".

src/astro/sql/operators/load_file.py
utkarsharma2: @tatiana I think this functionality spans the database as we are doing this for the dataframe operator as well. This PR handles this functionally for all the places a dataframe is created.

src/astro/sql/operators/dataframe.py
utkarsharma2: @tatiana aren't we overpopulating the database class?

src/astro/sql/operators/load_file.py
utkarsharma2: As per discussion with @tatiana we agreed that this functionality spans across the lib, wherever we are creating the dataframe and this param introduces a way to control cols case. Hence resolving this comment.

src/astro/sql/operators/dataframe.py
utkarsharma2: As per discussion with @tatiana we agreed that this functionality spans across the lib, wherever we are creating the dataframe and this param introduces a way to control cols case. Hence resolving this comment.

src/astro/utils/dataframe.py
utkarsharma2: As per discussion with @tatiana since the logic is specific to dataframes it makes sense to keep them in Utils.

src/astro/utils/dataframe.py
utkarsharma2: Done.

sunank200 (approved):

sunank200 (commented):
Deepsource error needs to be fixed.

src/astro/databases/base.py
```diff
@@ -183,13 +184,16 @@ def create_table_using_schema_autodetection(
         table: Table,
         file: Optional[File] = None,
         dataframe: Optional[pd.DataFrame] = None,
+        columns_names_capitalization: ColumnCapitalization = "lower",  # skipcq
```
tatiana, line 187: @utkarsharma2 another strategy here (I added this, so DeepSource does not complain for now) could be to raise an exception if this value is passed to the base database - since it is not being used. That said, this shouldn't be a blocker for the PR to be merged.

tatiana (approved):

PR #568, opened 2022-07-25
Add benchmarking results for S3 to Bigquery transfer

utkarsharma2:
# Description
Benchmarking results for the S3 to bigquery is missing

related: #429

## What is the new behavior?
Added benchmarking results for S3 to Bigquery

No

### Checklist

### DIFF
--- a/tests/benchmark/results.md
+++ b/tests/benchmark/results.md
@@ -108,6 +108,7 @@ The benchmark was run as a Kubernetes job in GKE:
 | postgres_conn_benchmark | ten_gb     | 45.64min     | 61.71MB      | 2.37min         | 11.48s            |
 
 ### Database S3 to Bigquery using default path
+Note - These results are generated manually, there is a issue added for the same [#574](https://github.com/astronomer/astro-sdk/issues/574)
 
 |  database | dataset    | total_time   | memory_rss   | cpu_time_user   | cpu_time_system   |
 |:----------|:-----------|:-------------|:-------------|:----------------|:------------------|
@@ -117,6 +118,7 @@ The benchmark was run as a Kubernetes job in GKE:
 | bigquery  | ten_mb     | 47.02s       | 201.8MB      | 32.79ms         | 11.09ms           |
 
 ### Database S3 to Bigquery using native path
+Note - These results are generated manually, there is a issue added for the same [#574](https://github.com/astronomer/astro-sdk/issues/574)
 
 | database   | dataset    | total_time   | memory_rss   | cpu_time_user   | cpu_time_system   |
 |:-----------|:-----------|:-------------|:-------------|:----------------|:------------------|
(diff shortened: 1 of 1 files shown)


sunank200 (commented):
I think the results are missing for 5 GB and 10 GB

tests/benchmark/results.md
```diff
@@ -106,3 +106,21 @@ The benchmark was run as a Kubernetes job in GKE:
...
+| bigquery  | ten_mb     | 47.02s       | 201.8MB      | 32.79ms         | 11.09ms           |
+
+### Database S3 to Bigquery using native path
+
+| database   | dataset    | total_time   | memory_rss   | cpu_time_user   | cpu_time_system   |
+|:-----------|:-----------|:-------------|:-------------|:----------------|:------------------|
+| bigquery   | hundred_kb | 2.92min      | 56.57MB      | 41.94ms         | 35.83ms           |
+| bigquery   | one_gb     | 4.25min      | 66.32MB      | 54.88ms         | 48.01ms           |
+| bigquery   | ten_kb     | 2.9min       | 55.51MB      | 40.74ms         | 34.94ms           |
+| bigquery   | ten_mb     | 4.17min      | 57.6MB       | 46.64ms         | 46.83ms           |
```
tatiana, line 126: @utkarsharma2, please, if possible, document why we didn't use the standard benchmarking script on K8s in this document.

Also, please,  log a ticket so we can further investigate the NDJSON issues you faced while running it.

utkarsharma2:
Note - 

1. We didn't publish the results with all the datasets because not all the datasets were working with native as well as the default path. The ones that were working are published in this PR.
2. The idea with this benchmarking results was to check at what file size we should use the native path VS the default path.
3. For larger files we needed to make changes in the dataset and benchmarking script, due to limitations in pandas and Bigquey DB for that reason we ran these tests in VM and not K8 Cluster.

cc: @kaxil @tatiana @sunank200

tests/benchmark/results.md
tatiana: @utkarsharma2 please, document in the `results.md file` the exact problems you had with which datasets and what changes someone would need to be able to make to reproduce these results.

tatiana: If possible, please, share here the link to the Github issue related to these problems - so we can address them after 1.0.0 is released.

utkarsharma2:
Since we are seeing significant disadvantages of using the native path for smaller file sizes. Added following ticket as one possible way of dealing with the issue - https://github.com/astronomer/astro-sdk/issues/573

cc: @kaxil @tatiana

utkarsharma2:
@tatiana - https://github.com/astronomer/astro-sdk/issues/574

tatiana (approved):

Issue #570, opened 2022-07-26
Bug distutils.errors.DistutilsOptionError: No configuration found for dynamic 'description'.


tatiana:
**Version**
* Astro Python SDK: 0.11.0
* Astro Python SDK: main

**Describe the bug**
During the last couple of hours, we started getting the error:
```
         File "/home/runner/work/astro-sdk/astro-sdk/.nox/type_check-3-8/lib/python3.8/site-packages/setuptools/config/pyprojecttoml.py", line 301, in _ensure_previously_set
          raise OptionError(msg)
      distutils.errors.DistutilsOptionError: No configuration found for dynamic 'description'.
      Some dynamic fields need to be specified via `tool.setuptools.dynamic`
      others must be specified via the equivalent attribute in `setup.py`.
      [end of output]
```

When running the following command:
```
python -m pip install -e '.[all]' f
```
In unchanged branches, just by retriggering the job to run in the CI.

**To Reproduce**

```
nox > Command python -m pip install -e '.[all]' failed with exit code 1:
Obtaining file:///home/runner/work/astro-sdk/astro-sdk
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 2
  ╰─> [45 lines of output]
      Collecting flit_core~=3.2
        Downloading flit_core-3.7.1-py3-none-any.whl (60 kB)
           ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 60.2/60.2 kB 3.2 MB/s eta 0:00:00
      Installing collected packages: flit_core
      ERROR: Exception:
      Traceback (most recent call last):
        File "/home/runner/work/astro-sdk/astro-sdk/.nox/type_check-3-8/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 167, in exc_logging_wrapper
          status = run_func(*args)
        File "/home/runner/work/astro-sdk/astro-sdk/.nox/type_check-3-8/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 247, in wrapper
          return func(self, options, args)
        File "/home/runner/work/astro-sdk/astro-sdk/.nox/type_check-3-8/lib/python3.8/site-packages/pip/_internal/commands/install.py", line 461, in run
          installed = install_given_reqs(
[...]

PR #571, opened 2022-07-26
Fix DistutilsOptionError #570

tatiana:
Co-authored-with: @uranusjr
Fix: #570

Aims to solve this issue in the CI:
```
File "/home/runner/work/astro-sdk/astro-sdk/.nox/type_check-3-8/lib/python3.8/site-packages/setuptools/config/pyprojecttoml.py", line 301, in _ensure_previously_set
          raise OptionError(msg)
      distutils.errors.DistutilsOptionError: No configuration found for dynamic 'description'.
      Some dynamic fields need to be specified via `tool.setuptools.dynamic`
      others must be specified via the equivalent attribute in `setup.py`.
      [end of output]
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,6 +36,7 @@ jobs:
     env:
       MYPY_FORCE_COLOR: 1
       TERM: xterm-color
+      SETUPTOOLS_USE_DISTUTILS: stdlib
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3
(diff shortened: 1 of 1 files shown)


pankajastro (approved):

.github/workflows/ci.yaml
```diff
@@ -36,6 +36,7 @@ jobs:
     env:
       MYPY_FORCE_COLOR: 1
       TERM: xterm-color
+      SETUPTOOLS_USE_DISTUTILS: stdlib
```
sunank200, line 39: What does this line do? Just for my knowledge.

sunank200 (approved):

.github/workflows/ci.yaml
tatiana: @sunank200 I believe this comment explains the key issue:
https://github.com/pypa/setuptools/issues/2965#issuecomment-1002178847

```
The way the distutils hack works, it makes Setuptools' own copy of distutils take precedence over the standard library, so causes import distutils to put sys.modules['distutils'] = <setuptools._distutils>. So it's not a big surprise to me that running the tests from the standard library against code not in the standard library will have problems.
```

More information in the official docs:
https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html

Issue #377, opened 2022-05-19
Astro SDK + Task Mapping


kaxil:
Check the possibility of using task mapping with Astro SDK, something like

```python
@sql
def get_campaigns():
    return """select campaign_id from active_campaigns"""

summarize_campaign.expand(capaign_id=get_campaigns())
```

Add an example DAG or add a task in existing Example DAG to show this usage

pankajastro:
> Any progress on this @pankajastro ?

No @kaxil last week I came to know about this but will conclude by this week.

kaxil:
Sounds good, let's try to get it done on Monday so we can cut 1.0.0b1 on Mon/Tue

phanikumv:
@pankajastro is facing the below issue while spinning up airflow through docker container. Debugging through it

`ERROR: You need to upgrade the database. Please run airflow db upgrade. Make sure the command is run using Airflow version 2.3.3+astro.1.`

pankajastro:
`@sql` decorator directly does not and throw the below error. I'm using `from astro import sql` import.

```
Broken DAG: [/usr/local/airflow/dags/task_map_test.py] Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/local/airflow/dags/task_map_test.py", line 31, in <module>
    def get_campaigns(table: Table):
TypeError: 'module' object is not callable  
```

pankajastro:
The potential decorator that I can use for this is either `sql.transform` or `sql.run_raw_sql`.

Getting below error when `GOOGLE_APPLICATION_CREDENTIALS` env is not set and using airflow connection only

```
[2022-07-26, 07:52:55 UTC] {base.py:68} INFO - Using connection ID 'google_cloud_default' for task execution.
[2022-07-26, 07:52:55 UTC] {base.py:68} INFO - Using connection ID 'google_cloud_default' for task execution.
[2022-07-26, 07:52:57 UTC] {bigquery.py:1053} INFO - Dataset Resource: Dataset(DatasetReference('astronomer-airflow-providers', 'tmp_astro'))
[2022-07-26, 07:52:57 UTC] {base.py:68} INFO - Using connection ID 'google_cloud_default' for task execution.
[2022-07-26, 07:53:01 UTC] {_metadata.py:99} WARNING - Compute Engine Metadata server unavailable on attempt 1 of 3. Reason: timed out
[...]

PR #575, opened 2022-07-26
Example Dynamic Map Task with Astro-SDK

pankajastro:
# Description
Add Example DAG for Dynamic Map Task with Astro-SDK. 
DAG does
- load data from s3 to bigquery table
- get_campaigns task execute SQL query and get a Table object 
-  expanding summarize_campaign task so if the bigquery table has n row then n copy of summarize_campaign will be created 
- clean table

closes: #377 

### Checklist

### DIFF
--- a/example_dags/example_bigquery_dynamic_map_task.py
+++ b/example_dags/example_bigquery_dynamic_map_task.py
@@ -1,8 +1,8 @@
 from datetime import datetime
-from pandas import DataFrame
 
 from airflow import DAG
 from airflow.decorators import task
+from pandas import DataFrame
 from astro import sql as aql
 from astro.sql import Table
 from astro.sql.table import Metadata
--- a/example_dags/example_bigquery_dynamic_map_task.py
+++ b/example_dags/example_bigquery_dynamic_map_task.py
@@ -3,10 +3,11 @@ from datetime import datetime
 
 from airflow import DAG
 from airflow.decorators import task
+
 from astro import sql as aql
+from astro.files import File
 from astro.sql import Table
 from astro.sql.table import Metadata
-from astro.files import File
 
 ASTRO_BIGQUERY_DATASET = os.getenv("ASTRO_BIGQUERY_DATASET", "dag_authoring")
 ASTRO_GCP_LOCATION = os.getenv("ASTRO_GCP_LOCATION", "us")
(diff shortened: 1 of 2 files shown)


sunank200 (approved):
Overall LGTM. But the comment on the code would be helpful.

example_dags/example_bigquery_dynamic_map_task.py
```diff
@@ -0,0 +1,102 @@
...
+
+    delete_dataset = BigQueryDeleteDatasetOperator(
+        task_id="delete_dataset",
+        dataset_id=ASTRO_BIGQUERY_DATASET,
+        delete_contents=True,
+        gcp_conn_id=ASTRO_GCP_CONN_ID,
+        trigger_rule="all_done",
+    )
+
+    t1 = get_campaigns(bq_table)
```
utkarsharma2, line 100: @pankajastro Can you please add comments explaining the Dynamic task mapping? I think it can be valuable for users.

example_dags/example_bigquery_dynamic_map_task.py
```diff
@@ -0,0 +1,102 @@
...
+        task_id="insert_query_job",
+        configuration={
+            "query": {
+                "query": INSERT_ROWS_QUERY,
+                "useLegacySql": False,
+            }
+        },
+        location=ASTRO_GCP_LOCATION,
+        gcp_conn_id=ASTRO_GCP_CONN_ID,
+    )
```
tatiana, line 78: @pankajastro What about using `aql.load_file` for loading this data into BQ, as opposed to traditional Airflow operators?

example_dags/example_bigquery_dynamic_map_task.py
```diff
@@ -0,0 +1,102 @@
...
+        task_id="delete_dataset",
+        dataset_id=ASTRO_BIGQUERY_DATASET,
+        delete_contents=True,
+        gcp_conn_id=ASTRO_GCP_CONN_ID,
+        trigger_rule="all_done",
+    )
+
+    t1 = get_campaigns(bq_table)
+    summarize_campaign.expand(capaign_id=t1)
+    create_dataset >> create_table >> insert_query_job >> t1 >> delete_dataset
```
tatiana, line 102: One of the goals of the Astro Python SDK is to promote Task Flow API and discourage the usage of the bit shift operator to write DAGs and define task dependency. I believe it is very important that we stick to this principle in all our examples - at least for 1.0.0.

example_dags/example_bigquery_dynamic_map_task.py
```diff
@@ -0,0 +1,102 @@
...
+    schedule_interval=None,
+    start_date=datetime(2022, 1, 1),
+    catchup=False,
+) as dag:
+    create_dataset = BigQueryCreateEmptyDatasetOperator(
+        task_id="create_dataset",
+        dataset_id=ASTRO_BIGQUERY_DATASET,
+        location=ASTRO_GCP_LOCATION,
+        gcp_conn_id=ASTRO_GCP_CONN_ID,
+    )
```
tatiana, line 57: In order to simplify the example, it would be great if we reused an existing dataset - so we could avoid having to add steps for creation and deletion.

example_dags/example_bigquery_dynamic_map_task.py
```diff
@@ -0,0 +1,102 @@
...
+
+from airflow import DAG
+from airflow.decorators import task
+from airflow.providers.google.cloud.operators.bigquery import (
+    BigQueryCreateEmptyDatasetOperator,
+    BigQueryCreateEmptyTableOperator,
+    BigQueryDeleteDatasetOperator,
+    BigQueryInsertJobOperator,
+)
+
```
tatiana, line 12: Do we need to use all these operators? Would it be an option to promote using existing Astro SDK operators/decorators?

pankajastro: Now I'm using aql.loads it would not require

example_dags/example_bigquery_dynamic_map_task.py
pankajastro: Added on top of file please let me know how it is looking

example_dags/example_bigquery_dynamic_map_task.py
pankajastro: Make sense now I'm using `load_file` to load data in bigquery.

example_dags/example_bigquery_dynamic_map_task.py
pankajastro: hmm, changed it, now I'm not using tradition operator

utkarsharma2:
@pankajastro can you change/delete the image added to PR?

example_dags/example_bigquery_dynamic_map_task.py
pankajastro: I liked the idea and initial testing I did on the existing table. But I feel it good to create a table and clean it so that people can test it with just connection in their env

tatiana: I agree, @pankajastro , I believe we need to have a better plan moving forward on how we would like to setup/teardown resources of the example DAGs. If we could find a way which allows them to be simple, while making it easy for any developer to run them - it would be amazing.

tatiana (approved):
Thank you very much for addressing all the comments, @pankajastro , I was impressed with how the example code size was reduced!

Issue #569, opened 2022-07-25
Add the ability set task dependency without a data dependency


dimberman:
**Please describe the feature you'd like to see**
Currently with the Astro SDK there is no easy way to chain together tasks without having a data dependency. Let's take an example where someone wants to wait until a table is truncated before adding in new data. Currently the only way to do that is to use the traditional bitshift operators like so:

```python
@aql.run_raw_sql(conn_id='sqlite_conn')
def truncate_foo():
    return """TRUNCATE TABLE foo;
    )"""

@dag(schedule_interval='0 * * * *', start_date=datetime(2022, 7, 13, 21, 52, 12), catchup=False, tags=[])
def example_sqlite_astro_build_comment():
    truncated_table = truncate_foo()
    loaded_data = aql.load_file(File("https://my/file.csv"), output_table=Table("foo", conn_id="my_conn"))
    truncated_table >> loaded_data
dag_obj = example_sqlite_astro_build_comment()
```

This of course pulls users out of the promise that we are building a more pythonic functional library. Currently even if we were to TRY to use functional composition like this:

```python
@aql.run_raw_sql(conn_id='sqlite_conn')
def truncate_foo():
    return """TRUNCATE TABLE foo;
    )"""

def truncate_bar(table: Table):
    return "TRUNCATE TABLE bar"

@dag(schedule_interval='0 * * * *', start_date=datetime(2022, 7, 13, 21, 52, 12), catchup=False, tags=[])
def example_sqlite_astro_build_comment():
    truncated_table = truncate_foo()
    truncate_bar(truncated_table)
    loaded_data = aql.load_file(File("https://my/file.csv"), output_table=Table("foo", conn_id="my_conn"))
    truncate_bar >> loaded_data
```

We would get an error because the `run_raw_sql` function does not place anything in xcom, so there's nothing to render.

This issue leads to both a bad story when passing dependencies functionally, and when trying to tie astro-sdk tasks with traditional Airflow operators.

```python
@dag(schedule_interval='0 * * * *', start_date=datetime(2022, 7, 13, 21, 52, 12), catchup=False, tags=[])
def example_sqlite_astro_build_comment():
    b = BashOperator(...)
    truncated_table = truncate_foo()
    b >> truncated_table
```

**Describe the solution you'd like**

I see a two-part solution to this issue.

[...]

PR #585, opened 2022-07-28
Create `upstream_tasks` parameter for dependencies independent of data transfers

dimberman:
# Description
Currently there is "native" way to set dependencies between tasks without passing either a table or a dataframe between said tasks. This is especially apparent when chaining raw_sql functions such as here: 

https://github.com/astronomer/astro-sdk/issues/569

In order to have a solution that works we need somehting that feels pythonic and doesn't rely on traditional airflow `>>` operators

closes: #569 

## What is the new behavior?

This PR adds in the ability to set dependencies using a parameter `upstream_tasks`. This parameter can be used at runtime to set deps without passing data. Notably we don't want to actually render the data (imagine a use-case where a task relies on 10 dataframe tasks)

No

### Checklist
- [ ] Add documentation around `upstream_tasks`
- [ ] solution should not render the data since the data is never used.


docs/getting-started/GETTING_STARTED.md
```diff
@@ -392,3 +392,41 @@ or
 
 In all scenarios, even if the user gives a non-temporary table, only temporary
 tables will actually be deleted.
+
+## Tying Astro SDK decorators to traditional Airflow Operators
```
tatiana, line 396: Is the getting started tutorial the best place for this documentation?

I feel it could be best placed within our Sphinx documentation as a reference page. I believe the idea of the tutorial was to walk users through a first example of the Python SDK without introducing too many possibilities. Could you confirm this, @mikeshwe ?

docs/getting-started/GETTING_STARTED.md
```diff
@@ -392,3 +392,41 @@ or
...
+    return "SELECT * FROM {{table}} LIMIT {{num_rows}}"
+
+
+with dag:
+    bash_command = BashOperator(...)
+    get_rows(table=Table(), num_rows=5, upstream_tasks=[bash_command])
+```
+When tying traditional tasks to astro-sdk decorators, you might run into a situation where the original operators
+might not pass any data. In
+One potential friction point when adopting astro-sdk mig
```
tatiana, line 432: Seems there is something missing here.

src/astro/sql/operators/append.py
```diff
@@ -4,12 +4,13 @@
...
 
 from astro.databases import create_database
+from astro.sql.operators.upstream_tasks import UpstreamTaskMixin
 from astro.sql.table import Table
 
 APPEND_COLUMN_TYPE = Optional[Union[List[str], Tuple[str], Dict[str, str]]]
 
 
-class AppendOperator(BaseOperator):
+class AppendOperator(UpstreamTaskMixin, BaseOperator):
```
tatiana, line 13: Does this PR make all of the Python Astro SDK decorators and tasks inherit from `UpstreamTaskMixin`?

Would it make more sense to define an Astro SDK `BaseOperator` class which would both:
* inherit from `BaseOperator`
* implement the `UpstreamTaskMixing` behaviour that should be available to all operators?

src/astro/sql/operators/upstream_tasks.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class UpstreamTaskMixin:
+    def __init__(self, **kwargs):
+        upstream_tasks = kwargs.pop("upstream_tasks", [])
+
+        super().__init__(**kwargs)
+
+        for task in upstream_tasks:
+            if isinstance(task, XComArg):
```
tatiana, line 11: What if the task is not an instance of `XComArg`, do we need to do any additional steps to set the dependency?

tatiana (commented):
@dimberman, I believe this PR does not close the original problem reported by #585, which is the comments at the beginning of the SQL statement were not working.

It does solve the issue of chaining `run_raw_sql` statements and operators which return `None` in the `XComArg`. We should probably log a Github issue for this and link it to this PR.

src/astro/sql/operators/upstream_tasks.py
dimberman: @tatiana I've added a condition for if it's a BaseOperator, if the task isn't one of those two then there's nothing for us to upstream.

src/astro/sql/operators/append.py
dimberman: @tatiana I've made an AstroSQLBaseOperator class that takes in the UpstreamTaskMixin. This will look cleaner for those implementing operators, but we have to keep the mixin because the `BaseSQLDecoratedOperator` also depends on the `DecoratedOperator` class so we would run the risk of double inheritance.

docs/getting-started/GETTING_STARTED.md
dimberman: @tatiana this is one of the most common questions we get asked (how can I use astro-sdk with traditional airflow operators) so it seemed like the getting started page was a good place to do it. Glad to discuss other options though.

docs/getting-started/GETTING_STARTED.md
dimberman: fixed

docs/getting-started/GETTING_STARTED.md
feluelle: I think this is fair. 👍

docs/getting-started/GETTING_STARTED.md
```diff
@@ -392,3 +392,46 @@ or
...
+
+```python
+@aql.transform
+def get_rows(table: Table, num_rows: int):
+    return "SELECT * FROM {{table}} LIMIT {{num_rows}}"
+
+
+with dag:
+    bash_command = BashOperator(...)
+    get_rows(table=Table(), num_rows=5, upstream_tasks=[bash_command])
```
feluelle, line 436:
```suggestion
    bash_command >> get_rows(table=Table(), num_rows=5)
```
So this doesn't work? Why?

src/astro/sql/operators/drop.py
```diff
@@ -1,13 +1,14 @@
...
-from airflow.models import BaseOperator
+from airflow.models.baseoperator import BaseOperator
 
 from astro.databases import create_database
+from astro.sql.operators.base_operator import AstroSQLBaseOperator
 from astro.sql.table import Table
 
 
-class DropTableOperator(BaseOperator):
+class DropTableOperator(AstroSQLBaseOperator, BaseOperator):
```
feluelle, line 11:
```suggestion
class DropTableOperator(AstroSQLBaseOperator):
```
We don't need that as `AstroSQLBaseOperator` is already based on `BaseOperator`.

src/astro/sql/operators/export_file.py
```diff
@@ -1,17 +1,18 @@
...
 from astro.constants import ExportExistsStrategy
 from astro.databases import create_database
 from astro.files import File
+from astro.sql.operators.base_operator import AstroSQLBaseOperator
 from astro.sql.table import Table
 from astro.utils.task_id_helper import get_task_id
 
 
-class ExportFile(BaseOperator):
+class ExportFile(AstroSQLBaseOperator, BaseOperator):
```
feluelle, line 15: Same here.

src/astro/sql/operators/load_file.py
```diff
@@ -2,18 +2,19 @@
...
 from astro.databases import BaseDatabase, create_database
 from astro.exceptions import IllegalLoadToDatabaseException
 from astro.files import File, check_if_connection_exists, resolve_file_path_pattern
+from astro.sql.operators.base_operator import AstroSQLBaseOperator
 from astro.sql.table import Table
 from astro.utils.task_id_helper import get_task_id
 
 
-class LoadFile(BaseOperator):
+class LoadFile(AstroSQLBaseOperator, BaseOperator):
```
feluelle, line 17: And here.

src/astro/sql/operators/merge.py
```diff
@@ -5,12 +5,13 @@
...
 from astro.constants import MergeConflictStrategy
 from astro.databases import create_database
+from astro.sql.operators.base_operator import AstroSQLBaseOperator
 from astro.sql.table import Table
 
 MERGE_COLUMN_TYPE = Union[List[str], Tuple[str], Dict[str, str]]
 
 
-class MergeOperator(BaseOperator):
+class MergeOperator(AstroSQLBaseOperator, BaseOperator):
```
feluelle, line 14: And here.

docs/getting-started/GETTING_STARTED.md
dimberman: @feluelle it's not that it doesn't work, it's that it breaks with the new more functional paradigm.

feluelle (approved):
LGTM 👍

Issue #507, opened 2022-07-04
Implement `get_file_list` (Dynamic Task Templates)


kaxil:
This builds on top of [Dynamic Tasks Mapping](https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html) released in Airflow 2.3

Iterating over a set of files matching a pattern in a file system:
- Look for all files matching a pattern in a file system such as S3 
- And then, to iterate over a set of tasks for each file found matching this pattern. 
- Again, the number of files found is unknown at DAG creation time, but it is not uncommon for this to be dozens of files. 

The `get_file_list` operation would abstract away the underlying Operator setup and boilerplate code, similar to the existing `load_file` operation in the Astro Python SDK. It would then get back a list of files suitable for the ensuing “map” operation. 

The idea here is to minimize the “boiler plate” code which needs to be written by a DAG Author and to institutionalize common patterns. Effectively, the above when coupled with a TaskGroup which works on each value represents the FileIterator pattern for DAG Authoring.

pankajastro:
initial investigation: https://docs.google.com/document/d/1qaxJ5cP7Cu_lVMDIUMWEYoDdDt63PpISWczqTwBcjXo/edit?usp=sharing

utkarsharma2:
@pankajastro We already have similar method - https://github.com/astronomer/astro-sdk/blob/b77e22e27ad60aa86073b59bcf37ac942e355981/src/astro/files/base.py#L99

pankajastro:
> @pankajastro We already have similar method -
> 
> https://github.com/astronomer/astro-sdk/blob/b77e22e27ad60aa86073b59bcf37ac942e355981/src/astro/files/base.py#L99

thanks @utkarsharma2 I'll take a look and try to use it

pankajastro:
Leveraging  `create_file_location` from  `astro.files.locations` to get the location type and then getting `paths` property on the location. currently, it is expected that the user will pass path and pattern in a single param `path` for example in the case of a local filesystem if I want to list all CSV files which inside the logs folder I'll do something like `dtt.get_file_list('conn', logs/*.csv)`

PR #596, opened 2022-07-29
Add `get_file_list` for dynamic task template

pankajastro:
# Description
Add an internal operator `_ListFileOperator` which call existing `create_file_location` method from  `astro.files.locations`. 
Add a wrapper `get_file_list` on this operator to return the output of this operator. This will help to use the airflow dynamic task map feature in a more convenient way. 

example 
```
from astro.files import get_file_list

files = get_file_list(conn_id=ASTRO_GCP_CONN_ID,  path=GCS_BUCKET)
```

You can also pass the underlying operator param in kwargs 

```
from astro.files import get_file_list

files = get_file_list(conn_id=ASTRO_GCP_CONN_ID,  path=GCS_BUCKET, task_id="task_1")
```

closes: #507


example_dags/example_dymaic_task_template.py
```diff
@@ -0,0 +1,28 @@
...
+
+with DAG(
+    dag_id="example_dynamic_task_template",
+    schedule_interval=None,
+    start_date=datetime(2022, 1, 1),
+    catchup=False,
+) as dag:
+    print_file_path.expand(path=dtt.get_file_list("aws_default", f"{S3_BUCKET}/*.csv"))
+
+    print_file_path.expand(
```
utkarsharma2, line 26: @pankajastro Can we do something like ingesting file data to a database table here? by probably using an aql.load_file op instead of just printing them?

tests/test_example_dags.py
```diff
@@ -32,6 +32,7 @@ def session():
         "example_snowflake_partial_table_with_append",
         "example_sqlite_load_transform",
         "example_dynamic_map_task",
+        "example_dynamic_task_template",
```
utkarsharma2, line 35: @kaxil @tatiana @dimberman  - we should find a way to split these dags and run them in parallel/parallel groups since we will be adding a lot of them going forwards and this can slow CI. WDYT?

example_dags/example_dymaic_task_template.py
pankajastro: hmm, we can do this. The suggestion looks reasonable to me I'll include this.

tests/test_example_dags.py
kaxil: agreed. Can you create a ticket on that please

src/astro/dtt/__init__.py
```diff
@@ -0,0 +1,16 @@
+from typing import List
```
kaxil, line 1: Should we have a separate module for files and databases?

src/astro/dtt/files/__init__.py
```diff
@@ -0,0 +1,16 @@
+from typing import List
```
tatiana, line 1: The name `dtt` makes me uncomfortable. It doesn't mean much, and it is focused on the implementation (dynamic task templates) instead of the problem it's solving (listing files). It also feels like we're trying to mimic a folder structure that we already have within the Python SDK (`files`).

@pankajastro What if we introduced dynamic task templates within previously existing packages?
In this case, we could add the implementation to `astro/files/dynamic_tasks.py` and expose it within `astro/files/__init__.py`.

This way,  the import statement would be focused on the usage and not implementation, e.g.: `astro.files import get_file_list` - WDYT?

example_dags/example_dymaic_task_template.py
```diff
@@ -0,0 +1,42 @@
+import os
```
tatiana, line 1: Should we also add a sphinx page explaining Dynamic Task templates and listing the ones which are available, linking to this example?

pankajastro: Yes, we will add documentation for it. just waiting for the 1st documentation PR related to example to land so that we minimise rework.

src/astro/dtt/files/__init__.py
pankajastro: Actually, right now I'm sure how many methods we will have to expose to the users. if we keep this in the existing module it might populate it in future when we will add methods for more patterns. But I would like to hear a better replacement for `dtt` name. 
@kaxil WDYT about @tatiana suggestion ??

tatiana: @pankajastro based on the conversation with @kaxil yesterday, there are vert few dynamic task templates planned at the moment (four at most). This particular one could be implemented in `astro/files/dynamic_tasks.py`, but be exposed in `astro/files/__init__.py` - so users can import all features related to files just by doing:

```
from astro.files import File, get_file_list
```

kaxil: Shared in our slack thread

example_dags/example_dymaic_task_template.py
pankajastro: Added, please let me know how it is looking now.

Screenshot 2022-08-11 at 2 33 32 PM

src/astro/dtt/files/__init__.py
pankajastro: @tatiana @kaxil I have changed accordingly please let me know how it is looking now?

docs/astro/files/operators/get_file_list.rst
```diff
@@ -0,0 +1,14 @@
+=============
+get_file_list
+=============
+
+.. _get_file_list:
```
utkarsharma2, line 5: @pankajastro We need to move this to the top of the page before H1, otherwise when we use this **:ref:`get_file_list`** in doc the sphinx doc will print "When to use the ``get_file_list`` operator" and not "get_file_list".

src/astro/files/__init__.py
```diff
@@ -14,3 +16,26 @@ def check_if_connection_exists(conn_id: str) -> bool:
...
+    Supported filesystem: Local, HTTP, S3, GCS
+
+    :param conn_id: Airflow connection id
+    :param path: Path pattern to the file in the filesystem/Object stores
+    """
+    if "task_id" in kwargs:
+        task_id = kwargs["task_id"]
+        del kwargs["task_id"]
+    else:
+        task_id = get_unique_task_id(task_id="get_file_list")
```
utkarsharma2, line 34: We don't need this method of airflow handles this internally. Will append number if there are duplicate task names.

src/astro/files/__init__.py
```diff
@@ -14,3 +16,26 @@ def check_if_connection_exists(conn_id: str) -> bool:
     """
     BaseHook.get_connection(conn_id)
     return True
+
+
+def get_file_list(path: str, conn_id: str, **kwargs) -> TaskDecorator:
```
utkarsharma2, line 21: @pankajastro Why are we returning `TaskDecorator` shouldn't we be returning list of string/file objects?

tatiana, line 21: Have you considered using:
```get_file_list(files_prefix:File)```

Since `File` already abstracts both properties:
```File(path="s3://some-bucket/prefix/*", conn_id="some_conn")```

File: example_dags/example_dynamic_task_template.py [hunk 9]
```diff
@@ -0,0 +1,43 @@
...
+
+GCS_BUCKET = os.getenv("GCS_BUCKET", "gs://dag-authoring")
+ASTRO_GCP_CONN_ID = os.getenv("ASTRO_GCP_CONN_ID", "google_cloud_default")
+ASTRO_BIGQUERY_DATASET = os.getenv("ASTRO_BIGQUERY_DATASET", "dag_authoring")
+
+
+# [START howto_operator_get_file_list]
+@task
+def load_to_bigquery(path):
+    aql.load_file(
```
tatiana, line 20: Would this also work?
```
load_to_bigquery = aql.load_file(...)
```

src/astro/files/operators/files.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class _ListFileOperator(BaseOperator):
+    """List the file available at path and storage
+
+    :param task_id: The task id for uniquely identify a task in a DAG
+    :param path: A path pattern for which you want to get a list of file
+    :param conn_id: connection id for the services
+    """
+
```
tatiana, line 16: These two properties should probably be `template_fileds` - we may avoid this if we use `File`

tests/files/test_file.py
```diff
@@ -223,3 +223,12 @@ def test_resolve_file_path_pattern_raise_exception(invalid_path, caplog):
...
     assert expected_error in str(e.value)
+
+
+def test_get_file_list():
+    """Assert that get_file_list handle kwargs correctly"""
+    resp = get_file_list(path="path", conn_id="conn")
+    assert resp.operator.task_id == "get_file_list"
+
+    resp = get_file_list(path="path", conn_id="conn", task_id="test")
+    assert resp.operator.task_id == "test"
```
tatiana, line 234: It is probably worth testing the unique name scenario is working as expected as well:
```
resp = get_file_list(path="path", conn_id="conn", task_id="test")]
assert resp.operator.task_id != "test"
```

tatiana (approved):
Hey @pankajastro , as we spoke, there are minor comments which would be great if you could address! Overall, it looks great, thank you for addressing the initial requests!

src/astro/files/operators/files.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any
+
+from airflow.models import BaseOperator
+from airflow.utils.context import Context
+
+from astro.files.locations import create_file_location
+
+
+class _ListFileOperator(BaseOperator):
```
tatiana, line 9: Since we're not prefixing the other operators, it may be worth to not have it now - and to log a micro ticket for us to prefix all of them with _. This would make the code consistent.

example_dags/example_dynamic_task_template.py
pankajastro: ok, I ``path``  param to initialize ``File`` object. How would I pass that?

src/astro/files/__init__.py
pankajastro: hmm, I think it would make the public interface more consistent. but getting error when creating below object 
```
Broken DAG: [/usr/local/airflow/dags/task_flow.py] Traceback (most recent call last):
  File "/usr/local/airflow/astro_sdk/src/astro/files/types/__init__.py", line 25, in create_file_type
    filetype = get_filetype(path)
  File "/usr/local/airflow/astro_sdk/src/astro/files/types/__init__.py", line 53, in get_filetype
    raise ValueError(
ValueError: Missing file extension, cannot automatically determine filetype from path 'gs://pankaj_test_dtt/'. Please pass the 'filetype' param with the explicit filetype (e.g. csv, ndjson, etc.).
```
here I want to list files from ``pankaj_test_dtt``

src/astro/files/__init__.py
pankajastro: No, if you place your task with a duplicate id in a DAG airflow throw exception and do not resolve it.

src/astro/files/__init__.py
pankajastro: Not list, A task output is actually `XComArg` I'll change accordingly.

docs/astro/files/operators/get_file_list.rst
pankajastro: actually, I was looking at other examples that we have added https://github.com/astronomer/astro-sdk/blame/main/docs/astro/sql/operators/append.rst#L7 .

do you want to move  `.. _get_file_list` on top of file?

how does it resolve when we have multiple objects in the same rst file?

tests/files/test_file.py
pankajastro: Added a test case for this.

src/astro/files/operators/files.py
```diff
@@ -0,0 +1,27 @@
...
+from airflow.utils.context import Context
+
+from astro.files.locations import create_file_location
+
+
+class ListFileOperator(BaseOperator):
+    """List the file available at path and storage
+
+    :param task_id: The task id for uniquely identify a task in a DAG
+    :param path: A path pattern for which you want to get a list of file
```
kaxil, line 14: It should probably contain examples of what are the things that you can pass e.g bucket like: `s3://bucket/object` or even a pattern like `gs://bucket/object*.csv`

example_dags/example_dynamic_task_template.py
```diff
@@ -0,0 +1,43 @@
...
+        input_file=File(path=path),
+        output_table=Table(
+            metadata=Metadata(
+                schema=ASTRO_BIGQUERY_DATASET,
+            ),
+            conn_id=ASTRO_GCP_CONN_ID,
+        ),
+        use_native_support=False,
+    )
```
kaxil, line 29:
```suggestion
load_to_bigquery = aql.load_file(
    input_file=File(path=GCS_BUCKET),
    output_table=Table(
        metadata=Metadata(
            schema=ASTRO_BIGQUERY_DATASET,
        ),
        conn_id=ASTRO_GCP_CONN_ID,
    ),
    use_native_support=False,
)
```

src/astro/files/__init__.py
```diff
@@ -14,3 +17,30 @@ def check_if_connection_exists(conn_id: str) -> bool:
...
+    """
+    if "task_id" in kwargs:
+        task_id = kwargs["task_id"]
+        del kwargs["task_id"]
+    else:
+        task_id = get_unique_task_id(
+            task_id="get_file_list",
+            dag=kwargs.get("dag"),
+            task_group=kwargs.get("task_group"),
+        )
```
kaxil: this logic should be moved to operator I feel

kaxil (commented):
I just pushed a commit for moving `task_id` part

example_dags/example_dynamic_task_template.py
kaxil: check https://github.com/astronomer/astro-sdk/pull/596#discussion_r943745295

kaxil:
```
load_to_bigquery = aql.load_file(
    input_file=File(path=GCS_BUCKET),
    output_table=Table(
        metadata=Metadata(
            schema=ASTRO_BIGQUERY_DATASET,
        ),
        conn_id=ASTRO_GCP_CONN_ID,
    ),
    use_native_support=False,
)
```

src/astro/files/__init__.py
kaxil: hmm thinking of this more, this would be a problem from load_file too I think i.e it might fail if we pass `File(path="gs://pankaj_test_dtt/folder1/*", conn_id="bigquery_conn_id")`

cc @utkarsharma2 @sunank200 ^^ -- Can we test this before 1.0 please

example_dags/example_dynamic_task_template.py
pankajastro: but @kaxil GCS_BUCKET path is dynamic here, isn't it?

kaxil: Aaah yes, I get it now, we probably need docstring to explain that

pankajastro: ok

kaxil: ok after looking at this closely, there are 2 issue:

1) If you have the following, you are running a task within a task, I don't think it will return anything.
```
@task
def load_to_bigquery(path):
    aql.load_file(
```

2) The `load_file` function or `LoadFile` operator expects a `File` object whereas `get_file_list` returns a "list of string". We have 2 options: 1) Make `get_file_list` return a list of File objects and 2) Make `LoadFile`/`load_file` accept a "string" along with the `File` objects so users can use it with non-Astro SDK tasks and Dynamic Task Mapping.

kaxil: We should make the following example work to address the larger issue mentioned in the last comment:

```
with DAG(
    "example_dynamic_task_template",
    schedule_interval=None,
    start_date=datetime(2000, 1, 1),
    catchup=False,
) as dag:

    imdb_movies = LoadFile.partial(
        task_id="load_csv",
        output_table=Table(name="imdb_movies_2", conn_id="sqlite_default"),
    ).expand(input_file=get_file_list(path=GCS_BUCKET, conn_id=ASTRO_GCP_CONN_ID))
```

dimberman: @kaxil to your first point, I think get_file_list should absolutely return a list of File objects. Each of the tasks created need that context to properly process the file. Files should be the defacto way to pass files between tasks, anything else would be off.

To your second point, maybe there is a larger discussion about adding dynamic task mapping functionality to xcomargs? I think that makes sense if people expect to treat them like operator objects.

src/astro/sql/operators/load_file.py
```diff
@@ -182,7 +190,7 @@ def replace_illegal_columns_chars(char: str, database: BaseDatabase) -> str:
 
 
 def load_file(
-    input_file: File,
+    input_file: Union[File, str],
```
dimberman, line 193: @kaxil @tatiana should we actually allow people to just pass a string in? I feel like forcing File objects will be a good way to teach people to write things "properly" where they can easily change systems later

src/astro/files/__init__.py
kaxil: The problem was because the list returned by `FileLocation.path` contained "folders" (they are not folder per see on Cloud Storage) too, example:

```
In [13]: [urlunparse((url.scheme, url.netloc, keys, "", "", "")) for keys in prefixes]
Out[13]:
['gs://dag-authoring/dynamic_task/',
 'gs://dag-authoring/dynamic_task/homes.csv']
```

This is why the operator failed to determine filetype for first path.

Now, this is valid for Native support, but specifically not needed for `get_file_list`, `ListFileOperator` so I have pushed https://github.com/astronomer/astro-sdk/pull/596/commits/adf94c5163bff151ecca58cae6f022566708df5c for this PR. But we should still think about `load_file` operator, if somehow the "files" are loaded twice, as path 1 'gs://dag-authoring/dynamic_task/' is superset of 'gs://dag-authoring/dynamic_task/homes.csv'.

Can one of you check this tomorrow please?

example_dags/example_dynamic_task_template.py
kaxil: re: 1) Yup updated the PR with `get_file_list` returning list of `File` objects.

re: 2) Since `XComArg` already instantiates an operator, we can't have `partial` and `expand` as some of the params/args are required and will raise error. For example, `load_file` function already instantiates `LoadFileOperator` so we "can't not" pass `input_file` or else it will raise a required param not passed error. We can only do that on "class level" so not sure if it is feasible.

example_dags/example_dynamic_task_template.py
```diff
@@ -0,0 +1,37 @@
...
+# [START howto_operator_get_file_list]
+
+
+with DAG(
+    dag_id="example_dynamic_task_template",
+    schedule_interval=None,
+    start_date=datetime(2022, 1, 1),
+    catchup=False,
+) as dag:
+    LoadFile.partial(
```
pankajastro, line 25: @kaxil here ```LoadFile``` is a like traditional operator.  yesterday I was discussing with @tatiana to make make the traditional private in this repo. What is your thought on this?

src/astro/files/operators/files.py
```diff
@@ -0,0 +1,36 @@
...
+    :param conn_id: connection id for the services
+    """
+
+    template_fields = ("path", "conn_id")
+
+    def __init__(self, path: str, conn_id: str, task_id: str = "", **kwargs):
+        task_id = task_id or get_unique_task_id(
+            "get_file_list", dag=kwargs.get("dag"), task_group=kwargs.get("task_group")
+        )
```
feluelle, line 24: This should be set on `ListFileOperator` creation from outside i.e. inside the `get_file_list` function. With that being said, you can then also remove `task_id: str = ""` from the `__init__`.

example_dags/example_dynamic_task_template.py
feluelle: Only if we then have a non "private" replacement for it in examples. We don't want to advocate for calling a private operator, do we?

example_dags/example_dynamic_task_template.py
```diff
@@ -0,0 +1,37 @@
...
+        task_id="load_gcs_to_bq",
+        output_table=Table(
+            metadata=Metadata(
+                schema=ASTRO_BIGQUERY_DATASET,
+            ),
+            conn_id=ASTRO_GCP_CONN_ID,
+        ),
+        use_native_support=False,
```
feluelle, line 33: Why don't we use native support here?

example_dags/example_dynamic_task_template.py
kaxil: Yeah like Felix mentioned - we can't prefix the `LoadFile` with `_` because we need it for Dynamic Task Mapping to work. Dynamic Task Mapping works with "an Operator" class (not instance of the operator) not "task" whereas `load_file` returns an XComArg (output of task) - (`.output` on instantiated operator)

docs/astro/files/operators/get_file_list.rst
pankajastro: as discussed fixed it.

example_dags/example_dynamic_task_template.py
pankajastro: actually no reason let me change it to use native

src/astro/sql/operators/load_file.py
pankajastro: We have reverted this changes

src/astro/files/__init__.py
pankajastro: got it. So @kaxil does it mean currently we can't change

```def get_file_list(path: str, conn_id: str, **kwargs)``` ->  ```def get_file_list(file: File, **kwargs)```

here file, can be a bucket folder path?

Comment (pankajastro) [reply src/astro/files/__init__.py]: discussed with @utkarsharma2 currently, if want user to pass file object then they will have to pass file type too like below 
```
get_file_list(File(path=GCS_BUCKET, conn_id=ASTRO_GCP_CONN_ID, filetype=FileType.CSV))
```

Comment (pankajastro) [reply src/astro/files/__init__.py]: created an issue to fix pattern searching https://github.com/astronomer/astro-sdk/issues/642

Comment (pankajastro) [reply src/astro/files/__init__.py]: @tatiana for now we will go wiouthout file object because would restrict us and user will have to pass filetype also

I have created a ticket to make file object more flexible https://github.com/astronomer/astro-sdk/issues/644

File: src/astro/files/operators/files.py (hunk 18 above)
Comment (pankajastro) [reply src/astro/files/operators/files.py:24]: it was earlier in `get_file_list` but we decided to move it to the operator.

Comment (feluelle) [reply src/astro/files/operators/files.py:24]: Okay. Could you explain why you decided to move it? Because how should `ListFileOperator` know that it gets called within `get_file_list` function? Only then, it makes sense to name the task `get_file_list`. In all other cases `get_file_list` here as task id does not make sense.

File: docs/astro/files/operators/get_file_list.rst [hunk 20]
```diff
@@ -0,0 +1,14 @@
...
+=============
+
+When to use the ``get_file_list`` operator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We can use ``get_file_list`` when you want a list file available in storage based on storage path pattern and Airflow connection. This might useful when you want to create dynamic task on the response of this operator.
+
+.. literalinclude:: ../../../../example_dags/example_dynamic_task_template.py
+   :language: python
+   :start-after: [START howto_operator_get_file_list]
+   :end-before: [END howto_operator_get_file_list]
```
kaxil, line 14: @utkarsharma2 thoughts on docs?

kaxil (approved):

tests/test_example_dags.py
```diff
@@ -35,6 +35,7 @@ def session():
         "example_append",
         "example_load_file",
         "example_merge_bigquery",
+        "example_dynamic_task_template",
```
kaxil, line 38: Btw this test will not succeed on Airflow 2.2.5 job as Dynamic Task Mapping was added in 2.3 @pankajastro

feluelle: How do we handle these cases? 🤔 Should we add a version there and only if this version is installed run the test?

src/astro/files/operators/files.py
pankajastro: just curious, why do ``ListFileOperator`` should know who calls it?
If task_id logic would be in operator then ``get_file_list`` would be a little bit lite and also below code would work I.e you can directly call ListFileOperator without passing `task_id`

```
op1 = ListFileOperator(conn, path)
op2 = ListFileOperator(conn, path)
```

but if you move it in `get_file_list` then you will have pass task_id if you directly use ``ListFileOperator``

regarding name  ``get_file_list`` is just a thin wrapper to return response of ListFileOperator task.

tests/test_example_dags.py
pankajastro: Added a new test case let see if it work e04ab283b439b16247bd81bcc240f432c3a19fb8

tests/test_example_dags.py
```diff
@@ -46,3 +46,23 @@ def test_example_dag(session, dag_id):
     if dag is None:
         raise NameError(f"The DAG with dag_id: {dag_id} was not found")
     test_utils.run_dag(dag, account_for_cleanup_failure=True)
+
+
+@pytest.mark.skip(
+    airflow.__version__ >= "2.3.0", reason="Require Airflow version > 2.3.0"
+)
```
kaxil, line 53:
```suggestion
@pytest.mark.skipif(
    airflow.__version__ < "2.3.0", reason="Require Airflow version > 2.3.0"
)
```

kaxil: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1

kaxil: i.e. you want to skip if the version < 2.3

pankajastro: yes :)

tests/test_example_dags.py
```diff
@@ -46,3 +46,23 @@ def test_example_dag(session, dag_id):
     if dag is None:
         raise NameError(f"The DAG with dag_id: {dag_id} was not found")
     test_utils.run_dag(dag, account_for_cleanup_failure=True)
+
+
+@pytest.mark.skip(
+    airflow.__version__ < "2.3.0", reason="Require Airflow version > 2.3.0"
```
pankajastro, line 52:
```suggestion
    airflow.__version__ < "2.3.0", reason="Require Airflow version >= 2.3.0"
```

kaxil:
Good work 👏 @pankajastro

src/astro/files/operators/files.py
feluelle: > just curious, why do ListFileOperator should know who calls it?

That's what I mean. It should not, but to me it seems that it is aware of who is intended to be the caller as `get_file_list` matches the function name of the function that calls it i.e. `get_file_list`. If it is not clear, we can also quickly jump on a call :)

> but if you move it in get_file_list then you will have pass task_id if you directly use ListFileOperator

Isn't `task_id` a requirement for every operator inheriting from `BaseOperator`, if so you don't have to explicitly add it as an argument.

docs/astro/files/operators/get_file_list.rst
utkarsharma2: Looking good to me.

Issue #572, opened 2022-07-26
Add ability to test example DAG in UI


pankajastro:
**Please describe the feature you'd like to see**
It would be nice if we can add the ability to test example DAG in airflow UI on the latest local astro-sdk code

**Describe the solution you'd like**
Spin the airflow instance using docker and we can keep make file to manage it like we have in astronomer-provider repo

**Are there any alternatives to this feature?**
Is there another way we could solve this problem or enable this use-case?

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**
NA

pankajastro:
I tried to do it here https://github.com/astronomer/astro-sdk/tree/spin_local_airflow_container it working but installation of astro-sdk in container looks bit hacky

PR #597, opened 2022-07-29
Spin Local Airflow Container to Test DAG

pankajastro:
# Description
Locally spin-up airflow container to test the DAG 

closes: #572


Makefile
```diff
@@ -50,23 +50,36 @@ config:  ## Create sample configuration files related to Snowflake, Amazon and G
...
+
+clean: ## Remove all the containers along with volumes
+	docker-compose -f dev/docker-compose.yaml down  --volumes --remove-orphans
+	rm -rf dev/logs
+
+build-run: ## Build the Docker Image & then run the containers
+	docker-compose -f dev/docker-compose.yaml up --build -d
+
+help: ## Prints this message
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-41s\033[0m %s\n", $$1, $$2}'
```
tatiana, line 85: Would it be possible to introduce the new Makefile tasks without renaming all the existing ones? 
It is usually handy to have short-named tasks, and I believe the introduction to container operations shouldn't discourage the other commands

dev/README.md
```diff
@@ -0,0 +1,39 @@
+## Development Environment
```
tatiana, line 1: We already have some development documentation:
https://github.com/astronomer/astro-sdk/tree/main/docs

It may be worth extending it

Makefile
pankajastro: Created a new makefile please let me know if you have a better name for it and also how it is looking now.

dev/README.md
pankajastro: moved the documentation in [docs](https://github.com/astronomer/astro-sdk/tree/main/docs) folder

Makefile.mk
```diff
@@ -0,0 +1,19 @@
+PHONY: build-run clean logs stop help
```
tatiana, line 1: @pankajastro it looks a bit weird that we have both a `Makefile` and a `Makefile.mk` in the project's root.
What do you think of the following directory structure:
```
astro-sdks/
    Makefile
    mk/
        local.mk
        container.mk
```
The Makefile could include includes to both local and container subsets, something similar to what is described in:
https://downloads.haskell.org/~ghc/6.4/docs/html/building/sec-makefile-arch.html

docs/development/DEVELOPMENT.md
```diff
@@ -129,3 +129,57 @@ Publish a release to PyPI:
 * Pass `-R` to skip environment setup, e.g. `nox -Rs lint`
 * Pass `-r` to skip environment creation but re-install packages, e.g. `nox -rs dev`
 * Find more automation commands with `nox -l`
+
+## Test DAG in Airflow UI
```
tatiana, line 133:
```suggestion
## Using a container to run Airflow DAGs
```

.pre-commit-config.yaml
```diff
@@ -35,8 +35,6 @@ repos:
     hooks:
       - id: forbid-crlf
       - id: remove-crlf
-      - id: forbid-tabs
```
pankajastro, line 38: Not sure about this change but is causing issue in makefile

Makefile.mk
pankajastro: I have tried here df5c6616100121cf7f33a5e1fa4f423f352a97c1 please let me know how it is looking.

dev/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apt-get update -y && apt-get install -y git
+RUN apt-get install -y --no-install-recommends \
+        build-essential \
+        libsasl2-2 \
+        libsasl2-dev \
+        libsasl2-modules
+ENV SETUPTOOLS_USE_DISTUTILS=stdlib
+
+COPY ../ .
+COPY . ${AIRFLOW_HOME}/astro_sdk/
```
kaxil, line 13: Can this not be clubbed into one?
```suggestion
COPY ../  ${AIRFLOW_HOME}/astro_sdk/
```

mk/container.mk
```diff
@@ -0,0 +1,19 @@
+PHONY: build-run clean logs stop help
```
kaxil, line 1: I see the following error when running `make container clean` :

```
❯ make container clean
build-run                                 Build the Docker Image & then run the containers
clean                                     Remove all the containers along with volumes
help                                      Prints this message
logs                                      View logs of the all the containers
stop                                      Stop all the containers
make: *** No rule to make target `clean'.  Stop.
```

pankajastro: it would be like 
`make container target=clean`
https://github.com/astronomer/astro-sdk/pull/597/files#diff-4fd9f56e707d4bf8c35dab69efacc13669aa6cbf743ee73409233a9bc6fc6630R170

kaxil:
```
❯ make container target=clean
container.mk:6: *** missing separator.  Stop.
make: *** [container] Error 2
```

kaxil: I now get the above error @pankajastro ^^

pankajastro: weird, let me take the latest code and check it.

Screenshot 2022-08-09 at 7 52 07 PM

tatiana (approved):
Looks great, thanks @pankajastro ! Feel free to merge once the checks are passing!

pankajastro:
Flake8 pre-commit checks were failing in PR, G200 violation. Yes, we violate this check at some places but are still unsure why it is not falling in others PR. for time being I'm disabling the G200 check for flake

dev/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get install -y --no-install-recommends \
+        build-essential \
+        libsasl2-2 \
+        libsasl2-dev \
+        libsasl2-modules
+ENV SETUPTOOLS_USE_DISTUTILS=stdlib
+
+COPY ../  ${AIRFLOW_HOME}/astro_sdk/
+
+RUN pip install -e "${AIRFLOW_HOME}/astro_sdk/"
```
feluelle, line 14: Does it make sense to use the docker multistage here i.e. only copying the installed pip package?

feluelle (commented):
Instead of Make, wdyt of using [task](https://taskfile.dev/)? I find it much more readable. As long as we only use Make for our development, I would say it makes more sense to use something more readable for us. It only requires one `brew install go-task/tap/go-task`.

pankajastro:
> Instead of Make, wdyt of using [task](https://taskfile.dev/)? I find it much more readable, simpler and easier to use. As long as we only use Make for our development, I would say it makes more sense to use something more convenient for us. It only requires one `brew install go-task/tap/go-task`.

I'm not familiar with it and also we already have makefiles for a different purpose in the repo so keeping makefile, for now, make sense.  Created an issue to explore it https://github.com/astronomer/astro-sdk/issues/635

dev/Dockerfile
pankajastro: this would reduce the final image size but it would also increase the build time a little bit given that this Dockefile would be used for local dag testing purposes only so I'm sure we shall introduce multistage or not here.

kaxil:
> Flake8 pre-commit checks were failing in PR, G200 violation. Yes, we violate this check at some places but I'm still unsure why it is not falling in others PR. for time being I'm disabling the G200 check for flake8

I think it wasn't failing on other PRs because of cache, this PR updates the `.pre-commit-config.yaml` file which is used for cache by [pre-commit action](https://github.com/pre-commit/action):

https://github.com/astronomer/astro-sdk/blob/5e9c9d39aa1c6bb917ad52bc26d6cdf09b1611dd/.github/workflows/ci.yaml#L24

Source:
https://github.com/pre-commit/action/blob/4835dd97624f38176d7f9f87fcce089836248ee9/action.yml#L15-L18

.flake8
```diff
@@ -3,5 +3,5 @@ min_python_version = 3.7
 exclude = .git,__pycache__,docs/source/conf.py,old,build,dist
 max-complexity = 6
 max-line-length = 120
-ignore = A002,W503
```
kaxil, line 6: I think you need the following diff to actually fix the `flake8` error:

```diff
diff --git a/src/astro/databases/base.py b/src/astro/databases/base.py
index 60e558e..827e960 100644
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -379,7 +379,11 @@ class BaseDatabase(ABC):
             )
         # Catching NATIVE_LOAD_EXCEPTIONS for fallback
         except self.NATIVE_LOAD_EXCEPTIONS as exe:  # skipcq: PYL-W0703
-            logging.warning(exe)
+            exe: DatabaseCustomError
+            logging.warning(
+                "Loading files failed with Native Support. Falling back to Pandas-based load",
+                exc_info=True
+            )
             if enable_native_fallback:
                 self.load_pandas_dataframe_to_table(
                     source_file.export_to_dataframe(),
diff --git a/src/astro/databases/google/bigquery.py b/src/astro/databases/google/bigquery.py
index ac54aee..04179d9 100644
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -335,7 +335,6 @@ class BigqueryDatabase(BaseDatabase):
         try:
             return str(self.hook.project_id)
         except AttributeError as exe:
-            logging.warning(exe)
             raise DatabaseCustomError(
                 f"conn_id {target_table.conn_id} has no project id"
             ) from exe
diff --git a/src/astro/databases/snowflake.py b/src/astro/databases/snowflake.py
index 3340dd7..49b2122 100644
--- a/src/astro/databases/snowflake.py
+++ b/src/astro/databases/snowflake.py
@@ -423,7 +423,6 @@ class SnowflakeDatabase(BaseDatabase):
         try:
             self.hook.run(sql_statement)
         except (ValueError, AttributeError) as exe:
-            logging.warning(exe)
             raise DatabaseCustomError from exe
         self.drop_stage(stage)
```

kaxil: @feluelle is fixing it in a separate PR

.flake8
```diff
@@ -3,5 +3,5 @@ min_python_version = 3.7
...
+ignore = A002,W503,G200
```
kaxil, line 6:
```suggestion
ignore = A002,W503
```

kaxil (approved):

Issue #594, opened 2022-07-29
Add example DAG and documentation for cleanup


sunank200:
**Please describe the feature you'd like to see**
In the past, we had a tutorial which illustrated how to use each of our operators/decorators:
https://github.com/astronomer/astro-sdk/blob/be6280df00ccff0d7a1c0dfb099b2065303dbe88/REFERENCE.md

**Describe the solution you'd like**
Have a reference page per operator/decorator similar to
https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/ecs.html#howto-operator-ecsoperator

In which we reference parts of an (automated tested) example DAG which illustrates the usage of that operator/decorator.

Many of these use cases already exist in our example DAGs - we should reference them.

* [ ] cleanup 
    - single example

If no example DAGs or operators are illustrating the functionalities, we should create them.

**Acceptance Criteria**
* [ ] cleanup
    - single example
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

Issue #191, opened 2022-03-10
Enable test coverage


tatiana:
**Description**

To improve the testing health of the project, we will add a [codecov](https://about.codecov.io/) integration s.t. all PRs with passing tests will recieve a codecov report. We will also fail any PR where code coverage goes down more than 2%.

**Acceptance criteria**
* Users should have the ability to locally run a code coverage plugin to get reports without submitting to CI
* CI should automatically generate a code coverage report when users submit a PR with passing tests
* As part of the pull request, evaluate the coverage and fail if the coverage goes down by more than -2%
* Enable [`code-cov`](https://about.codecov.io/) on the Github repo

PR #609, opened 2022-08-05
Document the cleanup operator

utkarsharma2:
# Description

closes: #594

## What is the new behavior?

cleanup
        single example

If no example DAGs or operators are illustrating the functionalities, we should create them.

Nope

**Acceptance Criteria**
* [X] cleanup
    - single example
- [X] Example DAG
- [X] Improve the documentation (README, Sphinx, and any other relevant)

### DIFF
--- a/docs/astro/sql/operators/cleanup.rst
+++ b/docs/astro/sql/operators/cleanup.rst
@@ -12,3 +12,5 @@ The ``cleanup`` operator is used to clean temporary tables(:ref:`table`). It mon
    :language: python
    :start-after: [START cleanup_example]
    :end-before: [END cleanup_example]
+
+Users can also specify the temporary tables they want to delete by passing a list of tables in the parameter ``tables_to_cleanup``. If non temporary tables are passed they won't be deleted.
--- a/docs/astro/sql/operators/cleanup.rst
+++ b/docs/astro/sql/operators/cleanup.rst
@@ -4,7 +4,7 @@ cleanup operator
 
 .. _cleanup_operator:
 
-When to Use cleanup operator
+When to use the ``cleanup`` operator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Cleanup operator is used to clean temporary tables(:ref:`table`) created. It monitors the status of all the other tasks and execute at the end to clear all the temporary tables(:ref:`table`) created in a dag run. It is recommended to add ``cleanup`` operator in every DAG that uses the Astro SDK.
 
--- a/docs/astro/sql/operators/cleanup.rst
+++ b/docs/astro/sql/operators/cleanup.rst
@@ -4,7 +4,7 @@ cleanup operator
 
 When to Use cleanup operator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Cleanup operator is used to clean :ref:`temp_table` created. It monitors the status of all the other tasks and execute at the end to clear all the :ref:`temp_table` created in a dag run. It is recommended to add cleanup operator in every DAG that uses AstroSDK.
+Cleanup operator is used to clean :ref:`temp_table` created. It monitors the status of all the other tasks and execute at the end to clear all the :ref:`temp_table` created in a dag run. It is recommended to add ``cleanup`` operator in every DAG that uses the Astro SDK.
 
 .. literalinclude:: ../../../../example_dags/example_amazon_s3_postgres.py
    :language: python
--- a/docs/astro/sql/operators/cleanup.rst
+++ b/docs/astro/sql/operators/cleanup.rst
@@ -2,9 +2,11 @@
 cleanup operator
 ======================================
 
+.. _cleanup_operator:
+
 When to Use cleanup operator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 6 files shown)


docs/astro/sql/operators/cleanup.rst
```diff
@@ -0,0 +1,12 @@
+======================================
+cleanup operator
+======================================
+
+When to Use cleanup operator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Cleanup operator is used to clean :ref:`temp_table` created. It monitors the status of all the other tasks and execute at the end to clear all the :ref:`temp_table` created in a dag run. It is recommended to add cleanup operator in every DAG that uses AstroSDK.
```
tatiana, line 7:
```suggestion
Cleanup operator is used to clean :ref:`temp_table` created. It monitors the status of all the other tasks and execute at the end to clear all the :ref:`temp_table` created in a dag run. It is recommended to add ``cleanup`` operator in every DAG that uses the Astro SDK.
```

docs/concepts.rst
```diff
@@ -0,0 +1,28 @@
+============
+Concepts
+============
+
```
tatiana, line 4: I think we should have a section "Tables" and within it, have an introduction before detailing both table types, something along the lines of:

Tables represent the location and, optionally, the column types of a SQL Database table. They are used in most Astro SDK tasks and decorators.

There are two types of tables:
1. Persistent tables
2. Temporary tables

docs/concepts.rst
```diff
@@ -0,0 +1,28 @@
...
+.. _named_table:
+
+Named Tables
```
tatiana, line 7: If a user did something along the lines:
```
table = Table(conn_id="sample_conn_id")
table.temp = False
```
The table would be "unnamed" (the Python SDK would be generating its name), but a persistent table. Therefore, I think it is safer to call it Persistent table as opposed to Named.

docs/concepts.rst
```diff
@@ -0,0 +1,28 @@
...
+~~~~~~~~~~~~~
+
+These are tables that are of some importance to users and will we persist in a database even after a DAG run is finished. You can create these tables by passing in a ``name`` parameter while creating a ``astro.sql.table.Table`` object.
```
tatiana, line 10: Persistent Table instances will persist after the DAG run and are not affected by the :ref:`aql.cleanup` task. Users can still drop them by using :ref:`aql.drop_table`.

docs/concepts.rst
```diff
@@ -0,0 +1,28 @@
...
+   :start-after: [named_table_example_start]
+   :end-before: [named_table_example_end]
+
+
+.. _temp_table:
+
+Temporary Tables
+~~~~~~~~~~~~~~~~~~~
+
+These are the tables that only exist while a dag is running and will be deleted from the database post that. There is an assumption that these tables are not considered important to the user and only hold intermediate data. You can create these by not passing the ``name`` parameter while creating a ``astro.sql.table.Table`` object.
```
tatiana, line 23:
```suggestion
It is a common pattern to create intermediate tables during a workflow that don't need to be persisted afterwards. To accomplish this, users can use Temporary Tables in conjunction with the :ref:`aql.cleanup` task.

There are two approaches to create temporary tables:
1. Explicit: instantiate a ``astro.sql.table.Table`` using the argument  `temp=True`
2. Implicit: instantiate a ``astro.sql.table.Table`` without giving it a name, and without specifying the `temp` argument

```

docs/concepts.rst
utkarsharma2: This is nice, I'll add it.

utkarsharma2: This is nice, I'll add it.

docs/concepts.rst
utkarsharma2: I missed this case, will change accordingly.

docs/concepts.rst
utkarsharma2: Added both points.

docs/concepts.rst
utkarsharma2: Added the changes.

docs/astro/sql/operators/cleanup.rst
```diff
@@ -0,0 +1,14 @@
+======================================
+cleanup operator
+======================================
+
+.. _cleanup_operator:
+
+When to Use cleanup operator
```
tatiana, line 7:
```suggestion
When to use the ``cleanup`` operator
```

docs/astro/sql/operators/cleanup.rst
```diff
@@ -0,0 +1,14 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Cleanup operator is used to clean temporary tables(:ref:`table`) created. It monitors the status of all the other tasks and execute at the end to clear all the temporary tables(:ref:`table`) created in a dag run. It is recommended to add ``cleanup`` operator in every DAG that uses the Astro SDK.
```
tatiana, line 9:
```suggestion
The ``cleanup`` operator is used to clean temporary tables(:ref:`table`). It monitors the status of all the tasks within a DAG and deletes the created temporary tables(:ref:`table`) by the end of the DAG run. It is recommended to add ``cleanup`` operator in every DAG that uses the Astro SDK.
```

It is probably worth mentioning that the cleanup operator can be used to clean all or a subset of temporary tables, depending on how it is used, and have examples of how it can be used to clean up a subset. Something like:
```
cleanup(temp_table, temp_table)
```

tatiana (approved):

PR #617, opened 2022-08-08
Add schema settings on database level

utkarsharma2:
# Description
When a user is using Astro-SDK for multiple databases. We impose a requirement of having the same-named SCHEMA in all the databases. Which is unrealistic.

related: #304

## What is the new behavior?
Now we have defined config variables to define SCHEMA per-database level.

Users can either configure env var:
AIRFLOW_ASTRO_SDK__POSTGRES_SQL_SCHEMA
AIRFLOW_ASTRO_SDK__BIGQUERY_SQL_SCHEMA
AIRFLOW_ASTRO_SDK__SNOWFLAKE_SQL_SCHEMA
or define them in airflow's config.

Nope

### Checklist

### DIFF
--- a/docs/configurations.rst
+++ b/docs/configurations.rst
@@ -23,7 +23,7 @@ or by updating Airflow's configuration
    [astro_sdk]
    schema = "tmp"
 
-We can also configure schema on database level.
+We can also configure the default schema specific to the database type (example: specific to Snowflake, BigQuery, Postgres). If both the default and database-specific schemas are defined, the preference is given to the database-specific value.
 
 .. code:: python
 
--- a/docs/configurations.rst
+++ b/docs/configurations.rst
@@ -1,0 +1,31 @@
+Configuration Reference
+=======================
+
+Configuring database default schema
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We can configure the default schema that will be used for all operation involving database.
+environment variable :
+
+.. code:: shell
+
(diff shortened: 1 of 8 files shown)


docs/configurations.rst
```diff
@@ -0,0 +1,31 @@
+Configuration
+=============
+
+Configuring database default schema
```
tatiana, line 4:
```suggestion
Configuring the database default schema
```

docs/configurations.rst
```diff
@@ -0,0 +1,31 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We can configure the default schema that will be used for all operation involving database.
```
tatiana, line 6:
```suggestion
We can configure the default schema that will be used for all operations involving databases.
```

docs/configurations.rst
```diff
@@ -0,0 +1,31 @@
...
+   AIRFLOW__ASTRO_SDK__SCHEMA="tmp"
+
+or by updating Airflow's configuration
+
+.. code:: shell
+
+   [astro_sdk]
+   schema = "tmp"
+
+We can also configure schema on database level.
```
tatiana, line 20:
```suggestion
We can also configure the default schema specific to the database type (example: specific to Snowflake, BigQuery, Postgres). If both the default and database-specific schemas are defined, the preference is given to the database-specific value.
```

tatiana (approved):

Issue #599, opened 2022-08-01
Postgres tables being created in default schema 


phanikumv:
When a postgres connection is created with a particular schema , say "postgres", the table gets created in a schema called "tmp_astro" which ideally shouldnt be the case.

Schema name should be read from connection properties and the table should be created accordingly.

kaxil:
@tatiana mentioned that there is already a ticket for it. She will post it here

kaxil:
https://github.com/astronomer/astro-sdk/issues/304 this was it

pankajkoti:
@utkarsharma2 Please link the correct existing ticket

sunank200:
As @tatiana mentioned :
- In Postgres, databases and schema are different concepts: https://www.postgresql.org/docs/current/ddl-schemas.html
- the PostgresHook only exposes schema: https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html
However, implementation-wise, it seems that if the PostgresHook receives a schema during initialisation, but it uses it as a database in  the connection to Postgres: https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L96


So for this change is required on OSS airflow in class `PostgresHook`

phanikumv:
AFAIK it should not use `tmp_astro`  schema, when user provides a schema in connection

sunank200:
That's correct but the way airflow hook handles schema is wrong right now. It treats the database as schema from connections as per https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L96.

This would require change on the OSS airflow side which is a breaking change on the airflow side as well. Also, we shouldn’t use database and schema as the same value. The problem is: that Airflow Postgres Hook exposes the Postgres Database property incorrectly with the name schema. That’s why we’re doing this in the Astro SDK till its fixed on OSS airflow side

PR #618, opened 2022-08-08
Add doc string with details about schema and database for Postgres

sunank200:
- In Postgres, databases and schema are different concepts: https://www.postgresql.org/docs/current/ddl-schemas.html
- the PostgresHook only exposes schema: https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html
- However, implementation-wise, it seems that if the PostgresHook receives a schema during initialisation, but it uses it as a database in the connection to Postgres: https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L96

So for this change is required on OSS airflow in class PostgresHook. Add doc-string mentioning that.

closes: #599 

### Checklist

### DIFF
--- a/src/astro/databases/postgres.py
+++ b/src/astro/databases/postgres.py
@@ -40,7 +40,18 @@ class PostgresDatabase(BaseDatabase):
 
     @property
     def default_metadata(self) -> Metadata:
-        """Fill in default metadata values for table objects addressing postgres databases"""
+        """
+        Fill in default metadata values for table objects addressing postgres databases.
+
+        Currently, Schema is not being fetched from airflow connection for Postgres because in Postgres, databases and
+        schema are different concepts: https://www.postgresql.org/docs/current/ddl-schemas.html
+        The PostgresHook only exposes schema:
+        https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html
+        However, implementation-wise, it seems that if the PostgresHook receives a schema during initialisation,
+        but it uses it as a database in the connection to Postgres:
+        https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L96
+        """
+        # TODO: Change airflow PostgresHook to fetch database and schema separately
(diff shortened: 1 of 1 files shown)


src/astro/databases/postgres.py
```diff
@@ -40,7 +40,18 @@ def hook(self) -> PostgresHook:
...
+        """
+        Fill in default metadata values for table objects addressing postgres databases.
+
+        Currently, Schema is not being fetched from airflow connection for Postgres because in Postgres, databases and
+        schema are different concepts: https://www.postgresql.org/docs/current/ddl-schemas.html
+        The PostgresHook only exposes schema:
+        https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html
+        However, implementation-wise, it seems that if the PostgresHook receives a schema during initialisation,
+        but it uses it as a database in the connection to Postgres:
+        https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L96
```
utkarsharma2, line 52:
```suggestion
        Fill in default metadata values for table objects addressing Postgres databases.
        
        Currently, Schema is not being fetched from airflow connection for Postgres because, in Postgres, databases and
        schema are different concepts: https://www.postgresql.org/docs/current/ddl-schemas.html
        The PostgresHook only exposes schema:
        https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html
        However, implementation-wise, it seems that if the PostgresHook receives a schema during initialization,
        but it uses it as a database in the connection to Postgres:
        https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L96
```

sunank200: What is the difference @utkarsharma2

utkarsharma2: @sunank200 Some suggestions via Grammarly.

utkarsharma2 (approved):

PR #619, opened 2022-08-09
Document for append operator

utkarsharma2:
# Description

closes: #590 

## What is the new behavior?

Nope

### Checklist


tests/test_example_dags.py
```diff
@@ -25,13 +25,14 @@ def session():
...
-        "example_snowflake_partial_table_with_append",
-        "example_sqlite_load_transform",
-        "example_dynamic_map_task",
+        # "example_amazon_s3_postgres",
+        # "example_amazon_s3_postgres_load_and_save",
+        # "example_amazon_s3_snowflake_transform",
+        # "example_google_bigquery_gcs_load_and_save",
+        # "example_snowflake_partial_table_with_append",
+        # "example_sqlite_load_transform",
+        # "example_dynamic_map_task",
```
pankajkoti, line 34: Does this need to remain commented?

example_dags/example_append.py
```diff
@@ -0,0 +1,48 @@
...
+    dag_id="example_append",
+    start_date=datetime(2019, 1, 1),
+    max_active_runs=3,
+    schedule_interval=timedelta(minutes=30),
+    default_args=default_args,
+)
+
+with dag:
+    load_main = aql.load_file(
+        input_file=File(path=str(CWD) + "/data/homes.csv"),
```
pankajkoti, line 28: We could also create a DATA_DIR as a variable with `DATA_DIR = CWD + "/data/"` and reuse it here and in the below `load_append` task

docs/astro/sql/operators/append.rst
```diff
@@ -0,0 +1,43 @@
+======================================
+append operator
+======================================
+
+.. _append_operator:
+
+When to use the ``append`` operator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
pankajkoti, line 8: Can we restrict the tilde(s) to the text length above? How does it appear? I mean to see do we see extra tildes below the text? And shouldn't those be equal to the text above?

docs/astro/sql/operators/append.rst
```diff
@@ -0,0 +1,43 @@
...
+
+When used without a columns parameter, AstroSDK assumes that both tables have the same schema.
+
+When tables have same schema
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#. **Case 1:** When complete table needs to be merged. You can skip ``columns`` parameter.
+    .. literalinclude:: ../../../../example_dags/example_append.py
+       :language: python
+       :start-after: [START append_example]
+       :end-before: [END append_example]
```
pankajkoti, line 24: This block is same as the block on line 11-14. Do we need the block on line 11-14 since we're covering it in 2-3 cases, one without columns here and one without columns below.

docs/astro/sql/operators/append.rst
```diff
@@ -0,0 +1,43 @@
...
+When tables have different schema, we can map different column names by passing a ``dict`` of *source cols to target cols*.
+
+    .. literalinclude:: ../../../../example_dags/example_append.py
+       :language: python
+       :start-after: [START append_example_col_dict]
+       :end-before: [END append_example_col_dict]
+
+Conflicts
+~~~~~~~~~
+``append operator`` don't handle the conflicts that may arise while appending data. If you want to handle those scenarios you can use ``merge operator``
```
pankajkoti, line 43:
```suggestion
``append operator`` doesn't handle the conflicts that may arise while appending data. If you want to handle those scenarios, you can use ``merge operator``
```

tests/test_example_dags.py
utkarsharma2: Nope, added a commit to fix it.

docs/astro/sql/operators/append.rst
utkarsharma2: @pankajkoti It doesn't need to be equal but not less than the heading text. But yes this looks odd, I will change it.

example_dags/example_append.py
utkarsharma2: Done.

docs/astro/sql/operators/append.rst
utkarsharma2: @pankajkoti  It's there to explain the basic and use-case based usage. I think both make sense in their own place. WDYT?

pankajkoti: No specific opinion as such. Just felt that the code block that would appear will be the same and won't have any difference, so could be a duplicate.

kaxil (approved):

example_dags/example_append.py
```diff
@@ -0,0 +1,50 @@
...
+
+DATA_DIR = str(CWD) + "/data/"
+
+with dag:
+    load_main = aql.load_file(
+        input_file=File(path=DATA_DIR + "homes.csv"),
+        output_table=Table(conn_id="postgres_conn"),
+    )
+    load_append = aql.load_file(
+        input_file=File(path=DATA_DIR + "/homes2.csv"),
```
pankajkoti, line 34:
```suggestion
        input_file=File(path=DATA_DIR + "homes2.csv"),
```
We have an extra `/` here

pankajkoti (approved):

docs/astro/sql/operators/append.rst
utkarsharma2: @pankajkoti I see your point, I'll look if there is a better way to deal with this.

Issue #604, opened 2022-08-03
Investigate if we can unpin pandas


kaxil:
We have pinned Pandas library https://github.com/astronomer/astro-sdk/blob/6b90c92029c7470f5e383345fcdaf77861c8cb94/pyproject.toml#L19

This was done in https://github.com/astronomer/astro-sdk/pull/130 but a lot of things have changed since. Let's investigate if this pin (upper bound) is still required.

Related Slack thread: https://astronomer.slack.com/archives/C03868KGF2Q/p1659546153698379?thread_ts=1659545419.511649&cid=C03868KGF2Q

**Acceptance Criteria**:
Run example DAGs with unpinned pandas and see if it runs fine and create a PR on this repo to see if CI passes.

PR #620, opened 2022-08-09
Unpin pandas upperbound dependency

utkarsharma2:
# Description
We have pinned Pandas library

[astro-sdk/pyproject.toml](https://github.com/astronomer/astro-sdk/blob/6b90c92029c7470f5e383345fcdaf77861c8cb94/pyproject.toml#L19)

Line 19 in [6b90c92](https://github.com/astronomer/astro-sdk/commit/6b90c92029c7470f5e383345fcdaf77861c8cb94)
 "pandas>=1.3.4,<=1.3.5", 

This was done in https://github.com/astronomer/astro-sdk/pull/130 but a lot of things have changed since. Let's investigate if this pin (upper bound) is still required.

Related Slack thread: https://astronomer.slack.com/archives/C03868KGF2Q/p1659546153698379?thread_ts=1659545419.511649&cid=C03868KGF2Q

closes: #604 

## What is the new behavior?
unpinned the upper bound of pandas dependency

Nope

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ requires-python = ">=3.7"
 dependencies = [
     "apache-airflow>=2.0",
     "markupsafe>=1.1.1,<2.1.0",
-    "pandas>=1.3.4,<=1.3.5",
+    "pandas>=1.3.4",
     "pyarrow",
     "python-frontmatter",
     "smart-open",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -16,7 +16,7 @@ requires-python = ">=3.7"
 dependencies = [
     "apache-airflow>=2.0",
     "markupsafe>=1.1.1,<2.1.0",
```
kaxil, line 18: Worth investigating in a separate PR/issue if we can remove `markupsafe`

utkarsharma2: @kaxil I'll create a ticket for the same. Should it be part of this sprint?

kaxil: Yes please

kaxil (approved):

Issue #516, opened 2022-07-07
Fix flaky test_example_dag[example_snowflake_partial_table_with_append]


tatiana:
**Version**
* Astro: 0.11.0

**Describe the bug**

Quite often this test group fails to run in the CI:
```
nox -s "test-3.8(airflow='2.3')" -- --splits 12 --group 1 --cov=src --cov-report=xml --cov-branch
```
With:
```
=========================== short test summary info ============================
FAILED tests/test_example_dags.py::test_example_dag[example_snowflake_partial_table_with_append]
==== 1 failed, 29 passed, 329 deselected, 22 warnings in 294.40s (0:04:54) =====
```

Even if there were no changes which affected this test/our code-base, for instance:
https://github.com/astronomer/astro-sdk/pull/480
https://github.com/astronomer/astro-sdk/runs/7231300739

**Expected behaviour**
This test should not fail unless there is an issue with the feature tested.

PR #626, opened 2022-08-10
Added 'extend_existing' to Sqla Table object

utkarsharma2:
# Description
Quite often this test group fails to run in the CI:

nox -s "test-3.8(airflow='2.3')" -- --splits 12 --group 1 --cov=src --cov-report=xml --cov-branch

With:

=========================== short test summary info ============================
FAILED tests/test_example_dags.py::test_example_dag[example_snowflake_partial_table_with_append]
==== 1 failed, 29 passed, 329 deselected, 22 warnings in 294.40s (0:04:54) =====

Even if there were no changes that affected this test/our code-base, for instance:
https://github.com/astronomer/astro-sdk/pull/480
https://github.com/astronomer/astro-sdk/runs/7231300739

closes: #516 

## What is the new behavior?
The issue was with SQLA's reflection cache which keeps track of the tables created and maintains all the tables in a map 
`schema_columns[self.normalize_name(table_name)]`. When this test was individually run there are no issues on local since SQLA cache is just initialized just for this test and because of which this was an intermittent issue CI. The fix is to introduce a parameter [extend_existing](https://docs.sqlalchemy.org/en/14/core/metadata.html?highlight=table#sqlalchemy.schema.Table.params.extend_existing) which basically ensures the cache is not used.

Ran test multiple time to ensure this is working:
https://github.com/astronomer/astro-sdk/actions/runs/2832783413

Nope

### DIFF
--- a/example_dags/example_snowflake_partial_table_with_append.py
+++ b/example_dags/example_snowflake_partial_table_with_append.py
@@ -5,15 +5,7 @@ import pandas as pd
 from airflow.decorators import dag
 
 from astro.files import File
-from astro.sql import (
-    append,
-    cleanup,
-    dataframe,
-    drop_table,
-    load_file,
-    run_raw_sql,
-    transform,
-)
+from astro.sql import append, cleanup, dataframe, load_file, run_raw_sql, transform
 from astro.sql.table import Metadata, Table
 
 """
@@ -66,10 +58,10 @@ def filter_data(homes_long: Table):
 
 
 @run_raw_sql
-def create_table():
+def create_table(table: Table):
     """Create the reporting data which will be the target of the append method"""
     return """
-    CREATE TABLE IF NOT EXISTS homes_reporting (
+    CREATE TABLE IF NOT EXISTS {{table}} (
       sell number,
       list number,
       variable varchar,
@@ -85,7 +77,6 @@ def example_snowflake_partial_table_with_append():
     homes_data1 = load_file(
         input_file=File(path=FILE_PATH + "homes.csv"),
         output_table=Table(
-            name="homes",
             conn_id=SNOWFLAKE_CONN_ID,
             metadata=Metadata(
                 database=os.getenv("SNOWFLAKE_DATABASE"),
@@ -97,7 +88,6 @@ def example_snowflake_partial_table_with_append():
     homes_data2 = load_file(
         input_file=File(path=FILE_PATH + "homes2.csv"),
         output_table=Table(
-            name="homes2",
             conn_id=SNOWFLAKE_CONN_ID,
             metadata=Metadata(
                 database=os.getenv("SNOWFLAKE_DATABASE"),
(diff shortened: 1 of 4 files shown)


pankajkoti (approved):

feluelle (approved):
SGTM 👍

tatiana:
@utkarsharma2 isn't the root cause of the issue the fact that concurrent tests are trying to create a table with the same name (`home` and `homes2`)?
https://github.com/astronomer/astro-sdk/blob/367f94b3cbcfd48a3a1f8ce84c13fe5fcd41e08d/example_dags/example_snowflake_partial_table_with_append.py#L88

Wouldn't it be better to change the example DAG to create tables with unique names?

src/astro/databases/base.py
```diff
@@ -477,7 +477,10 @@ def get_sqla_table(self, table: Table) -> SqlaTable:
         :param table: Astro Table to be converted to SQLAlchemy table instance
         """
         return SqlaTable(
-            table.name, table.sqlalchemy_metadata, autoload_with=self.sqlalchemy_engine
+            table.name,
+            table.sqlalchemy_metadata,
+            autoload_with=self.sqlalchemy_engine,
+            extend_existing=True,
```
tatiana, line 483: Without these arguments, the Python SDK would raise an exception if concurrent processes tried to update/create the same table. With this option, we may be hiding this type of issue - making it harder for users to troubleshoot this type of anti-pattern.

utkarsharma2: @tatiana This was my first guess as well I converted persistent tables to temp tables, but the issue still persisted - https://github.com/astronomer/astro-sdk/runs/7765749178?check_suite_focus=true

And I don't think busting the cache will hide the issue if anything this will bring up the issue sooner since now you will be referring to the database directly and not the cache. I don't understand how this is this antipattern.

utkarsharma2: @tatiana  But I do understand your point, having temp would be better, and chances of collision would be reduced. I can make this change on top of existing changes. WDYT?

utkarsharma2: As per the discussion on a call with @tatiana, we concluded the following:

1. The `extend_existing` is required and was the root cause of the issue in CI.
2.  Persistent tables in example dag can also lead to conflicts when there are parallel test suites running. 

Based on the above points I have updated the PR to reflect those changes.

tatiana (approved):

example_dags/example_snowflake_partial_table_with_append.py
```diff
@@ -119,31 +109,24 @@ def example_snowflake_partial_table_with_append():
...
 
-    # We truncate this table only to avoid wasting Snowflake resources
-    # Why? Between 2022-03-25 and 2022-04-11 it accumulated 301G (89 million rows) because
-    # this example DAG used to append rows without deleting them
-    # [START drop_table_example]
-    truncate_results = drop_table(
-        table=Table(name="homes_reporting", conn_id=SNOWFLAKE_CONN_ID)
-    )
-    # [END drop_table_example]
-    truncate_results.set_upstream(record_results)
```
tatiana, line 146: Won't this break our truncate documentation?

src/astro/databases/base.py
tatiana: Amazing, thank you very much, @utkarsharma2 !

example_dags/example_snowflake_partial_table_with_append.py
utkarsharma2: @tatiana  Nope it won't I added a separate example for this case in example_dags/example_sqlite_load_transform.py file.

tatiana (approved):

PR #629, opened 2022-08-11
Add document for merge operator

utkarsharma2:
# Description

closes: #591 

## What is the new behavior?

merge
    with conflicts (what options the user has)
    without conflicts

Nope

### Checklist

### DIFF
--- a/docs/astro/sql/operators/merge.rst
+++ b/docs/astro/sql/operators/merge.rst
@@ -12,7 +12,7 @@ Merge operator is used when you expect some conflicts while merging two tables d
 Prerequisite
 ------------
 
-Merge operator internally run different SQL queries based on the databases and some databases only allow you to run these queries if there are constraints on columns specified in parameter ``target_conflict_columns``. Below is the list of databases and their constraint requirement.
+Merge operator internally runs different SQL queries based on the databases and some databases only allow you to run these queries if there are constraints on columns specified in parameter ``target_conflict_columns``. Below is the list of databases and their constraint requirement.
 
 .. list-table::
    :widths: auto
--- a/docs/astro/sql/operators/merge.rst
+++ b/docs/astro/sql/operators/merge.rst
@@ -1,3 +1,5 @@
+.. _merge_operator:
+
 ======================================
 merge operator
 ======================================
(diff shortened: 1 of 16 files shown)


docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+Prerequisite
+------------
+
+Merge operator internally run different SQL queries based on the databases and some databases only allow you to run these queries if there are constraints on columns specified in parameter ``target_conflict_columns``. Below is the list of databases and their constraint requirement.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Database
```
tatiana, line 21: It is a bit confusing to have Database as one of the items of this list at the same level of the others.

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+   * - Bigquery
+     - No
+   * - Postgres
+     - Yes
+   * - Snowflake
+     - Yes
+   * - SQLite
+     - Yes
+
+
```
tatiana, line 32: I think it would be more helpful if you gave an example of the SQL statement that should be run (eg. an example of creating a constraint in Snowflake/postgres, and an example of creating an index in Sqlite).

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+Merge operator internally run different SQL queries based on the databases and some databases only allow you to run these queries if there are constraints on columns specified in parameter ``target_conflict_columns``. Below is the list of databases and their constraint requirement.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Database
+     - Constraint required
+   * - Bigquery
+     - No
```
tatiana, line 24: No requirements

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+   * - Postgres
+     - Yes
```
tatiana, line 26: Depends on previously created constraint

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+   * - Snowflake
+     - Yes
```
tatiana, line 28: Depends on previously created constraint

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+   * - SQLite
+     - Yes
+
```
tatiana, line 31: Depends on previously created index

docs/astro/sql/operators/merge.rst
utkarsharma2: It's header row actually. Will be rendered like -
Screenshot 2022-08-11 at 6 23 09 PM

docs/astro/sql/operators/merge.rst
utkarsharma2: @tatiana I think that would be a bit misleading since it would imply that users need to create constraints before using the ``Merge operator``. But that's not the case and users might have different constraints altogether.

docs/astro/sql/operators/merge.rst
tatiana: 🙈 that's great, @utkarsharma2, thanks!

docs/astro/sql/operators/merge.rst
tatiana: @utkarsharma2 My understanding was that, without constraints, the merge operator wouldn't work for Postgres and Snowflake. Did I get it all wrong? 🙈

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+   * - Database
+     - Constraint required
+   * - Bigquery
+     - No
+   * - Postgres
+     - Yes
+   * - Snowflake
+     - Yes
+   * - SQLite
+     - Yes
```
tatiana, line 30: Doesn't SQLite expect an index as opposed to constraint?

docs/astro/sql/operators/merge.rst
utkarsharma2: @tatiana That is correct, but we don't have to create constraints to use the Merge operator. But if we have constraints then we use Merge. If we don't we use Append.

docs/astro/sql/operators/merge.rst
utkarsharma2: @tatiana I think it's sqlite way of creating constraints, based on the following link 
https://stackoverflow.com/questions/15497985/how-to-add-unique-constraint-to-existing-table-in-sqlite

docs/astro/sql/operators/merge.rst
tatiana: @utkarsharma2 as we agreed, please, adapt a complete end-to-end example of loading with the Python SDK setting columns constraints and then merging - and highlight in the docs this part of the table creation, please.

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
...
+
+When to use the ``merge`` operator
+----------------------------------
+
+Merge operator is used when you expect some conflicts while merging two tables due to underline constraints. If you don't expect any conflicts you can use :ref:`append_operator`. There are two common cases when merging tables.
+
+Prerequisite
+------------
+
+Merge operator internally run different SQL queries based on the databases and some databases only allow you to run these queries if there are constraints on columns specified in parameter ``target_conflict_columns``. Below is the list of databases and their constraint requirement.
```
kaxil, line 15:
```suggestion
Merge operator internally runs different SQL queries based on the databases and some databases only allow you to run these queries if there are constraints on columns specified in parameter ``target_conflict_columns``. Below is the list of databases and their constraint requirement.
```

docs/astro/sql/operators/merge.rst
```diff
@@ -0,0 +1,143 @@
+.. _merge_operator:
+
+======================================
+merge operator
+======================================
+
+When to use the ``merge`` operator
+----------------------------------
+
+Merge operator is used when you expect some conflicts while merging two tables due to underline constraints. If you don't expect any conflicts you can use :ref:`append_operator`. There are two common cases when merging tables.
```
dimberman, line 10:
```suggestion
Unlike the :ref:`append_operator`, which expects data to be unique and conflict free, the Merge operator allows users to add data with conflict resolution techniques like `ignore` and `update`.
```

docs/astro/sql/operators/merge.rst
dimberman: I think we should mention that while you don't need to create a constraint, a constraint DOES need to exist to guarantee uniqueness. That is an important point for when to use append/merge.

utkarsharma2: Added an example dag to demonstrate the table creation with constraints.

utkarsharma2:
Closing this one as this had conflicts. Raised a new PR - https://github.com/astronomer/astro-sdk/pull/636

PR #634, opened 2022-08-12
Make the operator names consistent

kaxil:
`LoadFile` and `ExportFile` didn't have operator suffix. While it is not needed, I thought it would be good to make them consistent.


feluelle (approved):
Definitely! 💯

pankajkoti (approved):

PR #637, opened 2022-08-12
Move decorators to specific Operator files

kaxil:
This PR/commit moves the decorator code to individual files instead of `src/astro/sql/__init__.py`, while still importing them in `src/astro/sql/__init__.py` so the import paths will still remain the same but atleast now we can just look at the operator and decorator in the same file like we already do for `load_file` and `export_file`.

The decorator/functions are in `src/astro/sql/__init__.py` but Operator is in separate file

## What is the new behavior?

Move the functions to the specific files and still expose the same API to users by importing them in `src/astro/sql/__init__.py`

NO


src/astro/sql/operators/append.py
```diff
@@ -54,3 +54,28 @@ def execute(self, context: dict) -> Table:  # skipcq: PYL-W0613
...
         return self.target_table
+
+
+def append(
+    *,
+    source_table: Table,
+    target_table: Table,
+    columns: APPEND_COLUMN_TYPE = None,
+    **kwargs: Any,
+):
```
feluelle, line 65: Let's please add the return type for these functions.

src/astro/sql/operators/drop.py
```diff
@@ -31,3 +31,17 @@ def execute(self, context: Dict) -> Table:  # skipcq: PYL-W0613
...
+
+def drop_table(
+    table: Table,
+    **kwargs: Any,
+):
+    """
+    Drops a table.
+
+    :param table: Table to be dropped
+    :param kwargs:
```
feluelle, line 44:
```suggestion
```
Are those the BaseOperator ones? If yes, we don't necessary need to document them. If they contain other kwargs as well, let us document them :)

src/astro/sql/__init__.py
```diff
@@ -1,270 +1,10 @@
...
-except ImportError:
-    from airflow.decorators.base import task_decorator_factory
-    from airflow.decorators import _TaskDecorator as TaskDecorator
-
-from astro.constants import ColumnCapitalization, MergeConflictStrategy
-from astro.sql.operators.append import APPEND_COLUMN_TYPE, AppendOperator
-from astro.sql.operators.cleanup import CleanupOperator
-from astro.sql.operators.dataframe import DataframeOperator
-from astro.sql.operators.drop import DropTableOperator
+from astro.sql.operators.append import AppendOperator, append  # noqa: F401
```
feluelle, line 1:
```suggestion
from astro.sql.operators.append import AppendOperator, append  # noqa: F401
```
How about we add `# noqa: F401` at the top of the file or in the config for this specific file?

src/astro/sql/operators/cleanup.py
```diff
@@ -213,3 +213,19 @@ def resolve_tables_from_tasks(
...
+    The cleanup operator allows for two possible scenarios: Either a user wants to clean up a specific set of tables
+    during the DAG run, or the user wants to ensure that all temporary tables are deleted once the DAG run is finished.
+    The idea here is to ensure that even if a user doesn't have access to a "temp" schema, that astro does not leave
+    hanging tables once execution is done.
+
+    :param tables_to_cleanup: A list of tables to cleanup, defaults to waiting for all upstream tasks to finish
+    :param kwargs:
+    :return:
+    """
+    return CleanupOperator(tables_to_cleanup=tables_to_cleanup, **kwargs)
```
feluelle, line 231: Does it make sense to also return the `.output` here too?

src/astro/sql/__init__.py
kaxil: There isn't a way in flake8 to ignore a block of code as per the maintainer of flake8

Your options are:

1. ignore each line that produces an error by putting # noqa: E501 on it
2. ignore the entire file (but this turns off all other errors as well) with a # flake8: noqa on a line by itself
3. ignore E501 in the entire file by using per-file-ignores:
```ini
[flake8]
per-file-ignores =
     path/to/file.py: E501
```

kaxil: will do the 3rd option

src/astro/sql/operators/cleanup.py
feluelle: If not, what does returning the operator bring us?

src/astro/sql/__init__.py
kaxil: Done in https://github.com/astronomer/astro-sdk/pull/637/commits/98fd8a561cb136b865525b33b5b1207dffd1d122

src/astro/sql/operators/drop.py
kaxil: Done in https://github.com/astronomer/astro-sdk/pull/637/commits/98fd8a561cb136b865525b33b5b1207dffd1d122

src/astro/sql/operators/append.py
kaxil: Done in https://github.com/astronomer/astro-sdk/pull/637/commits/98fd8a561cb136b865525b33b5b1207dffd1d122

src/astro/sql/operators/cleanup.py
kaxil: 🤷 - Probably just syntactic sugar to have "functions". We can return table objects that represent the deleted table but I wouldn't like to change the implementation in this PR as this PR just moves objects around

feluelle (approved):

PR #650, opened 2022-08-12
Use lazy evaluated Type Annotations from PEP 563

kaxil:
Details: https://peps.python.org/pep-0563/ and https://discuss.python.org/t/type-annotations-pep-649-and-pep-563/11363/20

tl_dr: This PEP lazy evaluates Annotations so they are not evaluated at parse time and allows forward references as well.

This also includes https://peps.python.org/pep-0604/ which allows writing §typing.Union[int, str]` as `int | str` and includes https://peps.python.org/pep-0585/ which allows us to write `def f(x: List[str]) -> None` to `def f(x: list[str]) -> None`

### DIFF
--- a/src/astro/sql/operators/append.py
+++ b/src/astro/sql/operators/append.py
@@ -1,15 +1,13 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
 from airflow.decorators.base import get_unique_task_id
 from airflow.models.baseoperator import BaseOperator
+from airflow.models.xcom_arg import XComArg
 
 from astro.databases import create_database
 from astro.sql.table import Table
-
-if TYPE_CHECKING:
-    from airflow.models.xcom_arg import XComArg
 
 
 class AppendOperator(BaseOperator):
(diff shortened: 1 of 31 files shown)


kaxil:
Also credits to @tatiana she brought it up first in https://github.com/astronomer/astro-sdk/pull/384#discussion_r878956136

kaxil:
There are some caveats though as listed in https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563.

@uranusjr Is it safe to use this or should we wait till we drop support for Python 3.7 and 3.8?

src/astro/sql/operators/append.py
```diff
@@ -9,8 +11,6 @@
 if TYPE_CHECKING:
     from airflow.models.xcom_arg import XComArg
```
uranusjr, line 12: This `if TYPE_CHECKING` guard seems unnecessary?

uranusjr (approved):

src/astro/sql/operators/append.py
kaxil: Fixed

feluelle (approved):

pankajkoti:
There are a couple of suggestion from DeepSource:Python: https://deepsource.io/gh/astronomer/astro-sdk/run/8c8ca0ea-fa9c-48f3-901b-d8e2c8ea476a/python/

They are unrelated to this PR but should we fix them as part of this PR?

PR #651, opened 2022-08-15
Fix development environments

feluelle:
# Description

- fix local.mk to respect the correct directory
- fix container.mk to use native docker compose from Docker Desktop
- improve development docs

### Checklist

### DIFF
--- a/docs/development/DEVELOPMENT.md
+++ b/docs/development/DEVELOPMENT.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
-* Python 3.9
+* At least Python 3.7, 3.8 or 3.9
 * (Optional but highly recommended) [pyenv](https://github.com/pyenv/pyenv)
 
 _On **Apple M1** it is [currently required](https://github.com/psycopg/psycopg2/issues/1286#issuecomment-914286206) to install `postgresql` package. Once [compatible wheels](https://github.com/psycopg/psycopg2/issues/1482) are released, you can remove it._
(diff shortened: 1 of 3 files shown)


docs/development/DEVELOPMENT.md
```diff
@@ -1,18 +1,18 @@
 # Developing the package
 
-Prerequisites
+## Prerequisites
 
-* (Optional but highly recommended) [pipx](https://pypa.github.io/pipx/).
-* [Nox](https://nox.thea.codes/) for all automation.
+* Python 3.9
```
tatiana, line 5: We support Python 3.7, 3.8 and 3.9 ATM

feluelle, line 5:
```suggestion
* At least Python 3.7, 3.8 or 3.9
```

feluelle: WDYT @tatiana?

kaxil: Fine for me

docs/development/DEVELOPMENT.md
```diff
@@ -140,50 +151,50 @@ You can configure the Docker-based testing environment to test your DAG
...
-2. Run ``make container target=build-run``
+1. Run ``make container target=build-run``
 
-3. Put the DAGs you want to run in the dev/dags directory:
+1. Put the DAGs you want to run in the dev/dags directory:
 
-4. If you want to add Connections, create a connections.yaml file in the dev directory.
+1. If you want to add Connections, create a connections.yaml file in the dev directory.
 
    See the `Connections Guide <https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html>`_ for more information.
```
kaxil, line 160: Not related to this PR, but this link is RST not MD, we should fix that too

mk/container.mk
```diff
@@ -3,17 +3,17 @@ PHONY: build-run clean logs stop help
 .DEFAULT_GOAL:= help
 
 logs: ## View logs of the all the containers
-	docker-compose -f ../dev/docker-compose.yaml logs --follow
+	docker compose -f ../dev/docker-compose.yaml logs --follow
```
kaxil, line 6: Why this change?

kaxil: TIL: Looks like "docker compose" is the new way approach and written in GO as compared to `docker-composer`. 

https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command

cc @astronomer/airflow-providers - We should update it in astronomer-providers repo too :)

kaxil (approved):

mk/container.mk
feluelle: Yes, exactly. It is much nicer also in terms of output. I believe the v1 docker-compose did not have coloured output and is not collapsible.

Issue #579, opened 2022-07-27
Make the tasks and operators ids consistent


tatiana:
**Please describe the feature you'd like to see**
At the moment, if the user does not define `task_id`, we are naming tasks inconsistently across different operators and decorators. Some are prefixed or suffixed with `_`, others are not.

**Describe the solution you'd like**
* By default, name based on the functionality, without `_`, eg. `load_file`
* If the name is already used by another task, suffix with `_` and a unique identifier

**Additional context**
Example of how the [tutorial](https://github.com/astronomer/astro-sdk/blob/main/docs/getting-started/GETTING_STARTED.md) currently renders:
<img width="1228" alt="Screenshot 2022-07-27 at 10 58 00" src="https://user-images.githubusercontent.com/272048/181233379-d0cecec0-916d-4a1f-9330-8a40b22a1901.png">

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] All decorators and tasks are named consistently (both the tutorial and the other ones)

PR #663, opened 2022-08-16
Make default operator task_ids consistent

utkarsharma2:
# Description
At the moment, if the user does not define `task_id`, we are naming tasks inconsistently across different operators and decorators. Some are prefixed or suffixed with `_`, others are not.

closes: #579

## What is the new behavior?
* By default, name based on the functionality, without `_`, eg. `load_file` 
* If the name is already used by another task, suffix with `_` and a unique identifier - used `get_unique_task_id` which always generates a unique task id for tasks.

Nope

### Checklist
- [x] Created tests that fail without the change

### DIFF
--- a/src/astro/sql/operators/export_file.py
+++ b/src/astro/sql/operators/export_file.py
@@ -91,7 +91,7 @@ def export_file(
     :param task_id: task id, optional
     """
 
-    task_id = task_id if task_id is not None else get_unique_task_id("export_file")
+    task_id = task_id or get_unique_task_id("export_file")
 
     return ExportFileOperator(
         task_id=task_id,
--- a/src/astro/sql/operators/export_file.py
+++ b/src/astro/sql/operators/export_file.py
@@ -1,6 +1,7 @@
 from typing import Any, Optional, Union
 
 import pandas as pd
+from airflow.decorators.base import get_unique_task_id
 from airflow.models import BaseOperator
 from airflow.models.xcom_arg import XComArg
 
@@ -8,7 +9,6 @@ from astro.constants import ExportExistsStrategy
 from astro.databases import create_database
 from astro.files import File
 from astro.sql.table import Table
-from astro.utils.task_id_helper import get_task_id
 
 
 class ExportFileOperator(BaseOperator):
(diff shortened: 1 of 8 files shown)


src/astro/sql/operators/export_file.py
```diff
@@ -91,9 +91,7 @@ def export_file(
     :param task_id: task id, optional
     """
 
-    task_id = (
-        task_id if task_id is not None else get_task_id("export_file", output_file.path)
-    )
+    task_id = task_id if task_id is not None else get_unique_task_id("export_file")
```
sunank200, line 94: What happens when `task_id` provided by user is not unique?

sunank200 (approved):
Overall LGTM. Had a question - what happens when `task_id` provided by user is not unique?

src/astro/sql/operators/export_file.py
utkarsharma2: @sunank200  get_unique_task_id() appends a suffix and convert it to a unique task id

pankajastro: If the user provides task_id then it is expected that they would provide unique task_id if they user won't provide task_id then it will generate unique task_id

pankajastro:
LGTM

pankajastro (approved):

src/astro/sql/operators/export_file.py
utkarsharma2: @sunank200 it will raise the following error:
Screenshot 2022-08-16 at 5 01 09 PM

But I'm not sure if we should update the task_id provided by users or not. 
@kaxil @tatiana @feluelle thoughts?

feluelle: I think it is okay like that.

Btw you can also make it a bit simpler:
```suggestion
    task_id = task_id or get_unique_task_id("export_file")
```

feluelle: If the user does not provider a unique id an exception should be raised to the user.

utkarsharma2: I'm also of the same opinion that we should raise the exception.

feluelle (approved):

kaxil (approved):

Issue #664, opened 2022-08-16
Add `astro_orders` to example dags


utkarsharma2:
**Please describe the feature you'd like to see**
Right now we have to manually check for the breaking changes in dag listed in https://github.com/astronomer/astro-sdk/blob/main/docs/getting-started/GETTING_STARTED.md

**Describe the solution you'd like**
Add 'astro_orders' das as part of example dags so we don't have to manually check for breaking changes.

**Additional context**
It would be nice if we can refer to the example dag code in the GETTING_STARTED.md file.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass

PR #665, opened 2022-08-16
Add getting-started dag to example dags

utkarsharma2:
# Description
Right now we have to manually check for the breaking changes in dag listed in https://github.com/astronomer/astro-sdk/blob/main/docs/getting-started/GETTING_STARTED.md

closes: #664 

## What is the new behavior?
Add 'astro_orders' das as part of example dags so we don't have to manually check for breaking changes.

Nope

### Checklist
- [ ] Created tests that fail without the change (if possible)

### DIFF
--- a/example_dags/astro_orders.py
+++ b/example_dags/astro_orders.py
@@ -1,0 +1,85 @@
+from datetime import datetime
+
+from airflow.models import DAG
+from pandas import DataFrame
+
+from astro import sql as aql
+from astro.files import File
+from astro.sql.table import Table
+
+S3_FILE_PATH = "s3://tmp9"
+S3_CONN_ID = "aws_default"
+SNOWFLAKE_CONN_ID = "snowflake_conn"
+SNOWFLAKE_ORDERS = "orders_table"
+SNOWFLAKE_FILTERED_ORDERS = "filtered_table"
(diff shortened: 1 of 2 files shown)


utkarsharma2:
@mikeshwe - I was working on adding dag in https://github.com/astronomer/astro-sdk/blob/main/docs/getting-started/GETTING_STARTED.md to example dags so that we can test it as we make changes. But I observed that this is not a self-sufficient dag, and requires a File in S3 and a table in the Snowflake database, For me to add it to integration tests I need to reduce the external dependencies it has. So would it be okay If we load file orders_data_header.csv  from the local path and load table customers_table  using the local file instead of running queries?

Will DAG loses its value as a tutorial if we do the changes suggested above?

utkarsharma2:
@kaxil is there a way for us to refer to a python code from the .md file?

mikeshwe:
> Thanks for pointing out this issue, Utkarsh.

Is there value in having your example DAG load the file orders_data_header.csv from the local path and load table customers_table using the local file, while the GETTING_STARTED.md remain the same? I would like to keep database references in example because they are much more representative of actual usage, AFAIK. Also, I believe that Jake Witz was going to create another version that uses postgres, so in general, we are going to have examples in docs.astronomer.io that won't be in our automated test dags.

Issue #605, opened 2022-08-03
Add an Example DAG / Docs for SQL file support in transform


kaxil:
Currently, we allow users to pass a SQL file to the `sql` param in the transform decorator and `TransformOperator`.

This is handled by the following code:

https://github.com/astronomer/astro-sdk/blob/6b90c92029c7470f5e383345fcdaf77861c8cb94/src/astro/sql/operators/base_decorator.py#L111-L113

We should add an example DAG that uses a SQL file or document it and also mention it in docstring that users can pass a SQL file too.

PR #667, opened 2022-08-16
Update docstring and example DAG for transform operator with SQL file

sunank200:
# Description

Currently, we allow users to pass a SQL file to the`sql` param in the transform decorator and TransformOperator.

closes: #605 

## What is the new behavior?

- Add an example DAG that uses a SQL file
- Document it and mention in docstring that users can pass a SQL file too.

No

### Checklist

### DIFF
--- a/docs/astro/sql/operators/transform.rst
+++ b/docs/astro/sql/operators/transform.rst
@@ -11,7 +11,7 @@ The transform function of the SQL decorator is the ``T`` of the ELT system. Each
 
 The ``transform`` function also treats values in the double brackets as Airflow jinja templates. More details on templating can be found at :ref:`templating`.
 
-There are two use cases of the ``transform`` operator.
+There are three use cases of the ``transform`` operator.
 
 Case 1: When we pass tables between tasks while completing a data transformation.
     The following example applies a SQL ``SELECT`` statement to a ``imdb_movies`` table with templating and saves the result to a ``top_animation`` table.
@@ -45,3 +45,11 @@ Case 2: When we pass a Pandas dataframe while completing a data transformation.
        :language: python
        :start-after: [START transform_example_4]
        :end-before: [END transform_example_4]
+
+Case 3: When we pass SQL file in the transform decorator.
+    The following example DAG shows how we can quickly pass SQL file for the data transformation.
+
+    .. literalinclude:: ../../../../example_dags/example_transform.py
+       :language: python
+       :start-after: [START transform_example_5]
+       :end-before: [END transform_example_5]
--- a/docs/astro/sql/operators/transform.rst
+++ b/docs/astro/sql/operators/transform.rst
@@ -45,11 +45,3 @@ Case 2: When we pass a Pandas dataframe while completing a data transformation.
        :language: python
        :start-after: [START transform_example_4]
        :end-before: [END transform_example_4]
-
-Case 3: When we pass SQL file in the transform decorator.
-    The following example DAG shows how we can quickly pass SQL file for the data transformation.
-
-    .. literalinclude:: ../../../../example_dags/example_transform.py
-       :language: python
-       :start-after: [START transform_example_5]
-       :end-before: [END transform_example_5]
--- a/docs/astro/sql/operators/transform.rst
+++ b/docs/astro/sql/operators/transform.rst
@@ -43,3 +43,5 @@ Case 2: Passing a Pandas dataframe between tasks while completing data transform
        :language: python
        :start-after: [START transform_example_4]
        :end-before: [END transform_example_4]
+
+Please note that in case you want to pass SQL file in the transform decorator, use :ref:`transform_file_operator`
--- a/docs/astro/sql/operators/transform.rst
+++ b/docs/astro/sql/operators/transform.rst
@@ -4,20 +4,18 @@
 transform operator
 ==================
 
-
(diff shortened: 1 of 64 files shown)


docs/astro/sql/operators/transform.rst
```diff
@@ -45,3 +45,11 @@ Case 2: When we pass a Pandas dataframe while completing a data transformation.
        :language: python
        :start-after: [START transform_example_4]
        :end-before: [END transform_example_4]
+
+Case 3: When we pass SQL file in the transform decorator.
```
feluelle, line 49: Wouldn't it make more sense to `return` the file in the function instead of adding another argument to the decorator? What was the reasoning behind that decision?

example_dags/example_transform.py
```diff
@@ -69,6 +71,21 @@ def union_table_and_dataframe(input_table: Table, input_dataframe: pd.DataFrame)
...
 
+# [START transform_example_5]
+@aql.transform(
+    sql=str(pathlib.Path(CWD).parents[0])
+    + "/example_dags/demo_parse_directory/last_ten_animations.sql"
+)  # skipcq: PTC-W0049
+def last_ten_animations(input_table: Table):  # skipcq: PTC-W0049
+    """
+    Pass SQL file ending with .sql
+    """
```
feluelle, line 82: I think this is confusing for users, because the documentation is in the function.
```suggestion
    """
    Run a transformation using a sql file.
    """
```
WDYT?

docs/astro/sql/operators/transform.rst
utkarsharma2: @feluelle Yes, it would. but the code we have suggests we pass it in decorator, I need to confirm this behavior with daniel once.

example_dags/example_transform.py
utkarsharma2: @feluelle makes sense.

dimberman:
@utkarsharma2 I have created a PR for adding the `transform_file` functionality to the astro SDK. Please take a look and finish it up tomorrow so we can merge it in before the release https://github.com/astronomer/astro-sdk/pull/677

@kaxil as discussed, this is a bit of a "quick and dirty" solution, but it should work and we can change how it is formatted in time for 1.0.1 or 1.1.

feluelle:
@sunank200 after @dimberman PR is approved and merged, could you please update the docs in your PR accordingly? 🙏

sunank200:
Sure. Making the changes in the docs now

src/astro/sql/operators/transform.py
```diff
@@ -62,6 +62,14 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
               "exec_date": execution_date
           }
 
+    You can also pass SQL file to the sql parameter in the transform decorator
```
utkarsharma2, line 65: @sunank200 do we want to show this feature here?

utkarsharma2: To me this looks weird, and since we have a different operator for this. Maybe we can remove this?

example_dags/example_transform.py
```diff
@@ -1,3 +1,4 @@
+import pathlib
```
utkarsharma2, line 1: @sunank200 do we need this?

example_dags/example_transform.py
```diff
@@ -9,6 +10,7 @@
 
 START_DATE = datetime(2000, 1, 1)
 LAST_ONE_DF = pd.DataFrame(data={"Title": ["Random movie"], "Rating": [121]})
+CWD = pathlib.Path(__file__).parent
```
utkarsharma2, line 13: @sunank200 Do we need this?

example_dags/demo_parse_directory/last_ten_animations.sql
```diff
@@ -0,0 +1 @@
+SELECT * FROM {{input_table}} WHERE Genre1=='Animation' ORDER BY Rating asc LIMIT 5;
```
utkarsharma2, line 1: @sunank200 Do we need this file?

docs/astro/sql/operators/transform.rst
```diff
@@ -11,7 +11,7 @@ The transform function of the SQL decorator is the ``T`` of the ELT system. Each
 
 The ``transform`` function also treats values in the double brackets as Airflow jinja templates. More details on templating can be found at :ref:`templating`.
 
-There are two use cases of the ``transform`` operator.
+There are three use cases of the ``transform`` operator.
```
utkarsharma2, line 14:
```suggestion
There are two use cases of the ``transform`` operator.
```

feluelle:
Not sure I understand.. why do we need this PR if we don't want users to use `sql` ?

example_dags/example_transform.py
sunank200: Added another example DAG

sunank200:
> Not sure I understand.. why do we need this PR if we don't want users to use `sql` ?

@feluelle this is documentation for https://github.com/astronomer/astro-sdk/pull/677 along with example DAG

feluelle (approved):
LGTM 👍

example_dags/example_transform.py
kaxil: Do we need this?

kaxil (approved):

Issue #603, opened 2022-08-02
Unable to use standard Google conn_id for aql.load_file into BigQuery


TheCodyRich:
**Describe the bug**
When using aql.load_file to load data from GCS to BQ, task fails due to authentication error.  The GCSToBigQueryOperator is able to run successfully using the Google Cloud Platform connection, however the astro sdk does not.  I was able to get the astro sdk task to run successfully by adding `RUN curl -sSL https://sdk.cloud.google.com | bash` to the docker file then manually authenticating in the docker container and running `gcloud auth application-default login` (i.e. not using the conn_id).

**Version**
* Astro SDK 0.11.0
* Astro CLI 1.02
* Astro run-time 5.0.6

**To Reproduce**
1. Create a google cloud connection utilizing the connection type, "Google Cloud Platform" (Figure 1).  Message me for connection details
2. Run attached DAG as is with connection details provided by me, or if using your own connection details update paths, schemas,  and variables to reflect your own data and connections
3. See the astro sdk task fails due to an auth error while the GCSToBigQueryOperator tasks run successfully (Figure 2)

**Expected behavior**
astro sdk should be able to authenticate to Google BigQuery by providing the Google Cloud Platform connection id using the conn_id `google_cloud_default`

**Screenshots**

Figure 1: Google Cloud connection
![image](https://user-images.githubusercontent.com/83995939/182436367-7d643086-b12e-4f2d-8cd3-02a71c65fd58.png)

Figure 2: astro sdk task fails (due to auth error) with GCS BQ operators succeed.
![image](https://user-images.githubusercontent.com/83995939/182443928-5564665f-066b-473a-a06c-0dd87a5f2e22.png)

**Additional context**
[Full logs from the astro sdk task (gcs_to_bq_sdk)](https://github.com/astronomer/astro-sdk/files/9244653/Task_Instance__gcs_to_bq_sdk.log)

[test DAG](https://github.com/astronomer/astro-sdk/files/9244666/astro_sdk_bq_test.py.zip)

feluelle:
Hey @TheCodyRich,

I have managed to fix your issue in #668, I believe. At least I have tested it with your connection details and it worked.

To test it with your astro stack, do the following:

Edit your `requirements.txt` to install from my branch:
```txt
https://github.com/astronomer/astro-sdk/archive/feature/enable-google-cloud-service-account-authentication.zip
```

[...]

PR #668, opened 2022-08-16
Use Service Account authentication for loading from pandas to bigquery

feluelle:
# Description

Currently we are using the default authentication method for loading from pandas to bigquery. This has the disadvantage of having to manually authenticating through a browser.

closes: #603 

## What is the new behavior?

This PR proposes a change which makes the loading functionality use the service account authentication which will work without having to open a browser window. It will automatically authenticate with the information provided in the airflow connection.

No

### Checklist

### DIFF
--- a/src/astro/databases/google/bigquery.py
+++ b/src/astro/databases/google/bigquery.py
@@ -161,6 +161,7 @@ class BigqueryDatabase(BaseDatabase):
             if_exists=if_exists,
             chunksize=chunk_size,
             project_id=self.hook.project_id,
+            credentials=self.hook._get_credentials(),
         )
 
     def merge_table(
--- a/tests/databases/test_bigquery.py
+++ b/tests/databases/test_bigquery.py
@@ -478,3 +478,31 @@ def test_get_run_id():
         S3ToBigqueryDataTransfer.get_run_id(config)
         == "62d6a4df-0000-2fad-8752-d4f547e68ef4"
     )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "database_table_fixture",
+    [
+        {
+            "database": Database.BIGQUERY,
+            "table": Table(metadata=Metadata(schema=SCHEMA)),
+        },
+    ],
+    indirect=True,
+    ids=["bigquery"],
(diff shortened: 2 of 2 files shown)


src/astro/databases/google/bigquery.py
```diff
@@ -140,6 +141,23 @@ def get_merge_initialization_query(parameters: Tuple) -> str:
...
+                conn_extra["extra__google_cloud_platform__keyfile_dict"]
+            )
+
+        if "extra__google_cloud_platform__key_path" in conn_extra:
+            return Credentials.from_service_account_file(
+                conn_extra["extra__google_cloud_platform__key_path"]
+            )
+
+        # Falling back to Default Authentication Methods,
+        # see https://pandas-gbq.readthedocs.io/en/latest/howto/authentication.html for details.
```
kaxil, line 158: Can we not replace this with the following?:

```python
self.hook._get_credentials(),
```

feluelle: Yes, we can! Good catch! 🥇 

> I know it is an internal method but wdyt?

🤔 🤞 that it won't change. As it is a _private_ function it can be changed even in patch versions without a changelog note. Are you a code owner for this file in the airflow project? We somehow need to subscribe to changes so we can react quickly 😅

feluelle: Actually.. It already has a lot of references in other hooks:
Screenshot 2022-08-18 at 07 22 31

I think it should be public actually. WDYT? I can open a PR in airflow core.

feluelle: I have changed my code to use the existing `_get_credentials` function and have rewritten the tests.

utkarsharma2 (approved):

src/astro/databases/google/bigquery.py
feluelle: Btw, I have created a PR [here](https://github.com/apache/airflow/pull/25785).

kaxil: Awesome

kaxil (approved):

Issue #658, opened 2022-08-15
Implement `get_value_list` (Dynamic Task Templates)


kaxil:
This builds on top of [Dynamic Tasks Mapping](https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html) released in Airflow 2.3 and is similar to https://github.com/astronomer/astro-sdk/issues/507

```python
get_value_list(‘select product from products’, db_conn) 
```

Iterating over a set of files matching a pattern in a file system:
- Query a table in a database to get a list of values.
- And then, to iterate over a set of tasks for each value in this list.
- The number of values in this list is unknown at the time of DAG creation, but it is not uncommon for this list to be hundreds of values.

The `get_value_list` operation would abstract away the underlying Operator setup and boilerplate code, similar to the existing `load_file` operation in the Astro Python SDK. 

This would enable the DAG author to get a list of products from the database, accessed using the specified connection. In the spirit of the existing DAG authoring framework, the underlying API takes care of all the boilerplate for getting the right database Operator, setting it up, executing the SQL, and getting the data back into a list suitable for the ensuing “map” operation.

PR #673, opened 2022-08-17
Add `get_value_list` for Dynamic Task 

pankajastro:
# Description
Add operator RunQueryOperator which takes a SQL stmt as input and returns its result

closes: https://github.com/astronomer/astro-sdk/issues/658


src/astro/sql/__init__.py
```diff
@@ -1,10 +1,25 @@
...
+
+
+def get_value_list(sql_statement: str, conn_id: str, **kwargs) -> XComArg:
+    """
+    Execute a sql statement and return the result
+
+    :param sql_statement: sql query to execute
+    :param conn_id: Airflow connection id
+    """
+    return LoadTableOperator(
```
feluelle, line 23: `LoadTableOperator` seems confusing to me. It is not loading it is exporting, no?

docs/astro/sql/operators/get_value_list.rst
```diff
@@ -0,0 +1,14 @@
+.. _get_value_list:
+
+==============
+get_value_list
+==============
+
+When to use the ``get_value_list`` operator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We can use ``get_value_list`` when you want to execute a SQL query on a database table get the result. This might useful when you want to create dynamic task on the response of this operator.
```
feluelle, line 9:
```suggestion
We can use ``get_value_list`` when you want to execute a SQL query on a database table and get the result. This might be useful when you want to create dynamic tasks on the response of this operator.
```

src/astro/sql/__init__.py
pankajastro: hmm, It actually execute a SQL query and return result so what about `RunQueryOperator`?

feluelle: Or `FetchQueryOperator` ? :)

utkarsharma2: Suppose it is an executing query and returning results. Can we name it executeQuery? because there is no restriction to what query the user can execute.

src/astro/sql/operators/load_table.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    template_fields = ("sql_statement", "conn_id")
+
+    def __init__(
+        self,
+        sql_statement: str,
+        conn_id: str,
+        handler: Optional[Callable] = None,
+        task_id: str = "",
```
utkarsharma2, line 26: We don't need to define task_id explicitly, do we?

src/astro/sql/operators/load_table.py
```diff
@@ -0,0 +1,44 @@
...
+        **kwargs
+    ):
+        task_id = task_id or get_unique_task_id(
```
utkarsharma2, line 29: We can extract it from kwargs here.

src/astro/sql/__init__.py
sunank200: I like `RunQueryOperator`.

src/astro/sql/operators/load_table.py
```diff
@@ -0,0 +1,44 @@
...
+from airflow.decorators.base import get_unique_task_id
+from airflow.models import BaseOperator
+from airflow.utils.context import Context
+
+from astro.databases import create_database
+
+
+class LoadTableOperator(BaseOperator):
+    """
+    Load the result of a SQL statement
```
sunank200, line 12: We should mention what it returns. WDYT?

src/astro/sql/operators/load_table.py
```diff
@@ -0,0 +1,44 @@
...
+            "get_value_list", dag=kwargs.get("dag"), task_group=kwargs.get("task_group")
+        )
+        self.sql_statement = sql_statement
+        self.conn_id = conn_id
+        self.handler = handler
+        super().__init__(task_id=task_id, **kwargs)
+
+    def execute(self, context: Context) -> Any:
+        db = create_database(self.conn_id)
+        result_set = db.run_sql(self.sql_statement)
```
utkarsharma2, line 39: @pankajastro can you try this with apache-airflow-providers-snowflake=3.2.0? We faced a [`bug`](https://github.com/astronomer/astro-sdk/pull/686) yesterday. Just wanted to make sure this works with it.

tests/sql/operators/test_load_table.py
```diff
@@ -0,0 +1,53 @@
...
+    op = LoadTableOperator(
+        task_id="task1", sql_statement="select * from 1", conn_id="conn"
+    )
+    op.execute(None)
+
+
+@patch("astro.databases.google.bigquery.BigqueryDatabase.sqlalchemy_engine")
+@patch("airflow.hooks.base.BaseHook.get_connection")
+def test_load_table_bigquery(mock_get_conn, mock_sqlalchemy_engine):
+    mock_get_conn.return_value = Connection(conn_id="conn", conn_type="gcpbigquery")
```
sunank200, line 36: Whats this test doing? Its not asserting anything.

tests/sql/operators/test_load_table.py
```diff
@@ -0,0 +1,53 @@
...
+        task_id="task1", sql_statement="select * from 1", conn_id="conn"
+    )
+    op.execute(None)
+    mock_sqlalchemy_engine.assert_called_once_with("sqlite:///localhost")
+
+
+@patch("astro.databases.base.BaseDatabase.sqlalchemy_engine")
+@patch("airflow.hooks.base.BaseHook.get_connection")
+def test_load_table_postgres(mock_get_conn, sqlalchemy_engine):
+    mock_get_conn.return_value = Connection(
```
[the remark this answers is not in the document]
pankajastro: hmm fixed it

tests/sql/operators/test_load_table.py
pankajastro: fixed it

src/astro/sql/operators/load_table.py
pankajastro: Rephrase it please let me know WDYT?

src/astro/sql/__init__.py
pankajastro: `executeQuery` and `RunQueryOperator` look me more close to its behaviour let me know WDYT

utkarsharma2: @pankajastro I'm okay with `RunQueryOperator`

src/astro/sql/operators/load_table.py
pankajastro: tested it worked.

tests/sql/test_table.py
```diff
@@ -43,3 +48,17 @@ def test_table_name_with_temp_prefix():
     """Check that the table is no longer considered temp when the name is set after initialization."""
     table = Table(conn_id="some_connection")
     assert table.name.startswith("_tmp_")
+
+
+def test_get_value_list():
+    """Assert that get_file_list handle kwargs correctly"""
+    dag = DAG(dag_id="dag1", start_date=datetime(2022, 1, 1))
+
+    resp = get_value_list(sql_statement="path", conn_id="conn", dag=dag)
```
utkarsharma2, line 57: @pankajastro Why is the `path` added as `sql_statement` here?

utkarsharma2 (approved):
@pankajastro LGTM

tests/sql/test_table.py
pankajastro, line 57:
```suggestion
    resp = get_value_list(sql_statement="select * from 1", conn_id="conn", dag=dag)
```

python-sdk/src/astro/sql/operators/run_query.py
```diff
@@ -0,0 +1,43 @@
+from typing import Any, Callable, Optional
+
+from airflow.decorators.base import get_unique_task_id
+from airflow.models import BaseOperator
+from airflow.utils.context import Context
+
+from astro.databases import create_database
+
+
+class RunQueryOperator(BaseOperator):
```
kaxil, line 10: @tatiana @dimberman  What do you think about the naming?

python-sdk/src/astro/sql/operators/run_query.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Run a SQL statement and return the result of execution
+
+    :param sql_statement: sql query to execute
+    :param conn_id: Airflow connection id
```
kaxil, line 15: Can you expand what this connection would be used for?

python-sdk/src/astro/sql/operators/run_query.py
```diff
@@ -0,0 +1,43 @@
...
+
+    template_fields = ("sql_statement", "conn_id")
+
+    def __init__(
+        self,
+        sql_statement: str,
+        conn_id: str,
+        handler: Optional[Callable] = None,
+        **kwargs
+    ):
```
kaxil, line 27: Should we have a limit on number of records? Or do we expect users to pass it "limit" in SQL.

Regardless of what we expect, we should document what might happen if they try to pull 1000s records as an extreme example

python-sdk/src/astro/sql/__init__.py
```diff
@@ -6,5 +8,18 @@
...
 from astro.sql.operators.transform import TransformOperator, transform, transform_file
 from astro.sql.table import Metadata, Table
+
+
+def get_value_list(sql_statement: str, conn_id: str, **kwargs) -> XComArg:
+    """
+    Execute a sql statement and return the result
+
+    :param sql_statement: sql query to execute
+    :param conn_id: Airflow connection id
```
kaxil, line 21: Same as https://github.com/astronomer/astro-sdk/pull/673/files#r962154439

python-sdk/example_dags/example_dynamic_task_template.py
```diff
@@ -39,4 +44,15 @@
     ).expand(input_file=get_file_list(path=GCS_BUCKET, conn_id=ASTRO_GCP_CONN_ID))
 
     # [END howto_operator_get_file_list]
+
+    # [START howto_operator_get_value_list]
+    @task
+    def custom_task(row):
+        print(row)
```
kaxil, line 51: Do we have a better example, than just printing it?

python-sdk/src/astro/sql/__init__.py
pankajastro: Improved this param description please let me know it looks now

python-sdk/src/astro/sql/operators/run_query.py
pankajastro: Improved this param description please let me know it looks now

python-sdk/src/astro/sql/operators/run_query.py
pankajastro: I was assuming that the user would pass the limit in the SQL query string for now I have improved the docstring with possible issues if they try to fetch huge number of rows.

I'm ok with adding limit param in this function let me know what you think.

python-sdk/example_dags/example_dynamic_task_template.py
pankajastro: Doing the sum of a table's column if the value is valid float.

python-sdk/src/astro/sql/operators/run_query.py
pankajastro: putting the previous discussion about this naming here for reference https://github.com/astronomer/astro-sdk/pull/673#discussion_r949811361

python-sdk/example_dags/example_dynamic_task_template.py
```diff
@@ -39,4 +44,29 @@
...
+    # [START howto_operator_get_value_list]
+    @task
+    def custom_task(rating_val):
+        try:
+            print("rating_val ", rating_val)
+            print("rating_val.__dict__ ", rating_val.__dict__)
+            print("rating_val.__dict__.get()", rating_val.__dict__.get("rating"))
+            return float(rating_val.__dict__.get("rating"))
+        except ValueError:
+            pass
```
tatiana, line 57: Could we simplify this block to become a single line? It feels it is more complex than needed and may distract from the focus of this example (illustrating how to use `get_file_list`.

python-sdk/src/astro/sql/operators/run_query.py
tatiana: @pankajastro @kaxil how does this operator relate to the RunRawSQL? It feels both are very similar. Do we want to keep both of them? Do we want to deprecate one of them?

python-sdk/src/astro/sql/operators/run_query.py
tatiana: @pankajastro what if we raised a warning if the user is not using a custom backend and attempts to run a query without limit?

Something along the lines of what we did for the `DataframeOperator` to avoid the user thrashing the Airflow metadata database:

https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/dataframe.py#L160-L163
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/exceptions.py#L5-L15

pankajastro: There are two problems that customer face when user try to fetch lots of data
- XCOM will be overloaded 
- since the main purpose of this function is to use it in the dynamic task map expand method so it can create a huge number of parallel tasks.

So, would be in favour of adding a warning if rows will be more than a certain number precisely 1024 since dynamic task map by default allow to expand these many task

python-sdk/src/astro/sql/operators/run_query.py
pankajastro: @tatiana Mainly ff we don't pass the handler in `RawSQLOperator` then it returns `None` but RunQueryOperator returns fetchall response.

pankajastro: @kaxil any thoughts on the above i.e using `RawSQLOperator` after some modifications?

python-sdk/example_dags/example_dynamic_task_template.py
pankajastro: We want to do some operations with on the result of expanded task so we need this.

python-sdk/src/astro/sql/operators/run_query.py
tatiana: @pankajastro and I agreed that we'll be using the `RawSQLOperator` to implement `get_value_list`, defining the handler

python-sdk/src/astro/sql/operators/run_query.py
tatiana: @pankajastro and I agreed we'll be introducing a parameter to limit the results outputted from `RawSQLOperator`, which also makes sense for the run_raw_sql decorator - and the `get_value_list` will be setting it to the same limit of Airflow's task template (which seems to be by default 1024, but he'll make sure we're using the same constant/config as Airflow)

feluelle: Makes sense to me 👍 Let's make sure we document it accordingly.

pankajastro: After making the changes tests are failing since `RawSQLOperator` is inherited from `DecoratedOperator` which expects `python_callable`. @tatiana  any thought how I can resolve this?

PR #676, opened 2022-08-17
Minimize Docker Image Rebuilding

pankajastro:
Copy only required files in the docker context to avoid rebuilding the image

This commit also:
- adds `.dockerignore`
- updated Dockerfile to use cache for development. Will only burst cache when `.pyproject.toml` changes
- adds 3 more make commands to build docs, restart some containers (that needs to reload airflow), and restart all containers
- adds missing deps for building docs
- remove the deprecated env `AIRFLOW__CORE__SQL_ALCHEMY_CONN` and use `AIRFLOW__DATABASE__SQL_ALCHEMY_CONN`


dev/Dockerfile
```diff
@@ -9,7 +9,9 @@ RUN apt-get install -y --no-install-recommends \
         libsasl2-modules
 ENV SETUPTOOLS_USE_DISTUTILS=stdlib
 
-COPY ../  ${AIRFLOW_HOME}/astro_sdk/
+COPY ../src  ${AIRFLOW_HOME}/astro_sdk/
+COPY ../pyproject.toml  ${AIRFLOW_HOME}/astro_sdk/
+COPY ../README.md  ${AIRFLOW_HOME}/astro_sdk/
```
pankajastro, line 14: without copying README getting an error. Most likely because pyproject.toml refer it.

```
 50.69   Traceback (most recent call last):
#0 50.69     File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
#0 50.69       main()
#0 50.69     File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
#0 50.69       json_out['return_val'] = hook(**hook_input['kwargs'])
#0 50.69     File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 144, in get_requires_for_build_editable
#0 50.69       return hook(config_settings)
#0 50.69     File "/tmp/pip-build-env-h_uqtbzk/overlay/lib/python3.9/site-packages/flit_core/buildapi.py", line 23, in get_requires_for_build_wheel
#0 50.69       info = read_flit_config(pyproj_toml)
#0 50.69     File "/tmp/pip-build-env-h_uqtbzk/overlay/lib/python3.9/site-packages/flit_core/config.py", line 70, in read_flit_config
#0 50.69       return prep_toml_config(d, path)
#0 50.69     File "/tmp/pip-build-env-h_uqtbzk/overlay/lib/python3.9/site-packages/flit_core/config.py", line 97, in prep_toml_config
#0 50.69       loaded_cfg = read_pep621_metadata(d['project'], path)
#0 50.69     File "/tmp/pip-build-env-h_uqtbzk/overlay/lib/python3.9/site-packages/flit_core/config.py", line 445, in read_pep621_metadata
#0 50.69       desc_content, mimetype = description_from_file(readme, path.parent)
#0 50.69     File "/tmp/pip-build-env-h_uqtbzk/overlay/lib/python3.9/site-packages/flit_core/config.py", line 277, in description_from_file
#0 50.69       raise ConfigError(
#0 50.69   flit_core.config.ConfigError: Description file README.md does not exist
```

dev/Dockerfile
```diff
@@ -9,7 +9,9 @@ RUN apt-get install -y --no-install-recommends \
         libsasl2-modules
 ENV SETUPTOOLS_USE_DISTUTILS=stdlib
 
-COPY ../  ${AIRFLOW_HOME}/astro_sdk/
+COPY ../src  ${AIRFLOW_HOME}/astro_sdk/
```
kaxil, line 12: This will still mean that you need to rebuild the entire image when we make any changes in source code as all the source code is in `src`

kaxil: Check how we do it in astronomer-providers - https://github.com/astronomer/astronomer-providers/blob/main/dev/Dockerfile

kaxil:
> I think it is better to add the files you dont want to have in the image to `.dockerignore`.

Yeah, like Felix mentioned we should have something like https://github.com/astronomer/astronomer-providers/blob/main/dev/.dockerignore

dev/Dockerfile
kaxil: https://github.com/astronomer/astro-sdk/pull/676/commits/af32a9da73a92a26f34e5ff8966f4d8e8537fdce should fix it :)

feluelle: With `.dockerignore` you can revert the change to the `COPY` command no? `COPY ../` is fine if we keep our `.dockerignore` up to date as we do with `.gitignore`.

pankajastro: right, we can do that.
Also, I can see we a `.dockerignore` at the root level so it was ignoring the newly added `dev/.dockerignore`. I have deleted the   `dev/.dockerignore` and updated  `.dockerignore` 3cb6012b76ee69742921d1318a28da2b7d864ef2

kaxil: The problem is the "src" folder that contains the entire source code, if we don't add it to .gitignore or specific files in Dockerfile, everytime I change something in `src/astro/database` for example, it will invalidate the cache and build it from scratch @pankajastro

kaxil: the following line is the problem: 

https://github.com/astronomer/astro-sdk/blob/3cb6012b76ee69742921d1318a28da2b7d864ef2/dev/Dockerfile#L12

We only want to install the dependencies in the docker-file, and then mount the sources via docker-compose file

kaxil: Validate this by changing a comment in https://github.com/astronomer/astro-sdk/blob/3cb6012b76ee69742921d1318a28da2b7d864ef2/src/astro/files/__init__.py and run `make container target=build-run` command again

pankajastro: Yes, I agree @kaxil. We can either add `src` also in the `.dockerignore` or we will have to copy only `README.md` and  `__init__.py` which are required by `pyproject.toml` I think the 2nd option is more clean and easy so I have reverted the `Dockerfile` changes here 736149e623f5fb7126cb85e4088b9f32d3dece5a

kaxil (approved):

PR #677, opened 2022-08-17
Create a transform_file operator

dimberman:
# Description
Currently there is no clean way to run `aql.transform` with a SQL file instead of a python function. This is highly limiting as SQL developers would not want to craft complex queries inside of a python function.

## What is the new behavior?

The new behavior is that we will now have an `aql.transform_file` function that can be called at runtime in the DAG. This is not a decorator, but is instead a python function that returns a task (similar to operators like `drop` and `merge`).

Instead of using function parameters, users would instead insert tables via the `parameters` kwarg.

It would look like this

```python
import pathlib

@aql.dataframe
def validate(df: pandas.DataFrame):
    assert df.columns.tolist() == ["title", "rating"]

with dag:
    target_table = Table(name="test_is_{{ ds_nodash }}", conn_id="sqlite_default")
    table_from_query = aql.transform_file(
        file_path=str(pathlib.Path(CWD).parents[0]) + "/transform/test.sql",
        parameters={"input_table": imdb_table, "output_table": target_table},
    )
    validate(table_from_query)
```

This function also offers a nice "less magic" endpoint for astro-sdk users who might not yet be comfortable with pytho decorators.

No

### Checklist

### DIFF
--- a/src/astro/sql/operators/transform.py
+++ b/src/astro/sql/operators/transform.py
@@ -107,7 +107,7 @@ def transform_file(
     database: Optional[str] = None,
     schema: Optional[str] = None,
     **kwargs: Any,
-) -> "XComArg":
+) -> XComArg:
     """
     :param file_path: File path for the SQL file you would like to parse. Can be an absolute path, or you can use a
         relative path if the `template_searchpath` variable is set in your DAG
--- a/src/astro/sql/operators/transform.py
+++ b/src/astro/sql/operators/transform.py
@@ -103,9 +103,9 @@ def transform(
 def transform_file(
     file_path: str,
     conn_id: str = "",
-    parameters: Optional[Union[Mapping, Iterable]] = None,
-    database: Optional[str] = None,
-    schema: Optional[str] = None,
+    parameters: Mapping | Iterable | None = None,
+    database: str | None = None,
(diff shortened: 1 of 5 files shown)


utkarsharma2:
@kaxil - me and @sunank200 discussed on this operator, can you please take a look at it.

feluelle (approved):
Looks so much cleaner and easier to understand for users! 👍  🚀

kaxil (approved):
@sunank200 you are working on adding docs for this I assume

src/astro/sql/operators/transform.py
```diff
@@ -96,3 +98,41 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
...
+
+def transform_file(
+    file_path: str,
+    conn_id: str = "",
+    parameters: Mapping | Iterable | None = None,
+    database: str | None = None,
+    schema: str | None = None,
+    **kwargs: Any,
+) -> XComArg:
+    """
```
kaxil, line 111: @sunank200 Can you add a 1-line description here in your PR too about what this function is doing

sunank200: sure

Issue #680, opened 2022-08-18
Move dag example in README.md in example_dags


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we have examples in the README.md file which are not getting tested as we update the code. 

**Describe the solution you'd like**
We can move our code to example_dags and refer to that file in README.md so that this file is tested with every change.

**Acceptance Criteria**
- [] All checks and tests in the CI should pass
- [] Example DAG
- [] Improve the documentation (README, Sphinx, and any other relevant)

PR #681, opened 2022-08-18
Add the readme.md example dag to example dags folder

utkarsharma2:
# Description
Currently, we have examples in the README.md file which are not getting tested as we update the code. 

closes: #680

## What is the new behavior?
We can move our code to example_dags and refer to that file in README.md so that this file is tested with every change.

Nope

### Checklist
- [X] All checks and tests in the CI should pass
- [X] Example DAG
- [X] Improve the documentation (README, Sphinx, and any other relevant)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,21 +61,26 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
 
 3. Copy the following workflow into a file named `calculate_popular_movies.py` and add it to the `dags` directory of your Airflow project:
 
-    https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L1-L37
+   https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L1-L37
 
-2. Ensure that your Airflow environment is set up correctly by running the following commands:
+   Alternatively you can download the file
+   ```shell
+    curl -O https://raw.githubusercontent.com/astronomer/astro-sdk/main/example_dags/calculate_popular_movies.py
+   ```
+
+4. Ensure that your Airflow environment is set up correctly by running the following commands:
 
         top_animations = Table(name="top_animation")
         top_animations = top_five_animations(input_table=imdb_movies, output_table=top_animations)
     ```
 
-4. Run the example DAG:
+5. Run the example DAG:
 
     ```sh
     airflow dags test calculate_popular_movies `date -Iseconds`
     ```
 
-5. Check the result of your DAG by running:
+6. Check the result of your DAG by running:
 
     ```shell
     sqlite3 "$SQL_TABLE_NAME" "select * from top_animation;" ".exit"
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
 
 4. Ensure that your Airflow environment is set up correctly by running the following commands:
 
-5. ```python
+   ```python
         top_animations = Table(name="top_animation")
         top_animations = top_five_animations(input_table=imdb_movies, output_table=top_animations)
     ```
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
 
    https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L1-L37
 
-   Alternatively you can download the file
+   Alternatively, you can download `calculate_popular_movies.py`
    ```shell
     curl -O https://raw.githubusercontent.com/astronomer/astro-sdk/main/example_dags/calculate_popular_movies.py
    ```
--- a/README.md
+++ b/README.md
@@ -61,32 +61,9 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
 
 3. Copy the following workflow into a file named `calculate_popular_movies.py` and add it to the `dags` directory of your Airflow project:
 
-    ```Python
-    from datetime import datetime
-    from airflow import DAG
-    from astro import sql as aql
-    from astro.files import File
-    from astro.sql.table import Table
+    https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L1-L37
 
-    @aql.transform()
-    def top_five_animations(input_table: Table):
-        return """
-            SELECT title, rating
-            FROM {{input_table}}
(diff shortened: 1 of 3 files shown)


example_dags/calculate_popular_movies.py
```diff
@@ -0,0 +1,37 @@
+from datetime import datetime
```
sunank200, line 1: Isn't this DAG just a subset of https://github.com/astronomer/astro-sdk/blob/main/example_dags/example_transform.py ?

We could just use that

sunank200 (commented):
Isn't this DAG just a subset of https://github.com/astronomer/astro-sdk/blob/main/example_dags/example_transform.py ?

We could just use that. WDYT?

feluelle (commented):
For me it is not showing the whole code at once. I would have to scroll down to see everything. :/

Screenshot 2022-08-18 at 11 57 37

utkarsharma2:
@feluelle I didn't see any way to change the size of the code scnipet.

example_dags/calculate_popular_movies.py
utkarsharma2: @sunank200 There was some other code added to the example_transform.py and I wasn't sure if we can use that in README.md. I think we can take @mikeshwe opinion on this.

README.md
```diff
@@ -50,42 +50,7 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
...
-            ),
-        )
-        top_five_animations(
-            input_table=imdb_movies,
-            output_table=Table(
-                name="top_animation"
-            ),
-        )
-    ```
+    https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L1-L37
```
kaxil, line 53: WDYT about the following instead, thoughts? cc @kentdanas 

```suggestion
    curl -O https://raw.githubusercontent.com/astronomer/astro-sdk/main/example_dags/calculate_popular_movies.py
```

README.md
```diff
@@ -50,42 +50,7 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
 
 1. Copy the following DAG into a file named `calculate_popular_movies.py` and add it to the `dags` directory of your Airflow project:
```
jwitz, line 51: I'd actually prefer if we kept the DAG code visible in the README. Even though it's longer, the reader won't have to click into another page/ run a command before they see how the SDK works. We can still keep the other DAG file, but this would serve as the window into what the SDK can do

README.md
```diff
@@ -50,42 +50,7 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
...
 
-    ```Python
-    from datetime import datetime
-    from airflow import DAG
-    from astro import sql as aql
-    from astro.files import File
-    from astro.sql.table import Table
-
-    @aql.transform()
```
jwitz, line 60: If we add this back to the README as code, it would be great to add some more inline comments about what the code/ SDK is doing here.

README.md
```diff
@@ -50,42 +50,7 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
...
-            ),
-        )
-        top_five_animations(
-            input_table=imdb_movies,
-            output_table=Table(
-                name="top_animation"
-            ),
-        )
-    ```
+    curl -O https://raw.githubusercontent.com/astronomer/astro-sdk/main/example_dags/calculate_popular_movies.py
```
tatiana, line 53: @utkarsharma2 why did you favour this approach instead of adding the [code snipped with permalink](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet#linking-to-code)? Is it because the permalink would have the version of the code hard-coded?

README.md
tatiana: If we could embed the code it would be the best scenario from both perspectives

README.md
feluelle: I would use the permalink, otherwise we could end up having a 404 (completely missing) in our README.

utkarsharma2: @tatiana initially permalinks was my approach - https://github.com/astronomer/astro-sdk/pull/681/commits/21f7bd8edaa7592098447969a7d8d84f5e849778#r949124948 

But I guess @kaxil felt the alternative would be better. I personally incline towards permalinks approach.

kaxil: I felt that mainly because users would otherwise have to copy/paste it and because of scrolling involved with code-snippet that might result in issues.

However, if all of you think snippet + permalink is a better approach then go for it.

tatiana: @utkarsharma2 What if we kept both of them? The code snipped & the curl command - so we would meet the requirement from the doc team while also simplifying for those who may want to download it straightway - without having to have the code hard-coded in the README?

utkarsharma2: @tatiana I think we can keep both if we are able to verify the curl link is working. Just a little worried if move the file somewhere else and the links are broken.

utkarsharma2: @tatiana Never mind - CI job takes cares of that - gaurav-nelson/github-action-markdown-link-check@v1

utkarsharma2: Updated: 

Screenshot 2022-09-07 at 2 34 48 AM

README.md
```diff
@@ -61,44 +61,26 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
...
-        schedule_interval=None,
-        start_date=datetime(2000, 1, 1),
-        catchup=False,
-    ) as dag:
-        imdb_src = File("https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb_v2.csv")
-        imdb_movies = Table(name="imdb_movies", conn_id="sqlite_default")
-        imdb_movies = aql.load_file(imdb_src, imdb_movies)
+   https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L1-L37
+
+   Alternatively you can download the file
```
tatiana, line 66:
```suggestion
   Alternatively, you can download the file by running:
```

README.md
```diff
@@ -61,44 +61,26 @@ pip install astro-sdk-python[amazon,google,snowflake,postgres]
...
+   ```shell
+    curl -O https://raw.githubusercontent.com/astronomer/astro-sdk/main/example_dags/calculate_popular_movies.py
+   ```
+
+4. Ensure that your Airflow environment is set up correctly by running the following commands:
 
         top_animations = Table(name="top_animation")
         top_animations = top_five_animations(input_table=imdb_movies, output_table=top_animations)
     ```
```
tatiana, line 75: I didn't fully understand this part. Do we need it?

utkarsharma2: @tatiana yes, we do, was a typo fixed it.

tatiana (approved):

README.md
feluelle: > Never mind - CI job takes care of that - gaurav-nelson/github-action-markdown-link-check@v1. I'm going ahead with adding both.

Right. That's great! +1 for using `main` version.

PR #683, opened 2022-08-18
Revert "Use lazy evaluated Type Annotations from PEP 563"

utkarsharma2:
Reverts astronomer/astro-sdk#650


utkarsharma2:
#563 is not causing the CI to fail. Closing this revert PR.

PR #686, opened 2022-08-18
Handle breaking changes for Snowflake provide version 3.2.0 and 3.1.0

utkarsharma2:
# Description
CI failing 
https://github.com/astronomer/astro-sdk/runs/7897173351?check_suite_focus=true
https://github.com/astronomer/astro-sdk/runs/7899645838?check_suite_focus=true

closes: #689

## What is the new behavior?
To handle breaking change in apache-airflow-providers-snowflake==3.2.0, we need to pass the handler param to get the rows. But in version apache-airflow-providers-snowflake==3.1.0 if we pass the handler, the provider raises an exception AttributeError 'sfid'. We added the Try Catch statement to handle the issue. 

Nope

### Checklist
- [X] Created tests which fail without the change

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ google = [
     "smart-open[gcs]>=5.2.1"
 ]
 snowflake = [
-    "apache-airflow-providers-snowflake==3.2.0",
+    "apache-airflow-providers-snowflake",
     "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
     "snowflake-connector-python[pandas]",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ google = [
     "smart-open[gcs]>=5.2.1"
 ]
 snowflake = [
-    "apache-airflow-providers-snowflake",
+    "apache-airflow-providers-snowflake==3.2.0",
     "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
     "snowflake-connector-python[pandas]",
 ]
--- a/src/astro/databases/snowflake.py
+++ b/src/astro/databases/snowflake.py
@@ -506,14 +506,23 @@ class SnowflakeDatabase(BaseDatabase):
 
         :param schema: DB Schema - a namespace that contains named objects like (tables, functions, etc)
         """
-        created_schemas = [
-            x["SCHEMA_NAME"]
-            for x in self.hook.run(
+
+        # Below code is added due to breaking change in apache-airflow-providers-snowflake==3.2.0,
+        # we need to pass handler param to get the rows. But in version apache-airflow-providers-snowflake==3.1.0
+        # if we pass the handler provider raises an exception AttributeError 'sfid'.
+        try:
+            schemas = self.hook.run(
                 "SELECT SCHEMA_NAME from information_schema.schemata WHERE LOWER(SCHEMA_NAME) = %(schema_name)s;",
                 parameters={"schema_name": schema.lower()},
-                handler=lambda x: x,
+                handler=lambda cur: cur.fetchall(),
             )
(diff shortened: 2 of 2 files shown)


src/astro/databases/snowflake.py
```diff
@@ -506,13 +506,23 @@ def schema_exists(self, schema: str) -> bool:
...
                 "SELECT SCHEMA_NAME from information_schema.schemata WHERE LOWER(SCHEMA_NAME) = %(schema_name)s;",
                 parameters={"schema_name": schema.lower()},
+                handler=lambda cur: cur.fetchall(),
             )
-        ]
+        except AttributeError:
+            schemas = self.hook.run(
+                "SELECT SCHEMA_NAME from information_schema.schemata WHERE LOWER(SCHEMA_NAME) = %(schema_name)s;",
+                parameters={"schema_name": schema.lower()},
+            )
```
feluelle, line 523: Why do we still need this if we upgrade to `3.2.0` ?

src/astro/databases/snowflake.py
```diff
@@ -506,13 +506,23 @@ def schema_exists(self, schema: str) -> bool:
 
         :param schema: DB Schema - a namespace that contains named objects like (tables, functions, etc)
         """
-        created_schemas = [
-            x["SCHEMA_NAME"]
-            for x in self.hook.run(
+
+        # Below code is added due to breaking change in apache-airflow-providers-snowflake==3.2.0,
+        # we need to pass handler param to get the rows. But in version apache-airflow-providers-snowflake==3.1.0
+        # if we pass the handler provider raises an exception AttributeError 'sfid'.
```
feluelle, line 512: I think this should be rather in the PR or commit message. But I like that you commented it in general :)

src/astro/databases/snowflake.py
utkarsharma2: @feluelle I'm not sure about this part, but I was trying to avoid placing a lower bound on snowflake dependency. To be more backward compatible.

utkarsharma2: @kaxil Thoughts?

utkarsharma2: If we don't wanna support this becomes a lot easier as well :)

feluelle: I would say we should not support prior version (3.1.0).

However if you think or kaxil that it makes sense use `>=3.1.0,<=3.20` or similar to allow installing 3.1.0 if necessary.

pyproject.toml
```diff
@@ -52,7 +52,7 @@ google = [
     "smart-open[gcs]>=5.2.1"
 ]
 snowflake = [
-    "apache-airflow-providers-snowflake",
+    "apache-airflow-providers-snowflake==3.2.0",
```
pankajastro, line 55: I prefer to keep `>=` instead of `==`

feluelle: Fine for me.

feluelle: The question is do we want to still allow `>=3.1.0` or `>=3.2.0`.

utkarsharma2: My bad, we don't really need this pin. Removing it.

src/astro/databases/snowflake.py
kaxil: Yeah, I would prefer to have backwards compatibility as users wouldn't update to 3.2.0 as it was just released a day or two ago.

kaxil (approved):

src/astro/databases/snowflake.py
utkarsharma2: @feluelle 3.2.0 is released recently and not many people will be having the newer version. So I think we should be supporting 3.1.0 as well.

Issue #690, opened 2022-08-18
Expose `delimiter` parameter for file object for S3/GCS pattern matching 


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we are not exposing the `delimiter` parameter for pattern resolution for S3 and GCS file locations. Which significantly limits the use of external files.

**Describe the solution you'd like**
Expose a `delimiter` which then is passed to the upstream API calls.

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Add an example in docs with the usage

kaxil:
Can we use / repurpose the "filetype" arg we have for that? thoughts?

phanikumv:
This is not required anymore as discussed with the team. Hence closing for now.

PR #691, opened 2022-08-18
Expose S3 and GCS `delimiter` parameter for listing objects

utkarsharma2:
# Description
Currently, we are not exposing the `delimiter` parameter for pattern resolution for S3 and GCS file locations. Which significantly limits the use of external files.

closes: #690 

## What is the new behavior?
Expose a `delimiter` which then is passed to the upstream API calls.

Nope

### Checklist

### DIFF
--- a/src/astro/files/base.py
+++ b/src/astro/files/base.py
@@ -24,22 +24,24 @@ class File:
         conn_id: str | None = None,
         filetype: constants.FileType | None = None,
         normalize_config: dict | None = None,
+        delimiter: str | None = None
     ):
         """Init file object which represent a single file in local/object stores
 
         :param path: Path to a file in the filesystem/Object stores
         :param conn_id: Airflow connection ID
         :param filetype: constant to provide an explicit file type
-        :param normalize_config: parameters in dict format of pandas json_normalize() function.
+        :param normalize_config: parameters in dict format of pandas json_normalize() function
+        :param delimiter: the delimiter marks key hierarchy
         """
-        self.location: BaseFileLocation = create_file_location(path, conn_id)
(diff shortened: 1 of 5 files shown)


feluelle (commented):
LGTM, but don't we need to mention it in the docs?

tatiana (commented):
@utkarsharma2 please, could you add some tests to highlight the difference in the behaviour before and after this change?

sunank200 (commented):
LGTM. Adding test for the test would be useful

src/astro/files/base.py
```diff
@@ -24,22 +24,24 @@ def __init__(
         conn_id: str | None = None,
         filetype: constants.FileType | None = None,
         normalize_config: dict | None = None,
+        delimiter: str | None = None,
```
tatiana, line 27: We'll probably need to align this with #772, depending on which PR is merged first.
Should we have a test to cover this new parameter support?

src/astro/files/base.py
```diff
@@ -24,22 +24,24 @@ def __init__(
...
-        self.location: BaseFileLocation = create_file_location(path, conn_id)
+        self.location: BaseFileLocation = create_file_location(path, conn_id, delimiter)
         self.type = create_file_type(
             path=path, filetype=filetype, normalize_config=normalize_config
         )
 
     @property
     def path(self) -> str:
-        return self.location.path
+        return str(self.location.path)  # ToDo: look into typing issue
```
feluelle, line 44: Could we try to fix that before merging? Let me know if you need help.

utkarsharma2:
We don't need this now.

PR #692, opened 2022-08-18
Re-order docs sections

kaxil:
This reorders doc sections and groups similar sections.

- Moves dev docs at the end to focus on Concepts and Guides
- Moves references to second last

**Before**:
image

**After**:

image

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -72,7 +72,7 @@ This is a faster way for datasets of larger size as there is only one network ca
 .. _templating:
 
 Templating
-~~~~~~~~~~
+----------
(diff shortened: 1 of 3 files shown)


utkarsharma2 (approved):
LGTM

jwitz (approved):
Looks a lot cleaner! I don't feel strongly either way, but in the future it might make sense to add "Operators" to "Reference" and to have "Concepts" include some more high level documentation. Knowing what each operator does _seems_ more referency to me. 

But because "Concepts" is only one doc for now, I think this structure works for the content we have.

kaxil:
> Looks a lot cleaner! I don't feel strongly either way, but in the future it might make sense to add "Operators" to "Reference" and to have "Concepts" include some more high level documentation. Knowing what each operator does _seems_ more referency to me.
> 
> But because "Concepts" is only one doc for now, I think this structure works for the content we have.

That's a nice feedback Jake, thanks, we will sync up with you over the coming weeks to change that.

PR #696, opened 2022-08-18
Remove disclaimer & bump version

kaxil:
We are now production-ready

### DIFF
--- a/src/astro/__init__.py
+++ b/src/astro/__init__.py
@@ -1,6 +1,6 @@
 """A decorator that allows users to run SQL queries natively in Airflow."""
 
-__version__ = "1.0.0b1"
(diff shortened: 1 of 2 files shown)


pankajastro (approved):

src/astro/__init__.py
```diff
@@ -1,6 +1,6 @@
 """A decorator that allows users to run SQL queries natively in Airflow."""
 
-__version__ = "1.0.0b1"
+__version__ = "1.1.0a1"
```
utkarsharma2, line 3: @kaxil it should be 1.0.0, right?

kaxil: No, this is for main branch -- so it should be the next minor version so we can differentiate between what is released and what is next

PR #700, opened 2022-08-19
Add redshift native path

feluelle:
# Description

We currently do not have a native path available for transferring data to redshift.

closes: #613

## What is the new behavior?

This PR adds native path for transferring data to redshift.

No

### Checklist


dev/Dockerfile
```diff
@@ -9,10 +9,12 @@ RUN apt-get install -y --no-install-recommends \
         libsasl2-modules
 ENV SETUPTOOLS_USE_DISTUTILS=stdlib
 
+
```
pankajkoti, line 12:
```suggestion
```

dev/Dockerfile
```diff
@@ -9,10 +9,12 @@ RUN apt-get install -y --no-install-recommends \
...
 COPY ../pyproject.toml  ${AIRFLOW_HOME}/astro_sdk/
 # The following file are needed because version they are referenced from pyproject.toml
 COPY ../README.md  ${AIRFLOW_HOME}/astro_sdk/
 COPY ../src/astro/__init__.py  ${AIRFLOW_HOME}/astro_sdk/src/astro/__init__.py
 
 RUN pip install -e "${AIRFLOW_HOME}/astro_sdk[all,tests,doc]"
+
```
pankajkoti, line 19:
```suggestion
```

pankajkoti:
Guide for creating IAM role with needed permissions for COPY command authorization: https://www.dataliftoff.com/iam-roles-for-loading-data-from-s3-into-redshift/

src/astro/databases/aws/redshift.py
```diff
@@ -1,19 +1,30 @@
...
 from astro.databases.base import BaseDatabase
+from astro.exceptions import DatabaseCustomError
 from astro.files import File
 from astro.settings import REDSHIFT_SCHEMA
 from astro.sql.table import Metadata, Table
 
 DEFAULT_CONN_ID = RedshiftSQLHook.default_conn_name
+NATIVE_PATHS_SUPPORTED_FILE_TYPES = {
+    FileType.CSV: "CSV",
+    FileType.JSON: "JSON 'auto'",
```
utkarsharma2, line 25: @pankajkoti Can you add a comment explaining what `auto` means here?

src/astro/databases/aws/redshift.py
```diff
@@ -113,4 +127,100 @@ def is_native_load_file_available(
...
+            sql_statement = (
+                f"COPY {table_name} "
+                f"FROM '{source_file.path}' "
+                f"IAM_ROLE '{iam_role}' "
+                f"{file_type} "
+                f"{copy_options}"
+            )
+        else:
+            aws_creds = source_file.location.hook.get_credentials()
+            sql_statement = (
```
utkarsharma2, line 214: @pankajkoti isn't this a security issue? should we do this?

cc: @kaxil

tests/databases/test_redshift.py
```diff
@@ -189,10 +189,12 @@ def test_load_file_from_cloud_to_table(database_table_fixture):
...
     database.load_file_to_table(
         File(
-            "s3://astro-sdk/data_redshift/", conn_id="aws_conn", filetype=FileType.CSV
+            "s3://astro-sdk-redshift/data_redshift/",
+            conn_id="aws_conn",
+            filetype=FileType.CSV,
         ),
         target_table,
-        {},
+        use_native_support=False,
```
utkarsharma2, line 197: @pankajkoti why is this - `use_native_support=False` here? what are we testing?

pankajkoti: This is an existing test which was needed for testing load from a S3 bucket containing files and sub_folders; where the with the default option dataframe infers the schema and loads data into corresponding table.

src/astro/databases/aws/redshift.py
pankajkoti: Which is a security issue you mean?

pankajkoti: btw had discussed this with @utkarsharma2 on what he meant by security issue. The implementation provides two options for authorization, one using IAM role and the other using IAM user. If IAM role is not provided, it assumes the creds are for IAM user and then uses the access_key_id and secret_access_key pair for authentication. Utkarsh feels that using this pair is a security issue and in general we should not provide this option of IAM user and only provide option of IAM role.

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -113,4 +131,96 @@ def is_native_load_file_available(
...
+        else:
+            aws_creds = source_file.location.hook.get_credentials()
+            sql_statement = (
+                f"COPY {table_name} "
+                f"FROM '{source_file.path}' "
+                f"CREDENTIALS 'aws_access_key_id={aws_creds.access_key};"
+                f"aws_secret_access_key={aws_creds.secret_key}' "
+                f"{file_type} "
+                f"{copy_options}"
+            )
```
kaxil, line 221: @utkarsharma2 - I think there are just 2 ways of loading from S3 to Redshift

1. using IAM role
2. using Creds

In docs, we can say that we highly recommend using IAM role

src/astro/databases/aws/redshift.py
kaxil: sure, Yeah I am happy with that

pankajkoti:
Based on the offline discussion, we're removing the options to allow using `creds`. We're only supporting `IAM_ROLE` option now.

kaxil:
> Based on the offline discussion, we're removing the options to allow using `creds`. We're only supporting `IAM_ROLE` option now.

We should cover it in the docs on why we don't support creds though

pankajkoti:
> > Based on the offline discussion, we're removing the options to allow using `creds`. We're only supporting `IAM_ROLE` option now.
> 
> We should cover it in the docs on why we don't support creds though

We have enhanced the docs now. It appears as below . Thanks @kaxil for directing to use the Sphinx note directive.
Screenshot 2022-09-01 at 9 29 56 PM


cc: @utkarsharma2

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -264,4 +283,89 @@ def is_native_load_file_available(
...
+            <https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html>`_
+        """
+        native_support_kwargs = native_support_kwargs or {}
+
+        table_name = self.get_table_qualified_name(target_table)
+        file_type = NATIVE_PATHS_SUPPORTED_FILE_TYPES.get(source_file.type.name)
+
+        iam_role = native_support_kwargs.pop("IAM_ROLE", None)
+        if not iam_role:
+            raise TypeError(
```
utkarsharma2, line 353:
```suggestion
            raise ValueError(
```

utkarsharma2: @pankajkoti I think this should be ValueError since the value is expected as input and is not passed.

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -7,13 +7,29 @@
...
 from astro.sql.table import Metadata, Table
 
 DEFAULT_CONN_ID = RedshiftSQLHook.default_conn_name
+NATIVE_PATHS_SUPPORTED_FILE_TYPES = {
+    FileType.CSV: "CSV",
+    # By default, COPY attempts to match all columns in the target table to JSON field name keys.
+    # With this option, matching is case-sensitive. Column names in Amazon Redshift tables are always lowercase,
+    # so when you use the 'auto' option, matching JSON field names must also be lowercase.
+    # Refer: https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-format.html#copy-json
+    FileType.JSON: "JSON 'auto'",
```
utkarsharma2, line 30: @pankajkoti can you add a test case where we have mixed casing in cols names for load_file operator? use - tests/data/homes_upper.csv. 

I just want to make sure this works with mixed col casing.

python-sdk/src/astro/databases/aws/redshift.py
pankajkoti: I believe this should be TypeError. I too was confused on whether it should be ValueError or TypeError, but value error is when an invalid value is passed. Here, Python gives a Typ
Screenshot 2022-09-02 at 12 12 39 PM
eError when an expected argument itself is not passed.

python-sdk/src/astro/databases/aws/redshift.py
pankajkoti: Added a test. Also using` json 'auto ignorecase' ` instead of `json 'auto'` option to allow non-lower case columns as mentioned here https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-format.html#copy-json

utkarsharma2 (approved):
LGTM 👍

Issue #702, opened 2022-08-19
Fix Biguery native load for 10GB benchmark dataset


utkarsharma2:
**Describe the bug**
When running benchmarking for 10GB files we faced issues in loading and running benchmarking.

**Version**
* Astro: 1.0.0

**To Reproduce**
Try running benchmarking script for Bigquery for any native paths. 

**Expected behavior**
Data should load without any issues.

PR #707, opened 2022-08-21
Add benchmarking result for Bigquery 10gb datasets

utkarsharma2:
# Description
When running benchmarking for 10GB files we faced issues in loading and running benchmarking.

closes: #702

## What is the new behavior?
Added 10GB dataset benchmarking numbers. Added a new dataset that works with big queries.

Nope

### Checklist

### DIFF
--- a/tests/benchmark/results.md
+++ b/tests/benchmark/results.md
@@ -68,7 +68,7 @@ For Machine types: n2-standard-4
 | imdb/title_ratings_10mb.csv                                              | 10MB | 19.40  |
 | stackoverflow/stackoverflow_posts_1g.ndjson                              | 1GB  | 30.26  |
 | trimmed/pypi/*                                                           | 5GB  | 59.90  |
-| gs://astro-sdk/benchmark/trimmed/stackoverflow/10gb/ - 10 Files 1gb each | 10GB |11.88min|
+| gs://astro-sdk/benchmark/trimmed/stackoverflow/10gb/ - 10 Files 1gb each | 10GB |1.94min|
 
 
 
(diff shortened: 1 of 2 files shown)


tests/benchmark/results.md
```diff
@@ -61,13 +61,15 @@ For Machine types: n2-standard-4
...
-|stackoverflow/stackoverflow_posts_1g.ndjson|1GB  | 30.26          |
-|trimmed/pypi/*                             |5GB  | 59.90          |
+| Dataset                                                                  | Size | Duration(seconds) |
+|--------------------------------------------------------------------------|------|--------|
+| covid_overview/covid_overview_10kb.csv                                   | 10 KB | 13.57  |
+| tate_britain/artist_data_100kb.csv                                       | 100KB | 16.10  |
+| imdb/title_ratings_10mb.csv                                              | 10MB | 19.40  |
+| stackoverflow/stackoverflow_posts_1g.ndjson                              | 1GB  | 30.26  |
+| trimmed/pypi/*                                                           | 5GB  | 59.90  |
+| gs://astro-sdk/benchmark/trimmed/stackoverflow/10gb/ - 10 Files 1gb each | 10GB |1.94min|
```
sunank200, line 71: is this a new dataset? May be we should add it to the list in Dataset.md

sunank200 (approved):

tests/benchmark/results.md
tatiana: Why is this a new dataset..?
Should we also run the benchmark for the other databases to use it, so we can have a fair comparison?

utkarsharma2: @tatiana 
1. invalid rows with null values 
2. There was a type mismatch as well

Issue #672, opened 2022-08-17
Run tests on release branch


kaxil:
Currently, our tests don't run on PRs that are not from the main, even if it is release branch because of 

https://github.com/astronomer/astro-sdk/blob/main/.github/workflows/ci.yaml#L2-L10

PR #720, opened 2022-08-23
Run tests on release branch

sunank200:
# Description

Currently, our tests don't run on PRs that are not from the main, even if it is a release branch because of

https://github.com/astronomer/astro-sdk/blob/main/.github/workflows/ci.yaml#L2-L10

close: #672 

## What is the new behavior?

- Add CI/CD triggers on release branches

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,12 +1,12 @@
 name: Build and test astro
 on:
   push:
-    branches: [ 'main' ]
+    branches: [ 'main', 'release-**' ]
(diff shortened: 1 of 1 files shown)


kaxil (commented):
There is a separate issue, where if you create a Github release it will just run https://github.com/astronomer/astro-sdk/blob/main/.github/workflows/release.yaml

Can we add a change in this PR so that the release step only runs after all the test succeeds?

kaxil (approved):

Issue #722, opened 2022-08-23
load_file operator: filetype param is not considered for path pattern 


utkarsharma2:
**Describe the bug**
When passing the pattern to the load_file operator the filetype is not considered by the pattern resolution function.

**Version**
* Astro: 1.0.0

**To Reproduce**
1. pass a pattern to the load_file operator without containing the file extensions and pass in the extension in filetype param. 
2. This will fail even if you have passed with filetype.

PR #723, opened 2022-08-23
Add missing filetype argument to "resolve_file_path_pattern" function

utkarsharma2:
# Description
When passing the pattern to the load_file operator the filetype is not considered by the pattern resolution function.

closes: #722 

## What is the new behavior?
Pass the new param

No

### Checklist

### DIFF
--- a/src/astro/databases/base.py
+++ b/src/astro/databases/base.py
@@ -308,6 +308,7 @@ class BaseDatabase(ABC):
             input_file.path,
             input_file.conn_id,
             normalize_config=normalize_config,
+            filetype=input_file.type.name,
         )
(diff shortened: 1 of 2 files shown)


sunank200:
@utkarsharma2 code coverage needs to be fixed.

src/astro/databases/base.py
```diff
@@ -308,6 +308,7 @@ def load_file_to_table(
             input_file.path,
             input_file.conn_id,
             normalize_config=normalize_config,
+            filetype=input_file.type.name,
```
kaxil, line 311: Should we add a unit-test for this case or not needed, wdyt?

utkarsharma2: Added.

Issue #729, opened 2022-08-23
if_exists='append' in not respected in Snowflake database


utkarsharma2:
**Describe the bug**
Even if we have passed if_exists='append' we are still trying to create a table. 

**Version**
* Astro: 1.0.0

**To Reproduce**
Try load_file with if_exists='append' the operator will still try to create a table.

**Expected behavior**
When if_exists='append' AstroSDK should not create a table.

PR #730, opened 2022-08-23
Added a check on "replace" in snowflake's "load_pandas_dataframe_to_table" 

utkarsharma2:
# Description
Even if we have passed if_exists='append' we are still trying to create a table.

closes: #729 

## What is the new behavior?
We check for if_exists=='replace' and only create a table in this case.

Nope

### Checklist


sunank200 (approved):
LGTM. Is this only in snowflake? We should check in other storage as well.

Deepsource and tests needs to pass

utkarsharma2:
@sunank200 Yes, I checked for all DBs and it's only in snowflake.

src/astro/databases/snowflake.py
```diff
@@ -445,9 +445,9 @@ def load_pandas_dataframe_to_table(
         :param if_exists: Strategy to be used in case the target table already exists.
         :param chunk_size: Specify the number of rows in each batch to be written at a time.
         """
-        self.create_table(target_table, dataframe=source_dataframe)
+        if if_exists == "replace":
```
kaxil, line 448: Does it do the check to see if the table already exists?

tests/databases/test_snowflake.py
```diff
@@ -153,6 +153,38 @@ def test_load_pandas_dataframe_to_table(database_table_fixture):
...
+    database, table = database_table_fixture
+
+    pandas_dataframe = pd.DataFrame(data={"id": [1, 2]})
+
+    with mock.patch("snowflake.connector.pandas_tools.write_pandas"):
+        with mock.patch("astro.databases.base.BaseDatabase.create_table") as method:
+            database.load_pandas_dataframe_to_table(
+                pandas_dataframe, table, if_exists="replace"
+            )
+            assert method.called
```
kaxil, line 179:
```suggestion
            assert method.called_once_with()
```

kaxil: This will be a better test if you can assert what's passed to `create_table`

tests/databases/test_snowflake.py
```diff
@@ -153,6 +153,38 @@ def test_load_pandas_dataframe_to_table(database_table_fixture):
...
+
+        with mock.patch("astro.databases.base.BaseDatabase.create_table") as method:
+            database.load_pandas_dataframe_to_table(
+                pandas_dataframe, table, if_exists="append"
+            )
+            assert not method.called
```
kaxil, line 185: Did you verify that this test fails without this change?

src/astro/databases/snowflake.py
utkarsharma2: create_table calls - 
1. create_table_using_schema_autodetection() - which calls prep_table() with if_exists="replace".
2. create_table_using_columns - I have added drop_table() to it.

tests/databases/test_snowflake.py
utkarsharma2: Yes, I was in a call with @bharanidharan14, when this was failing on multiple occasions.

kaxil (approved):

tests/databases/test_snowflake.py
```diff
@@ -153,6 +153,38 @@ def test_load_pandas_dataframe_to_table(database_table_fixture):
...
+    database, table = database_table_fixture
+
+    pandas_dataframe = pd.DataFrame(data={"id": [1, 2]})
+
+    with mock.patch("snowflake.connector.pandas_tools.write_pandas"):
+        with mock.patch("astro.databases.base.BaseDatabase.create_table") as method:
+            database.load_pandas_dataframe_to_table(
+                pandas_dataframe, table, if_exists="replace"
+            )
+            assert method.called_once_with(table, pandas_dataframe)
```
kaxil, line 179:
```suggestion
            assert method.assert_called_with(table, pandas_dataframe)
```

Issue #733, opened 2022-08-23
load_file failing while processing S3 pattern paths


utkarsharma2:
**Describe the bug**
When load_file is passed patterns paths of S3 the operator is falling.

**Version**
* Astro: 1.0.0

**To Reproduce**
1. Run load_file with pattern path of S3

**Expected behavior**
load_file should function with pattern path of S3

PR #740, opened 2022-08-24
Add check to skip folder when loading files

sunank200:
# Description

When load_file is passed patterns paths of S3 the operator is falling.

closes: #733 

## What is the new behavior?

- Check for folder while resolving the path pattern
- Add tests for the same.

### Checklist


src/astro/files/base.py
```diff
@@ -142,15 +144,21 @@ def resolve_file_path_pattern(
...
+        if path.endswith("/"):
+            continue
+
+        files.append(
+            File(
+                path=path,
+                conn_id=conn_id,
+                filetype=filetype,
+                normalize_config=normalize_config,
+            )
```
kaxil, line 160:
```suggestion

    files = [
        File(
            path=path,
            conn_id=conn_id,
            filetype=filetype,
            normalize_config=normalize_config,
        )
        for path in location.paths if not path.endswith("/")
    ]
```

kaxil: Logically same but uses list comprehension

sunank200: Ohh, I also added the same change there: https://github.com/astronomer/astro-sdk/pull/740/commits/124c42f624b6d7867a3e37c94a81335b04a67d5a. Sorry for that.

src/astro/files/base.py
```diff
@@ -38,11 +40,11 @@ def __init__(
         )
 
     @property
-    def path(self) -> str:
+    def path(self) -> Any:
```
kaxil, line 43: Why do we need to change this?

src/astro/files/base.py
```diff
@@ -38,11 +40,11 @@ def __init__(
...
         return self.location.path
 
     @property
-    def conn_id(self) -> str | None:
+    def conn_id(self) -> Any:
```
kaxil, line 47: Why do we need to change this?

tests/databases/test_snowflake.py
```diff
@@ -200,8 +200,8 @@ def test_if_exist_param_of__load_pandas_dataframe_to_table(database_table_fixtur
 def test_load_file_to_table(database_table_fixture):
     """Test loading on files to snowflake database"""
     database, target_table = database_table_fixture
-    filepath = str(pathlib.Path(CWD.parent, "data/sample.csv"))
-    database.load_file_to_table(File(filepath), target_table, {})
+    filepath = str(pathlib.Path(CWD.parent, "data/sub_folder/"))
```
kaxil, line 203: We need an integration test for checking this behaviour for remote storage like S3 or GCS too

src/astro/files/base.py
sunank200: Surprisingly, I had not changed that file but its throwing the following mypy error:
```
src/astro/files/base.py: note: In member "path" of class "File":
src/astro/files/base.py:44: error: Returning Any from function declared to
return "str"  [no-any-return]
```

But the type of path is `str`

src/astro/files/base.py
sunank200: I had not changed this code but getting this mypy error:
```
src/astro/files/base.py: note: In member "conn_id" of class "File":
src/astro/files/base.py:48: error: Returning Any from function declared to
return "Optional[str]"  [no-any-return]
```

Return type is `Optional[str]` which is equivalent to `str | None`

tests/databases/test_snowflake.py
sunank200: Added it here: https://github.com/astronomer/astro-sdk/pull/740/commits/544a21cb4982af2184337c4e461c258715b8c397

I have uploaded the similar folder structure on s3 bucket

src/astro/files/base.py
kaxil: The test passed when I reverted the changes @sunank200 in https://github.com/astronomer/astro-sdk/pull/740/commits/96a17d6bb035db24ee99d68bfbb09af9da01816a 

Where was it failing for you?

src/astro/files/base.py
kaxil: The test passed when I reverted the changes @sunank200 in https://github.com/astronomer/astro-sdk/pull/740/commits/96a17d6bb035db24ee99d68bfbb09af9da01816a 

Where was it failing for you?

kaxil (approved):

src/astro/files/base.py
```diff
@@ -40,11 +39,11 @@ def __init__(
         )
 
     @property
-    def path(self) -> Any:
+    def path(self) -> str:
```
sunank200, line 42: @kaxil does this work for you? Its throwing mypy error for me on my local

src/astro/files/base.py
sunank200: It failed for me in local while doing the commit @kaxil

src/astro/files/base.py
kaxil: Works for me

Issue #742, opened 2022-08-24
Limit Google Protobuf for compatibility with bigquery client


sunank200:
**Describe the bug**
The Biquery client does not work well with protobuf > 3.20.0 and it should be limited until we solve the problem.
Got the following error:
```
[2022-08-22T22:15:58.602045] Traceback (most recent call last):
[2022-08-22T22:15:58.602045] File "/app/api/run_task_utils.py", line 134, in run
[2022-08-22T22:15:58.602045] self.task_instance._run_raw_task()
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/airflow/utils/session.py", line 70, in wrapper
[2022-08-22T22:15:58.602045] return func(*args, session=session, **kwargs)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/airflow/models/taskinstance.py", line 1340, in _run_raw_task
[2022-08-22T22:15:58.602045] self._execute_task_with_callbacks(context)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/airflow/models/taskinstance.py", line 1477, in _execute_task_with_callbacks
[2022-08-22T22:15:58.602045] result = self._execute_task(context, self.task)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/airflow/models/taskinstance.py", line 1538, in _execute_task
[2022-08-22T22:15:58.602045] result = execute_callable(context=context)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/astro/sql/operators/transform.py", line 24, in execute
[2022-08-22T22:15:58.602045] super().execute(context)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/astro/sql/operators/base_decorator.py", line 60, in execute
[2022-08-22T22:15:58.602045] self.database_impl = create_database(self.conn_id)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/site-packages/astro/databases/__init__.py", line 31, in create_database
[2022-08-22T22:15:58.602045] module = importlib.import_module(module_path)
[2022-08-22T22:15:58.602045] File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
[2022-08-22T22:15:58.602045] return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #743, opened 2022-08-24
Limit Google Protobuf for compatibility with bigquery client 

sunank200:
# Description

The Biquery client does not work well with protobuf > 3.20.0

closes: #742 
Also fixed in https://github.com/apache/airflow/pull/25886

## What is the new behavior?

- Limit `protobuf<=3.20.*` in pyproject.toml

No

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ tests = [
     "sqlalchemy-stubs"  # Change when sqlalchemy is upgraded https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
 ]
 google = [
-    "protobuf==3.20.*",
+    "protobuf<=3.20.*",
     "apache-airflow-providers-google>=6.4.0",
     "sqlalchemy-bigquery>=1.3.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -51,6 +51,7 @@ tests = [
     "sqlalchemy-stubs"  # Change when sqlalchemy is upgraded https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
 ]
 google = [
+    "protobuf<=3.20.*",
```
kaxil, line 54: Can you add a comment similar to the one on L51 -- on why we have this pin

kaxil, line 54:
```suggestion
    "protobuf<=3.20",
```

kaxil (approved):

pyproject.toml
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/743/commits/be643b513292f6e51821103bc0aaa11ae12ec2ea)

PR #744, opened 2022-08-24
Create subprojects for astro-sdk-python and astro-sql-cli

dimberman:
# Description
Currently, the astro-sdk project is set up only to handle one project. This has been fine when our only project was the Astro Python SDK, but now that we are expanding into more projects it is critical that we can run multiple projects in the same repo.

## What is the new behavior?
This PR takes all existing python-sdk work and puts it in a `python-sdk` directory. It then creates a `sql-cli` directory to ensure that we have a second working project (this process will make it easy to repeat across any other projects we want to add).

No

### Checklist


tatiana:
@dimberman great step towards the repo refactoring!

As we spoke yesterday, we'll need to make some adjustments in the main README, perhaps giving some context to the mono-repo. @mikeshwe, could you help us with this? @dimberman and I agreed to address this in a separate ticket/PR since this one is already massive!

Did you have a chance to run the makefile commands to confirm things are working as expected?

@pankajkoti @utkarsharma2 @sunank200 @pankajastro : it will be slightly disruptive to merge this change, but the sooner we can do that, I believe the better. Please, help reviewing it!

dimberman:
@tatiana https://github.com/astronomer/astro-sdk/issues/756

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
tatiana, line 1: We probably don't want this folder `.idea` and everything under it

sql-cli/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
tatiana, line 1: This should also be deleted

tatiana (approved):
Thanks a lot, @dimberman !

I'm sure there will be parts that may not be working since we did not test a couple of workflows (build docs, release) - but I think we should merge this sooner than later. We'll solve issues as we identify them.

Issue #749, opened 2022-08-25
When passing pattern to "File" object filetype param doesn't make sense


utkarsharma2:
**Please describe the feature you'd like to see**
If we pass pattern to a create a file object it may or may not have an extension in it
example -
`File(path="S3://tmp/")` - this won't work right now. Since it's mandatory to have a `filetype` param if the extensions is not specified in `path`

`filetype` param should not be mandatory, because a user can pass a folder path as well.

**Describe the solution you'd like**
When defining the file operator we should first try to resolve the location and then assign and then try to evaluate the file type.

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

pankajastro:
possible duplicate https://github.com/astronomer/astro-sdk/issues/644

sunank200:
Yes, @pankajastro. I am working on this

sunank200:
@utkarsharma2 I feel we need file type to be passed from the user's side in case of pattern or folder. 

What would be the filetype when the user passes a folder with both CSV and parquet file or other types as well?

tatiana:
I'm coming late to this discussion, so I'm probably out of context. Since the ticket is still open, however, I'll share my thoughts :)

I don't think we should support folders which contain a mixture of file types. What would be the use case for this?

My impression is that it also introduces more overhead and complexity without a clear benefit.

utkarsharma2:
@tatiana I agree, we are planning to address this by introducing glob patterns. So the user is in full control and filters out all the files he doesn't wish to load into a table. 

Example - 
S3 bucket has following files

s3://tmp/test.csv
s3://tmp/test2.csv
s3://tmp/test3.ndjson

[...]

PR #751, opened 2022-08-26
[WIP] Update `get_file_list` to accept File object

pankajastro:
Update `get_file_list` to accept File object instead of `path` and `conn_id` this would make the public interface more consistent.

depend on: https://github.com/astronomer/astro-sdk/issues/749


pankajastro:
closing this since it depends on https://github.com/astronomer/astro-sdk/issues/749 which is on hold.

PR #753, opened 2022-08-26
Add merge_operator for Redshift database

pankajkoti:
Redshift does not enforce constraints although it accepts those
in the table DDL. Ref: https://docs.aws.amazon.com/redshift/latest/dg/t_Defining_constraints.html
As a result, in case of conflicts, e.g. uniqueness constraint, it
does allow insertion of duplicate values and expects you to clean
up duplicated.

Additionally, for merge redshift documentation suggests making use
of a staging table. Ref: https://docs.aws.amazon.com/redshift/latest/dg/t_updating-inserting-using-staging-tables-.html
The merge implementation involves the following steps:
i. Create a staging table having schema identical to the target table
ii. Load data from source table into stage table by taking care of
    mapping source table column to stage table column using the given
    source_to_target_columns_map.
iii. Handle conflicts. There are two strategies to handle conflict and
     they are explained in detail as below:
     a. ignore: To ignore rows in case of conflicts, we remove the
                matching rows from stage table that are present in the
                target table.
     b. update: To update rows in case of conflicts, we remove the
		matching rows from the target table that are present
		in the stage table.
iv. Load the stage table data into target table

# Description
## What is the new behavior?

-
-
-

### Checklist


src/astro/databases/aws/redshift.py
```diff
@@ -104,6 +107,96 @@ def load_pandas_dataframe_to_table(
...
+        :param if_conflicts: The strategy to be applied if there are conflicts.
+        """
+        source_schema = source_table.metadata.schema
+        source_table_name = f"{source_schema}.{source_table.name}"
+        target_schema = target_table.metadata.schema
+        target_table_name = f"{target_schema}.{target_table.name}"
+        stage_table_name = random.choice(string.ascii_lowercase) + "".join(
+            random.choice(string.ascii_lowercase + string.digits)
+            for _ in range(UNIQUE_HASH_SIZE - 1)
+        )
```
utkarsharma2, line 167:
```suggestion
        stage_table_name = self.get_table_qualified_name(Table(metadata=Metadata(schema=DEFAULT_SCHEMA)))
        )
```

src/astro/databases/aws/redshift.py
```diff
@@ -104,6 +107,96 @@ def load_pandas_dataframe_to_table(
...
+
+        begin_transaction = "BEGIN TRANSACTION"
+        create_temp_table = f"CREATE TEMP TABLE {stage_table_name} (LIKE {target_schema}.{target_table.name})"
```
utkarsharma2, line 170:
```suggestion
        create_temp_table = f"CREATE TEMP TABLE {stage_table_name} (LIKE {target_table_name})"
```

src/astro/databases/aws/redshift.py
```diff
@@ -104,6 +107,96 @@ def load_pandas_dataframe_to_table(
...
+        end_transaction = "END TRANSACTION"
+
+        statements = [begin_transaction, create_temp_table, insert_into_stage_table]
+        if conflict_statement:
+            statements.append(conflict_statement)
+        statements.extend([insert_into_target_table, drop_stage_table, end_transaction])
+
+        with self.hook.get_cursor() as cursor:
+            for statement in statements:
+                cursor.execute(statement)
```
utkarsharma2, line 198: @pankajkoti Are all the statements executed in the same session? I just want to make sure `BEGIN TRANSACTION` and `END TRANSACTION` are respected when running statements in the different `cursor.execute(statement)` or they should be part of the same query and executed with the single cursor.execute( ";".join(statement)) ??

pankajkoti: Yes @utkarsharma2 . They are part of the same session and until the point `END TRANSACTION` is executed, nothing is committed to the DB. Also, I had investigated previously on whether we can execute multiple queries in one go, but it is a limitation on Redshift connector.
Ref: https://github.com/aws/amazon-redshift-python-driver/issues/82 & https://github.com/aws/amazon-redshift-python-driver/issues/39

src/astro/databases/aws/redshift.py
```diff
@@ -104,6 +107,94 @@ def load_pandas_dataframe_to_table(
...
+        :param source_to_target_columns_map: Dict of target_table columns names to source_table columns names
+        :param target_conflict_columns: List of cols where we expect to have a conflict while combining
+        :param if_conflicts: The strategy to be applied if there are conflicts.
+        """
+        source_schema = source_table.metadata.schema
+        source_table_name = self.get_table_qualified_name(source_table)
+        target_schema = target_table.metadata.schema
+        target_table_name = self.get_table_qualified_name(target_table)
+        stage_table_name = self.get_table_qualified_name(Table(metadata=Metadata(schema=DEFAULT_SCHEMA)))
+        )
```
pankajkoti, line 165:
```suggestion
```

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -104,6 +104,157 @@ def load_pandas_dataframe_to_table(
...
+
+        :param source_table: Contains the rows to be merged to the target_table
+        :param target_table: Contains the destination table in which the rows will be merged
+        :param source_to_target_columns_map: Dict of target_table columns names to source_table columns names
+        :param target_conflict_columns: List of cols where we expect to have a conflict while combining
+        :param if_conflicts: The strategy to be applied if there are conflicts.
+        """
+        source_table_name = self.get_table_qualified_name(source_table)
+        target_table_name = self.get_table_qualified_name(target_table)
+        stage_table_name = self.get_table_qualified_name(Table())
```
utkarsharma2, line 187: @pankajkoti I think it would be a good idea to create this table in the temp schema. right now we are creating this in the default database scheme. WDYT?

utkarsharma2 (approved):
LGTM! Nice work! :)

python-sdk/src/astro/databases/aws/redshift.py
pankajkoti: Redshift doe not accept schema option for temp tables. As per the docs here: https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html

TEMPORARY | TEMP
Keyword that creates a temporary table that is visible only within the current session. The table is automatically dropped at the end of the session in which it is created. The temporary table can have the same name as a permanent table. The temporary table is created in a separate, session-specific schema. (You can't specify a name for this schema.) This temporary schema becomes the first schema in the search path, so the temporary table will take precedence over the permanent table unless you qualify the table name with the schema name to access the permanent table. For more information about schemas and precedence, see [search_path](https://docs.aws.amazon.com/redshift/latest/dg/r_search_path.html).

PR #766, opened 2022-09-02
Remove duplicate dependency

kaxil:
`protobuf` is mentioned twice


python-sdk/pyproject.toml
```diff
@@ -74,7 +74,6 @@ all = [
     "apache-airflow-providers-google>=6.4.0",
     "apache-airflow-providers-postgres",
     "apache-airflow-providers-snowflake",
-    "protobuf<=3.20",
```
kaxil, line 77: It is also present on L83

Issue #611, opened 2022-08-08
Map File & Table classes to Datasets (AIP-48)


kaxil:
There will be an alpha build of 2.4 very soon (Sync with the Ash/Jed to get the tag), so to make sure Python SDK can use 2.4 dataset feature (and allows viewing dataset dependency in UI), let's map the File and Table to AIP-48 datasets to show dependencies between tasks and Dags.

kaxil:
Astro Runtime `6.0.0-alpha1` image is out where I will do the mapping

kaxil:
WIP branch at https://github.com/astronomer/astro-sdk/compare/main...aip-48-mapping

kaxil:
Both Table and File object have been mapped with 6.0 alpha2 runtime image but the code needs to be cleaned up

kaxil:
Pending items:

- [ ] Add inlets and outlets on all operators
- [ ] Add docs

pankajkoti:
PR is in review and @utkarsharma2 is going to review it. Potentially merge it if looks all good today.

kaxil:
Another PR: https://github.com/astronomer/astro-sdk/pull/808

Only doc PR is pending now

kaxil:
I will create separate task for docs so that I can close this out

PR #768, opened 2022-09-02
Make Table and Metadata class to attrs

kaxil:
part of https://github.com/astronomer/astro-sdk/issues/611

Since Airflow's Dataset object uses `attrs`, we need to convert Table and File object to attrs so we can inherit from Airflow's dataset in the following PRs.

This PR only updates Table and Metadata, I will do File in a follow-up PR

No

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
         exclude: ^mk/.*\.mk$|^python-sdk/docs/Makefile|^python-sdk/Makefile$|^python-sdk/tests/benchmark/Makefile$
 
   - repo: https://github.com/psf/black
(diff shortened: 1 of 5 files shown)


kaxil:
I am merging this without approval but please review and let me know if you have any thoughts, but we need this for 1.1

.pre-commit-config.yaml
```diff
@@ -1,4 +1,6 @@
 ---
+ci:
```
utkarsharma2, line 2: @kaxil why we did these changes? just curious :)

kaxil: Since we are already running mypy checks separately in CI -- separate from pre-commit hook

Issue #769, opened 2022-09-03
Cannot use `cleanup` when running Airflow <2.2.3 due to `airflow.utils.context` import error


josh-fell:
**Describe the bug**
When attempting to start an Astro CLI project running Airflow 2.2.0, the following exception is thrown in the the Scheduler, Triggerer, and Webserver containers:
  ```
File "/usr/local/lib/python3.9/site-packages/airflow/providers_manager.py", line 278, in initialize_providers_list
    self._discover_all_providers_from_packages()
  File "/usr/local/lib/python3.9/site-packages/airflow/providers_manager.py", line 345, in _discover_all_providers_from_packages
    provider_info = entry_point.load()()
  File "/usr/local/lib/python3.9/site-packages/importlib_metadata/__init__.py", line 203, in load
    module = import_module(match.group('module'))
  File "/usr/local/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/usr/local/lib/python3.9/site-packages/astro/__init__.py", line 19, in <module>
    import astro.sql  # noqa: F401
  File "/usr/local/lib/python3.9/site-packages/astro/sql/__init__.py", line 2, in <module>
    from astro.sql.operators.cleanup import CleanupOperator, cleanup
  File "/usr/local/lib/python3.9/site-packages/astro/sql/operators/cleanup.py", line 12, in <module>
    from airflow.utils.context import Context
ModuleNotFoundError: No module named 'airflow.utils.context'
```

The `airflow.utils.context` module was not introduced to OSS Airflow until 2.2.3. Importing this module as a [top-level import in the `CleanupOperator`](https://github.com/astronomer/astro-sdk/blob/bd67729a029bf0c3234059677404092eddd4bfdb/python-sdk/src/astro/sql/operators/cleanup.py#L12) sets an implicit, minimum requirement for Airflow 2.2.3 which is higher than the [minimum `apache-airflow` requirement](https://github.com/astronomer/astro-sdk/blob/bd67729a029bf0c3234059677404092eddd4bfdb/python-sdk/pyproject.toml#L20) set for Astro SDK.

**Version**
* Astro: astro-runtime:4.0.0
* OS: Debian GNU/Linux 10 (buster)

**To Reproduce**
1. Create a DAG using the `cleanup()` function within an Astro CLI project running `astro-runtime:4.0.0`
2. Run `astro dev start`
[...]

PR #770, opened 2022-09-03
Move `airflow.utils.context.Context` import under `TYPE_CHECKING`

josh-fell:
Closes: #769

# Description
The `airflow.utils.context` module was not introduced to OSS Airflow until 2.2.3. Importing this module as a top-level import sets an implicit, minimum requirement for Airflow 2.2.3 which is higher than the minimum `apache-airflow` requirement set for Astro SDK.

## What is the new behavior?

- The `from airflow.utils.context import Context` import is moved under `TYPE_CHECKING`

It should not.

### Checklist


python-sdk/src/astro/files/operators/files.py
```diff
@@ -1,12 +1,15 @@
...
+from typing import TYPE_CHECKING, Any
 
 from airflow.decorators.base import get_unique_task_id
 from airflow.models import BaseOperator
-from airflow.utils.context import Context
 
 from astro.files.base import File
 from astro.files.locations import create_file_location
 
+if TYPE_CHECKING:
```
kaxil, line 9: same as https://github.com/astronomer/astronomer-providers/pull/619#discussion_r962155341

thoughts?

kaxil (approved):

python-sdk/src/astro/files/operators/files.py
tatiana: @josh-fell & @kaxil I like @kaxil 's proposal of having a compatibility module.
It would also be nice to see tests covering this issue - since we are using nox, it should be straightforward to add a test.

python-sdk/src/astro/sql/operators/cleanup.py
```diff
@@ -18,6 +17,10 @@
 from astro.sql.operators.load_file import LoadFileOperator
 from astro.sql.table import Table
 
+if TYPE_CHECKING:
+    # TODO: This can be removed from TYPE_CHECKING once there is a minimum requirement of Airflow 2.2.3+
+    from airflow.utils.context import Context
```
feluelle, line 22: Let's make sure to change to `"Context"` in all places of the current file as you have done it for the other file.

python-sdk/src/astro/files/operators/files.py
josh-fell: Actually, any preference on having a pre-commit hook to check that the typing compat module is used vs. a unit test?

kaxil: Either is fine :)

josh-fell:
Added a pre-commit hook along with a `typing_compat` module in https://github.com/astronomer/astro-sdk/pull/770/commits/68e8522fc8c9f37f79aa681b1cbd3a09352a648a.

python-sdk/src/astro/utils/typing_compat.py
```diff
@@ -0,0 +1,19 @@
...
+# typically used as the typing for the ``context`` arg in operators and sensors. However, using this class for
+# typing outside of TYPE_CHECKING in modules sets an implicit, minimum requirement for Apache Airflow 2.2.3
+# which is currently more recent than the current minimum requirement of Apache Airflow 2.2.0.
+#
+# TODO: Remove this once the repo has a minimum Apache Airflow requirement of 2.2.3+.
+try:
+    from airflow.utils.context import Context
+except ModuleNotFoundError:
+
+    class Context(MutableMapping[str, Any]):  # type: ignore[no-redef]
```
josh-fell, line 13: With the approach of `Context = MutableMapping[str, Any]`, mypy was complaining about not being able to assign to a type and the variable assignment types not matching. Instead of ignoring those mypy errors (which felt icky), I decided to redefine the typing class and ignore the redefinition mypy error.

Does this seem reasonable? Any other mypy workarounds?

kaxil: Sounds good

python-sdk/src/astro/utils/typing_compat.py
```diff
@@ -0,0 +1,19 @@
...
+        """Placeholder typing class for ``airflow.utils.context.Context``."""
+
+        ...
```
kaxil, line 16: you don't need this if you have docstrings
```suggestion
```

josh-fell: 🤦 Yeah, you're right. Removed.

josh-fell:
Actually, since this repo implements PEP563 typing evaluations, we could still use `TYPE_CHECKING` with `airflow.utils.context.Context` _and_ not have to use a stringified type annotation for `Context`. Then, when this repo has a minimum Airflow requirement of at least 2.2.3, it's only a matter of removing the `TYPE_CHECKING` rather than removing the compat module and changing all of the imports.

Just another option to consider and I'll defer to the group on maintenance preference.

kaxil (approved):

PR #771, opened 2022-09-03
Allow passing a dict to `Table.metadata`

kaxil:
This will allow passing a dict to `Table.metadata`:

```python
In [5]: Table("xyz", metadata={'schema': "das", 'database': "dsa"})
Out[5]: Table(conn_id='xyz', _name='', metadata=Metadata(schema='das', database='dsa'), columns=[], temp=True)
```

This is needed for serialization and de-serialization of the Table class in Airflow as Airflow 2.4 serializes the Dataset classes to json/dict for lineage

part of https://github.com/astronomer/astro-sdk/issues/611

### DIFF
--- a/python-sdk/src/astro/sql/table.py
+++ b/python-sdk/src/astro/sql/table.py
@@ -46,7 +46,11 @@ class Table:
     # Some ideas: TableRef, TableMetadata, TableData, TableDataset
     conn_id: str = field(default="")
     _name: str = field(default="")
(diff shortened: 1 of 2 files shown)


utkarsharma2 (approved):
LGTM! :)

utkarsharma2:
@kaxil Why do we need this?

kaxil:
> @kaxil Why do we need this?

Check the PR description :)

PR #772, opened 2022-09-03
Make `File` class attrs object

kaxil:
Part of https://github.com/astronomer/astro-sdk/issues/611

Since Airflow's Dataset object using attrs, we need to convert File object to attrs so we can inherit from Airflow's dataset in the following PRs.

Similar to https://github.com/astronomer/astro-sdk/pull/768

No

### DIFF
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -4,14 +4,16 @@ import io
 
 import pandas as pd
 import smart_open
+from attrs import define
 
 from astro import constants
 from astro.files.locations import create_file_location
 from astro.files.locations.base import BaseFileLocation
-from astro.files.types import create_file_type
+from astro.files.types import FileType, create_file_type
 
 
-class File:  # skipcq: PYL-W1641
+@define
+class File:
     """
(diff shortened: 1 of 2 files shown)


utkarsharma2 (approved):
LGTM!

python-sdk/src/astro/files/base.py
```diff
@@ -22,27 +24,27 @@ class File:  # skipcq: PYL-W1641
     :param normalize_config: parameters in dict format of pandas json_normalize() function.
     """
 
-    template_fields = ("location",)
```
tatiana, line 25: @kaxil is this change to `template_fields` a breaking change in any sense?

I noticed it is now:
```
   template_fields = (
        "path",
        "conn_id",
    )
```
That makes sense, but I was wondering if there was a use case for a `location` to be a template field in the first place.

Should we update the changelog as part of this change to inform the users?

kaxil: Since `location` object just has `path` and `conn_id`, this will have no impact i.e should not change any behavior as now instead of `location`, we template `path` and `conn_id`.

PR #776, opened 2022-09-06
Add test for File & Location equality check

kaxil:
Follow-up of https://github.com/astronomer/astro-sdk/pull/772 to just add more test coverage

### DIFF
--- a/python-sdk/src/astro/files/locations/base.py
+++ b/python-sdk/src/astro/files/locations/base.py
@@ -133,3 +133,6 @@ class BaseFileLocation(ABC):
         if not isinstance(other, self.__class__):
             return NotImplemented
         return self.path == other.path and self.conn_id == other.conn_id
+
+    def __hash__(self) -> int:
+        return hash((self.path, self.conn_id))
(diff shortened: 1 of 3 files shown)


python-sdk/src/astro/files/locations/base.py
```diff
@@ -133,3 +133,6 @@ def __eq__(self, other) -> bool:
         if not isinstance(other, self.__class__):
             return NotImplemented
         return self.path == other.path and self.conn_id == other.conn_id
+
+    def __hash__(self) -> int:
```
utkarsharma2, line 137: @kaxil Where are we planning to use this? For URl?

utkarsharma2 (approved):

python-sdk/src/astro/files/locations/base.py
kaxil: https://deepsource.io/directory/analyzers/python/issues/PYL-W1641

> In order to conform to the object model, classes that define their own equality method should also define their own hash method, or be unhashable. If the hash method is not defined then the hash of the super class is used. This is unlikely to result in the expected behavior.
> 
> A class can be made unhashable by setting its __hash__ attribute to None.
> 
> In Python 3, if you define a class-level equality method and omit a __hash__ method then the class is automatically marked as unhashable.
> 
> It is recommended to implement a __hash__ method or set __hash__ = None, whenever you define an __eq__ method for a class.

Also explained in https://www.attrs.org/en/stable/hashing.html

feluelle (approved):
Nice 👍

PR #777, opened 2022-09-06
Add example DAG using Datasets with astro-sdk

phanikumv:
# Description
There is no example DAG which showcases scheduling a DAG on a Dataset

closes: [#761]

## What is the new behavior?

- New example DAG created which contains a micropipeline DAG getting scheduled on the arrival of a upstream Dataset

No

### Checklist


phanikumv:
Waiting on https://github.com/astronomer/astro-sdk/issues/611 to test the DAG

python-sdk/example_dags/example_datasets.py
```diff
@@ -0,0 +1,49 @@
+from datetime import datetime
+
+from airflow import DAG
+
+from astro import sql as aql
+from astro.files import File
+from astro.sql.table import Table
+
+input_file = File(path="gs://dag-authoring/movies/imdb_v2.csv")
```
kaxil, line 9: Should we add this dag as part of integration tests too?

thoughts @pankajastro @utkarsharma2 ?

dimberman (changes requested):
Looking at this example without context I would have no idea how this relates to airflow datasets. can you please add a description to the example DAG to explain how this ties to AIP-48?

kaxil:
We probably need description like https://github.com/astronomer/astro-sdk/blob/main/python-sdk/example_dags/example_google_bigquery_gcs_load_and_save.py#L1-L14 with what the dag is doing and how you can resuse it

phanikumv:
> Looking at this example without context I would have no idea how this relates to airflow datasets. can you please add a description to the example DAG to explain how this ties to AIP-48?

Hi @dimberman - could you please check now. I made the changes.

python-sdk/example_dags/example_datasets.py
feluelle: In my opinion yes, if we don't have any other DAG testing datasets compatibility.

feluelle (approved):
This is really nice! :) This will be a game changer for airflow etl'ing.

python-sdk/example_dags/example_datasets.py
```diff
@@ -0,0 +1,68 @@
...
+from astro import sql as aql
+from astro.files import File
+from astro.sql.table import Table
+
+input_file = File(
+    path="https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb_v2.csv"
+)
+imdb_movies_table = Table(name="imdb_movies", conn_id="sqlite_default")
+top_animations_table = Table(name="top_animation", conn_id="sqlite_default")
+START_DATE = datetime(2000, 1, 1)
```
kaxil, line 33:
```suggestion
START_DATE = datetime(2022, 9, 1)
```

kaxil:
We should still do this - https://github.com/astronomer/astro-sdk/pull/777#discussion_r966046616

python-sdk/example_dags/example_datasets.py
phanikumv: Will address this in https://github.com/astronomer/astro-sdk/issues/816

phanikumv:
> We should still do this - [#777 (comment)](https://github.com/astronomer/astro-sdk/pull/777#discussion_r966046616)

We will address this in https://github.com/astronomer/astro-sdk/issues/816

PR #779, opened 2022-09-06
Fix ReadTheDocs for Python SDK

kaxil:

Currently the docs are broken for "main" as shown in https://readthedocs.org/projects/astro-sdk-python/builds/

## Changes
`.readthedocs.yaml` should be at the root of repo - https://docs.readthedocs.io/en/stable/config-file/index.html


kaxil:
The docs were built - https://astro-sdk-python--779.org.readthedocs.build/en/779/ -- merging the PR

Issue #709, opened 2022-08-21
Use native autodetect schema


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, If the user does not specify the table columns property, we currently use pandas to autodetect the table column types automatically. This can sometimes lead to a mismatch in type inferred by pandas and one that the database expects.

**Describe the solution you'd like**
We can let the database infer schema when it is supported by the database and if it is use native support this reduces the chances of mismatch.

**Acceptance Criteria**
- [ ] Use native autodetection for Redshift, BQ, and Snowflake 
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions, and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

tatiana:
@utkarsharma2 It may be worth detailing a bit more:
> Currently, we are using pandas to create tables based on files passed. 

To something like:
> If the user does not specify the table columns property, we currently use pandas to autodetect the table column types automatically.

Is the acceptance criteria for this ticket to use native autodetection for all databases (Redshift, BQ and Snowflake)?

utkarsharma2:
@tatiana @feluelle This ticket is an alternative to pandas schema detection. But for a specific case. So the decision tree will be something like this:
```
user-provided schema?
Yes: Use is to create a table
No -
     - Are we using the native path?
          - Yes
              - does the destination database provide schema detection?
                  - Yes: Use native autodetect
                  - No: Use pandas
          - No: Use pandas
```

utkarsharma2:
@sunank200 and I consolidated the databases that support auto-detection: 

[...]

PR #780, opened 2022-09-06
Add native autodetect schema feature

feluelle:
# Description

Currently for inferring the schema we are always using pandas.

For more information, check the issue below..

closes: #709

## What is the new behavior?

Add native autodetect schema feature for snowflake

- add snowflake file format data class
- add option to use native autodetect schema for databases
- implement snowflake specific autodetect schema

Add native autodetect schema for bigquery

- implement bigquery native autodetect schema
- add tests for bigquery

No

### Checklist


.pre-commit-config.yaml
```diff
@@ -92,7 +92,7 @@ repos:
       rev: 'v0.971'
       hooks:
       - id: mypy
-        additional_dependencies: [types-PyYAML, types-attrs]
+        additional_dependencies: [types-PyYAML, types-attrs, attrs]
```
feluelle, line 95: @kaxil did pre-commit mypy work for you without adding that?

feluelle:
The failed checks are unrelated.

dimberman (approved):
This is super cool! Huge +1

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -333,6 +435,59 @@ def drop_stage(self, stage: SnowflakeStage) -> None:
     # Table load methods
     # ---------------------------------------------------------
 
+    def is_native_autodetect_schema_available(  # skipcq: PYL-R0201
```
utkarsharma2, line 438: @feluelle can we keep this method at the base database?

feluelle: It is, but without implementation as the implementation could slightly differ, I guess.

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -60,6 +60,9 @@
 }
 BIGQUERY_WRITE_DISPOSITION = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
 
+NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_TYPES = (FileType.CSV, FileType.NDJSON)
```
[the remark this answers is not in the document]
sunank200: Also isn't schema detection supported for parquet files in bigquery?

sunank200 (commented):
Overall looks good. Added few comments.

python-sdk/src/astro/databases/google/bigquery.py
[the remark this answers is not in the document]
feluelle: > Schema auto-detection is not used with Avro files, Parquet files, ORC files, Firestore export files, or Datastore export files. When you load these files into BigQuery, the table schema is automatically retrieved from the self-describing source data.
> 
> To use schema auto-detection when you load JSON or CSV data:
> ...
> 

from [here](https://cloud.google.com/bigquery/docs/schema-detect)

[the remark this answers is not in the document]
feluelle: > You don't need to enable schema auto-detection for Avro, Parquet, ORC, Firestore export, or Datastore export files. These file formats are self-describing, so BigQuery automatically infers the table schema from the source data. For Parquet, Avro, and Orc files, you can optionally provide an explicit schema to override the inferred schema.

I guess there is but the user cannot impact it.

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -60,6 +60,9 @@
...
+NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_LOCATIONS = (FileLocation.GS,)
```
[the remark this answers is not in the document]
feluelle: Did you test this? 😅 I have not tested this. If `FileLocation` is hashable it should work :)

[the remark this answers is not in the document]
feluelle: It works. Thanks @sunank200 👍

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -60,6 +60,9 @@
 }
 BIGQUERY_WRITE_DISPOSITION = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
 
+NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_TYPES = {FileType.CSV, FileType.NDJSON}
```
utkarsharma2, line 63: @feluelle bigquery has no support for parquet?

feluelle: See https://github.com/astronomer/astro-sdk/pull/780#discussion_r967004658

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -60,6 +60,9 @@
...
+NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_LOCATIONS = {FileLocation.GS}
```
utkarsharma2, line 64: @feluelle How would we represent cases like:
S3 - Parquet
GCS - CSV, JSON, Parquet

feluelle: > GCS - CSV, JSON, Parquet

`GS` is GCS; CSV & JSON is supported; For parquet the native inference within bigquery will be used automatically. See comment https://github.com/astronomer/astro-sdk/pull/780#discussion_r967004658

feluelle: > S3 - Parquet

I guess, there is not native path available which means it will use pandas instead.

utkarsharma2: So currently what this mapping implies is that for every supported location(GCS/S3) we support all the filetypes - CVS/JSON, right? but is it the case always? 

```
is_file_type_supported = (
    file.type.name in NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_TYPES
)
is_file_location_supported = (
    file.location.location_type
    in NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_LOCATIONS
)
```
Would be a good idea to check for location and then support fileType by that location? Something like - 
```
{
"GCS": [CSV, NDJSON],
"S3": [PARQUET]  # just an example. 
}
```

feluelle: but we do - in the `return` statement:
```python
return is_file_type_supported and is_file_location_supported
```

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -212,6 +215,56 @@ def merge_table(
...
+        )
+        return is_file_type_supported and is_file_location_supported
+
+    def create_table_using_native_schema_autodetection(
+        self,
+        table: Table,
+        file: File,
+    ) -> None:
+        """
+        Create a SQL table, automatically inferring the schema using the given file via native database support.
```
utkarsharma2, line 242: @feluelle I think this would populate the file in the table as well right?

If the user has given a pattern that results in two files

example - 
pattern : s3://tmp/

which results in files:
1. s3://tmp/test1.csv
2. s3://tmp/test2.csv

With the existing code of autodetecting schema, we will end up with:

Step 1: Autoschema detect -  Table + 1st file load
Step 2: Load data in a table - 1st file load + 2nd file load

Would it result in 1st file being loaded twice? Can we confirm if this is not the case?

cc: @sunank200

python-sdk/src/astro/databases/google/bigquery.py
feluelle: ah, okay I got what you mean now. Yes, this could make sense. I just used the same functionality we use for file loading. Is file loading different from schema autodetection? 🤔

python-sdk/src/astro/databases/google/bigquery.py
feluelle: > Step 1: Autoschema detect - Table + 1st file load

The autoschema detect should not load the file actually..

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -61,6 +61,80 @@
...
+
+    @staticmethod
+    def _create_unique_name() -> str:
+        """
+        Generate a valid Snowflake file format name.
+
+        :return: unique file format name
+        """
+        return (
+            "file_format_"
```
utkarsharma2, line 101: Looks like we are having a duplication of sorts, would it be a good idea to have a single function to generate a random string of the desired length and take prefix into account? 

https://github.com/astronomer/astro-sdk/blob/f1030fdeceed2d7de857d044c1e3773b05bf9ef3/python-sdk/src/astro/sql/table.py#L82

python-sdk/src/astro/databases/google/bigquery.py
feluelle: But good point, I will check that.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Or we can use this function in table.py

python-sdk/src/astro/databases/google/bigquery.py
feluelle: I tested it for snowflake by adding:
```python
    statement = f"SELECT COUNT(*) FROM {database.get_table_qualified_name(table)}"
    count = database.run_sql(statement).scalar()
    assert count == 0
```
which passed.

python-sdk/src/astro/databases/google/bigquery.py
utkarsharma2: @feluelle No, I guess not. We should change that as well. I just realized that we can do that in a separate PR.

python-sdk/src/astro/databases/snowflake.py
feluelle: I don't have a strong preference here. But if we want to make it more elegant and truly unique, I would prefer using [uuid](https://docs.python.org/3/library/uuid.html) (maybe uuid4) to create a unique name and store the function in a utils file? WDYT?

python-sdk/tests/databases/test_bigquery.py
```diff
@@ -131,6 +131,62 @@ def test_bigquery_create_table_with_columns(database_table_fixture):
...
+            "database": Database.BIGQUERY,
+            "table": Table(
+                metadata=Metadata(schema=SCHEMA),
+            ),
+        }
+    ],
+    indirect=True,
+    ids=["bigquery"],
+)
+def test_bigquery_create_table_using_native_schema_autodetection(
```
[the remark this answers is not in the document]
feluelle: I can do that, but if that is documented that it only creates the schema imo this is enough.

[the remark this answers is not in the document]
feluelle: bigquery indeed loads data into the table 🙄

Screenshot 2022-09-12 at 12 31 13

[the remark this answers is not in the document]
feluelle: It seems we cannot change that. So we have to `DELETE` the rows afterwards? 😅 Or wdyt?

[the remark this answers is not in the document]
feluelle: https://github.com/astronomer/astro-sdk/pull/780/commits/8530ffc61ea08d5c6bb88e0f6004c2061a9b9ae4 - Let me know what you think, please.

python-sdk/src/astro/databases/google/bigquery.py
utkarsharma2: @feluelle Nice, Can we add one for bigquery as well?

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: We can use it. But I guess we considered UUI for table names, but I cannot recall why we didn't. @tatiana might have better context there.

python-sdk/src/astro/databases/google/bigquery.py
feluelle: I have added it see https://github.com/astronomer/astro-sdk/pull/780#discussion_r968293149

python-sdk/src/astro/databases/google/bigquery.py
kaxil: @utkarsharma2 Can you create a separate issue on the separate PR you are talking about?

python-sdk/tests/databases/test_bigquery.py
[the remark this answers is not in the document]
utkarsharma2: Yes, we need to delete the rows, not ideal but we can improve on this later.

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -212,6 +215,60 @@ def merge_table(
...
+            "jobType": "LOAD",
+            "load": load_job_config,
+            "labels": {"target_table": table.name},
+        }
+
+        self.hook.insert_job(
+            configuration=job_config,
+        )
+
+        # We have to clear the table afterwards as bigquery automatically loads the data when creating the table.
```
kaxil, line 268: If so, are we loading the files twice?
1) Load entire file and then truncate the table
2) Load the files again

python-sdk/tests/databases/test_bigquery.py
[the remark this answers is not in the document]
kaxil: Does that mean file is loaded twice?

python-sdk/src/astro/databases/google/bigquery.py
feluelle: In this PR, yes, if we would not truncate the table. Bigquery does not let us create only.

kaxil:
I have merged this PR, but we should take care of the double-loading in the next PR

python-sdk/src/astro/databases/google/bigquery.py
sunank200: @kaxil I think @feluelle is truncating the table: https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/google/bigquery.py#L279

The test only needs fix

python-sdk/src/astro/databases/google/bigquery.py
sunank200: @kaxil, @utkarsharma2  I think @feluelle is truncating the table: https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/google/bigquery.py#L279

The test only needs fix

python-sdk/tests/databases/test_bigquery.py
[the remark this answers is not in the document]
sunank200: @kaxil, @utkarsharma2 I think @feluelle is truncating the table: https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/google/bigquery.py#L279

The test only needs fix.

python-sdk/src/astro/databases/google/bigquery.py
utkarsharma2: Added - https://github.com/astronomer/astro-sdk/issues/853

PR #786, opened 2022-09-06
Map `File` & `Table` classes to Dataset (AIP-48)

kaxil:
Part of https://github.com/astronomer/astro-sdk/issues/611

This PR will map `File` and `Table` classes to inherit from `Dataset` object (if it is available - will be released in Airflow 2.4). This will allow users to schedule DAGs on a `File` and `Table` object when used with Airflow 2.4 and above.

As a follow-up to this PR, I will create a PR to automatically add "inlets" and "outlets" to all of the operators in SDK so that users can see the datasets that are produced or consumed by tasks.

No

### Checklist
- [ ] Extended the README / documentation, if necessary -- will do that in a follow-up PR once the entire feature is added

### DIFF
--- a/python-sdk/docs/conf.py
+++ b/python-sdk/docs/conf.py
@@ -51,6 +51,10 @@ autoapi_options = [
     "show-inheritance",
     "show-module-summary",
 ]
+
+suppress_warnings = [
+    "autoapi.python_import_resolution",
+]
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_autoapi_templates"]
 
--- a/python-sdk/src/astro/airflow/datasets.py
+++ b/python-sdk/src/astro/airflow/datasets.py
@@ -1,0 +1,9 @@
+try:
+    # Airflow >= 2.4
+    from airflow.datasets import Dataset
+
+    DATASET_SUPPORT = True
+except ImportError:
+    # Airflow < 2.4
+    Dataset = object
+    DATASET_SUPPORT = False
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -135,9 +135,9 @@ class File(Dataset):
 
         parsed_url = urlparse(url=self.path)
         netloc = parsed_url.netloc
-        parsed_scheme = parsed_url.scheme
         # Local filepaths do not have scheme
-        scheme = f"astro+{parsed_scheme}" if parsed_scheme else "astro+file"
+        parsed_scheme = parsed_url.scheme or "file"
+        scheme = f"astro+{parsed_scheme}"
         extra = {}
         if self.filetype:
             extra["filetype"] = str(self.filetype)
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -4,16 +4,17 @@ import io
 
 import pandas as pd
 import smart_open
-from attrs import define
+from attr import define, field
 
 from astro import constants
+from astro.airflow.datasets import Dataset
 from astro.files.locations import create_file_location
 from astro.files.locations.base import BaseFileLocation
 from astro.files.types import FileType, create_file_type
 
 
 @define
-class File:
+class File(Dataset):
     """
     Handle all file operations, and abstract away the details related to location and file types.
     Intended to be used within library.
@@ -28,6 +29,9 @@ class File:
     conn_id: str | None = None
     filetype: constants.FileType | None = None
     normalize_config: dict | None = None
+
+    uri: str = field(init=False)
+    extra: dict | None = field(init=False, factory=dict)
 
     template_fields = (
         "path",
@@ -123,6 +127,27 @@ class File:
 
     def __hash__(self) -> int:
         return hash((self.path, self.conn_id, self.filetype))
+
+    @uri.default
+    def _path_to_dataset_uri(self) -> str:
+        """Build a URI to be passed to Dataset obj introduced in Airflow 2.4"""
+        from urllib.parse import urlencode, urlparse
+
+        parsed_url = urlparse(url=self.path)
+        netloc = parsed_url.netloc
+        parsed_scheme = parsed_url.scheme
+        # Local filepaths do not have scheme
+        scheme = f"astro+{parsed_scheme}" if parsed_scheme else "astro+file"
+        extra = {}
+        if self.filetype:
+            extra["filetype"] = str(self.filetype)
+
+        new_parsed_url = parsed_url._replace(
+            netloc=f"{self.conn_id}@{netloc}" if self.conn_id else netloc,
+            scheme=scheme,
+            query=urlencode(extra),
+        )
+        return new_parsed_url.geturl()
 
 
 def resolve_file_path_pattern(
--- a/python-sdk/src/astro/sql/table.py
+++ b/python-sdk/src/astro/sql/table.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 import random
 import string
 
-from attrs import define, field, fields_dict
+from attr import define, field, fields_dict
 from sqlalchemy import Column, MetaData
+
+from astro.airflow.datasets import Dataset
 
 MAX_TABLE_NAME_LENGTH = 62
 TEMP_PREFIX = "_tmp_"
@@ -33,7 +35,7 @@ class Metadata:
 
 
 @define
-class Table:
+class Table(Dataset):
     """
     Withholds the information necessary to access a SQL Table.
     It is agnostic to the database type.
@@ -61,6 +63,9 @@ class Table:
     )
     columns: list[Column] = field(factory=list)
     temp: bool = field(default=False)
+
+    uri: str = field(init=False)
+    extra: dict | None = field(init=False, factory=dict)
 
     def __attrs_post_init__(self) -> None:
         if not self._name or self._name.startswith("_tmp"):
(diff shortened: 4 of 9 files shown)


kaxil:
Current test failures are because we exceeded rate limit, will re-run in few hours:

>pandas_gbq.exceptions.GenericGBQException: Reason: 403 Exceeded rate limits: too many table update operations for this table. For more information, see https://cloud.google.com/bigquery/docs/troubleshoot-quotas

utkarsharma2:
@kaxil I'm seeing the below issue in the main branch, I think is related to `attrs` dep we added as part of https://github.com/astronomer/astro-sdk/issues/611 ticket 

Screenshot 2022-09-07 at 3 24 54 AM
Screenshot 2022-09-07 at 3 24 18 AM

utkarsharma2:
@kaxil I think we need to pin attrs to the correct version

Screenshot 2022-09-07 at 3 37 52 AM

kaxil:
> @kaxil I think we need to pin attrs to the correct version
> 
> Screenshot 2022-09-07 at 3 37 52 AM

aah yes, looks like Airflow has a limit set for 2.3.3 - https://github.com/apache/airflow/blob/2.3.3/setup.cfg#L88-L89 but not for 2.3.4.

If you upgrade to Airflow 2.3.4 it will work -- but yeah I will change it to using `attr` -- not a pblm

kaxil:
@utkarsharma2 Updated in https://github.com/astronomer/astro-sdk/pull/786/commits/ee1389fbe3ec0f577713c923d36c4b8a90cff8d0

python-sdk/src/astro/sql/table.py
```diff
@@ -112,3 +117,19 @@ def name(self, value: str) -> None:
         if not isinstance(value, property) and value != self._name:
             self._name = value
             self.temp = False
+
+    @uri.default
+    def _path_to_dataset_uri(self) -> str:
+        """Build a URI to be passed to Dataset obj introduced in Airflow 2.4"""
+        from urllib.parse import urlencode, urlparse
+
+        path = f"astro://{self.conn_id}@"
```
utkarsharma2, line 126:
```suggestion
        path = f"astro+Table://{self.conn_id}@"
```

utkarsharma2: @kaxil Would it be a good idea to add a Table or File in the URL? would be more explicit and will give a nice namespace to Table and File objects. Also, prevent any collisions among File URI and Table URI or any other data placeholder we may add.

utkarsharma2: If we agree we can do this for Files as well.

utkarsharma2:
@kaxil Assuming this is the last PR for #611. Should we add a test having dags with inlets and outlets?

kaxil:
> @kaxil Assuming this is the last PR for #611. Should we add a test having dags with inlets and outlets? which can run for airflow 2.4 >=

Not yet, check the PR description - "As a follow-up to this PR, I will create a PR to automatically add "inlets" and "outlets" to all of the operators in SDK so that users can see the datasets that are produced or consumed by tasks."

python-sdk/src/astro/sql/table.py
kaxil: That doesn't give any new info, "scheme" for files will be `astro+gs` or `astro+s3` so just change it to files doesn't make sense in my opinion

tatiana: @kaxil I believe that we should not add "Table" in the URI prefix since it does not give much information and is not consistent with how we intend to represent the `File` URI prefix (`astro+gs` or `astro+s3`, as you described).

It feels it would be more symmetric if we added to the URI prefix the prefix used to access the dataset in question. For databases, this could be:
* PostgresSQL: astro+[postgresql](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
* SQLite: "astro+[file](https://www.sqlite.org/uri.html)"

Another alternative could be to use only `astro` as the URI prefix - since the remaining information is part of `conn_id`.

I also believe we should update the docs with the logic used to represent datasets.

python-sdk/src/astro/files/base.py
```diff
@@ -124,6 +128,27 @@ def __eq__(self, other) -> bool:
...
+    @uri.default
+    def _path_to_dataset_uri(self) -> str:
+        """Build a URI to be passed to Dataset obj introduced in Airflow 2.4"""
+        from urllib.parse import urlencode, urlparse
+
+        parsed_url = urlparse(url=self.path)
+        netloc = parsed_url.netloc
+        parsed_scheme = parsed_url.scheme
+        # Local filepaths do not have scheme
+        scheme = f"astro+{parsed_scheme}" if parsed_scheme else "astro+local"
```
tatiana, line 140:
```suggestion
        scheme = f"astro+{parsed_scheme}" if parsed_scheme else "astro+file"
```
Shouldn't we use `file` for local files, following the URI convention?
https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -0,0 +1,9 @@
+try:
```
tatiana, line 1: Are we planning to introduce more things inside the package `airflow` inside Astro?

Since this is a configuration `DATASET_SUPPORT`), to some extent, have you considered adding this to `settings.py` or `constants.py` (`AIRFLOW_DATASET_SUPPORT`)?

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -0,0 +1,9 @@
...
+    # Airflow >= 2.4
+    from airflow.datasets import Dataset
+
+    DATASET_SUPPORT = True
```
tatiana, line 5: It is probably worth to run a test in Nox, so we cover this branch

python-sdk/src/astro/files/base.py
tatiana: Also, what are we gaining by not representing files using their original URIs, since they are already unique?

python-sdk/docs/conf.py
```diff
@@ -51,6 +51,10 @@
     "show-inheritance",
     "show-module-summary",
 ]
+
+suppress_warnings = [
+    "autoapi.python_import_resolution",
```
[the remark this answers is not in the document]
kaxil: Yes, I should have added a comment -- this is because `airflow.dataset` is still not available for Airflow -- so Sphinx fails to resolve it -- which is why I am suppressing that error.

python-sdk/src/astro/airflow/datasets.py
kaxil: What sort of test are you thinking about for it?

python-sdk/src/astro/airflow/datasets.py
kaxil: Yeah I am thinking of adding `kwargs_with_datasets` --> same as https://github.com/astronomer/astro-sdk/blob/aip-48-mapping/src/astro/utils/compat.py

kaxil:
>PostgresSQL: astro+[postgresql](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
SQLite: "astro+[file](https://www.sqlite.org/uri.html)"

@tatiana To identify the type of connection from `conn_id`, we need to make a DB call to Airflow metadata DB --- this is done to avoid it and so that users see the a similar thing to what they pass to Table or File object -- as below

https://github.com/astronomer/astro-sdk/blob/f8cb25963819ede051d4cc663e0df713f98400d4/python-sdk/tests/sql/test_table.py#L78-L107

python-sdk/src/astro/files/base.py
```diff
@@ -4,16 +4,17 @@
 
 import pandas as pd
 import smart_open
-from attrs import define
+from attr import define, field
```
feluelle, line 7: It changed from `attrs` to `attr` or is this a typo? 🤔

python-sdk/src/astro/files/base.py
kaxil: Done, changed in https://github.com/astronomer/astro-sdk/pull/786/commits/a783557f7c3bb532a1852d3a713955a5981dd6d3

python-sdk/src/astro/sql/table.py
```diff
@@ -120,3 +125,19 @@ def name(self, value: str) -> None:
...
+        db_extra = {}
+        if self.metadata.schema:
+            db_extra["schema"] = self.metadata.schema
+        if self.metadata.database:
+            db_extra["database"] = self.metadata.database
+        db_extra["table"] = self.name
```
feluelle, line 140:
```suggestion
        db_extra = {"table": self.name}
        if self.metadata.schema:
            db_extra["schema"] = self.metadata.schema
        if self.metadata.database:
            db_extra["database"] = self.metadata.database
```

python-sdk/src/astro/files/base.py
```diff
@@ -124,6 +128,27 @@ def __eq__(self, other) -> bool:
...
+    @uri.default
+    def _path_to_dataset_uri(self) -> str:
+        """Build a URI to be passed to Dataset obj introduced in Airflow 2.4"""
+        from urllib.parse import urlencode, urlparse
+
+        parsed_url = urlparse(url=self.path)
+        netloc = parsed_url.netloc
+        parsed_scheme = parsed_url.scheme
+        # Local filepaths do not have scheme
+        scheme = f"astro+{parsed_scheme}" if parsed_scheme else "astro+file"
```
feluelle, line 140:
```suggestion
        parsed_scheme = parsed_url.scheme or "file"
        scheme = f"astro+{parsed_scheme}"
```

python-sdk/src/astro/sql/table.py
utkarsharma2: @kaxil I'm not suggesting a replacement of `gs/s3` instead `file+file+gs` or `file+file+s3` that was with string we can know that it's coming from a file object or table. Just a suggestion to be a little verbose.

sunank200: But @utkarsharma2 doesn't `gs` or `s3` suggest the file itself?

utkarsharma2 (approved):

python-sdk/src/astro/sql/table.py
kaxil: That starts looking too verbose in the UI once you have many datasets:

image

utkarsharma2: oh, I didn't think of that.

sunank200 (approved):

Issue #784, opened 2022-09-06
Add param docstrings for Table object


kaxil:
Our Table object doesn't contain docstring for `params` similar to our `File` object.

https://github.com/astronomer/astro-sdk/blob/d83499cc390493e5c95effaeb026ed3158de577c/python-sdk/src/astro/sql/table.py#L32-L40

https://github.com/astronomer/astro-sdk/blob/d83499cc390493e5c95effaeb026ed3158de577c/python-sdk/src/astro/files/base.py#L15-L25

PR #788, opened 2022-09-07
Add missing docs for Table class

pankajastro:
# Description
closes: #784 

## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/table.py
+++ b/python-sdk/src/astro/sql/table.py
@@ -44,9 +44,7 @@ class Table:
     :param conn_id: The Airflow connection id. This will be used to identify the right database type at the runtime
     :param name: The name of the database table. If name not provided then it would create a temporary name
     :param metadata: A metadata object which will have database or schema name
-    :param columns: Table columns names.
-        If columns values provided then it would create table having provided columns names
-        otherwise it autodetect columns name from dataset
+    :param columns: columns which define the database table schema.
     """
 
     template_fields = ("name",)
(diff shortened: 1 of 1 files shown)


python-sdk/src/astro/sql/table.py
```diff
@@ -37,6 +40,13 @@ class Table:
     If no name is given, it auto-generates a name for the Table and considers it temporary.
 
     Temporary tables are prefixed with the prefix TEMP_PREFIX.
+
+    :param conn_id: The Airflow connection id. This will be used to identify the right database type at the runtime
+    :param name: The name of the database table. If name not provided then it would create a temporary name
+    :param metadata: A metadata object which will have database or schema name
+    :param columns: Table columns names.
```
tatiana, line 47:
```suggestion
    :param columns: columns which define the database table schema.
```

python-sdk/src/astro/sql/table.py
```diff
@@ -37,6 +40,13 @@ class Table:
...
+        If columns values provided then it would create table having provided columns names
+        otherwise it autodetect columns name from dataset
```
tatiana, line 49:
```suggestion
```
I believe this information shouldn't be here since to use columns to create a table is a behaviour of an operator (`load_file`) and not of the `table` definition itself. In other operators `columns` is not used to modify the table schema (e.g. `append` and `merge`).

phanikumv (approved):

tatiana (approved):

PR #797, opened 2022-09-08
Add testcase to verify table creation in load_file for append/replace

utkarsharma2:
# Description
Currently, we don't have a testcase to verify the behavior of load_file() operator where if the user passes `if_exists='append'` or` if_exists='replace'` and the table doesn't exist, the table is created.

## What is the new behavior?
Add a test case to verify this behavior

Nope


python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -1048,3 +1048,28 @@ def test_aql_load_column_name_mixed_case_json_file_to_dbs(
...
+    ],
+    indirect=True,
+    ids=["Sqlite"],
+)
+@pytest.mark.parametrize("if_exists", ["replace", "append"])
+def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
+    """
+    Verify creation of new tables in case we pass if_exists=replace/append if they don't exists.
+    """
+    path = str(CWD) + "/../../data/homes_main.csv"
```
phanikumv, line 1068: We should have a constant which refers to `"/../../data/homes_main.csv"` , I see in multiple places we are referring to this path.

python-sdk/src/astro/files/base.py
```diff
@@ -145,7 +145,7 @@ def resolve_file_path_pattern(
     location = create_file_location(path_pattern, conn_id)
 
     files = [
-        File(
+        File(  # type: ignore
```
kaxil, line 148: hmm why?

python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -1048,3 +1048,28 @@ def test_aql_load_column_name_mixed_case_json_file_to_dbs(
...
+    [
+        {
+            "database": Database.SQLITE,
+        },
+    ],
+    indirect=True,
+    ids=["Sqlite"],
+)
+@pytest.mark.parametrize("if_exists", ["replace", "append"])
+def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
```
kaxil, line 1064: Can we not mock this instead of an integration test?

python-sdk/src/astro/files/base.py
sunank200: Removed this https://github.com/astronomer/astro-sdk/pull/797/commits/cedb4a4c681d9fd7643eb5352b4543d16cd3a255

python-sdk/src/astro/files/operators/files.py
```diff
@@ -32,7 +32,7 @@ def execute(self, context: Context) -> Any:  # skipcq: PYL-W0613
         location = create_file_location(self.path, self.conn_id)
         # Get list of files excluding folders
         return [
-            File(path=path, conn_id=location.conn_id)
+            File(path=path, conn_id=location.conn_id)  # type: ignore
```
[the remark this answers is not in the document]
sunank200: Removed this https://github.com/astronomer/astro-sdk/pull/797/commits/cedb4a4c681d9fd7643eb5352b4543d16cd3a255

python-sdk/tests/sql/operators/test_load_file.py
sunank200: This is quite contextual. Other places where this file is used is either a test for HTTP, s3 or GCS as file location.

tatiana (approved):
@sunank200 @utkarsharma2 Since this is covering a feature that is not covered elsewhere, I think it is still valuable for us to merge this change - once it is rebased - and we can improve and make our tests more consistent over time.

Issue #708, opened 2022-08-21
Use native paths pattern support


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, in `load_file()` the algorithm to resolve patterns in paths

1. Resolve patterns in the paths, get all the files that fit the pattern
2. iterate over each file from the file list
3. Perform transfer - native/dataframe

Ref: https://github.com/astronomer/astro-sdk/blob/f00c4b5386b825c143b52dd1ea6ab002ddb6a6c0/src/astro/databases/base.py#L322

**Describe the solution you'd like**
many native transfers support the patterns - it is be better to use the native support to load multiple files as opposed above algo.

1. Explore Databases for pattern support
2. Check if the file path passed has a pattern
3. if pattern is passed and database support pattern, use native support for pattern

**Acceptance Criteria**
- [x] All checks and tests in the CI should pass
- [x] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [x] Integration tests (if the feature relates to a new database or external service)
- [x] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [x] Exception handling in case of errors
- [x] Logging (are we exposing useful information to the user? e.g. source and destination)
- [x] Improve the documentation (README, Sphinx, and any other relevant)
- [x] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

pankajkoti:
@sunank200 worked on GCS for this and PR will be ready today.

How to resolve schema is a question though. Need to check if we have to create a table for certain databases & also need to check if schema auto-detection works for all supported file types & for file patterns approach.

utkarsharma2:
@sunank200 can you add points from yesterday's discussion here?

sunank200:
<img width="989" alt="Screenshot 2022-09-09 at 5 28 52 PM" src="https://user-images.githubusercontent.com/8670962/189342655-89d5017b-d0c2-4ce1-b9e1-aec4697e3eb1.png">
This is the flow to be checked.

pankajkoti:
[...]

PR #802, opened 2022-09-08
Directly pass the file to native library when native support is enabled

sunank200:
# Description

Currently, in load_file() the algorithm to resolve patterns in paths

- Resolve patterns in the paths, get all the files that fit the pattern
- Iterates over each file from the file list
- Perform transfer - native/dataframe

closes: #708 

## What is the new behaviour?

- Check if autodetection of schema is supported if yes directly pass the file to the library
- If autodetection is not supported, detect and create the table, then pass the path to the library
- Screenshot 2022-09-09 at 5 28 52 PM
This is the flow to be checked.

No

### Checklist


python-sdk/src/astro/databases/base.py
```diff
@@ -276,6 +276,61 @@ def drop_table(self, table: Table) -> None:
     # Table load methods
     # ---------------------------------------------------------
 
+    def create_schema_table_if_needed(
```
utkarsharma2, line 279: Would renaming this to `create_schema_and_table_if_needed` make it more explicit?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -73,6 +73,15 @@ class BigqueryDatabase(BaseDatabase):
         FileLocation.S3: "load_s3_file_to_table",
         FileLocation.LOCAL: "load_local_file_to_table",
     }
+    AUTODETECT_SCHEMA_SUPPORTED: set[FileLocation] = {
```
utkarsharma2, line 76: @sunank200  I think there is some conflict between this implementation and https://github.com/astronomer/astro-sdk/pull/780/files for the bigquery. Can we check this?

python-sdk/src/astro/files/base.py
```diff
@@ -69,6 +70,21 @@ def is_binary(self) -> bool:
         result: bool = self.type.name == constants.FileType.PARQUET
         return result
 
+    def is_pattern(self) -> bool:
+        """
```
utkarsharma2, line 74: @sunank200 Can we add a test case for this?

python-sdk/src/astro/databases/base.py
```diff
@@ -305,50 +360,65 @@ def load_file_to_table(
...
+                target_table=output_table,
+                if_exists="append",
+                normalize_config=normalize_config,
+                native_support_kwargs=native_support_kwargs,
+                enable_native_fallback=enable_native_fallback,
+                chunk_size=chunk_size,
+            )
+        else:
+            self.load_file_to_table_using_pandas(
+                input_file=input_file,
```
utkarsharma2, line 386: @sunank200 What will happen in pandas path? I'm not seeing the entire code here but all I see it the input_fiel getting passed to `load_file_to_table_using_pandas` will this work?

utkarsharma2: never mind got the implementation and you are resolving the path there.

python-sdk/src/astro/databases/base.py
```diff
@@ -305,50 +360,65 @@ def load_file_to_table(
...
-                    chunk_size=chunk_size,
-                )
-            else:
-                self.load_pandas_dataframe_to_table(
-                    file.export_to_dataframe(),
-                    output_table,
-                    chunk_size=chunk_size,
-                    if_exists="append",  # We've already created a new table in this case
-                )
+            self.load_pandas_dataframe_to_table(
```
utkarsharma2, line 409: @sunank200 What will happen when don't get results after pattern resolution? Do we raise an error or not? If not, I think we should. WDYT?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -225,6 +234,32 @@ def is_native_load_file_available(
         location_type = self.NATIVE_PATHS.get(source_file.location.location_type)
         return bool(location_type and file_type)
 
+    def check_schema_autodetection_is_supported(self, source_file: File) -> bool:
+        """
```
utkarsharma2, line 238: @sunank200 Can we push these method in database/base.py and just populate `AUTODETECT_SCHEMA_SUPPORTED` and `FILE_PATTERN_BASED_AUTODETECT_SCHEMA_SUPPORTED` in specific databases? 
in databse/base.py can be empty - 
AUTODETECT_SCHEMA_SUPPORTED  = []
FILE_PATTERN_BASED_AUTODETECT_SCHEMA_SUPPORTED = []

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -177,6 +177,9 @@ class SnowflakeDatabase(BaseDatabase):
         RequestTimeoutError,
     )
     DEFAULT_SCHEMA = SNOWFLAKE_SCHEMA
+    AUTODETECT_SCHEMA_SUPPORTED: set[Any] = set()
```
utkarsharma2, line 180: I like the use of the set, nice! :)

python-sdk/src/astro/files/base.py
utkarsharma2: Also, if we can validate that this works with the glob pattern would be nice, since we will be working on that going forward.

python-sdk/src/astro/files/base.py
```diff
@@ -128,7 +144,7 @@ def __eq__(self, other) -> bool:
     def __hash__(self) -> int:
         return hash((self.path, self.conn_id, self.filetype))
 
-    @uri.default
+    @uri.default  # type: ignore
```
utkarsharma2, line 147: @sunank200 is this still happening on local? have you tried rebasing from main?

python-sdk/src/astro/databases/base.py
```diff
@@ -276,6 +276,61 @@ def drop_table(self, table: Table) -> None:
...
+            and is_file_pattern_based_schema_autodetection_supported
+        ):
+            return
+
+        self.create_schema_if_needed(table.metadata.schema)
+        if if_exists == "replace" or not self.table_exists(table):
+            self.drop_table(table)
+            self.create_table(
+                table,
+                resolve_file_path_pattern(
```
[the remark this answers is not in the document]
sunank200: @utkarsharma2 result wont be empty as it will raise value error in https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/files/base.py#L155

python-sdk/src/astro/databases/google/bigquery.py
sunank200: I am have merged that code to my branch and its running fine.

python-sdk/src/astro/databases/base.py
```diff
@@ -51,6 +51,8 @@ class BaseDatabase(ABC):
     illegal_column_name_chars_replacement: list[str] = []
     NATIVE_LOAD_EXCEPTIONS: Any = DatabaseCustomError
     DEFAULT_SCHEMA = SCHEMA
+    AUTODETECT_SCHEMA_SUPPORTED: set[Any] = set()
```
utkarsharma2, line 54: @sunank200 I think it would be helpful if we add a comment about the values we expect in set.

utkarsharma2: Also, is this ticket applies here too? 

https://github.com/astronomer/astro-sdk/issues/853

python-sdk/src/astro/databases/base.py
```diff
@@ -318,6 +321,62 @@ def drop_table(self, table: Table) -> None:
...
+            self.create_table(
+                table,
+                resolve_file_path_pattern(
+                    file.path,
+                    file.conn_id,
+                    normalize_config=normalize_config,
+                    filetype=file.type.name,
+                )[0],
+                columns_names_capitalization=columns_names_capitalization,
+            )
```
feluelle, line 378:
```suggestion
            files = resolve_file_path_pattern(
                file.path,
                file.conn_id,
                normalize_config=normalize_config,
                filetype=file.type.name,
            )
            self.create_table(
                table,
                # We only use the first file for inferring the table schema
                files[0],
                columns_names_capitalization=columns_names_capitalization,
            )
```
Let us make it a bit clearer that we only use the first file for the table schema.

python-sdk/src/astro/databases/base.py
```diff
@@ -318,6 +321,64 @@ def drop_table(self, table: Table) -> None:
...
+        is_schema_autodetection_supported = (
+            self.check_schema_autodetection_is_supported(source_file=file)
+        )
+        is_file_pattern_based_schema_autodetection_supported = (
+            self.check_file_pattern_based_schema_autodetection_is_supported(
+                source_file=file
+            )
+        )
+        if if_exists == "replace":
+            self.drop_table(table)
```
feluelle, line 351: This is a duplicate. We have that already on line 366/367.
```
        if if_exists == "replace" or not self.table_exists(table):
            self.drop_table(table)
```

utkarsharma2: @sunank200 Why do we have this? isn't this duplicate?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -75,6 +75,15 @@ class BigqueryDatabase(BaseDatabase):
         FileLocation.S3: "load_s3_file_to_table",
         FileLocation.LOCAL: "load_local_file_to_table",
     }
+    AUTODETECT_SCHEMA_SUPPORTED: set[FileLocation] = {
+        FileLocation.GS,
+        FileLocation.LOCAL,
```
utkarsharma2, line 80: @sunank200 can we change the Mapping to something like this - https://github.com/astronomer/astro-sdk/blob/ddf7f12b723595d88991d286fc39612cec6de455/python-sdk/src/astro/databases/google/bigquery.py#L62

utkarsharma2: Isn't this dependent on filetypes as well?

utkarsharma2:
@sunank200 Would it be a good idea to have a test case resolving the pattern and autodetection schema? WDYT?

sunank200:
> @sunank200 Would it be a good idea to have a test case resolving the pattern and autodetection schema? WDYT?

There is already test case for this: https://github.com/astronomer/astro-sdk/blob/use-native-pattern-support/python-sdk/example_dags/example_load_file.py#L147

feluelle (approved):
LGTM

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -82,6 +73,24 @@ class BigqueryDatabase(BaseDatabase):
...
     }
 
+    NATIVE_AUTODETECT_SCHEMA_CONFIG: Mapping[
+        FileLocation, Mapping[str, list[FileType] | Callable]
+    ] = {
+        FileLocation.GS: {
+            "filetype": [FileType.CSV, FileType.NDJSON, FileType.PARQUET],
+            "method": lambda table, file: None,
+        },
+        FileLocation.LOCAL: {
```
utkarsharma2, line 83: @sunank200 Please add a test case for the local path being a pattern.

utkarsharma2 (approved):

Issue #803, opened 2022-09-08
Avoid loading whole file into memory with load_operator while schema detection


pankajkoti:
I believe we're loading whole file into memory and converting it into a byte stream while schema auto-detection.
As a result, jobs with large files e.g. 5GB and higher are getting killed.
Check if we really need to load the whole file into memory and if not needed just return the stream so that pandas dataframe can fetch only the initial rows of a file for schema autodetection.

pankajkoti:
We have an approval for the approach from Kaxil. Will need to address few comments for Felix and merge the PR.

PR #804, opened 2022-09-08
Avoid loading whole file into memory with load_operator for schema detection

pankajkoti:
# Description
We're loading whole file into memory and converting it into a byte stream
for schema auto-detection. As a result, jobs with large files e.g. 5GB and
higher are getting killed.
Opening this for discussion if we really need to load the whole file into
memory and if not needed just return the stream so that pandas dataframe
can fetch only the initial rows of a file for schema auto-detection.

closes: #803 

## What is the new behavior?

Do not load whole file into memory but only return the stream.

No, tested with postgres and Redshift.


pankajkoti:
Closing, cut from wrong branch. Will create another PR.

Issue #748, opened 2022-08-25
Add Benchmarking data for REDSHIFT


utkarsharma2:
**Please describe the feature you'd like to see**
Right now we don't benchmark data for the redshift database.

**Describe the solution you'd like**
We need to run benchmarking script on the redshift database.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

pankajkoti:
The benchmarking script is failing due to multiple failures. I see it has mostly failed in the last few days on the Kubernetes cluster. 
I tried making changes to the script, was able to make a few fixes related to path caused after the folder restructuring.
The image is getting built well now locally. Upon running locally, it starts but gets killed after a while. Also, it fails to launch on Kubernetes cluster. Will seek help from @utkarsharma2 & @tatiana

pankajkoti:
Will pair up with @sunank200 post stand up call to resolve this.

PR #805, opened 2022-09-08
Avoid loading whole file into memory with load_operator for schema detection

pankajkoti:
# Description
## What is the current behavior?

We're loading whole file into memory and converting it into a byte stream for schema auto-detection. As a result, jobs with large files e.g. 5GB and higher are getting killed.
Opening this for discussion if we really need to load the whole file into memory and if not needed just return the stream so that pandas dataframe can fetch only the initial rows of a file for schema autodetection.

closes: #803 
closes: #826 
## What is the new behavior?

Do not load the whole file into memory but just return the smart_open stream to be used by the dataframe.
## Does this introduce a breaking change?

No. Tested with postgres & Redshift.

PR #806, opened 2022-09-08
Publish Redshift benchmark results

pankajkoti:
Publish Redshift benchmark results for native and default approach.

The existing dataset failed with schema mismatch error while inserting rows.
The pandas auto-detection created schema with columns as `varchar(256)` limiting
the values to be 256 bytes long. However, some row contained a value larger than 256
bytes and then it complained with `value too long` error.
Hence, we have created new fake data set and kept in S3 with data
of required various sizes for benchmarking purposes and have also updated
the `datasets.md` to provide details of this new files.

blocked by: https://github.com/astronomer/astro-sdk/pull/805
closes: #748


python-sdk/tests/benchmark/results.md
```diff
@@ -248,3 +248,54 @@ Note - These results are generated manually, there is a issue added for the same
...
+  * Memory: 10 Gi
+
+| database   | dataset    | total_time   | memory_rss   | cpu_time_user   | cpu_time_system|
+|:-----------|:-----------|:-------------|:-------------|:----------------|:---------------|
+| redshift   | ten_kb     | 21.1s        | 59.57MB      | 910.0ms         | 100.0ms        |
+| redshift   | hundred_kb | 1.15min      | 59.22MB      | 1.16s           | 180.0ms        |
+| redshift   | one_mb     | 11.01min     | 62.89MB      | 3.46s           | 810.0ms        |
+
+The `pandas.to_sql` function uses highly suboptimal INSERT statement for Redshift. Hence, you're not recommended
+to use this approach for large files, but instead use the native approach.
```
feluelle, line 275: Let's have a look at the current pandas implementation tomorrow and how we can improve it with batch `INSERT`ion, before publishing them. WDYT?

sunank200 (approved):

python-sdk/tests/benchmark/results.md
pankajkoti: The release is to be cut tonight. I will try the approach that you suggested later tonight, but it looks a bit unclean to me. We can also connect tomorrow. Just wanted to draft the findings as per the current implementation if a beta has to be cut today.

dimberman:
@pankajkoti why did we have to use a fake dataset for redshift?

pankajkoti:
> @pankajkoti why did we have to use a fake dataset for redshift? Can you please add this to the description?

@dimberman That is already added to the description.

pankajkoti:
> @pankajkoti why did we have to use a fake dataset for redshift? Can you please add this to the description?

@dimberman Updated the description. Please check

utkarsharma2:
@pankajkoti Let's not merge this PR until we resolve - https://github.com/astronomer/astro-sdk/pull/805

utkarsharma2 (approved):

PR #807, opened 2022-09-08
Fix sql injection issues

feluelle:
# Description

Currently, we have some sql injection issues in our code which DeepSource detects for us, but some of them we ignored.

## What is the new behavior?

This PR should fix them.

No.

### Checklist


python-sdk/src/astro/databases/base.py
```diff
@@ -490,15 +490,13 @@ def export_table_to_pandas_dataframe(self, source_table: Table) -> pd.DataFrame:
 
         :param source_table: An existing table in the database
         """
-        table_qualified_name = self.get_table_qualified_name(source_table)
         if self.table_exists(source_table):
-            return pd.read_sql(
-                # We are avoiding SQL injection by confirming the table exists before this statement
-                f"SELECT * FROM {table_qualified_name}",  # skipcq BAN-B608
```
dimberman, line 497: Thank you for catching this.

kaxil (approved):

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -197,20 +214,31 @@ def merge_table(
...
+            if not self.columns_exist(source_table, source_columns):
+                raise ValueError(
+                    f"Not all the columns provided exist for {source_table_name}!"
+                )
+            if not self.columns_exist(target_table, target_columns):
+                raise ValueError(
+                    f"Not all the columns provided exist for {target_table_name}!"
+                )
+            # Note: Ignoring below sql injection warning, as we validate that the table columns exist beforehand.
+            update_statement = f"UPDATE SET {update_statement_map}"  # skipcq BAN-B608
```
feluelle, line 239: @kaxil wdyt of this? Is this enough for sql injection validation?

Unfortunately, I could not find a better way of handling the sql injection issue reported by DeepSource. Formatting the query (via parameters) did not work.

kaxil: Yeah fine enough. Most of it will be coming from DAGs rather than an API -- so we should be good

PR #808, opened 2022-09-08
Set inlets and outlets for operators

kaxil:
This PR adds inlets and outlets to the operators where it is trivial and clear on what the input and output datasets are.

If users pass their inlets and outlets to the operator, SDK will use those

part of: https://github.com/astronomer/astro-sdk/issues/611

image

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/airflow/datasets.py
+++ b/python-sdk/src/astro/airflow/datasets.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 try:
     # Airflow >= 2.4
     from airflow.datasets import Dataset
@@ -7,3 +9,26 @@ except ImportError:
     # Airflow < 2.4
     Dataset = object
     DATASET_SUPPORT = False
+
+
+def kwargs_with_datasets(
+    kwargs: dict | None = None,
+    input_datasets: list[Dataset] | Dataset | None = None,
+    output_datasets: list[Dataset] | Dataset | None = None,
+):
+    """
+    Extract inlets and outlets from kwargs if users have passed it. If not, set input datasets as inlets and
+    set output dataset as outlets
+    """
+    kwargs = kwargs or {}
+    if "inlets" in kwargs or DATASET_SUPPORT and input_datasets:
+        inlets = kwargs.get("inlets", input_datasets)
+        inlets = inlets if isinstance(inlets, list) else [inlets]
+        kwargs.update({"inlets": inlets})
+
+    if "outlets" in kwargs or DATASET_SUPPORT and output_datasets:
+        outlets = kwargs.get("outlets", output_datasets)
+        outlets = outlets if isinstance(outlets, list) else [outlets]
+        kwargs.update({"outlets": outlets})
+
+    return kwargs
--- a/python-sdk/src/astro/sql/operators/append.py
+++ b/python-sdk/src/astro/sql/operators/append.py
@@ -5,6 +5,7 @@ from typing import Any
 from airflow.decorators.base import get_unique_task_id
 from airflow.models.xcom_arg import XComArg
 
+from astro.airflow.datasets import kwargs_with_datasets
 from astro.databases import create_database
 from astro.sql.operators.base_operator import AstroSQLBaseOperator
 from astro.sql.table import Table
@@ -43,7 +44,12 @@ class AppendOperator(AstroSQLBaseOperator):
         self.columns = columns or {}
         task_id = task_id or get_unique_task_id("append_table")
 
-        super().__init__(task_id=task_id, **kwargs)
+        super().__init__(
+            task_id=task_id,
+            **kwargs_with_datasets(
+                kwargs=kwargs, input_datasets=source_table, output_datasets=target_table
+            ),
+        )
 
     def execute(self, context: dict) -> Table:  # skipcq: PYL-W0613
         db = create_database(self.target_table.conn_id)
--- a/python-sdk/src/astro/sql/operators/merge.py
+++ b/python-sdk/src/astro/sql/operators/merge.py
@@ -5,6 +5,7 @@ from typing import Any
 from airflow.decorators.base import get_unique_task_id
 from airflow.models.xcom_arg import XComArg
 
+from astro.airflow.datasets import kwargs_with_datasets
 from astro.constants import MergeConflictStrategy
 from astro.databases import create_database
 from astro.sql.operators.base_operator import AstroSQLBaseOperator
@@ -50,8 +51,14 @@ class MergeOperator(AstroSQLBaseOperator):
         self.columns = columns or {}
         self.if_conflicts = if_conflicts
         task_id = task_id or get_unique_task_id("merge")
-
-        super().__init__(task_id=task_id, **kwargs)
+        super().__init__(
+            task_id=task_id,
+            **kwargs_with_datasets(
+                kwargs=kwargs,
+                input_datasets=source_table,
+                output_datasets=target_table,
+            ),
+        )
 
     def execute(self, context: dict) -> Table:
         db = create_database(self.target_table.conn_id)
--- a/python-sdk/tests/airflow/test_datasets.py
+++ b/python-sdk/tests/airflow/test_datasets.py
@@ -1,0 +1,79 @@
+from unittest import mock
+
+import pytest
+
+from astro.airflow.datasets import kwargs_with_datasets
+from astro.sql.table import Table
+
+
+@pytest.mark.parametrize(
+    "kwargs,input_datasets,output_datasets,dataset_support,expected_kwargs",
+    [
+        (None, None, None, True, {}),
+        ({"task_id": "ex1"}, None, None, True, {"task_id": "ex1"}),
+        (
+            {"task_id": "ex1"},
+            Table("inlet", conn_id="con1"),
+            Table("outlet", conn_id="con2"),
+            True,
+            {
+                "task_id": "ex1",
+                "inlets": [Table("inlet", conn_id="con1")],
+                "outlets": [Table("outlet", conn_id="con2")],
+            },
+        ),
+        (
+            {
+                "task_id": "ex1",
+                "inlets": Table("inlet", conn_id="con1"),
+                "outlets": Table("outlet"),
+            },
+            Table("input_dataset"),
+            Table("output_dataset"),
+            True,
+            {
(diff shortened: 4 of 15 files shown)


python-sdk/src/astro/sql/operators/append.py
```diff
@@ -43,7 +44,12 @@ def __init__(
         self.columns = columns or {}
         task_id = task_id or get_unique_task_id("append_table")
 
-        super().__init__(task_id=task_id, **kwargs)
+        super().__init__(
+            task_id=task_id,
+            **kwargs_with_datasets(
+                kwargs=kwargs, input_ds=source_table, output_ds=target_table
```
tatiana, line 50: Shouldn't the `input_ds` be both `target_table` and `source_table` in this case?
It feels `target_table` is also an input of append (it is a dataset which is expected to exist before calling this method).

python-sdk/src/astro/sql/operators/merge.py
```diff
@@ -50,8 +51,14 @@ def __init__(
         self.columns = columns or {}
         self.if_conflicts = if_conflicts
         task_id = task_id or get_unique_task_id("merge")
-
-        super().__init__(task_id=task_id, **kwargs)
+        super().__init__(
+            task_id=task_id,
+            **kwargs_with_datasets(
+                kwargs=kwargs,
+                input_ds=source_table,
```
tatiana, line 58: Shouldn't the `input_ds` be both `target_table` and `source_table` in this case?
It feels that `target_table` is also a `merge` input (it is a dataset expected to exist before calling this method).

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -7,3 +9,26 @@
     # Airflow < 2.4
     Dataset = object
     DATASET_SUPPORT = False
+
+
+def kwargs_with_datasets(
+    kwargs: dict | None = None,
+    input_ds: list[Dataset] | Dataset | None = None,
```
tatiana, line 16: Have you considered having arguments with more explicit names `input_datasets` and `output_datasets`?

python-sdk/src/astro/sql/operators/append.py
kaxil: I don't think so, `target_table` is just the destination

kaxil: Otherwise, it won't make sense on what happened, 2 inputs and the output is the same as one of the input ds

python-sdk/src/astro/sql/operators/merge.py
kaxil: same comment as append

python-sdk/src/astro/airflow/datasets.py
kaxil: I did but it was just too verbose. I can change them though

kaxil: done in https://github.com/astronomer/astro-sdk/pull/808/commits/3e2ab23827dddbe3d2850e631d3d052ca38e7dbb

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -7,3 +9,26 @@
...
+
+
+def kwargs_with_datasets(
+    kwargs: dict | None = None,
+    input_datasets: list[Dataset] | Dataset | None = None,
+    output_datasets: list[Dataset] | Dataset | None = None,
+):
+    """
+    Extract inlets and outlets from kwargs if users have passed it. If not, set input datasets as inlets and
+    set output dataset as outlets
```
phanikumv, line 21: Do we have a note somewhere in the documentation that we support Datasets feature with python sdk on Airflow 2.4 ?

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -7,3 +9,26 @@
...
+    kwargs = kwargs or {}
+    if "inlets" in kwargs or (DATASET_SUPPORT and input_datasets):
+        inlets = kwargs.get("inlets", input_datasets)
+        inlets = inlets if isinstance(inlets, list) else [inlets]
+        kwargs.update({"inlets": inlets})
+
+    if "outlets" in kwargs or (DATASET_SUPPORT and output_datasets):
+        outlets = kwargs.get("outlets", output_datasets)
+        outlets = outlets if isinstance(outlets, list) else [outlets]
+        kwargs.update({"outlets": outlets})
```
feluelle, line 32:
```suggestion
    if "inlets" in kwargs or DATASET_SUPPORT and input_datasets:
        inlets = kwargs.get("inlets", input_datasets)
        inlets = inlets if isinstance(inlets, list) else [inlets]
        kwargs.update({"inlets": inlets})

    if "outlets" in kwargs or DATASET_SUPPORT and output_datasets:
        outlets = kwargs.get("outlets", output_datasets)
        outlets = outlets if isinstance(outlets, list) else [outlets]
        kwargs.update({"outlets": outlets})
```
You shouldn't need those as `and` has precedence.

python-sdk/tests/airflow/test_datasets.py
```diff
@@ -0,0 +1,72 @@
...
+            Table("input_ds"),
+            Table("output_ds"),
+            True,
+            {
+                "task_id": "ex1",
+                "inlets": [Table("inlet", conn_id="con1")],
+                "outlets": [Table("outlet")],
+            },
+        ),
+        # If File is not supported
```
feluelle, line 40: If you think it is worth commenting, I would rather add `ids=[...]`. Then we also have proper naming for those tests :)

python-sdk/tests/airflow/test_datasets.py
```diff
@@ -0,0 +1,72 @@
...
+                "inlets": [Table("inlet", conn_id="con1")],
+                "outlets": [Table("outlet", conn_id="con2")],
+            },
+        ),
+    ],
+)
+def test_kwargs_with_datasets(
+    kwargs, input_ds, output_ds, dataset_support, expected_kwargs
+):
+    """Test"""
```
feluelle, line 70: You don't say 😮 😅

feluelle (approved):
LGTM in general 👍

python-sdk/src/astro/airflow/datasets.py
sunank200: +1

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -7,3 +9,26 @@
...
+    kwargs: dict | None = None,
+    input_ds: list[Dataset] | Dataset | None = None,
+    output_ds: list[Dataset] | Dataset | None = None,
+):
+    """
+    Extract inlets and outlets from kwargs if users have passed it. If not, set input datasets as inlets and
+    set output dataset as outlets
+    """
+    kwargs = kwargs or {}
+    if "inlets" in kwargs or (DATASET_SUPPORT and input_ds):
```
sunank200:
```suggestion
    if "inlets" in kwargs or DATASET_SUPPORT and input_ds:
```

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -7,3 +9,26 @@
...
+        inlets = kwargs.get("inlets", input_ds)
+        inlets = inlets if isinstance(inlets, list) else [inlets]
+        kwargs.update({"inlets": inlets})
+
+    if "outlets" in kwargs or (DATASET_SUPPORT and output_ds):
```
sunank200:
```suggestion
    if "outlets" in kwargs or DATASET_SUPPORT and output_ds:
```

python-sdk/tests/airflow/test_datasets.py
sunank200, line 70:
```suggestion
    """Test Datasets supports with keyword arguments."""
```

sunank200 (approved):
Overall LGTM. Added few suggestions

python-sdk/tests/airflow/test_datasets.py
kaxil: lol 😄 -- haha Forgot to add the line

python-sdk/tests/sql/operators/test_append.py
```diff
@@ -182,3 +183,33 @@ def test_append_on_tables_on_different_db(sample_dag, database_table_fixture):
...
         test_utils.run_dag(sample_dag)
+
+
+@pytest.mark.skipif(
+    not DATASET_SUPPORT, reason="Inlets/Outlets will only be added for Airflow >= 2.4"
+)
+def test_inlets_outlets_supported_ds():
+    """Test Datasets are set as inlets and outlets"""
+    input_file = File("gs://bucket/object.csv")
+    output_table = Table("test_name")
```
utkarsharma2, line 194: @kaxil is there a way for users to use File/Table in the operator and not produce an outlet? 

Use Case:
If the user wants to load a table and do multiple transformations on the same table and only in the last transformation it is ready to be consumed by another dag. How do we do this right now?

load_file(output_table=Table(name="test")) -> transform_table_1(test:Table) -> transform_table_2(test:Table)

I'm guessing that every transformation step will be producing the same outlet URI.

python-sdk/src/astro/airflow/datasets.py
kaxil: Yeah I have that planned for a separate PR as mentioned in https://github.com/astronomer/astro-sdk/issues/611#issuecomment-1241402135

python-sdk/tests/sql/operators/test_append.py
kaxil: >I'm guessing that every transformation step will be producing the same outlet URI and it will be treated as a different instance of datasets.

If the `uri` is same, Airflow should treat them as the same dataset.

python-sdk/src/astro/sql/operators/append.py
tatiana: The task does change one of the inputs (table being appended to) - so, from this perspective, the output is one of the inputs (likely modified, considering the `source_table` has content). Otherwise, from a conceptual level, it could seem this task only generates this dataset - without relying on its pre-existence.

python-sdk/src/astro/sql/operators/merge.py
tatiana: In the case of `load_file`, it would only make sense adding it if the mode is "append" - but that is a good point

PR #815, opened 2022-09-09
Add response_size to run_raw_sql and warn about db thrashing

### LINKED ISSUE #791: Set a configurable limit to run_raw_sql rows output
**Context**
By default, `aql.run_raw_sql` does not limit the number of rows it can return.

[Comment on commit](https://github.com/astronomer/astro-sdk/commit/5b93df870dac5787dc1ec4ff11f238b9c360ebb0#r83305025)

**Version**
* Astro: 1.0.2

**Acceptance criteria**

* [x] Have an Airflow config "astro_sdk.run_raw_sql_limit_rows" which allows users to limit the amount of rows returned when running `run_raw_sql`
* [x] Set the default value not to be limited
* [x] Raise a warning if unlimited, explaining the risks to the user

tatiana:
Fix #791

### DIFF
--- a/python-sdk/src/astro/settings.py
+++ b/python-sdk/src/astro/settings.py
@@ -31,3 +31,9 @@ SNOWFLAKE_STORAGE_INTEGRATION_GOOGLE = conf.get(
 LOAD_TABLE_AUTODETECT_ROWS_COUNT = conf.getint(
     section="astro_sdk", key="load_table_autodetect_rows_count", fallback=1000
 )
+
+
+#: Reduce responses sizes returned by aql.run_raw_sql to avoid trashing the Airflow DB if the BaseXCom is used.
+RAW_SQL_MAX_RESPONSE_SIZE = conf.getint(
+    section="astro_sdk", key="run_raw_sql_response_size", fallback=-1
+)
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -27,7 +27,7 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         handler: Function | None = None,
         database: str | None = None,
         schema: str | None = None,
-        response_limit: int = 0,
+        response_limit: int = -1,
         response_size: int = -1,
         sql: str = "",
         **kwargs: Any,
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -28,6 +28,7 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         database: str | None = None,
         schema: str | None = None,
         response_limit: int = 0,
+        response_size: int = -1,
         sql: str = "",
         **kwargs: Any,
     ):
@@ -42,6 +43,8 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         self.database = self.op_kwargs.pop("database", database)
         self.schema = self.op_kwargs.pop("schema", schema)
         self.response_limit = self.op_kwargs.pop("response_limit", response_limit)
+        self.response_size = self.op_kwargs.pop("response_size", response_size)
+
         self.op_args: dict[str, Table | pd.DataFrame] = {}
 
         # We purposely do NOT render upstream_tasks otherwise we could have a case where a user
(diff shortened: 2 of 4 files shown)


python-sdk/src/astro/settings.py
```diff
@@ -31,3 +31,9 @@
 LOAD_TABLE_AUTODETECT_ROWS_COUNT = conf.getint(
     section="astro_sdk", key="load_table_autodetect_rows_count", fallback=1000
 )
+
+
+#: Reduce responses sizes returned by aql.run_raw_sql to avoid trashing the Airflow DB if the BaseXCom is used.
+RAW_SQL_MAX_RESPONSE_SIZE = conf.getint(
+    section="astro_sdk", key="run_raw_sql_response_size", fallback=-1
```
feluelle, line 38:
```suggestion
    section="astro_sdk", key="run_raw_sql_response_size", fallback=0
```
WDYT of making it consistent with `response_limit` i.e. setting it to `0`?

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -28,11 +30,22 @@ def execute(self, context: dict) -> Any:
...
+                "avoid this problem by either setting `response_size` to a small integer or by using a custom XCom "
+                "backend."
+            )
+
         if self.handler:
             response = self.handler(result)
             if self.response_limit and len(response) > self.response_limit:
                 raise IllegalLoadToDatabaseException()  # pragma: no cover
-            return response
+            if self.response_size >= 0:
```
feluelle, line 45: ..then changing this to:
```suggestion
            if self.response_size > 0:
```

python-sdk/src/astro/settings.py
tatiana: @feluelle I find the proposed change more confusing to users. It seems more intuitive that `response_size` set to 0 means to return zero results, as opposed to disabling `response_size` checking. The -1 gives clarity: that we are not checking `response_size`.

But I'm all for consistency, so once we agree - I can change the behaviour in the correct place

tatiana: ATM, I can't think of an obvious situation when the user would like to set `response_size=0`, but I believe it is a fair assumption that if a user did this, they'd expect the decorator to return zero results, as opposed to not limit the response size.

I'd be happy with using None to represent no response to the code, including: https://github.com/astronomer/astro-sdk/blob/e5bc9bb7ca1de34e789c5441baa7091877e22de0/python-sdk/src/astro/settings.py#L37-L39
Do you think this change is worth the additional complexity?

python-sdk/src/astro/sql/operators/raw_sql.py
tatiana: (discussion in the previous thread)

python-sdk/src/astro/settings.py
utkarsharma2: @tatiana @feluelle For me, -1 is better than explicitly saying to return all results and 0 means don't send any response back.

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -27,7 +27,8 @@ def __init__(
         handler: Function | None = None,
         database: str | None = None,
         schema: str | None = None,
-        response_limit: int = 0,
+        response_limit: int = -1,
```
utkarsharma2, line 30: @tatiana Probably too late for the question, but what's the difference between response_limit and response_size?

utkarsharma2: Also, shouldn't we assign default values from RAW_SQL_MAX_RESPONSE_SIZE?

tatiana: @utkarsharma2:
* `response_limit`: raises an exception if the response limit is passed, used by the `get_value_list` dynamic task template. It was introduced by #673, to avoid passing the limit of allowed parallel tasks of Airflow itself. It is a hard limit which is not user-defined. If there are more values than this limit, the DTT will fail.
* `response_size`: "cuts" or "trims" the number of items in the response of the `run_raw_sql` decorator. This was done as a follow up from @kaxil's comment https://github.com/astronomer/astro-sdk/commit/5b93df870dac5787dc1ec4ff11f238b9c360ebb0#r83305025. This PR avoids introducing a breaking change, so we log a warning if the user uses the default backend and does not set this size. If the user sets this size, we cut the response without raising an exception.

Since we do not want to introduce a breaking change, I couldn't just reuse `response_limit`

PR #825, opened 2022-09-12
Solve issues with running the benchmark

tatiana:
Fix: #798

Before this PR, the commands (run from `python-sdk/tests/benchmark`):
```
make local
make container
```

Were failing, partially due to the mono-repo refactoring. This PR solves this by also running the benchmark on GCP.

One key component for the success of running the benchmark locally (inside or outside a container) is to make sure the service account credentials are valid. This can be tested by using:
```
gcloud auth activate-service-account --key-file=
```

These were the commands run to confirm if things are working as expected:

# Local run outside of docker
```
GCP_BUCKET=dag-authoring ./run.sh 
```
Which generated:
```
gsutil cat gs://dag-authoring/benchmark/results/results-2022-09-12T11:49:02.ndjson
```

# Local run inside docker
```
GOOGLE_APPLICATION_CREDENTIALS=/Users/al-cht01/Certificates/gcp-astro.json make local
```

# Run inside K8s cluster
```
make setup_gke
make run_job
```

This run is in progress, and it is outputting this file:
```
gsutil cat gs://dag-authoring/benchmark/results/results-2022-09-12T13:03:36.ndjson
```

Which already has the following content:
```
{"duration": 11.076506, "memory_full_info": {"rss": 58396672, "vms": 692084736, "shared": 18092032, "text": 0, "lib": 0, "data": 99639296, "dirty": 0, "uss": 59473920, "pss": 59531264, "swap": 0}, "cpu_time": {"user": 2.5700000000000003, "system": 0.12, "children_user": 0.0, "children_system": 0.0, "iowait": 0.03}, "disk_usage": 196608, "io_counters": [{"read_count": 4030, "write_count": 1155, "read_bytes": 0, "write_bytes": 2945024, "read_chars": 16521137, "write_chars": 2854728}], "dag_id": "load_file_ten_kb_into_snowflake", "execution_date": "2022-09-12 13:04:00.802242+00:00", "revision": "382c4a0", "chunk_size": 1000000}
[...]

PR #835, opened 2022-09-13
Fix broken test

tatiana:
Fix test which broke as part of #825 (PR was merged prematurely)

### DIFF
--- a/python-sdk/tests/databases/test_snowflake.py
+++ b/python-sdk/tests/databases/test_snowflake.py
@@ -266,7 +266,7 @@ def test_load_file_from_cloud_to_table(database_table_fixture):
 )
 @mock.patch("astro.databases.snowflake.SnowflakeDatabase.hook")
 @mock.patch("astro.databases.snowflake.SnowflakeDatabase.create_stage")
-def test_load_file_to_table_natively_for_fallback(
+def test_load_file_to_table_natively_for_fallback_raises_exception_if_not_enable_native_fallback(
     mock_stage, mock_hook, database_table_fixture
 ):
     """Test loading on files to bigquery natively for fallback."""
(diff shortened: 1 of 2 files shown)


python-sdk/tests/databases/test_snowflake.py
```diff
@@ -278,12 +278,12 @@ def test_load_file_to_table_natively_for_fallback(
     )
     database, target_table = database_table_fixture
     filepath = str(pathlib.Path(CWD.parent, "data/sample.csv"))
-    response = database.load_file_to_table_natively_with_fallback(
-        source_file=File(filepath),
-        target_table=target_table,
-        enable_native_fallback=False,
-    )
-    assert response is None
+    with pytest.raises(DatabaseCustomError):
```
utkarsharma2, line 281: @tatiana Should we also have a test where `load_file_to_table_natively_with_fallback` don't raise an exception? 

Also, we can verify the kind of exception we are getting for this file? like substring-based check?

tatiana: @utkarsharma2 there is a test in `test_bigquery`, which was also updated, which has the scenario in which we don't raise an exception.

In regards to the exception, at the moment the code does not raise any meaningful message to the users. SInce this is a bug fix because our master branch is broken, I don't think we should be doing further changes in this PR - but we can look into improving the user experience  in these scenarios in separate tickets/PRs.

kaxil:

python-sdk/tests/databases/test_snowflake.py
utkarsharma2: Sure @tatiana, we can address that in a separate PR.

utkarsharma2 (approved):

Issue #794, opened 2022-09-07
Analyze methods for parsing a SQL directory as a DAG


dimberman:
**Please describe the feature you'd like to see**
To create a SQL CLI, we need to offer SQL as a native integration into Airflow. This would mean having a system where a SQL engineer can create a directory of SQL files and have Airflow automatically turn the output into a DAG. We can assume that the SQL tasks are in individual files and use a similar templating scheme to the astro python-sdk to communicate inheritance.

**Describe the solution you'd like**
The solution we'd like to see would involve the following UX

1. SQL engineers should not need to know that there is an Airflow DAG being generated (and should have minimal exposure to airflow overall
2. Should be easy for data engineers to deploy and manage in their airflow cluster
3. Should not lead to any speed degradation

**Are there any alternatives to this feature?**
None I can think of

**Additional context**

No
**Acceptance Criteria**

- [x] Create a document that compares different methods of parsing a SQL directory into an airflow readable DAG
- [x] Document should compare pros and cons of rendering the DAG at parse time vs. compiling the directory into a python-sdk DAG that is airflow readable
- [ ] Result of document should lead to an AEP

tatiana:
As part of this task, we are aware of at least two strategies:
* Create a DAG file which has all the tasks expanded (e.g. multiple `aql.transform` statements)
* Create a DAG file which has an operator which points to a SQL directory (e.g. `aql.render`), similar to the prototype implemented in the past: https://github.com/astronomer/astro-sdk/blob/0.8.4/src/astro/sql/parsers/sql_directory_parser.py

We expect this ticket to be considering the pros and cons of at least these two approaches.

feluelle:
Similar to
> Create a DAG file which has all the tasks expanded (e.g. multiple aql.transform statements)

we can use multiple `aql.transform_file` statements

feluelle:
[Doc](https://www.notion.so/Parsing-a-SQL-directory-as-a-DAG-876112c7646349fdb09c654711d38328) available comparing the three methods with Pros and Cons for each of them.

tatiana:
We're only missing the AEP

PR #836, opened 2022-09-13
Add sql directory parser and dag generator

feluelle:
# Description

Currently, we do not have a way of parsing a directory of sql files and generating a DAG out of it.

closes: #794
related: #897

## What is the new behavior?

This PR adds a _SQL Directory Parser_ as well as a _DAG Generator_ with examples.

The current implementation adds minimal configurability. More can be handled in a separate PR.
The following assumptions have been made:

1. The sql files **directory** can be passed via cli arg
1. The **start_date** is `datetime.now()`
1. The **conn_id**, **schema** and **database** can be set via sql file header
1. The **output table name** equals the sql file name

No.

### Checklist


feluelle:
This is a **very early draft** of how it could look like using multiple `aql.transform_file` operators.

cc @dimberman

feluelle:
Failed checks are unrelated. I'll add unit tests once we agree to a specific strategy.

I tested it by installing the following deps:
* _Jinja2_ (very useful for templating and generating the dag)
* _fs_ (optional but useful for working with file systems)
* _astro-sdk_
* _networkx_ (for detecting the right order in which to call the transform steps)

You can then run the example via `AIRFLOW__CORE__ENABLE_XCOM_PICKLING=True python sql-cli/src/sql_cli/parser.py`.

feluelle:
@dimberman @tatiana PTAL

Note that we can add meta data to the sql files and use `frontmatter` as done by Daniel to read it. Or we use yaml file(s) for configuration and additional meta data. But please let me know first if this approach, using _jinja_ and _networkx_, could be a potential candidate OR we use what Daniel [has built](https://github.com/astronomer/astro-sdk/blob/0.8.4/src/astro/sql/parsers/sql_directory_parser.py) already (`aql.render`).

I like the jinja approach more as it generates a dag file from a template once (e.g. on cli call) which should in the end be faster for the airflow dag parser to process - instead of the `aql.render` which generates the dag/tasks every time the DAG file calling it gets parsed. Moreover the jinja approach allows someone to debug the DAG code if necessary whereas in the `aql.render` approach everything is dynamically being generated.

kaxil:

sql-cli/src/sql_cli/sql_directory_parser.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_sql_files(directory: Path) -> set[SqlFile]:
+    """
+    Get all sql files within a directory.
+
+    :params directory: The directory look in for sql files.
+
+    :returns: the sql files found in the directory.
+    """
+    return {SqlFile(path=sql_file) for sql_file in directory.glob("**/*.sql")}
```
tatiana, line 45: If the `SQL` files belong to subdirectories of the given path, are we happy to have all the tasks flattened in the DAG, disregarding the (sub-)directory structure?
Would it make sense to use task groups to represent those inner groups?

Example of SQL directory
```
sql_dir/
├── sub_dir_1
│       ├── a.sql
│       └── b.sql
├── sub_dir_2
│       └── c.sql
│       └── d.sql
├── e.sql
└──  f.sql
```

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,37 @@
...
+        graph = DiGraph()
+
+        for sql_file in sql_files:
+            graph.add_node(sql_file.get_variable_name())
+            for parameter in sql_file.get_parameters():
+                graph.add_edge(sql_file.get_variable_name(), parameter)
+
+        if not is_directed_acyclic_graph(graph):
+            raise Exception(
+                "Could not generate DAG! Please make sure your sql files do not have circular dependencies."
```
tatiana, line 29: Would it be possible for us to detail where the circular dependency is, to helping the user to troubleshoot the issue?

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+from networkx import DiGraph, depth_first_search, is_directed_acyclic_graph
```
tatiana, line 3: Networkx is a very mature library, and I think you used it well.
Do you know how Airflow identifies cyclic dependencies at the moment? I'd be curious to see how both approaches relate.

tatiana:
@feluelle I like how simple and how explicit this approach is. Let's wait for other's feedback :crossed_fingers:

sql-cli/src/sql_cli/sql_directory_parser.py
feluelle: I had the same idea. We can indeed reconstruct the directory structure in Airflow using TaskGroups. But I would say this can be a separate feature.

sql-cli/src/sql_cli/dag_generator.py
feluelle: I don't know, but I can try to find out. @kaxil do you know?

sql-cli/src/sql_cli/dag_generator.py
feluelle: Resolved in cb8f6de

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+from networkx import DiGraph, depth_first_search, find_cycle, is_directed_acyclic_graph
```
dimberman, line 3: Interesting touch to using a network analysis library for creating the SQL graphs. A few questions:

1. How big is this library? 
2. Are we doing complex enough DAG creation to require this library?

I don't have strong thoughts either way, just want to make sure we're asking the question. Certainly seems to make the code cleaner :).

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,38 @@
...
+
+        :returns: a list of sql files sorted by least dependencies first
+            so they can be called sequentially in python code.
+        """
+        graph = DiGraph()
+
+        for sql_file in sql_files:
+            graph.add_node(sql_file.get_variable_name())
+            for parameter in sql_file.get_parameters():
+                graph.add_edge(sql_file.get_variable_name(), parameter)
```
dimberman, line 25: This line seems... wrong? Like we need to make sure that the variable referenced is actually a table before we add it to the graph.

sql-cli/src/sql_cli/sql_directory_parser.py
dimberman: I think reconstructing using taskgroups makes sense, it works nicely in gusty

dimberman (changes requested):
@feluelle given the pros and cons we described, are we only going with the "generate transform_file function" option? Or would the idea be that the user can decide whether to generate or simply render at runtime?

I think given the speeds we've seen I don't think that "it renders faster" is a huge concern here. We were able to render 1000 tasks in pretty much the same amount of time. 

The debugging argument also seems a bit iffy to me. Ultimately the error either way is going to be "could not parse at ..../foo.sql.py", I don't think it really matters which line in the airflow DAG starts the task. Either way the user is going to need to debug the SQL file itself.

Also with the `aql.transform_file` approach the DAG would be essentially unreadable when you get too 100 or 1000 tasks. It would generate a monstrously large DAG that would be really painful to edit.

sql-cli/src/sql_cli/dag_generator.py
feluelle: In 2.4.0, [this](https://github.com/apache/airflow/blob/2.4.0/airflow/utils/dag_cycle_tester.py#L49-L83) is how it detects cycles. So it does not use any external package for that.

feluelle:
> @feluelle given the pros and cons we described, are we only going with the "generate transform_file function" option? Or would the idea be that the user can decide whether to generate or simply render at runtime?

I think it makes sense to offer both options:
* `aql.render` in python-sdk as it is a python function which can directly be added to an existing DAG
  * probably being favored by DE
* and the jinja approach in sql-cli as it does not require the user to know airflow or python at all to generate a DAG
  * probably being favored by DA

> I think given the speeds we've seen I don't think that "it renders faster" is a huge concern here. We were able to render 1000 tasks in pretty much the same amount of time.

This is correct. I have edited my comment in GitHub to link to the notion doc.

> The debugging argument also seems a bit iffy to me. Ultimately the error either way is going to be "could not parse at ..../foo.sql.py", I don't think it really matters which line in the airflow DAG starts the task. Either way the user is going to need to debug the SQL file itself.

Same here. Please let us have this discussion in [this comment](https://www.notion.so/Parsing-a-SQL-directory-as-a-DAG-876112c7646349fdb09c654711d38328?d=9e59a5273e144f39a435239d981d3375#b43f2720812f444d80e6e0a52b5e2c91) of the docs.

> Also with the `aql.transform_file` approach the DAG would be essentially unreadable when you get too 100 or 1000 tasks. It would generate a monstrously large DAG that would be really painful to edit.

The point is that you should not edit it as it gets overwritten every time we regenerate the DAG file. This is indeed one of the "Cons" of this approach, see this [block](https://www.notion.so/astronomerio/Parsing-a-SQL-directory-as-a-DAG-876112c7646349fdb09c654711d38328#01b39ca6254c4b9e8e01145e5908c7a2) in the docs. With the new dataset support in Airflow and astro-sdk, we can also easily split it into multiple DAGs 👍 (we should probably do that anyway, even if not meant to be edited.)

sql-cli/src/sql_cli/dag_generator.py
feluelle: Yes, totally fair to ask this question.

I personally dont trust my graph algorithm skills enough to reimplement this from scratch. 😅 I am afraid, I am missing edge cases etc. That is why I decided to use _networkx_. It has builtin functions for cycle detection and finding the right order to execute statements in using dfs.

sql-cli/src/sql_cli/dag_generator.py
feluelle: Not sure what you mean. Every `parameter` is an input table and `sql_file.get_variable_name()` [equals](https://github.com/astronomer/astro-sdk/blob/cb8f6de7c6c64130f8c04a11798bcf6d23a5e1d8/sql-cli/src/sql_cli/template.jinja2#L17) the output table name in the **current** implementation.

sql-cli/src/sql_cli/sql_directory_parser.py
feluelle: Added in 07da4f9

feluelle:
@dimberman @tatiana I changed the code to use `transform_file` but with raw sql code instead of file path, because of two reasons:

1. If you develop locally the file path will might be different to where you deploy to e.g. I generated the DAG locally but want to run it in astro-cli (in docker). I would have to copy the files or create a volume to be able to reference the files
2. Because you can see the raw sql code being deployed - no surprises! If I remember correctly, I think this was also a request from Mike. I also don't think we can see the sql code in the "Rendered" view of the task instance if we pass a file path. With the current implementation you can see the rendered sql code after #897 has been merged.

**Another note:**
I can also create a `transform_raw` function or similar (unfortunately `transform` we already have, so I would have to move it somewhere else if we want to use the same name `transform`) to make it clearer that we transform raw sql - not sql from a file.

sql-cli/examples/_dags/advanced.py
```diff
@@ -0,0 +1,69 @@
...
+                    ),
+                ),
+            },
+            task_id="b",
+        )
+    with sub_dir_1:
+        sub_dir_1_a = aql.transform_file(
+            file_path="""
+                select * from {{ sub_dir_1_sub_dir_1_1_b }}
+            """,
```
dimberman, line 36: @feluelle this isn't a file path, it's a SQL function, is this an error?

sql-cli/examples/_dags/advanced.py
```diff
@@ -0,0 +1,69 @@
...
+
+with DAG(
+    dag_id="advanced",
+    start_date=timezone.parse("2022-09-20"),
+) as dag:
+    sub_dir_1 = TaskGroup(group_id="sub_dir_1")
+    sub_dir_2 = TaskGroup(group_id="sub_dir_2")
+    sub_dir_1_1 = TaskGroup(group_id="sub_dir_1_1")
+    with sub_dir_1, sub_dir_1_1:
+        sub_dir_1_sub_dir_1_1_b = aql.transform_file(
```
dimberman, line 16: @feluelle if our ultimate goal with transform_file over render is readability, I think we need to generate better names. I don't think a user would be able to parse what any of this means.

dimberman: I think that the name should be generated based on the file name, not the taskgroup.

Also why are these tasks under two task groups?

sql-cli/examples/_dags/advanced.py
```diff
@@ -0,0 +1,69 @@
...
+                select * from {{ sub_dir_1_a }}
+                union
+                select * from {{ non_existing_reference }}
+            """,
+            parameters={
+                "sub_dir_1_a": sub_dir_1_a,
+            },
+            op_kwargs={
+                "output_table": Table(
+                    name="a",
```
dimberman, line 62: why are we outputing to a table named "a"? Either the tables should have names that mean something or we should use temporary tables.

sql-cli/examples/_dags/simple.py
```diff
@@ -0,0 +1,82 @@
...
+        op_kwargs={
+            "output_table": Table(
+                name="a",
+            ),
+        },
+        task_id="a",
+    )
+    b = aql.transform_file(
+        file_path="""
+            select * from {{ a }}
```
dimberman, line 26: again, if we're using "transform_file", why are we passing in SQL statements instead of file paths?

sql-cli/examples/_dags/simple.py
```diff
@@ -0,0 +1,82 @@
...
+from airflow import DAG
+from airflow.utils import timezone
+from astro import sql as aql
+from astro.sql.table import Table
+
+with DAG(
+    dag_id="simple",
+    start_date=timezone.parse("2022-09-20"),
+) as dag:
+    a = aql.transform_file(
```
dimberman, line 11: Assuming this is based on the file name, we should have better file names so that our variables have more interesting names. Single letter variables in our example DAGs won't give an impression of production readiness.

sql-cli/examples/advanced/sub_dir_1/a.sql
```diff
@@ -0,0 +1 @@
+select * from {{ sub_dir_1_sub_dir_1_1_b }}
```
dimberman, line 1: This is code that will be seen by our users if they visit our GitHub. We should come up with better file names than "a.sql", "b.sql", etc.

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+from networkx import DiGraph, depth_first_search, find_cycle, is_directed_acyclic_graph
+from sql_directory_parser import SqlFile
+
+
+class SqlFilesDAG:
```
dimberman, line 7: Can this same class be used as the backend for an eventual render function as well?

Maybe we can factor out certain elements of this and put it into the python-sdk module for that function?

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+from networkx import DiGraph, depth_first_search, find_cycle, is_directed_acyclic_graph
```
dimberman, line 3: Huge +1 for having an external library handle graph creation. We should scale test this with some different configurations to make sure there's no efficiency issues but otherwise I'm always a fan of offloading work.

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,46 @@
...
+    :param sql_files: The sql files to use for DAG generation.
+    """
+
+    def __init__(self, sql_files: set[SqlFile]) -> None:
+        self.sql_files = sql_files
+        self.nodes = [sql_file.get_variable_name() for sql_file in self.sql_files]
+        self.edges = [
+            (sql_file.get_variable_name(), parameter)
+            for sql_file in self.sql_files
+            for parameter in sql_file.get_parameters()
```
dimberman, line 20: We should only be creating edges for parameters that are of type "table"

sql-cli/src/sql_cli/sql_directory_parser.py
```diff
@@ -0,0 +1,132 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from utils import find_template_variables
+
+
+class SqlFile:
+    """
+    A SqlFile is equals a transform step in the Astro SDK.
```
dimberman, line 10:
```suggestion
    A SqlFile is equivalent to a transform step in the Astro SDK.
```

sql-cli/src/sql_cli/sql_directory_parser.py
```diff
@@ -0,0 +1,132 @@
...
+def get_sql_files(directory: Path) -> set[SqlFile]:
+    """
+    Get all sql files within a directory.
+
+    :param directory: The directory look in for sql files.
+
+    :returns: the sql files found in the directory.
+    """
+    return {
+        SqlFile(root_dir=directory, path=child)
```
dimberman, line 129: Should we create a taskgroup based on the directory name for nested directories?

dimberman (changes requested):

sql-cli/examples/_dags/advanced.py
feluelle: Please read my https://github.com/astronomer/astro-sdk/pull/836#issuecomment-1256174948 above (with the note).

sql-cli/examples/_dags/advanced.py
feluelle: > @feluelle if our ultimate goal with transform_file over render is readability, I think we need to generate better names. I don't think a user would be able to parse what any of this means.

This automatically generated code from the file names and the directory names. We can change it easily. I agree with you that we should not use this in official examples in our repository. Do you have some examples in mind? :)

> I think that the name should be generated based on the file name, not the taskgroup.

Then it won't be possible to use the same filename twice. That is what I want to show with this example. It contains two `a.sql` in different folders/task groups.

> Also why are these tasks under two task groups?

Because the file is in a **nested** directory, it will be in nested Task Groups. I think this is pretty useful when you have lots of tables to be able to create nested structures.

sql-cli/examples/_dags/advanced.py
feluelle: Because currently, the table name equals the file name as in dbt. I am completely open to use sql headers or yaml for other and more configuration options.

sql-cli/examples/_dags/simple.py
feluelle: See https://github.com/astronomer/astro-sdk/pull/836#discussion_r978882933

sql-cli/examples/_dags/simple.py
feluelle: I agree with you. This was only for testing purpose. I need to come up with some more realistic names. If you have any suggestion, please let me know :)

sql-cli/src/sql_cli/dag_generator.py
feluelle: Interesting idea. 🤔

sql-cli/src/sql_cli/dag_generator.py
feluelle: The "if" one line below checks that by doing this `if parameter in self.nodes  # only add edges for existing nodes`. `self.nodes` only contains file names / input tables.

sql-cli/src/sql_cli/sql_directory_parser.py
feluelle: Is this a feature request? Or a rhetorical question? 😅 The current implementation via `rglob("*.sql")` does a recursive traversal. So we find **all** directories.

I think yes. I think it makes sense to reflect the directory structure in the database as schemas.

feluelle:
@dimberman @tatiana please see [this slack thread](https://astronomer.slack.com/archives/C042ZUJ7QG2/p1664181739485199) about why I had to remove the TaskGroup feature.

sql-cli/examples/_dags/advanced.py
feluelle: fixed in 3372688

sql-cli/examples/_dags/advanced.py
feluelle: fixed in c6902aa

sql-cli/examples/_dags/simple.py
feluelle: fixed in c6902aa

sql-cli/examples/advanced/sub_dir_1/a.sql
feluelle: fixed in c6902aa

.pre-commit-config.yaml
```diff
@@ -6,6 +6,8 @@ default_language_version:
   # force all unspecified python hooks to run python3
   python: python3
 minimum_pre_commit_version: "1.20.0"
+# We want to have generated files in raw format
+exclude: ^sql-cli/examples/_dags/
```
tatiana, line 10: Is there anything about the DAGs we are generating which MyPy doesn't like?

sql-cli/examples/_dags/simple.py
```diff
@@ -0,0 +1,96 @@
+# NOTE: This is an auto-generated file. Please do not edit this file manually.
```
tatiana, line 1: It would be great to have tests running validating the DAG Generator works as expected. I wonder if the content of the `examples` directory should be within tests. We could have an integration test that inputs SQL files and outputs the DAG file and checks if the DAG file is what we want it to be.

sql-cli/examples/advanced/mart/union_top_and_last.sql
```diff
@@ -0,0 +1,3 @@
+SELECT title, rating from {{source_top_five_animations}}
```
tatiana, line 1: I see these more as test fixtures than examples, from a `sql-cli` project perspective, WDYT @feluelle ?

.pre-commit-config.yaml
feluelle: No, but I think it makes sense to show the raw generated example dags to users instead of formatting them.

But yeah, we can run lint, I guess.

sql-cli/examples/advanced/mart/union_top_and_last.sql
feluelle: But these are the same examples we have for `python-sdk`. I think it makes sense to have some examples in an `examples` directory - other than tests. For tests we can also use shorter variables such as `a`, `b`, ... 😄

sql-cli/src/sql_cli/dag_generator.py
feluelle: I am constantly refactoring the code. Let's have a look again when it is a bit more mature.

feluelle:
@dimberman @tatiana I have updated my code and also added lots of tests. PTAL when you have time. Thank you.

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+
+from networkx import DiGraph, depth_first_search, find_cycle, is_directed_acyclic_graph
+from sql_cli.sql_directory_parser import SqlFile
+
+
+class DagCycle(Exception):
+    """An exception raised when DAG contains a cycle."""
```
kaxil, line 8: Should we have a separate file for custom exceptions?

sql-cli/src/sql_cli/main.py
```diff
@@ -0,0 +1,53 @@
...
+from sql_cli.dag_generator import SqlFilesDAG
+from sql_cli.sql_directory_parser import get_sql_files
+from sql_cli.utils import render_jinja
+
+
+def cli() -> Namespace:
+    parser = ArgumentParser()
+    parser.add_argument("directory", type=Path)
+    args = parser.parse_args()
+    return args
```
kaxil, line 15: This is temporary until we switch to click?

sql-cli/src/sql_cli/main.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def init_project(directory: Path) -> Tuple[Path, Path]:
```
kaxil, line 18: Docstrings for all the function please

sql-cli/src/sql_cli/sql_directory_parser.py
```diff
@@ -0,0 +1,129 @@
...
+
+        :returns: True if this sql file is equal to the other one.
+        """
+        if isinstance(__o, SqlFile):
+            return self.root_directory == __o.root_directory and self.path == __o.path
+        return False
+
+    def __gt__(self, other: SqlFile) -> bool:
+        """
+        Check if this sql file is greater than the other sql file.
```
kaxil, line 44: curious on using the "name" for ">" -- what's the rationale behind it? Depending on that, we might want to use file-size

sql-cli/src/sql_cli/sql_directory_parser.py
```diff
@@ -0,0 +1,129 @@
...
+    ) -> None:
+        self.root_directory = root_directory
+        self.path = path
+        self.target_directory = target_directory
+
+        post = frontmatter.load(self.path)
+        self.content = post.content
+        self.metadata = post.metadata
+
+    def __eq__(self, __o: object) -> bool:
```
kaxil, line 30: why `__o` instead of just `other`?

kaxil: similar to what you already have in `__gt__`

kaxil (approved):
I have added some minor comments

sql-cli/src/sql_cli/dag_generator.py
feluelle: Makes sense. We could have more exceptions in the future.

sql-cli/src/sql_cli/main.py
feluelle: This is correct.

sql-cli/src/sql_cli/main.py
feluelle: Sure. :)

sql-cli/src/sql_cli/sql_directory_parser.py
feluelle: I will document it better even-though I pass the comparison to downstream _pathlib.Path_ comparison. It is about sorting based on name - not size.

sql-cli/src/sql_cli/sql_directory_parser.py
feluelle: This was auto-generated for me when I had typed `def __eq` in my vscode 😅. I have updated it.

feluelle:
The failing checks are unrelated. I will try to address them separately.

.pre-commit-config.yaml
```diff
@@ -22,6 +22,7 @@ repos:
       - id: check-yaml
       - id: debug-statements
       - id: end-of-file-fixer
+        exclude: ^sql-cli/examples/_dags/|^sql-cli/examples/_target/|^sql-cli/tests/_dags/|^sql-cli/tests/_target/
```
tatiana, line 25: Probably worth adding a comment why we're excluding this.

sql-cli/pyproject.toml
```diff
@@ -18,11 +18,9 @@ license = {file = "LICENSE"}
 requires-python = ">=3.7"
 dependencies = [
     "apache-airflow>=2.0",
```
tatiana, line 20: This can also go away, since it is a dependency of `astro-sdk-python`

sql-cli/src/sql_cli/dag_generator.py
```diff
@@ -0,0 +1,47 @@
...
+            so they can be called sequentially in python code.
+        """
+        graph = DiGraph()
+        graph.add_nodes_from(self.nodes)
+        graph.add_edges_from(self.edges)
+
+        if not is_directed_acyclic_graph(graph):
+            raise DagCycle(
+                "Could not generate DAG! "
+                f"A cycle between {' and '.join(' and '.join(edges) for edges in find_cycle(graph))} has been detected!"
```
tatiana, line 39: :raised_hands: amazing! thanks for adding this

dimberman (approved):
This looks fantastic @feluelle. well-documented, clean, love it.

sql-cli/src/sql_cli/macros/tasks.py.jinja2
```diff
@@ -0,0 +1,31 @@
...
+    {%- if sql_file.metadata and "conn_id" in sql_file.metadata %}
+    conn_id="{{ sql_file.metadata.conn_id }}",
+    {%- endif %}
+    op_kwargs={
+        "output_table": Table(
+            name="{{ sql_file.path.stem }}",
+            {%- if "database" in sql_file.metadata or "schema" in sql_file.metadata %}
+            metadata=Metadata(
+                {%- if "database" in sql_file.metadata %}
+                database="{{ sql_file.metadata.database }}",
```
tatiana, line 20: I like that we're being explicit at first, even though it may be verbose, it will help us - and the end-users - troubleshooting any issues.

sql-cli/src/sql_cli/sql_directory_parser.py
```diff
@@ -0,0 +1,131 @@
...
+
+    :param directory: The directory look in for sql files.
+    :param target_directory: The target directory path for the executable sql.
+
+    :returns: the sql files found in the directory.
+    """
+    return {
+        SqlFile(root_directory=directory, path=child, target_directory=target_directory)
+        for child in directory.rglob("*.sql")
+        if child.is_file()
```
tatiana, line 130: This is helpful to avoid cycles with symbolic links - it may be worth adding a test for this (it can be done ina  separate PR)

sql-cli/src/sql_cli/utils.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
```
tatiana, line 1: In future, we may be able to rename this module to something jinja related - so we avoid generic utils blowing up. I'm happy for us to do this in a separate PR as well :)

sql-cli/tests/_dags/sql_files.py
```diff
@@ -0,0 +1,58 @@
+# NOTE: This is an auto-generated file. Please do not edit this file manually.
```
tatiana, line 1: This could go to `.gitignore` (as well as `tests/target`)

tatiana (approved):
@feluelle minor comments, looking forward to seeing the feedback from our end-users on 0.1! :tada:

sql-cli/src/sql_cli/sql_directory_parser.py
feluelle: This did not check for symlinks, but [b7aeae8](https://github.com/astronomer/astro-sdk/pull/836/commits/b7aeae856e5d8e7e76e08976319e4ccae7bdedf1) does.

Issue #816, opened 2022-09-09
Add `example_datasets.py` DAG to unit tests


phanikumv:
**Please describe the feature you'd like to see**
Add example_datasets.py dag as part of unit tests

**Describe the solution you'd like**
Add example_datasets.py dag as part of unit tests

**Are there any alternatives to this feature?**
No

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #839, opened 2022-09-14
Add `example_datasets.py` DAG to unit tests

phanikumv:
closes: #816

### DIFF
--- a/python-sdk/tests/airflow/test_datasets.py
+++ b/python-sdk/tests/airflow/test_datasets.py
@@ -1,4 +1,5 @@
 from unittest import mock
+
 import airflow
 import pytest
 from astro.airflow.datasets import kwargs_with_datasets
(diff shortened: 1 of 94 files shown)


kaxil:

python-sdk/tests/test_example_dags.py
```diff
@@ -93,3 +93,22 @@ def test_example_dynamic_task_map_dag(session, dag_id):
...
+    ],
+)
+def test_example_dataset_dag(session, dag_id):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    db = DagBag(dir_path + "/../example_dags")
+    dag = db.get_dag(dag_id)
+
+    if dag is None:
+        raise NameError(f"The DAG with dag_id: {dag_id} was not found")
+    wrapper_run_dag(dag)
```
kaxil, line 114: Did you test it with 2.4.0b1 .. you can test it out on CI by bumping airflow dep to >=2.4.0b1 

I have a feeling this won't work as the second Dataset wont be triggered by the logic in `wrapper_run_dag`

kaxil: Bump the following to 2.4.0b1
https://github.com/astronomer/astro-sdk/blob/330c02a4d455759f9a164dde253a32747a7d2f5f/python-sdk/pyproject.toml#L20

phanikumv: No I havent tested with 2.4.0b1, doing it now

python-sdk/tests/test_example_dags.py
```diff
@@ -92,3 +92,22 @@ def test_example_dynamic_task_map_dag(session, dag_id):
...
     wrapper_run_dag(dag)
+
+
+@pytest.mark.skipif(
+    airflow.__version__ < "2.4.0", reason="Require Airflow version >= 2.4.0"
+)
+@pytest.mark.parametrize(
+    "dag_id",
+    [
+        "example_dataset_producer",
```
kaxil, line 103: Do we need to test that consumer DAG is triggered ?

python-sdk/tests/test_example_dags.py
phanikumv: somehow CI doesnt seem to run tests on 2.4.0b1

python-sdk/tests/test_example_dags.py
phanikumv: Yes , and this will need >=2.4.0

python-sdk/tests/test_example_dags.py
phanikumv: It has picked up the DAG successfully
`5.85s call     tests/test_example_dags.py::test_example_dataset_dag[example_dataset_producer]`

python-sdk/tests/test_example_dags.py
utkarsharma2: I think we can skip testing this since it's something specific to airflow and not AstroSDK.

phanikumv: Just wondering how can we ensure that if a developer changes the code relevant to mapping the Datasets to Astro-SDK, that it is not breaking any functionality?

python-sdk/tests/test_example_dags.py
phanikumv: As discussed, moved the test to `python-sdk/tests/airflow/test_datasets.py` . Its working fine with `2.4.0b1`

python-sdk/tests/airflow/test_datasets.py
```diff
@@ -76,3 +77,26 @@ def test_kwargs_with_datasets(
...
         )
+
+
+@pytest.mark.skipif(
+    airflow.__version__ < "2.4.0", reason="Require Airflow version >= 2.4.0"
+)
+def test_example_dataset_dag():
+    import os
+
+    from airflow.models.dagbag import DagBag
```
kaxil, line 88: This two can go at the top

kaxil (approved):

PR #843, opened 2022-09-14
Fix flaky redshift tests

utkarsharma2:
# Description

Redshift does accept constraints but does not enforce them. Adding the same name constraint is causing the CI to fail. The constraint was getting dropped after the table gets dropped: https://docs.aws.amazon.com/redshift/latest/dg/r_DROP_TABLE.html. 

## What is the new behavior?
Skip creating constraints for redshift.

Nope

### DIFF
--- a/python-sdk/tests/databases/test_bigquery.py
+++ b/python-sdk/tests/databases/test_bigquery.py
@@ -310,7 +310,7 @@ def test_load_file_to_table_natively_for_fallback(
     indirect=True,
     ids=["bigquery"],
 )
-def test_load_file_to_table_natively_for_fallback_wrong_file_location(
+def test_load_file_to_table_natively_for_fallback_wrong_file_location_with_enable_native_fallback(
     database_table_fixture,
 ):
     """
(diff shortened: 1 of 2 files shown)


sunank200:
Also python deepsource and type-check needs to be fixed.

pankajkoti (approved):

pankajkoti:
I think the description is incorrect. Redshift does accept constraint but does not enforce it. Also adding the constraint is not causing the CI to fail. But, adding a constraint with the same name is causing the CI to fail. I believe this constraint was getting dropped after the table gets dropped: https://docs.aws.amazon.com/redshift/latest/dg/r_DROP_TABLE.html. However, I believe something in the setup did not drop a table due to some probable abrupt error, and hence the constraint existed.

sunank200 (approved):

utkarsharma2:
@sunank200 Can't fix deepsource, it's failing with 500 - https://deepsource.io/gh/astronomer/astro-sdk/run/101fdb3f-d6b8-4ea4-9f5c-3bc04f31541e/python/PYL-W0613

PR #844, opened 2022-09-14
[WIP] Test workflow config

pankajastro:


python-sdk/tests/benchmark/Makefile
```diff
@@ -27,8 +27,9 @@ clean:
 # Takes approximately 7min
 setup_gke: clean
 	@cd infrastructure/terraform && \
-		terraform init && \
-		terraform apply
+		terraform init -input=false && \
+		terraform plan -out=tfplan -input=false && \
+		terraform apply -input=false tfplan
```
feluelle, line 32: We can set an env var to disable input (see [here](https://www.terraform.io/cli/config/environment-variables#tf_input)). Then you don't need to pass it every time.

tatiana: Why do we need to use `-input-false`? Are there any variables that did not have their values set?

python-sdk/tests/benchmark/Makefile
```diff
@@ -37,6 +38,7 @@ container: clean
 	@echo "Building and pushing container $(CONTAINER_REGISTRY)"
 	@gcloud auth configure-docker gcr.io
 	@docker build --build-arg=GIT_HASH=$(GIT_HASH) --tag=$(CONTAINER_REGISTRY):$(GIT_HASH) --tag=$(CONTAINER_REGISTRY):latest -f ./Dockerfile  ../../
+	@gcloud auth activate-service-account --key-file $(GOOGLE_APPLICATION_CREDENTIALS)
```
tatiana, line 41: Since we are introducing this dependency on the environment variable, it may be worth to make this dependency explicit in line 37, by adding `check_google_credentials`:
```
container: clean check_google_credentials
```

tatiana (commented):
@pankajastro this will help us with regression tests!

Do you have a link to an successful job where we run this Github action workflow? Did you have a chance of confirming the correspondent GCS file & bigquery were set as expected?

pankajastro:
> @pankajastro this will help us with regression tests!
> 
> Do you have a link to an successful job where we run this Github action workflow? Did you have a chance of confirming the correspondent GCS file & bigquery were set as expected?

this job succeeded https://github.com/astronomer/astro-sdk/actions/runs/3092321371/jobs/5011687626 but which is corresponding to commit `7884fcad6b5d53d10323aee724dc68d8b9096a2e ` but I'm not able to see the result I checked with ```SELECT * FROM `astronomer-dag-authoring.benchmark.load_files_to_database` ORDER BY execution_date DESC```

python-sdk/tests/benchmark/Makefile
pankajastro: Look like. otherwise, it throws an error 
```
Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: Releasing state lock. This may take a few moments...
Makefile:29: recipe for target 'setup_gke' failed
make: *** [setup_gke] Error 1
```

python-sdk/tests/benchmark/Makefile
```diff
@@ -61,7 +62,7 @@ run_job:
 	@kubectl create -f infrastructure/kubernetes/job.yaml
 
 # Takes approximately 2min
-teardown_gke:
-	@cd infrastructure/terraform && \
-		terraform init && \
-		terraform destroy --target google_container_node_pool.benchmark -auto-approve
+#teardown_gke:
```
utkarsharma2, line 65: is this temp change?

pankajastro: yes

pankajastro:
> @pankajastro this will help us with regression tests!
> 
> Do you have a link to an successful job where we run this Github action workflow? Did you have a chance of confirming the correspondent GCS file & bigquery were set as expected?

yes, I can see results in GCS and bigquery 
check biquery with
```
SELECT * FROM `astronomer-dag-authoring.benchmark.load_files_to_database` ORDER BY execution_date DESC
```
An GCS uploaded file
https://console.cloud.google.com/storage/browser/_details/dag-authoring/benchmark/results/results-2022-09-26T13:59:22.ndjson;tab=live_object?project=astronomer-dag-authoring

.github/ci-test-connections-benchmark.yaml
```diff
@@ -0,0 +1,53 @@
+connections:
```
tatiana, line 1: @pankajastro, I'm sure there is a good reason why we're not using the previously existing `ci-test-connections.yml`, but I'd love to know them from you. Could you share them, please?

python-sdk/tests/benchmark/Makefile
```diff
@@ -27,16 +27,17 @@ clean:
 # Takes approximately 7min
 setup_gke: clean
 	@cd infrastructure/terraform && \
-		terraform init && \
```
tatiana, line 30: @pankajastro on line 1, I suggest we remove `teardown_gke` for now. We'll need to wait for the job to be completed before tearing down the cluster.

pankajastro: Removed

.github/ci-test-connections-benchmark.yaml
pankajastro: We wanted to have connections with different parameters. But, as discussed offline removed the new configuration file and added a new connection to the existing one.

tatiana (approved):
Looks good, thanks a lot, @pankajastro !

PR #845, opened 2022-09-14
Fix isort via pre-commit

feluelle:
# Description

Currently, isort via pre-commit is not running correctly over all related files.

## What is the new behavior?

It runs properly over all related files.

No.

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -70,15 +70,13 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-        name: Run isort to sort imports in Python files
+        name: Run isort for python-sdk
         args: [ "--profile", "black" ]
-        files: .py$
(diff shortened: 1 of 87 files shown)


.pre-commit-config.yaml
```diff
@@ -70,15 +70,13 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-        name: Run isort to sort imports in Python files
+        name: Run isort for python-sdk
```
utkarsharma2, line 73: @feluelle Any reason we are doing this separately for Python/SQL SDK?

utkarsharma2 (approved):

.pre-commit-config.yaml
feluelle: Good question. I don't know. 🤷  @dimberman could you help us please?

kaxil (approved):

Issue #838, opened 2022-09-14
Fix duplicate load of data


phanikumv:
**Describe the bug**
In load_file() if the user has given a pattern that results in two files

example -
pattern : s3://tmp/

which results in files:

    s3://tmp/test1.csv
    s3://tmp/test2.csv

With the existing code and new autodetecting schema logic, we will end up with:

Step 1: Autoschema detect - Table + 1st file load
Step 2: Load data in a table - 1st file load + 2nd file load

will result in table 1 loading twice.

**Version**
* Astro: [e.g. 1.2.0]

**To Reproduce**
Load using a pattern and check the no. of rows.

**Expected behavior**
The number of rows should match with no. rows in files.

pankajkoti:
PR is ready for review

PR #851, opened 2022-09-14
Skip loading file for bigquery schema detection

utkarsharma2:
# Description
Currently, the way we are autodetecting schema for bigquery is loading an entire file, and then truncating the table loaded earlier is not an optimal approach.
 
related: https://github.com/astronomer/astro-sdk/issues/709
closes: #838 

## What is the new behavior?
We can skip autodetection altogether and this would prevent us from loading files twice as well.

No

### DIFF
--- a/python-sdk/src/astro/databases/google/bigquery.py
+++ b/python-sdk/src/astro/databases/google/bigquery.py
@@ -18,7 +18,6 @@ from astro.constants import (
 )
 from astro.databases.base import BaseDatabase
 from astro.exceptions import DatabaseCustomError
-from astro.files import File
 from astro.settings import BIGQUERY_SCHEMA
 from astro.sql.table import Metadata, Table
 from google.api_core.exceptions import (
@@ -50,6 +49,8 @@ from google.resumable_media import InvalidResponse
 from sqlalchemy import create_engine
 from sqlalchemy.engine.base import Engine
 from tenacity import retry, stop_after_attempt
+
+from astro.files import File
 
 DEFAULT_CONN_ID = BigQueryHook.default_conn_name
 NATIVE_PATHS_SUPPORTED_FILE_TYPES = {
@@ -247,7 +248,7 @@ class BigqueryDatabase(BaseDatabase):
             return True
         return False
 
-    def create_table_using_native_schema_autodetection(
+    def create_table_using_native_schema_autodetection(  # skipcq: PYL-R0201
         self,
         table: Table,
         file: File,
--- a/python-sdk/src/astro/databases/google/bigquery.py
+++ b/python-sdk/src/astro/databases/google/bigquery.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import time
-from typing import Any
+from typing import Any, Callable, Mapping
 
 import pandas as pd
 from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
@@ -18,7 +18,6 @@ from astro.constants import (
 )
 from astro.databases.base import BaseDatabase
 from astro.exceptions import DatabaseCustomError
-from astro.files import File
 from astro.settings import BIGQUERY_SCHEMA
 from astro.sql.table import Metadata, Table
 from google.api_core.exceptions import (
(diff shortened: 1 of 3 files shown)


python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -59,8 +59,14 @@
 }
 BIGQUERY_WRITE_DISPOSITION = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
 
-NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_TYPES = {FileType.CSV, FileType.NDJSON}
-NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_LOCATIONS = {FileLocation.GS}
+NATIVE_AUTODETECT_SCHEMA_CONFIG: Mapping[
```
kaxil, line 62: >Currently, the way we are autodetecting schema for bigquery is loading an entire file, and then truncating the table loaded earlier is not an optimal approach.

Do we not need to fix it for other DBs too?

utkarsharma2: @kaxil  Actually no, this is just a problem for Bigquery. Specifically GCS to Bigquery.

utkarsharma2: Snowflake does not load the data in the table, just create the table in the autodetection step.

cc: @sunank200 @tatiana

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -254,30 +258,10 @@ def create_table_using_native_schema_autodetection(
...
-
-        self.hook.insert_job(
-            configuration=job_config,
+        supported_config = NATIVE_AUTODETECT_SCHEMA_CONFIG.get(
+            file.location.location_type
         )
-
-        # We have to clear the table afterwards as bigquery automatically loads the data when creating the table.
-        statement = f"TRUNCATE TABLE `{self.get_table_qualified_name(table)}`"
-        self.run_sql(statement)
```
feluelle, line 280: Thanks @utkarsharma2. The truncate statement was pretty ugly and probably also error prone (potentially vulnerable to race conditions). And of course speed and cost plays an import role when loading data into the data warehouse.

feluelle: I am still wondering how I could come up with that idea 🤦

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -59,8 +59,14 @@
...
 BIGQUERY_WRITE_DISPOSITION = {"replace": "WRITE_TRUNCATE", "append": "WRITE_APPEND"}
 
-NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_TYPES = {FileType.CSV, FileType.NDJSON}
-NATIVE_AUTODETECT_SCHEMA_SUPPORTED_FILE_LOCATIONS = {FileLocation.GS}
+NATIVE_AUTODETECT_SCHEMA_CONFIG: Mapping[
+    FileLocation, Mapping[str, list[FileType] | Callable]
+] = {
+    FileLocation.GS: {
+        "filetype": [FileType.CSV, FileType.NDJSON, FileType.PARQUET],
+        "method": lambda table, file: None,
```
feluelle, line 67: What does this do? I mean the `None`.

What would `supported_config["method"](table=table, file=file)` look like for gcs to bq?

utkarsharma2: @feluelle We actually don't need to do anything in case of GCS to bigquery, the native load file takes care of this. So this is a dummy function we execute. To have a pluggable interface, down the line we can easily plugin any other location like
```
FileLocation.S3: {
 "filetype": [FileType.CSV, FileType.NDJSON, FileType.PARQUET],
 "method": someMethodThatHandlesForS3
}
def someMethodThatHandlesForS3():
 pass

```

python-sdk/src/astro/databases/google/bigquery.py
sunank200: That's correct this case is only valid for BigQuery. Other databases require the creation of tables explicitly.

sunank200 (approved):
Overall LGTM.

python-sdk/src/astro/databases/google/bigquery.py
kaxil: Worth adding that as comments in the code too

PR #854, opened 2022-09-15
Add documentation on how to support new database in Astro Python SDK

pankajkoti:
closes: #682

PR #852, opened 2022-09-15
Add docs on Dataset (AIP-48) support

pankajkoti:
The commit adds the following:
- Dataset concepts explained briefly and link to the Airflow documentation for Datasets
- Example Producer and Consumer Dataset usage
- Default input and output datasets mentioned for respective Astro SDK operators
- Python SDK Dataset URIs example: Table & File
- Detail about how to disable use of default datasets

closes:  #824

PR #858, opened 2022-09-15
Allow users to disable auto addition of inlets/outlets via airflow.cfg

kaxil:
Astro SDK automatically adds inlets and outlets for all the operators if DATASET is supported (Airflow >=2.4).

Users can over-ride it on a task level by adding inlets and outlets themselves. However, for users who want to use SDK with Airflow 2.4 and above but do not want to leverage Data-aware scheduling, auto-addition of inlets/outlets would be annoying so this new config setting will provide an escape hatch for such users.

closes https://github.com/astronomer/astro-sdk/issues/857

No

### Checklist
- [ ] Extended the README / documentation, if necessary. -- @pankajkoti can you add it in https://github.com/astronomer/astro-sdk/pull/854 plz

### DIFF
--- a/python-sdk/src/astro/settings.py
+++ b/python-sdk/src/astro/settings.py
@@ -36,3 +36,8 @@ LOAD_TABLE_AUTODETECT_ROWS_COUNT = conf.getint(
 RAW_SQL_MAX_RESPONSE_SIZE = conf.getint(
     section="astro_sdk", key="run_raw_sql_response_size", fallback=-1
 )
+
+# Should Astro SDK automatically add inlets/outlets to take advantage of Airflow 2.4 Data-aware scheduling
+AUTO_ADD_INLETS_OUTLETS = conf.getboolean(
+    "astro_sdk", "auto_add_inlets_outlets", fallback=True
+)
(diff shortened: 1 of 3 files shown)


pankajkoti:
@kaxil did you mean to link https://github.com/astronomer/astro-sdk/pull/852 in the description for extending the documentation ?

kaxil:
> @kaxil did you mean to link #852 in the description for extending the documentation ?

yup

python-sdk/src/astro/settings.py
```diff
@@ -36,3 +36,8 @@
 RAW_SQL_MAX_RESPONSE_SIZE = conf.getint(
     section="astro_sdk", key="run_raw_sql_response_size", fallback=-1
 )
+
+# Should Astro SDK automatically add inlets/outlets to take advantage of Airflow 2.4 Data-aware scheduling
+AUTO_ADD_INLETS_OUTLETS = conf.getboolean(
```
utkarsharma2, line 41: @kaxil Can we add this to docs?

utkarsharma2 (approved):
LGTM

python-sdk/src/astro/settings.py
kaxil: Yeah @pankajkoti is going to do it as part of #852 as mentioned in the PR description

Issue #856, opened 2022-09-15
Temp tables should not create an Airflow Dataset


utkarsharma2:
https://astronomer.slack.com/archives/C02B8SPT93K/p1663183720810149

All temporary tables generated with the same connection_id are treated as the same Dataset. from @collinmcnulty

kaxil:
PR: https://github.com/astronomer/astro-sdk/pull/862

PR #862, opened 2022-09-15
Stop generating Datasets for temp tables

kaxil:
closes https://github.com/astronomer/astro-sdk/issues/856

We don't want to generate datasets for temp tables so I have created a separate internal class called `TempTable` which will be used when users don't pass a name to the Table class.

```python
In [1]: from astro.sql.table import Table, TempTable

In [2]: Table()
Out[2]: TempTable(_name='', conn_id='', metadata=Metadata(schema=None, database=None), columns=[], temp=True)

In [3]: Table("name", conn_id="con1")
Out[3]: Table(_name='name', conn_id='con1', metadata=Metadata(schema=None, database=None), columns=[], temp=False, uri='astro://con1@?table=name', extra={})

```

No

### Checklist
- [x] Extended the README / documentation, if necessary - **No user-facing change is required -- maybe we just add a single line in ** https://github.com/astronomer/astro-sdk/pull/852 to say "Temp tables won't generate Datasets"

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,11 +73,11 @@ repos:
     hooks:
       - id: isort
         name: Run isort for python-sdk
-        args: ["--profile=black"]
+        args: [ "--profile", "black" ]
         files: ^python-sdk/
       - id: isort
         name: Run isort for sql-cli
-        args: ["--profile=black"]
+        args: [ "--profile", "black" ]
         files: ^sql-cli/
 
   - repo: https://github.com/codespell-project/codespell
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,11 +73,11 @@ repos:
     hooks:
       - id: isort
         name: Run isort for python-sdk
-        args: [ "--profile", "black" ]
+        args: ["--profile=black"]
         files: ^python-sdk/
       - id: isort
         name: Run isort for sql-cli
-        args: [ "--profile", "black" ]
+        args: ["--profile=black"]
         files: ^sql-cli/
 
   - repo: https://github.com/codespell-project/codespell
--- a/python-sdk/src/astro/airflow/datasets.py
+++ b/python-sdk/src/astro/airflow/datasets.py
@@ -29,8 +29,15 @@ def kwargs_with_datasets(
         and settings.AUTO_ADD_INLETS_OUTLETS
         and input_datasets
     ):
+        # Remove non-dataset objects like Temp tables (unless passed by user)
+        input_datasets = (
+            input_datasets if isinstance(input_datasets, list) else [input_datasets]
+        )
+        input_datasets = [
+            input_ds for input_ds in input_datasets if isinstance(input_ds, Dataset)
+        ]
+
         inlets = kwargs.get("inlets", input_datasets)
-        inlets = inlets if isinstance(inlets, list) else [inlets]
         kwargs.update({"inlets": inlets})
 
     if (
@@ -39,8 +46,15 @@ def kwargs_with_datasets(
         and settings.AUTO_ADD_INLETS_OUTLETS
         and output_datasets
     ):
+        # Remove non-dataset objects like Temp tables (unless passed by user)
+        output_datasets = (
+            output_datasets if isinstance(output_datasets, list) else [output_datasets]
+        )
+        output_datasets = [
+            output_ds for output_ds in output_datasets if isinstance(output_ds, Dataset)
+        ]
+
         outlets = kwargs.get("outlets", output_datasets)
-        outlets = outlets if isinstance(outlets, list) else [outlets]
         kwargs.update({"outlets": outlets})
 
     return kwargs
--- a/python-sdk/src/astro/sql/table.py
+++ b/python-sdk/src/astro/sql/table.py
@@ -152,7 +152,7 @@ class Table(BaseTable, Dataset):
     extra: dict | None = field(init=False, factory=dict)
 
     def __new__(cls, *args, **kwargs):
-        name = kwargs.get("name", "") or (args[0] if len(args) > 0 else "")
+        name = kwargs.get("name") or args and args[0] or ""
         temp = kwargs.get("temp", False)
         if temp or (not name or name.startswith("_tmp")):
             return TempTable(*args, **kwargs)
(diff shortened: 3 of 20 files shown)


dimberman (approved):
Oh man, this looks like a fun one... +1 LGTM @kaxil

python-sdk/src/astro/sql/table.py
```diff
@@ -125,6 +122,42 @@ def name(self, value: str) -> None:
...
+
+@define(slots=False)
+class TempTable(BaseTable):
+    """
+    Internal class to represent a Temporary table
+
+    :sphinx-autoapi-skip:
+    """
+
+    temp: bool = field(default=True)
```
ashb, line 134:
```suggestion
    temp: ClassVar[bool] = True
```

(And then along with a matching change to False in Table I think you can leave it slotted too?)

utkarsharma2 (approved):

python-sdk/src/astro/sql/table.py
kaxil: Didn't work unfortunately -- 
```
ImportError while loading conftest '/Users/kaxilnaik/Documents/GitHub/astronomer/astro-sdk/python-sdk/conftest.py'.
conftest.py:14: in <module>
    from astro.constants import Database, FileLocation, FileType
src/astro/__init__.py:17: in <module>
    import astro.sql  # noqa: F401
src/astro/sql/__init__.py:4: in <module>
    from astro.sql.operators.append import AppendOperator, append
src/astro/sql/operators/append.py:8: in <module>
    from astro.databases import create_database
src/astro/databases/__init__.py:5: in <module>
    from astro.databases.base import BaseDatabase
src/astro/databases/base.py:20: in <module>
src/astro/sql/table.py:139: in <module>
    class Table(BaseTable, Dataset):
E   TypeError: multiple bases have instance lay-out conflict
```

kaxil: >However, [it’s not possible](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots) to inherit from more than one class that has attributes in __slots__ (you will get an TypeError: multiple bases have instance lay-out conflict).

So probably since `airflow.Dataset` has slots and even `BaseTable` has slot It doesn't work :shrug:

https://www.attrs.org/en/stable/glossary.html?highlight=multiple%20bases%20have%20instance%20lay-out%20conflict#term-slotted-classes

.pre-commit-config.yaml
```diff
@@ -73,11 +73,11 @@ repos:
     hooks:
       - id: isort
         name: Run isort for python-sdk
-        args: [ "--profile", "black" ]
+        args: ["--profile=black"]
```
feluelle, line 76: Is this necessary?  I like the current way more tbh.

python-sdk/src/astro/airflow/datasets.py
```diff
@@ -29,8 +29,15 @@ def kwargs_with_datasets(
         and settings.AUTO_ADD_INLETS_OUTLETS
         and input_datasets
     ):
+        # Remove non-dataset objects like Temp tables (unless passed by user)
+        input_datasets = (
+            input_datasets if isinstance(input_datasets, list) else [input_datasets]
+        )
+        input_datasets = [
+            input_ds for input_ds in input_datasets if isinstance(input_ds, Dataset)
+        ]
```
feluelle, line 38: hmm 🤔 ? According to the type definition only `Dataset` type is supported anyway.

`input_datasets: list[Dataset] | Dataset | None = None`

python-sdk/src/astro/sql/table.py
```diff
@@ -125,6 +122,42 @@ def name(self, value: str) -> None:
...
+    :param conn_id: The Airflow connection id. This will be used to identify the right database type at the runtime
+    :param metadata: A metadata object which will have database or schema name
+    :param columns: columns which define the database table schema.
+    """
+
+    uri: str = field(init=False)
+    extra: dict | None = field(init=False, factory=dict)
+
+    def __new__(cls, *args, **kwargs):
+        name = kwargs.get("name", "") or (args[0] if len(args) > 0 else "")
```
feluelle, line 155:
```suggestion
        name = kwargs.get("name") or args and args[0] or ""
```

python-sdk/src/astro/sql/table.py
```diff
@@ -125,6 +122,42 @@ def name(self, value: str) -> None:
...
+        temp = kwargs.get("temp", False)
+        if temp or (not name or name.startswith("_tmp")):
```
feluelle, line 157:
```suggestion
        if temp or not name or name.startswith("_tmp"):
```

python-sdk/src/astro/airflow/datasets.py
kaxil: Yes and No :) --- It is mainly for removing `TempTable` from being added as inlets/outlets... and `TempTable` get generated by the hack in this PR by calling `Table()` without `name`.

Here `Table` is a Dataset class but `TempTable` isn't -- and instead of checking those in each operator, it would be better to check it here.

.pre-commit-config.yaml
kaxil, line 76:
```suggestion
        args: [ "--profile", "black" ]
```

.pre-commit-config.yaml
```diff
@@ -73,11 +73,11 @@ repos:
...
         files: ^python-sdk/
       - id: isort
         name: Run isort for sql-cli
-        args: [ "--profile", "black" ]
+        args: ["--profile=black"]
```
kaxil, line 80:
```suggestion
        args: [ "--profile", "black" ]
```

python-sdk/src/astro/sql/table.py
kaxil: I like the one with brackets better for readability

.pre-commit-config.yaml
kaxil: reverted

python-sdk/src/astro/airflow/datasets.py
feluelle: Okay, but then we should update the type definition to reflect that. From your comment it sounds like `TempTable` is supported but will be filtered out from `input_datasets`. So in my opinion it should be: `input_datasets: list[Dataset | TempTable] | Dataset | TempTable | None = None` or?

python-sdk/src/astro/sql/table.py
feluelle: Valid point! :)

Issue #866, opened 2022-09-16
[Dynamic Task Templates]:Change `sql_statement` param to `sql`


phanikumv:
**Describe the bug**
get_value_list has a kwarg called sql_statement which is too verbose, change it to sql

**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Write the DAG  '...'
2. Create connection '....'
3. Run using  '....'

**Expected behavior**

**Screenshots**

**Additional context**

phanikumv:
Duplicate of https://github.com/astronomer/astro-sdk/issues/865 . Hence closing

PR #867, opened 2022-09-16
Rename param sql_statement to sql in get_value_list

pankajastro:
# Description
Rename param `sql_statement` to `sql` in `get_value_list` dynamic task template pattern in order to make public interfaces more consistent. 
closes: #865 #866

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -112,7 +112,8 @@ class BaseDatabase(ABC):
                 DeprecationWarning,
                 stacklevel=2,
             )
-        sql = sql or kwargs.get("sql_statement")  # type: ignore
+        if kwargs.get("sql_statement"):
+            sql = kwargs.get("sql_statement")  # type: ignor
 
         if isinstance(sql, str):
             result = self.connection.execute(sqlalchemy.text(sql), parameters)
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -321,7 +321,7 @@ class BaseDatabase(ABC):
         statement = self._create_table_statement.format(
             self.get_table_qualified_name(target_table), statement
         )
-        self.run_sql_query(statement, parameters)
+        self.run_sql(statement, parameters)
 
     def drop_table(self, table: BaseTable) -> None:
         """
@@ -332,7 +332,7 @@ class BaseDatabase(ABC):
         statement = self._drop_table_statement.format(
             self.get_table_qualified_name(table)
         )
-        self.run_sql_query(statement)
+        self.run_sql(statement)
 
     # ---------------------------------------------------------
     # Table load methods
@@ -586,7 +586,7 @@ class BaseDatabase(ABC):
         # TODO: We should fix the following Type Error
         # incompatible type List[ColumnClause[<nothing>]]; expected List[Column[Any]]
         sql = insert(target_table_sqla).from_select(target_columns, sel)  # type: ignore[arg-type]
-        self.run_sql_query(sql=sql)
+        self.run_sql(sql=sql)
 
     def merge_table(
         self,
@@ -673,7 +673,7 @@ class BaseDatabase(ABC):
         # doesn't actually create a schema.
         if schema and not self.schema_exists(schema):
             statement = self._create_schema_statement.format(schema)
-            self.run_sql_query(statement)
+            self.run_sql(statement)
 
     def schema_exists(self, schema: str) -> bool:
         """
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import warnings
 from abc import ABC
-from typing import Any, Callable, Mapping, Union
+from typing import Any, Callable, Mapping
 
 import pandas as pd
 import sqlalchemy
@@ -112,7 +112,7 @@ class BaseDatabase(ABC):
                 DeprecationWarning,
                 stacklevel=2,
             )
-        sql: Union[str, ClauseElement] = sql or kwargs.get("sql_statement")
+        sql = sql or kwargs.get("sql_statement")  # type: ignore
 
         if isinstance(sql, str):
             result = self.connection.execute(sqlalchemy.text(sql), parameters)
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -105,14 +105,13 @@ class BaseDatabase(ABC):
         if parameters is None:
             parameters = {}
 
-        if "sql_statement" in kwargs:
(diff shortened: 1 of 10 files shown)


pankajkoti (commented):
There is a `run_sql` method in `databases/base.py` which still accepts `sql_statement` as an argument https://github.com/astronomer/astro-sdk/blob/ddf7f12b723595d88991d286fc39612cec6de455/python-sdk/src/astro/databases/base.py#L82
I think that may still cause a confusion and we might use `sql_statement` across other places later by considering this argument. I believe we called it `sql_statement` here  in `get_value_list` by referring to that in the first place. In my opinion, it makes sense to rename that argument to `sql` too and correct the usages in our codebase.

pankajastro:
> There is a `run_sql` method in `databases/base.py` which still accepts `sql_statement` as an argument
> 
> https://github.com/astronomer/astro-sdk/blob/ddf7f12b723595d88991d286fc39612cec6de455/python-sdk/src/astro/databases/base.py#L82
> 
> . I think that may still cause a confusion and we might use `sql_statement` across other places later by considering this argument. I believe we called it `sql_statement` here by referring to that in the first place. In my opinion, it makes sense to rename that argument to `sql` too and correct the usages in our codebase.

`run_sql` we have already released so not sure whether to deprecate that now or not. WDYT?

pankajkoti:
> > There is a `run_sql` method in `databases/base.py` which still accepts `sql_statement` as an argument
> > https://github.com/astronomer/astro-sdk/blob/ddf7f12b723595d88991d286fc39612cec6de455/python-sdk/src/astro/databases/base.py#L82
> > 
> > . I think that may still cause a confusion and we might use `sql_statement` across other places later by considering this argument. I believe we called it `sql_statement` here by referring to that in the first place. In my opinion, it makes sense to rename that argument to `sql` too and correct the usages in our codebase.
> 
> `run_sql` we have already released so not sure whether to deprecate that now or not. WDYT?

Since this is a base class method, I don't think external users are using it. My point to rename this was so that we have the argument consistent in our codebase and no one later uses `sql_statement` in other places by referring to this argument name.

What are your thoughts? @tatiana @phanikumv @utkarsharma2 @sunank200 @feluelle @kaxil ?

kaxil:
@pankajastro  PR description has some unfilled details !

kaxil:
>Since this is a base class method, I don't think external users are using it. My point to rename this was so that we have the argument consistent in our codebase and no one later uses sql_statement in other places by referring to this argument name.


No strong opinion on that since it is not a user-facing class.

kaxil (approved):

phanikumv:
> > > There is a `run_sql` method in `databases/base.py` which still accepts `sql_statement` as an argument
> > > https://github.com/astronomer/astro-sdk/blob/ddf7f12b723595d88991d286fc39612cec6de455/python-sdk/src/astro/databases/base.py#L82
> > > 
> > > . I think that may still cause a confusion and we might use `sql_statement` across other places later by considering this argument. I believe we called it `sql_statement` here by referring to that in the first place. In my opinion, it makes sense to rename that argument to `sql` too and correct the usages in our codebase.
> > 
> > 
> > `run_sql` we have already released so not sure whether to deprecate that now or not. WDYT?
> 
> Since this is a base class method, I don't think external users are using it. My point to rename this was so that we have the argument consistent in our codebase and no one later uses `sql_statement` in other places by referring to this argument name.
> 
> What are your thoughts? @tatiana @phanikumv @utkarsharma2 @sunank200 @feluelle @kaxil ?

We need to have consistency across the repo and risk is less since its a base class internal method. So my thought is to rename it there as well

feluelle:
I am all-in for deprecating sql_statement and making it consistent across the codebase. Note that renaming it is a breaking change. So we might want to deprecate this arg by creating a new function and calling the new function from the old function and raising a deprecation warning inbetween.

pankajastro:
> I am all-in for deprecating sql_statement and making it consistent across the codebase. Note that renaming it is a breaking change. So we might want to deprecate this arg by creating a new function and calling the new function from the old function and raising a deprecation warning inbetween.

Any thought on new function name?

pankajastro:
> We need to have consistency across the repo and risk is less since its a base class internal method. So my thought is to rename it there as well

renaming looks like a breaking change since there is not sign that this function is internal.

utkarsharma2:
I don't think we should have breaking changes in a minor release. 

cc: @kaxil

kaxil:
We don't need to create a separate `run_sql_query` method, we can just deprecate the kwarg `sql_statement`

python-sdk/src/astro/databases/base.py
```diff
@@ -81,23 +82,36 @@ def run_sql(
...
+            "Please use  `run_sql_query` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.run_sql_query(sql=sql_statement, parameters=parameters)
+
+    def run_sql_query(
+        self,
+        sql: str | ClauseElement,
+        parameters: dict | None = None,
```
kaxil, line 97:
```suggestion
        sql: str | ClauseElement,
        parameters: dict | None = None,
```

python-sdk/src/astro/databases/base.py
```diff
@@ -81,23 +82,36 @@ def run_sql(
...
-        :param sql_statement: Contains SQL query to be run against database
+        :param sql: Contains SQL query to be run against database
         :param parameters: Optional parameters to be used to render the query
         """
         if parameters is None:
             parameters = {}
 
-        if isinstance(sql_statement, str):
-            result = self.connection.execute(sqlalchemy.text(sql_statement), parameters)
+        if isinstance(sql, str):
```
kaxil, line 111:
```suggestion

        if "sql_statement" in kwargs:
            warnings.warn(
                "`sql_statement` is deprecated and will be removed in future release"
                "Please use  `sql` param instead.",
                DeprecationWarning,
                stacklevel=2,
            )
        sql = sql or kwargs.get("sql_statement")    
        if isinstance(sql, str):
```

kaxil: ^^ something like this (untested)

python-sdk/src/astro/databases/base.py
kaxil: We don't need a new method -- we just need to deprecate a kwarg

kaxil: something like https://github.com/apache/airflow/commit/b220fe60d5302546d85ee15cf6c8eaa5859316d3

python-sdk/src/astro/databases/base.py
utkarsharma2: Then it won't be a breaking change either

pankajastro:
I'm ignoring code coverage at few places in this PR created a followup ticket to fix it https://github.com/astronomer/astro-sdk/issues/870

utkarsharma2 (approved):

PR #871, opened 2022-09-16
Fix issues with pickling & unpickling the `Table` object

utkarsharma2:
# Description
Currently, Python-sdk is falling with pickling error.

```
airflow.exceptions.AirflowException: When using a synchronous executor (e.g. SequentialExecutor and DebugExecutor), the first run of this task will fail on purpose, so the single worker thread is unblocked to execute other tasks. The task is set up for retry and eventually works.
[2022-09-16 22:52:28,847] {backfill_job.py:190} ERROR - Task instance <TaskInstance: example_google_bigquery_gcs_load_and_save.cleanup backfill__2016-01-01T00:00:00+00:00 [failed]> failed
[2022-09-16 22:52:28,851] {backfill_job.py:370} INFO - [backfill progress] | finished run 0 of 1 | tasks waiting: 3 | succeeded: 1 | running: 0 | failed: 1 | skipped: 0 | deadlocked: 0 | not ready: 3
[2022-09-16 22:52:28,865] {base_executor.py:95} INFO - Adding to queue: ['<TaskInstance: example_google_bigquery_gcs_load_and_save.extract_top_5_movies backfill__2016-01-01T00:00:00+00:00 [queued]>']
[2022-09-16 22:52:28,872] {base_executor.py:95} INFO - Adding to queue: ['<TaskInstance: example_google_bigquery_gcs_load_and_save.save_file_to_gcs backfill__2016-01-01T00:00:00+00:00 [queued]>']
[2022-09-16 22:52:28,901] {debug_executor.py:85} ERROR - Failed to execute task: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte.
  File "/Users/utkarsharma/sandbox/astronomer/astro/.nox/dev/lib/python3.8/site-packages/airflow/models/xcom.py", line 618, in deserialize_value
    return pickle.loads(result.value)
_pickle.UnpicklingError: state is not a dictionary

```
## What is the new behavior?
To fix the above issue we need to add __getstate__() to the table class

Nope

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/table.py
+++ b/python-sdk/src/astro/sql/table.py
@@ -63,6 +63,7 @@ class BaseTable:
     columns: list[Column] = field(factory=list)
     temp: bool = field(default=False)
 
+    # We need this method to pickle Table object, without this we cannot push/pull this object from xcom.
     def __getstate__(self):
         return self.__dict__
 
--- a/python-sdk/src/astro/sql/table.py
+++ b/python-sdk/src/astro/sql/table.py
@@ -63,6 +63,9 @@ class BaseTable:
     columns: list[Column] = field(factory=list)
     temp: bool = field(default=False)
 
+    def __getstate__(self):
+        return self.__dict__
+
     def __attrs_post_init__(self) -> None:
         if not self._name or self._name.startswith("_tmp"):
             self.temp = True
--- a/python-sdk/tests/sql/test_table.py
+++ b/python-sdk/tests/sql/test_table.py
@@ -3,9 +3,8 @@ from datetime import datetime
 
 import pytest
 from airflow import DAG
-from astro.sql.table import Metadata, Table, TempTable
-
 from astro.sql import get_value_list
+from astro.sql.table import Metadata, Table, TempTable
 
 
 def test_table_with_explicit_name():
--- a/python-sdk/tests/sql/test_table.py
+++ b/python-sdk/tests/sql/test_table.py
@@ -3,8 +3,9 @@ from datetime import datetime
 
 import pytest
 from airflow import DAG
-from astro.sql import get_value_list
 from astro.sql.table import Metadata, Table, TempTable
+
+from astro.sql import get_value_list
 
 
 def test_table_with_explicit_name():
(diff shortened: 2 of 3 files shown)


python-sdk/tests/sql/test_table.py
```diff
@@ -145,3 +146,8 @@ def test_temp_table(table):
     assert table.temp
     assert isinstance(table, TempTable)
     assert not isinstance(table, Table)
+
+
+def test_if_table_object_can_be_pickled():
+    """Verify if we can pickle Table object"""
+    pickle.loads(pickle.dumps(Table(name="test")))
```
kaxil, line 153: Let's add this test for File object. The reason `self.__dict__` is not available for `File` is because it uses `slots`.

>A class whose instances have no [object.__dict__](https://docs.python.org/3/library/stdtypes.html#object.__dict__) attribute and [define](https://docs.python.org/3/reference/datamodel.html#slots) their attributes in a [object.__slots__](https://docs.python.org/3/reference/datamodel.html#object.__slots__) attribute instead. In attrs, they are created by passing slots=True to @attr.s (and are on by default in [attrs.define()](https://www.attrs.org/en/stable/api.html#attrs.define)/[attrs.mutable()](https://www.attrs.org/en/stable/api.html#attr.attrs.mutable)/[attrs.frozen()](https://www.attrs.org/en/stable/api.html#attr.attrs.frozen)).

utkarsharma2: Got it

kaxil: But atleast this test will verify that you can pickle and unpickle the object.

kaxil: Would the following be a better test?

```python
table = Table(name="test")
assert pickle.loads(pickle.dumps(table)) == table
```

cc @ashb

python-sdk/src/astro/sql/table.py
```diff
@@ -63,6 +63,9 @@ class BaseTable:
     columns: list[Column] = field(factory=list)
     temp: bool = field(default=False)
 
+    def __getstate__(self):
+        return self.__dict__
```
kaxil, line 67: Worth adding a comment on why this is required?

kaxil (approved):
Approving so you are not blocked but please address the review comments

python-sdk/tests/sql/test_table.py
utkarsharma2: @kaxil Ya this seems better.

PR #872, opened 2022-09-16
Check for filetype only for pattern and fix coverage

sunank200:
# Description

Currently is filetype is passed then pattern doesn't need to be resolved to check native autodetect

closes: #873

## What is the new behavior?

- Create file type in case not provided before checking for native autodetect schema

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -17,6 +17,8 @@ from astro.constants import (
     MergeConflictStrategy,
 )
 from astro.exceptions import DatabaseCustomError, NonExistentTableException
+from astro.files import File, resolve_file_path_pattern
+from astro.files.types import create_file_type
 from astro.files.types.base import FileType as FileTypeConstants
 from astro.settings import LOAD_TABLE_AUTODETECT_ROWS_COUNT, SCHEMA
 from astro.sql.table import BaseTable, Metadata
@@ -25,9 +27,6 @@ from sqlalchemy import column, insert, select
 from sqlalchemy.sql import ClauseElement
 from sqlalchemy.sql.elements import ColumnClause
 from sqlalchemy.sql.schema import Table as SqlaTable
-
-from astro.files import File, resolve_file_path_pattern
-from astro.files.types import create_file_type
 
 
 class BaseDatabase(ABC):
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -17,8 +17,6 @@ from astro.constants import (
     MergeConflictStrategy,
 )
 from astro.exceptions import DatabaseCustomError, NonExistentTableException
-from astro.files import File, resolve_file_path_pattern
-from astro.files.types import create_file_type
 from astro.settings import LOAD_TABLE_AUTODETECT_ROWS_COUNT, SCHEMA
 from astro.sql.table import BaseTable, Metadata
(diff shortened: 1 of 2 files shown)


python-sdk/src/astro/databases/base.py
```diff
@@ -744,11 +745,14 @@ def check_schema_autodetection_is_supported(self, source_file: File) -> bool:
...
-            path=source_file.path, filetype=source_file.type.name
+        source_filetype = (
+            source_file
+            if isinstance(source_file.type, FileTypeConstants)
+            else create_file_type(path=source_file.path, filetype=source_file.type)  # type: ignore
         )
+
         is_source_filetype_supported = (
-            (source_filetype in filetype_supported.get("filetype"))  # type: ignore
+            (source_filetype.type.name in filetype_supported.get("filetype"))  # type: ignore
```
kaxil, line 755: Can you add a test to avoid regression please?

sunank200: Added the test https://github.com/astronomer/astro-sdk/pull/872/commits/d572c8c9b5c1749b6bfe96e3bc61bb62f3060cb6

kaxil (approved):

python-sdk/tests/databases/test_base_database.py
```diff
@@ -59,6 +59,33 @@ def test_create_table_using_columns_raises_exception():
...
+    """
+    Test the condition native schema autodetection for files and prefixes
+    """
+    db = create_database("gcp_conn")
+    assert (
+        db.check_schema_autodetection_is_supported(
+            source_file=File(path="gs://bucket/prefix", filetype=FileType.CSV)
+        )
+        is True
+    )
```
kaxil, line 72:
```suggestion
    assert (
        db.check_schema_autodetection_is_supported(
            source_file=File(path="gs://bucket/prefix", filetype=FileType.CSV)
        )
    )
```

python-sdk/tests/databases/test_base_database.py
```diff
@@ -59,6 +59,33 @@ def test_create_table_using_columns_raises_exception():
...
+
+    assert (
+        db.check_schema_autodetection_is_supported(
+            source_file=File(path="gs://bucket/prefix/key.csv")
+        )
+        is True
+    )
```
kaxil, line 79:
```suggestion
    assert (
        db.check_schema_autodetection_is_supported(
            source_file=File(path="gs://bucket/prefix/key.csv")
        )
    )
```

python-sdk/tests/databases/test_base_database.py
```diff
@@ -59,6 +59,33 @@ def test_create_table_using_columns_raises_exception():
...
+
+    assert (
+        db.check_schema_autodetection_is_supported(
+            source_file=File(path="s3://bucket/prefix/key.csv")
+        )
+        is False
+    )
```
kaxil, line 86:
```suggestion
    assert not (
        db.check_schema_autodetection_is_supported(
            source_file=File(path="s3://bucket/prefix/key.csv")
        )
    )
```

python-sdk/tests/databases/test_base_database.py
kaxil: `is True` is redundant

python-sdk/tests/databases/test_base_database.py
kaxil: `is True` is redundant

Issue #574, opened 2022-07-26
Introduce multiple file type options for a dataset of a specific size.


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we have only one option per type and size, we should have multiple options for a particular size since there are cases when a dataset is not consumable by a native or default path, which causes issues in running benchmarking. 

**Describe the solution you'd like**
If we have multiple options for datasets of different sizes we can probably run benchmarking script smoothly.

**Ideal datasets**
1. Clean dataset  - Synthetic dataset
2. Available in all file formats supported - CSV, NDJSON, Parquet
3. For all Sizes 10KB, 100KB, 10MB, 100MB, 1GB, 5GB, 10GB
4.  For Single, Multiple files
5. This dataset needs to be available on GCS and S3

**Steps to Reproduce**

Case 1 - default path:

Using default path - pandas was complaining about below dataset
s3://astro-sdk-test/benchmark/trimmed/stackoverflow/stackoverflow_posts.ndjson
Worked with
s3://astro-sdk-test/benchmark/trimmed/test/stackoverflow_posts.ndjson

Case 2 - Native path:
Using the below dataset resulted in an invalid argument in the Data Transfer service - Bigquey
s3://astro-sdk-test/benchmark/trimmed/pypi/

Case 3 - Native path:
Using the below dataset resulted in an invalid argument in the Data Transfer service - Bigquey
s3://astro-sdk-test/benchmark/trimmed/github/github-archive/

Run benchmarking script with an above-mentioned dataset with Bigquery table as output table.

phanikumv:
Facing issue for 10GB Parquet file - Ankit is currently looking into this

PR #879, opened 2022-09-20
Generate synthetic dataset

sunank200:
# Description

Currently, we have only one option per type and size, we should have multiple options for a particular size since there are cases when a dataset is not consumable by a native or default path, which causes issues in running benchmarking.

closes: #574 

## What is the new behavior?

**Ideal datasets**
- Clean dataset - Synthetic dataset
- Available in all file formats supported - CSV, NDJSON, Parquet
- For all Sizes 10KB, 100KB, 10MB, 100MB, 1GB, 5GB, 10GB
- For Single, Multiple files
- This dataset is available on GCS and S3

No

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/datasets.md
+++ b/python-sdk/tests/benchmark/datasets.md
@@ -43,9 +43,11 @@ Within `s3://astro-sdk-test/benchmark/fake_data/`, following are the files with 
 
 ## Origin of the synthetic datasets
 
-To make the benchmarking results uniform, we have generated synthetic dataset for various size for all the file types:
+To make the benchmarking results uniform, we have generated synthetic dataset for various size for all the file types.
 
 1. CSV files
+
+The [script](synthetic_csv_generator.py) is used to generate the Parquet file.
 
 - GCS path
 
@@ -59,6 +61,8 @@ To make the benchmarking results uniform, we have generated synthetic dataset fo
 | two_gb     | 2 GB   | csv    | gs://astro-sdk/benchmark/synthetic-dataset/csv/two_gb.csv     |
 | five_gb    | 5 GB   | csv    | gs://astro-sdk/benchmark/synthetic-dataset/csv/five_gb.csv    |
 | ten_gb     | 10 GB  | csv    | gs://astro-sdk/benchmark/synthetic-dataset/csv/ten_gb.csv     |
+| five_gb    | 5 GB   | csv    | gs://astro-sdk/benchmark/synthetic-dataset/csv/five_gb/       |
+| ten_gb     | 10 GB  | csv    | gs://astro-sdk/benchmark/synthetic-dataset/csv/ten_gb/        |
 
 - AWS S3 path
 
@@ -72,8 +76,12 @@ To make the benchmarking results uniform, we have generated synthetic dataset fo
 | two_gb     | 2 GB   | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/two_gb.csv     |
 | five_gb    | 5 GB   | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/five_gb.csv    |
 | ten_gb     | 10 GB  | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/ten_gb.csv     |
+| five_gb    | 5 GB   | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/five_gb/       |
+| ten_gb     | 10 GB  | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/ten_gb/        |
 
 2. Parquet files
+
+The [script](synthetic_parquet_generator.py) is used to generate the Parquet file.
 
 - GCS path
 
@@ -104,6 +112,8 @@ To make the benchmarking results uniform, we have generated synthetic dataset fo
 | ten_gb     | 10 GB  | parquet    | s3://astro-sdk/benchmark/synthetic-dataset/parquet/ten_gb/            |
 
 3. NDJSON files
+
+The [script](synthetic_ndjson_generator.py) is used to generate the NDJSON file.
 
 - GCS path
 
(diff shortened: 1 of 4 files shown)


feluelle (approved):
Nice! 🤩

How are we uploading the files? Should we use IAC (e.g. terraform) to make sure the files exist?

python-sdk/tests/benchmark/synthetic_ndjson_generator.py
```diff
@@ -0,0 +1,64 @@
...
+    # Iterate the loop based on the input value and generate fake data
+    for n in range(0, records):
+        cus = {
+            "id": n,
+            "name": fake.name(),
+            "address": fake.address(),
+            "email": str(fake.email()),
+            "phone": str(fake.phone_number()),
+        }
+        customer.append(cus)
```
feluelle, line 23:
```suggestion
    # Declare an empty list
    customer = [
        {
            "id": n,
            "name": fake.name(),
            "address": fake.address(),
            "email": str(fake.email()),
            "phone": str(fake.phone_number()),
        }
        for n in range(0, records)
    ]
    # Iterate the loop based on the input value and generate fake data
    for n in range(0, records):
        cus = {
            "id": n,
            "name": fake.name(),
            "address": fake.address(),
            "email": str(fake.email()),
            "phone": str(fake.phone_number()),
        }
        customer.append(cus)
```

python-sdk/tests/benchmark/synthetic_ndjson_generator.py
```diff
@@ -0,0 +1,62 @@
...
+    ]
+    # Write the data into the NDJSON file
+    with open(file_name, "w") as fp:
+        ndjson.dump(customer, fp)
+
+    print(f"File {file_name} has been created.")
+
+
+# create ten_kb ndjson file
+ten_kb = "fake_dataset/ndjson/ten_kb.ndjson"
```
utkarsharma2, line 30: @sunank200 I think we can make this quicker for text-based data:
1. Create a file of 1MB or so
2. Make copy of the file generated
3. Then concrete it. `cat copy1.ndjson copy2.ndjson copy3.ndjson > newFile.ndjson`

python-sdk/tests/benchmark/datasets.md
```diff
@@ -41,6 +41,98 @@ Within `s3://astro-sdk-test/benchmark/fake_data/`, following are the files with
...
+
+|            | size   | format | path to csv file(s)                                           |
+|------------|--------|--------|---------------------------------------------------------------|
+| ten_kb     | 10 KB  | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/ten_kb.csv     |
+| hundred_kb | 100 KB | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/hundred_kb.csv |
+| ten_mb     | 10 MB  | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/ten_mb.csv     |
+| hundred_mb | 100 MB | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/hundred_mb.csv |
+| one_gb     | 1 GB   | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/one_gb.csv     |
+| two_gb     | 2 GB   | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/two_gb.csv     |
+| five_gb    | 5 GB   | csv    | s3://astro-sdk/benchmark/synthetic-dataset/csv/five_gb.csv    |
```
utkarsharma2, line 73: @sunank200 Are we adding multiple files for 5GB and 10GB files for CSV?

utkarsharma2: Maybe we can just copy 10 * 100MB files in a folder?

feluelle: I thought these are single files?! Multiple files should be tested separately or?

sunank200: For csv I have created a single large 10 GB and 5 GB files but for ndjson I have created multiple files (folder)

sunank200:
> Nice! 🤩
> 
> How are we uploading the files? Should we use IAC (e.g. terraform) to make sure the files exist?

@feluelle I was thinking we could create a volume and attach that volume every time while running the job. WDYT?

python-sdk/tests/benchmark/synthetic_ndjson_generator.py
sunank200: Well, there is an `id` which I wanted to keep unique.

python-sdk/tests/benchmark/datasets.md
feluelle: Okay, that's fine but we should test both cases - single file vs multi file or? If our sdk does not support 5gb file we should document it, if we have not already.

feluelle:
What kind of volume? Don't we need different volumes per cloud provider? Isn't it easier to upload them once to the buckets and just ensure we have them in code so they don't accidentally get removed/lost.

utkarsharma2:
@feluelle I think what @sunank200 meant by adding volume is for benchmarking local to the Database path. For the below-listed cases, we don't need any volume.
1. S3 to database
2. GCS to database

feluelle:
Ah okay, thanks.

python-sdk/tests/benchmark/datasets.md
sunank200: Added both single file and multi-file in this https://github.com/astronomer/astro-sdk/pull/879/commits/6d437918e0027f45527a3a01b9833c82bedd0901.

We do support 5 GB file in loadfile

utkarsharma2: I think we should create multiple files as well.

utkarsharma2:
@sunank200 I think we should also add multiple files in the case of CSV since different codebases handle different file types' paths. it would be nice to have these individual paths tested.

sunank200:
> @sunank200 I think we should also add multiple files in the case of CSV since different codebases handle different file types' paths. it would be nice to have these individual paths tested.

I have already added that and resolved your comment in this commit: https://github.com/astronomer/astro-sdk/pull/879/commits/6d437918e0027f45527a3a01b9833c82bedd0901

Issue #881, opened 2022-09-21
Load `.data` file into a pandas dataframe from a url


pankajastro:
**Describe the bug**
I have a .data file I am trying to load into a pandas dataframe from a url. The file is basically a CSV format but has the .data extension.

**To Reproduce**
Example code
```
data_url = "https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data"  # URL for Iris data API

dataframe = aql.load_file(
            input_file=File(path=data_url, filetype=FileType.CSV)
        )
```

**Expected behavior**
It should be able to load data in pandas dataframe

**Additional context**
error log

```
ValueError: Unsupported filetype 'data' from file 'https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data'.
```

- [x] To add a test case covering this functionality

phanikumv:
Ankit will start on this today

PR #891, opened 2022-09-22
Add filetype when resolving path in case of loading into dataframe

sunank200:
# Description

When a `.data` file is loaded into a pandas dataframe from a URL by specifying file type, it fails. The file is basically a CSV format but has the .data extension.

Example code
```
data_url = "https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data"  # URL for Iris data API

dataframe = aql.load_file(
            input_file=File(path=data_url, filetype=FileType.CSV)
        )
```

error log
```
ValueError: Unsupported filetype 'data' from file 'https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data'.
```

closes: #881

## What is the new behavior?

- Pass the filetype while resolving files for loading into data frame
- Add the corresponding test for this
-

No

### Checklist

### DIFF
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -4,11 +4,10 @@ from datetime import datetime, timedelta
 
 import sqlalchemy
 from airflow.models import DAG
-from astro.constants import FileType
-from astro.sql.table import Metadata, Table
-
 from astro import sql as aql
+from astro.constants import FileType
 from astro.files import File
+from astro.sql.table import Metadata, Table
 
 # To create IAM role with needed permissions,
 # refer: https://www.dataliftoff.com/iam-roles-for-loading-data-from-s3-into-redshift/
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -4,10 +4,11 @@ from datetime import datetime, timedelta
 
 import sqlalchemy
 from airflow.models import DAG
-from astro import sql as aql
 from astro.constants import FileType
-from astro.files import File
 from astro.sql.table import Metadata, Table
+
+from astro import sql as aql
+from astro.files import File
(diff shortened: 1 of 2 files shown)


feluelle (approved):
LGTM

python-sdk/example_dags/example_load_file.py
```diff
@@ -19,6 +19,7 @@
     "retries": 1,
     "retry_delay": 0,
 }
+data_url = "https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data"  # URL for Iris data API
```
utkarsharma2, line 22: @sunank200 Let's not add test cases dependency to external URL, can we move this to S3/GCS?

python-sdk/example_dags/example_load_file.py
```diff
@@ -221,4 +222,8 @@
     )
     # [END load_file_example_17]
 
+    # [START load_file_example_18]
```
utkarsharma2, line 225: @sunank200 Also should we move this to tests instead? and/or convert to Unit test instead? Since we are planning to move away from integration tests as much as possible.

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -128,6 +128,8 @@ def load_data_to_dataframe(self, input_file: File) -> pd.DataFrame | None:
         for file in resolve_file_path_pattern(
             input_file.path,
             input_file.conn_id,
+            normalize_config=self.normalize_config,
```
utkarsharma2, line 131: @sunank200  How adding `normalize_config=self.normalize_config` affected a CSV file?

utkarsharma2: This should have been part of earlier PRs right? the one with pattern resolution?

Issue #834, opened 2022-09-13
Benchmark script is failing to run on GKE with images built on Mac M1


pankajkoti:
We're getting the below error when images are built on Mac M1 and are submitted to a GKE cluster using the `make run_job` command.

```
(base) ➜  ~ kubectl -n benchmark logs benchmark-rb478-btljz
standard_init_linux.go:228: exec user process caused: exec format error
```

Refer https://blog.jaimyn.dev/how-to-build-multi-architecture-docker-images-on-an-m1-mac/ to try fixing

pankajkoti:
I will take a look at this tomorrow. @pankajastro in case if you get a chance to look at this today, please check.

PR #893, opened 2022-09-23
Fix benchmark script build on mac m1

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/834

Let's build the image with `--platform` option since if build on M1 Mac then it is not working in other architecture. For now, I have added the platform value `linux/amd64` if needed we can add multiple values in future like `linux/amd64,linux/arm64 `  

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/Dockerfile
+++ b/python-sdk/tests/benchmark/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bullseye AS build
+FROM --platform=linux/amd64 debian:bullseye AS build
 MAINTAINER Astronomer DAG Authoring Team <airflow_dag_authors@astronomer.io>
 
 ARG GIT_HASH=""
--- a/python-sdk/tests/benchmark/Dockerfile
+++ b/python-sdk/tests/benchmark/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 debian:bullseye AS build
+FROM debian:bullseye AS build
 MAINTAINER Astronomer DAG Authoring Team <airflow_dag_authors@astronomer.io>
 
 ARG GIT_HASH=""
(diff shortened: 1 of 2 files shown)


python-sdk/tests/benchmark/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:bullseye AS build
+FROM --platform=linux/amd64 debian:bullseye AS build
```
feluelle, line 1: You can also pass this option to the `docker` command line which I think is what you want as not everyone uses the same system.

feluelle: See [docker build](https://docs.docker.com/engine/reference/commandline/build/)

sunank200: +1

sunank200, line 1: Would be useful to pass this as a docker build option.

kaxil (approved):

pankajkoti:
Nice fix @pankajastro . Thank you!

python-sdk/tests/benchmark/Makefile
```diff
@@ -5,7 +5,7 @@ GCP_PROJECT ?= astronomer-dag-authoring
 GIT_HASH = $(shell git log -1 --format=%h)
 APP ?= benchmark
 CONTAINER_REGISTRY=gcr.io/$(GCP_PROJECT)/$(APP)
-
+PLATFORM="linux/amd64"
```
feluelle, line 8: This is not configurable right now, or? Everyone using this script has to run it on m1. Maybe we should make it configurable through an env var.

pankajastro: yes, it is not configurable. I think the platform option in the docker build is more about the target machine where we want to run this image?
I think it's ok for now but let me know if you want this value from env I'm happy to do that change and keep `linux/amd64` as the default value

feluelle: Yes, I think some of us do not have a m1.

feluelle: So it would be nice to be able to set it without editing the script.

pankajastro: PR should take care of this https://github.com/astronomer/astro-sdk/pull/1002

PR #897, opened 2022-09-23
Render sql code with parameters in BaseSQLDecoratedOperator

feluelle:
# Description

Currently, when using operators inheriting from `BaseSQLDecoratedOperator` we are unable to view the rendered SQL in the UI.

closes: #1003 
depends on: #1004 

## What is the new behavior?

- add sql and parameters to BaseSQLDecoratedOperator template_fields
- add .sql to BaseSQLDecoratedOperator template_ext
- add check for template_fields and template_ext

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -137,10 +137,9 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         :param context: Dict with values to apply on content
         :param jinja_env: Jinja environment
         """
-        output = super().render_template_fields(context, jinja_env)
         context = self._enrich_context(context)
 
-        return output
+        return super().render_template_fields(context, jinja_env)
 
     def execute(self, context: Context) -> None:
         self._enrich_context(context)
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -179,7 +179,7 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         return super().render_template_fields(context, jinja_env)
 
     def execute(self, context: Context) -> None:
-        self._enrich_context(context)
+        context = self._enrich_context(context)
 
         # TODO: remove pushing to XCom once we update the airflow version.
         context["ti"].xcom_push(key="base_sql_query", value=str(self.sql))
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -8,6 +8,7 @@ import pandas as pd
 from airflow.decorators.base import DecoratedOperator
 from airflow.exceptions import AirflowException
 from airflow.models import BaseOperator
+from airflow.models.xcom_arg import XComArg
 from sqlalchemy.sql.functions import Function
 
 from astro.airflow.datasets import kwargs_with_datasets
@@ -62,6 +63,40 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
             **kwargs_with_datasets(kwargs=kwargs, output_datasets=self.output_table),
         )
 
+    def _resolve_xcom_op_kwargs(self, context: Context) -> None:
+        """
+        Iterate through self.op_kwargs, resolving any XCom values with the given context.
+        Replace those values in-place.
+
+        :param context: The Airflow Context to be used to resolve the op_kwargs.
+        """
+        # TODO: confirm if it makes sense for us to always replace the op_kwargs or if we should
+        # only replace those that are within the decorator signature, by using
+        # inspect.signature(self.python_callable).parameters.values()
+        kwargs = {}
+        for kwarg_name, kwarg_value in self.op_kwargs.items():
+            if isinstance(kwarg_value, XComArg):
+                kwargs[kwarg_name] = kwarg_value.resolve(context)
+            else:
+                kwargs[kwarg_name] = kwarg_value
+        self.op_kwargs = kwargs
+
+    def _resolve_xcom_op_args(self, context: Context) -> None:
+        """
+        Iterates through self.op_args, resolving any XCom values with the given context.
+        Replace those values in-place.
+
+        :param context: The Airflow Context used to resolve the op_args.
+        """
+        args = []
+        for arg_value in self.op_args:
+            if isinstance(arg_value, XComArg):
+                item = arg_value.resolve(context)
+            else:
+                item = arg_value
+            args.append(item)
+        self.op_args = args  # type: ignore
+
     def _enrich_context(self, context: Context) -> Context:
         """
         Prepare the sql and context for execution.
@@ -75,6 +110,9 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
 
         :return: the enriched context with astro specific information.
         """
+        self._resolve_xcom_op_args(context)
+        self._resolve_xcom_op_kwargs(context)
+
         first_table = find_first_table(
             op_args=self.op_args,  # type: ignore
             op_kwargs=self.op_kwargs,
@@ -138,7 +176,6 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         :param jinja_env: Jinja environment
         """
         context = self._enrich_context(context)
-
         return super().render_template_fields(context, jinja_env)
 
     def execute(self, context: Context) -> None:
@@ -220,6 +257,8 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
                 ) = self.database_impl.get_sqlalchemy_template_table_identifier_and_parameter(v, k)
                 context[k] = jinja_table_identifier
                 self.parameters[k] = jinja_table_parameter_value
+            # elif isinstance(v, pd.DataFrame):
+            #    raise Exception("I should not be here")
             else:
                 context[k] = self.database_impl.parameterize_variable(k)
 
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -137,9 +137,10 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         :param context: Dict with values to apply on content
         :param jinja_env: Jinja environment
         """
+        output = super().render_template_fields(context, jinja_env)
         context = self._enrich_context(context)
 
-        return super().render_template_fields(context, jinja_env)
+        return output
 
     def execute(self, context: Context) -> None:
         self._enrich_context(context)
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -257,8 +257,6 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
                 ) = self.database_impl.get_sqlalchemy_template_table_identifier_and_parameter(v, k)
                 context[k] = jinja_table_identifier
                 self.parameters[k] = jinja_table_parameter_value
-            # elif isinstance(v, pd.DataFrame):
-            #    raise Exception("I should not be here")
             else:
                 context[k] = self.database_impl.parameterize_variable(k)
 
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -3,9 +3,11 @@ from __future__ import annotations
 import inspect
 from typing import Any, Callable, Sequence, cast
 
+import jinja2
 import pandas as pd
 from airflow.decorators.base import DecoratedOperator
 from airflow.exceptions import AirflowException
+from airflow.models import BaseOperator
 from sqlalchemy.sql.functions import Function
 
 from astro.airflow.datasets import kwargs_with_datasets
@@ -20,7 +22,8 @@ from astro.utils.table import find_first_table
 class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
     """Handles all decorator classes that can return a SQL function"""
 
-    template_fields: Sequence[str] = ("parameters", "op_args", "op_kwargs")
+    template_fields: Sequence[str] = ("sql", "parameters", "op_args", "op_kwargs")
+    template_ext: Sequence[str] = (".sql",)
 
     database_impl: BaseDatabase
 
@@ -59,7 +62,19 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
             **kwargs_with_datasets(kwargs=kwargs, output_datasets=self.output_table),
         )
 
-    def execute(self, context: Context) -> None:
+    def _enrich_context(self, context: Context) -> Context:
+        """
+        Prepare the sql and context for execution.
+
+        Specifically, it will do the following:
+        1. Preset database settings
+        2. Load dataframes into tables
+        3. Render sql as sqlalchemy executable string
+
+        :param context: The Airflow Context which will be extended.
+
+        :return: the enriched context with astro specific information.
+        """
         first_table = find_first_table(
             op_args=self.op_args,  # type: ignore
             op_kwargs=self.op_kwargs,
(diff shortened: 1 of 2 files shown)


python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -103,7 +103,7 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
 def transform_file(
     file_path: str,
     conn_id: str = "",
-    parameters: Mapping | Iterable | None = None,
+    parameters: dict | None = None,
```
feluelle, line 106: I don't consider this a breaking change. WDYT?

_mypy_ was complaining about `Mapping | Iterable`, which is why I changed that.

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -1,24 +1,30 @@
...
 from astro.utils.table import find_first_table
 from astro.utils.typing_compat import Context
+from jinja2 import DebugUndefined
 from sqlalchemy.sql.functions import Function
 
 
 class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
     """Handles all decorator classes that can return a SQL function"""
 
+    template_fields: Sequence[str] = ("sql", "parameters", "op_args", "op_kwargs")
```
dimberman, line 25: Have you tested on airflow to see if the SQL now shows up in the "rendered" section?

python-sdk/src/astro/sql/operators/transform.py
dimberman: I don't consider this breaking

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -56,6 +62,31 @@ def __init__(
...
+
+        # We don't resolve undefined variables such as input tables
+        # as this would require opening a connection to the database.
+        # E.g. calling `create_database(self.conn_id)`` etc.
+        jinja_env.undefined = DebugUndefined
+
+        self._do_render_template_fields(
+            self, self.template_fields, context, jinja_env, set()
+        )
+        return self
```
dimberman, line 88: Can you explain a bit more on why you're overriding this?

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -128,14 +128,16 @@ def transform_file(
...
         schema=schema,
+        sql=file_path,
+        python_callable=null_function,
         **kwargs,
-    )
-    def transform_file_func(file_path: str):
-        return file_path
-
-    return transform_file_func(file_path)
+    ).output
```
dimberman, line 143: Would it be possible to make an operator that just has the SQL stuff without all the decorator stuff? I feel like that gives us a lot of unnecessary complexity for the transform_file use-case.

dimberman (changes requested):

python-sdk/src/astro/sql/operators/base_decorator.py
feluelle: Yes. It does.

python-sdk/src/astro/sql/operators/base_decorator.py
feluelle: Of course :) We are overriding this because the context is not aware of our astro-sdk input tables in jinja. It cannot resolve the `{{ table_1 }}` variables. It raises an `Undefined` error, because we would have to update the context with our input tables. The problem is that we cannot easily update it with that, because this would require retrieving connections from the database and I am not sure if we want to hit the database every time we render the jinja.

We can also take this offline and I show it to you in a demo.

python-sdk/src/astro/sql/operators/transform.py
feluelle: ..but we are not using the `transform` decorator anymore. It uses the `TransformOperator` which is a normal Airflow operator. Could you elaborate on what you think is unnecessary, please?

python-sdk/src/astro/sql/operators/transform.py
kaxil: Yeah, not a breaking change

python-sdk/docs/astro/sql/operators/transform_sql.rst
```diff
@@ -0,0 +1,16 @@
+.. _transform_sql_operator:
+
+=======================
+transform_sql operator
+=======================
+
+When to use the ``transform_sql`` operator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``transform_sql`` operator allows you to implement the **T** of an ELT system by running a SQL query from specified SQL string. Each step of the transform pipeline creates a new table from the ``SELECT`` statement and enables tasks to pass those tables as if they were native Python objects.
```
tatiana, line 9: It is worth explaining when the user should be using this compared to `transform_file` or `transform`.

tatiana (commented):
@feluelle I think it could be confusing to end-users to have three methods for creating a table from a SQL statement:
* `transform`
* `transform_file`
* `transform_sql`

I can see the need for having a decorator (`transform`) x a task, and I understand you didn't want to introduce any breaking changes, but it can confuse end-users. What do you think of:
1, Adding a deprecation warning in `transform_file`
2. Name the new task `transform_task` (or something else) - so it is clear the difference in comparison to `transform` - and it could be used both to load from a file and a string?
3. Should we have some tests?

python-sdk/example_dags/example_transform_sql.py
```diff
@@ -0,0 +1,37 @@
+from datetime import datetime
```
tatiana, line 1: This example DAG won't be run with our test suite unless we add it to:
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/test_example_dags.py#L50

Perhaps we should log a separate ticket to run all the example DAGs as part of that test (as opposed to have to manually add them).

feluelle: Thanks 👍 I will add it to the test suite.

feluelle: > Perhaps we should log a separate ticket to run all the example DAGs as part of that test (as opposed to have to manually add them).

Not sure, because for some example dags it might require a lot of infra setup and imo example dags are for users to see how it works. It does not necessarily mean we are running them in our CI. Of course, it would be ideal if we test all our example dags. But I think this additional step makes sense to not having to block adding examples dags until we have tests setup.

feluelle: We also need to figure out costs. Not every DAG can be executed as often as others.

tatiana: Let's loop in more people to this discussion:
@kaxil @utkarsharma2 @phanikumv 

In the past we had issues with having DAGs which were no longer valid, and we had decided to run all of them using the `test_example_dags.py`. At the moment we are very likely missing DAGs and not running them in the suite. What are your thoughts?

feluelle:
> 1, Adding a deprecation warning in transform_file
> 2. Name the new task transform_task (or something else) - so it is clear the difference in comparison to transform - and it could be used both to load from a file and a string?
> 3. Should we have some tests?

Yes, actually this makes more sense.

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -100,10 +101,52 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
     )
 
 
+def transform_sql(
```
tatiana, line 104: @feluelle, you're introducing a sort of "feature" (to allow users to view the rendered SQL in the UI) - but we don't have any tests that check this.  This means that, probably, if someone rolled back the change, the CI wouldn't complain about this change, and we'd lose the feature. Do you have any idea for how we could test this?

python-sdk/docs/astro/sql/operators/transform_file.rst
```diff
@@ -14,3 +14,7 @@ The ``transform_file`` functions return a ``Table`` object that can be passed to
        :language: python
        :start-after: [START transform_file_example_1]
        :end-before: [END transform_file_example_1]
+
+.. note::
+
+    `transform_file` is deprecated and will be removed in future release. Please use `transform_sql` instead.
```
feluelle, line 20:
```suggestion
    ``transform_file`` is deprecated and will be removed in future release. Please use ``transform_sql`` instead.
```

python-sdk/src/astro/sql/operators/transform.py
feluelle: Do you think it is useful to add a test checking the `template_fields` and `template_ext`?

tatiana: I think it is worth it, @feluelle. From one perspective, it feels like we're copying the implementation and test, but at least we'd be testing the "feature" (assuming Airflow doesn't change their interface, which I think is a valid assumption for now)

feluelle: Alright, will add it then 👍

feluelle: Added in 1face6c.

python-sdk/example_dags/example_transform_sql.py
kaxil: Yes, I would for the example DAGs to be run in the CI -- to make it simple you could use SQLite

kaxil: and looks like this is using sqlite :) so should be easy enough to test

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -100,10 +130,53 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
...
+        return sql, parameters
+
+    return TransformOperator(
+        conn_id=conn_id,
+        parameters=parameters,
+        handler=None,
+        database=database,
+        schema=schema,
+        sql=sql,
+        python_callable=null_function,
```
kaxil, line 171:
```suggestion
        python_callable=lambda: sql, parameters,
```

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -100,10 +130,53 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
...
+    :param schema: Schema within the SQL instance you want to access. If left blank we will default to the
+        table.metatadata.schema in the first Table passed to the function (required if there are no table arguments)
+    :param kwargs: Any keyword arguments supported by the BaseOperator is supported (e.g ``queue``, ``owner``)
+    :return: Transform functions return a ``Table`` object that can be passed to future tasks.
+        This table will be either an auto-generated temporary table,
+        or will overwrite a table given in the `output_table` parameter.
+    """
+
+    def null_function():
+        return sql, parameters
```
kaxil, line 162:
```suggestion
```

Why not use lambda?

python-sdk/example_dags/example_transform_file.py
```diff
@@ -1,4 +1,3 @@
-import pathlib
 from datetime import datetime
```
kaxil, line 1: Should we remove this file since the function is deprecated and we ask users to use `transform_sql` instead?

kaxil: Also, why a new name `transform_sql`, why not just change the code of `transform_file`?

python-sdk/src/astro/sql/operators/transform.py
feluelle: Seems to work 🤔 I thought I had tried that. I remember that there was a type issue with lambda at least - that is why I changed that.

python-sdk/example_dags/example_transform_file.py
feluelle: I am not sure. We also keep the docs and just add a note about that it is deprecated. Also we use the example dag to test that the deprecated function still works. Deprecating a function does not mean we should not test it in my opinion.

python-sdk/example_dags/example_transform_sql.py
feluelle: Alright, I can add another PR for running all examples dags - instead of having to add them manually.

python-sdk/example_dags/example_transform_file.py
tatiana: I agree with @feluelle that we should continue testing the function even if it is deprecated - since it is still part of our interface, and we don't want for it to break until we remove it from the source-code.

My understanding was that this was a breaking change. If it is not a breaking change, I'm in favour of keeping `transform_file`. Did I get it wrong, @feluelle ? :see_no_evil:

feluelle: 🙈 all the work for nothing 🤦  I'll revert the `transform_sql` as we can also use `transform_file` in the sql-cli now. I added this because I thought we need a `transform_sql`.

python-sdk/tests/test_example_dags.py
```diff
@@ -61,6 +61,7 @@ def session():
         "example_transform",
         "example_merge_bigquery",
         "example_transform_file",
+        "example_transform_sql",
```
feluelle, line 64:
```suggestion
```

python-sdk/example_dags/example_transform_file.py
feluelle: Fixed. PTAL @kaxil @tatiana

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -103,13 +132,13 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
...
     conn_id: str = "",
-    parameters: Mapping | Iterable | None = None,
+    parameters: dict | None = None,
     database: str | None = None,
     schema: str | None = None,
     **kwargs: Any,
 ) -> XComArg:
     """
-    This function returns a ``Table`` object that can be passed to future tasks from specified SQL file.
+    This function returns a ``Table`` object that can be passed to future tasks from specified SQL code.
```
feluelle, line 141:
```suggestion
    This function returns a ``Table`` object that can be passed to future tasks from specified SQL file.
```

tatiana (approved):

python-sdk/example_dags/example_transform_file.py
dimberman: Yeah +1 on keeping the name `transform_file`, to me that's descriptive of what it's doing.

dimberman (dismissed):

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -56,6 +62,31 @@ def __init__(
             **kwargs_with_datasets(kwargs=kwargs, output_datasets=self.output_table),
         )
 
+    def render_template_fields(
```
kaxil, line 65: All looks good, one last comment: Based on our 1:1 discussion I remember you were going to look into `render_template_fields` vs the custom rendering we have in

https://github.com/astronomer/astro-sdk/blob/bf69fd7e2f053c440214e8cc5887a0f186c42d6d/python-sdk/src/astro/sql/operators/base_decorator.py#L126-L128

https://github.com/astronomer/astro-sdk/blob/bf69fd7e2f053c440214e8cc5887a0f186c42d6d/python-sdk/src/astro/sql/operators/base_decorator.py#L148-L182

Were you able to look at that? And see if we are not duplicating or if we can consolidate them?

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -128,14 +157,13 @@ def transform_file(
         or will overwrite a table given in the `output_table` parameter.
     """
 
-    @transform(
+    return TransformOperator(
```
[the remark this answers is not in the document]
feluelle: No, the `TransformOperator` (or `BaseDecorator`) is not the issue - this is what I compared it to. See the slack thread [here](https://astronomer.slack.com/archives/C03868KGF2Q/p1663233931881039).

The problem lays somewhere in the python decorator to python operators functionality in Airflow core.

python-sdk/src/astro/sql/operators/base_decorator.py
feluelle: Hey @kaxil, thanks but I did not deep dive into it yet. Do you think this is required for this ticket? This PR is mainly about enabling being able to see the rendered code at all.

The next step, imo, would be to try to consolidate them and allow rendering the sqlalchemy in the sql rendered view or we create a dedicated view as a plugin in Airflow.

kaxil: Sure, as long as you are ok with this change not being part of 1.1.1 of Python SDK and as long as you tackle that before 1.2 :) (mid-Oct)

python-sdk/example_dags/example_transform_sql.py
feluelle: PR https://github.com/astronomer/astro-sdk/pull/992 for running all examples dags.

python-sdk/src/astro/sql/operators/base_decorator.py
feluelle: Managed to get it working. See fbbc116 and c7482f7.

python-sdk/src/astro/utils/table.py
```diff
@@ -1,97 +1,122 @@
...
-from typing import Callable, Optional
+from typing import Callable
 
+from airflow.models.xcom_arg import XComArg
 from astro.sql.table import BaseTable
+from astro.utils.typing_compat import Context
 
 
-def _pull_first_table_from_parameters(parameters: dict) -> Optional[BaseTable]:
+def _have_same_conn_id(tables: list[BaseTable]) -> bool:
```
kaxil, line 11: `_have_same_conn_id` -> `_has_same_conn_id`

feluelle: I chose `have` because of `tables: list[BaseTable]` 😅

kaxil (approved):

kaxil:
Some tests are failing!

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -108,6 +161,25 @@ def execute(self, context: Context) -> None:
...
+        This mutates the attributes in-place and is irreversible.
+
+        :param context: Dict with values to apply on content
+        :param jinja_env: Jinja environment
+        """
+        context = self._enrich_context(context)
+        return super().render_template_fields(context, jinja_env)
+
+    def execute(self, context: Context) -> None:
+        self._enrich_context(context)
```
feluelle, line 182:
```suggestion
        context = self._enrich_context(context)
```
Shouldn't we also save this here?

tatiana (approved):

PR #924, opened 2022-09-27
Wait for benchmark completion

pankajastro:
# Description
This PR includes a script which runs as a part of the benchmark make command execution and waits for the benchmark job to complete.

## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/Makefile
+++ b/python-sdk/tests/benchmark/Makefile
@@ -1,4 +1,4 @@
-all: clean setup_gke container run_job
+all: clean setup_gke container run_job wait_for_completion
 .PHONY: all
 
 GCP_PROJECT ?= astronomer-dag-authoring
@@ -61,6 +61,9 @@ run_job:
 	@kubectl apply -f infrastructure/kubernetes/postgres_svc.yaml
 	@sed s/%VERSION%/${GIT_HASH}/g infrastructure/kubernetes/job.yaml > infrastructure/kubernetes/job_final.yaml
 	@kubectl create -f infrastructure/kubernetes/job_final.yaml
+
+wait_for_completion:
+	bash wait_for_completion.sh
 
 # Takes approximately 2min
 teardown_gke:
--- a/python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
+++ b/python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  generateName: benchmark-
+  name: benchmark-%VERSION%
   namespace: benchmark
 spec:
   template:
--- a/python-sdk/tests/benchmark/wait_for_completion.sh
+++ b/python-sdk/tests/benchmark/wait_for_completion.sh
@@ -1,0 +1,18 @@
+set -x  # print the shell command before being executed
+GIT_HASH=$(git log -1 --format=%h)
+complete_status=1
+failed_status=1
+while [[ $complete_status -ne 0 ]] && [[ $failed_status -ne 0 ]]; do
+  sleep 60
+  output=$(kubectl wait --for=condition=failed job/benchmark-"${GIT_HASH}" -n benchmark --timeout=0 2>&1)
+  failed_status=$?
+  output=$(kubectl wait --for=condition=complete job/benchmark-"${GIT_HASH}" -n benchmark --timeout=0 2>&1)
+  complete_status=$?
+done
+
+kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
+
+if [ $failed_status -eq 0 ]; then
+    echo "Job failed. Please check logs."
+    exit 1
+fi


.github/workflows/ci-python-sdk.yaml
```diff
@@ -384,3 +384,40 @@ jobs:
     env:
       TWINE_USERNAME: __token__
       TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+
+  Run-Benchmark:
```
pankajastro, line 388: Changes in this file are temporary for manual testing.

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  generateName: benchmark-
+  name: benchmark
```
utkarsharma2, line 4: @pankajastro I think we will need a unique name since we can have multiple runs concurrently. 
Maybe we can do something like - https://github.com/astronomer/astro-sdk/pull/927

pankajastro: ok, I'll wait for https://github.com/astronomer/astro-sdk/pull/927 to merge and reuse your changes i.e tag itself WDYT?

utkarsharma2: @pankajastro I think we can copy the code and merge/resolve conflicts later, in the interest of time.

python-sdk/tests/benchmark/Makefile
```diff
@@ -1,4 +1,4 @@
-all: clean setup_gke container run_job
+all: clean setup_gke container run_job wait_for_completion teardown_gke
```
tatiana, line 1: The `teardown_gke` step could be dreadful to parallel benchmark jobs triggered from other branches/PRs :see_no_evil: 
It is probably worth adding a separate ticket to address this (perhaps only teardown if no other jobs are running).

python-sdk/tests/benchmark/wait_for_completion.sh
```diff
@@ -0,0 +1,17 @@
...
+  output=$(kubectl wait --for=condition=failed job/benchmark -n benchmark --timeout=0 2>&1)
+  failed_status=$?
+  output=$(kubectl wait --for=condition=complete job/benchmark -n benchmark --timeout=0 2>&1)
+  complete_status=$?
+done
+
+kubectl delete job benchmark -n benchmark
+
+if [ $failed_status -eq 0 ]; then
+    echo "Job failed. Please check logs."
```
tatiana, line 15: What if we displayed the error from the K8s pod triggered by the job to make troubleshooting easier?

.github/workflows/ci-benchmark.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: rm -f test-connections.yaml
+      - run: ls -la # temporary to debug
```
tatiana, line 40: It may be worth changing this PR to draft until we change the current directory to be `python-sdk`.
The other change we discussed, of having a consistent way to override the env vars placeholders on the YAML (to make it more similar to `conftest.py`)  is optional.

tatiana (approved):

.github/workflows/ci-benchmark.yaml
pankajastro: @tatiana is it ok if I can take this in separate PR as improvement because I don't think will get time today to test it created an issue to track it https://github.com/astronomer/astro-sdk/issues/936

pankajastro: @tatiana is it ok if I can take this in separate PR as an improvement because I don't think will get time today to test it created an issue to track it https://github.com/astronomer/astro-sdk/issues/936

python-sdk/tests/benchmark/Makefile
pankajastro: I have created an issue to read cluster names from env maybe we can consider changing this in that ticket https://github.com/astronomer/astro-sdk/issues/934

python-sdk/tests/benchmark/wait_for_completion.sh
pankajastro: I think we need a pod name to fetch logs, right?
how we can get the pod name here, any thoughts?

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
pankajastro: I can see there is still some discussion going on in your PR.  The teardown is also problematic I feel we should take cluster name from env that way it would be completely isolated and even teardown will be fine I have created a ticket https://github.com/astronomer/astro-sdk/issues/934  WDYT?

python-sdk/tests/benchmark/Makefile
utkarsharma2: @pankajastro As per my discussion with @tatiana I think we should not teardown the cluster as of now. We can address it in https://github.com/astronomer/astro-sdk/issues/934

pankajastro: I'm ok to disable the teardown for now.

PR #925, opened 2022-09-27
Enable astro sdk without xcom pickling

dimberman:
# Description
## What is the new behavior?

-
-
-

### Checklist


python-sdk/src/astro/files/base.py
```diff
@@ -50,6 +51,25 @@ def type(self) -> FileType:  # noqa: A003
...
+        }
+
+    @classmethod
+    def from_json(cls, serialized_object: dict):
+        return File(
+            conn_id=serialized_object["conn_id"],
+            path=serialized_object["path"],
+            filetype=constants.FileType(serialized_object["filetype"]),
+            normalize_config=serialized_object["normalize_config"],
+        )
```
kaxil, line 71: Our File and Table object are attrs classes, so you can do the following to serialize them natively.

```python
In [6]: from astro.files import File
In [7]: import attrs

In [8]: attrs.asdict(File("/path"))
Out[8]:
{'path': '/path',
 'conn_id': None,
 'filetype': None,
 'normalize_config': None,
 'uri': 'astro+file:/path',
 'extra': {}}


In [9]: from astro.sql.table import Table, Metadata

In [11]: attrs.asdict(Table("ta", conn_id="gcp", metadata=Metadata(schema="s")))
Out[11]:
{'_name': 'ta',
 'conn_id': 'gcp',
 'metadata': {'schema': 's', 'database': None},
 'columns': [],
 'temp': False,
 'uri': 'astro://gcp@?table=ta&schema=s',
 'extra': {}}
```

dimberman: @kaxil  I'm not sure this solution is worth it for two reason:
1. Some of the fields are not usable for deserialization (for example in File, the uri field will break instantiation). If we can't fix this, then it might be worth it for us to manually control what fields we serialize.
2. There doesn't seem to be an attrs.fromdict so deserialization would need to be handled by hand
Do you have any thoughts on this? It's highly possible I'm missing something here

python-sdk/src/astro/settings.py
```diff
@@ -8,13 +8,13 @@
 REDSHIFT_SCHEMA = conf.get("astro_sdk", "redshift_default_schema", fallback=SCHEMA)
 
 
-ALLOW_UNSAFE_DF_STORAGE = conf.getboolean(
-    "astro_sdk", "dataframe_allow_unsafe_storage", fallback=False
+DATAFRAME_STORAGE_CONN_ID = conf.get(
```
tatiana, line 11: I'm slightly concerned about us enforcing Python SDK users to have to be on the cloud to run operations which use dataframes. This may not be the case for some of our customers.

What would you recommend for those users?

python-sdk/tests/sql/operators/test_drop_table.py
```diff
@@ -53,7 +53,7 @@ def test_drop_table_with_table_metadata(database_table_fixture, sample_dag):
     def do_nothing(df: pandas.DataFrame):
         return df
 
-    @task
```
tatiana, line 56: The Astro SDK wouldn't connect to standard Task Flow API tasks is the biggest downside of this approach, in my opinion.

I'd love to see how this approach compares to the Custom Backend solution. Would we have any other alternatives?

python-sdk/src/astro/settings.py
dimberman: @tatiana there's no safe way to store dataframes that doesn't involve some form of cloud storage. Storing on the local machine won't work unless all of your airflow workers are on that same machine, and storing in the metadata DB will cause performance degradation.

python-sdk/tests/sql/operators/test_drop_table.py
dimberman: @tatiana We can make the custom backend for taskflow parity, but that would need to be a separate PR as this PR is already quite large and changes a lot. I think we should turn `astro.dataframe` into `astro.python` so people understand it's not just for dataframes.

tatiana (commented):
@dimberman I see this PR addressing two issues:
1. Not using pickle to serialize objects inside the Airflow metadata DB (requirement from end-users, due to safety concerns)
2. Not storing objects inside the Airflow metadata DB (good practice to avoid trashing the Airflow DB)

The main downside of this approach is that we would no longer interoperate with the standard Task Flow API, as we do now - users would have only to use the Python SDK or handle the custom serialization as part of their tasks. I see this as a significant issue.

I'm keen on seeing other solutions, including the custom backend approach - before we move forward with this strategy.

python-sdk/src/astro/files/base.py
dimberman: @kaxil another issue I found :

```
TypeError: TempTable.__init__() got an unexpected keyword argument '_name'
```

This is when I use the 

```
        table_dict = attrs.asdict(obj, recurse=True, filter=lambda a, v: a.init)
```

thing you showed me. I suppose we could filter for the _name field but it's starting to feel like this is a round hole square peg situation.

Issue #888, opened 2022-09-21
Changes benchmarking/analyse.py script to use bigquery for markdown generation


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we are using `.ndjson` to generate markdown for results.md file.

**Describe the solution you'd like**
We should use bigquery as the main source of benchamrking data and modify the `analyse.py` to consume data from bigquery.

suggestion:
1. `benchmarking/analyse.py` can detect the git sha from the repo.
2. There can be duplication to resolve this we can:
   i. Take time as input

**Acceptance Criteria**

- [ ] `benchmarking/analyse.py` should generate markdown by querying bigquery.

PR #926, opened 2022-09-27
Modfied analyse.py script to read from bigquery

utkarsharma2:
# Description
Currently, we are using .ndjson to generate markdown for results.md file.

closes: #888 

## What is the new behavior?
We should use bigquery as the main source of benchamrking data and modify the analyse.py to consume data from bigquery.

1. Read data from bigquery. Filtering is done based on gitsha, revision col in the table.
2. Added --git-sha option
3. Added --output-filepath to output results to a file

Nope

### Checklist
- [X] benchmarking/analyse.py should generate markdown by querying bigquery.


kaxil (approved):

python-sdk/tests/benchmark/analyse.py
```diff
@@ -131,24 +167,55 @@ def analyse_results(results_filepath):
...
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument(
         "--results-filepath",
         "-r",
         type=str,
         help="NDJSON local path (/path/to/file.ndjson) or Google "
         "cloud storage path (gs://bucket/sample.ndjson) containing the results for a benchmark run",
     )
+    group.add_argument(
+        "--git-sha",
```
tatiana, line 196: @utkarsharma2 It may be worth giving an argument name which makes clear this relates to BQ. Something like: `--bq-git-sha` and `-b` WDYT?

Every result row has a git hash, so a generic argument `--git-sha` could make sense for analysing JSON files as well.

python-sdk/tests/benchmark/constants.py
```diff
@@ -0,0 +1,13 @@
+import os
```
tatiana, line 1: The convention we used for the Python SDK is to have a `settings.py` to retrieve things from environment variables and a `constants.py` file to static things. I suggest we name this `settings.py`.

python-sdk/tests/benchmark/analyse.py
```diff
@@ -131,24 +167,55 @@ def analyse_results(results_filepath):
...
 
+    content = ""
     # print Markdown tables per database
     for database_name in summary["database"].unique().tolist():
-        print(f"\n### Database: {database_name}\n")
-        print(summary[summary["database"] == database_name].to_markdown(index=False))
+        content = content + f"\n\n### Database: {database_name}\n"
+        content = content + str(
+            summary[summary["database"] == database_name].to_markdown(index=False)
+        )
```
feluelle, line 176: I think it is worth refactoring this into something like a `get_content` method.

python-sdk/tests/benchmark/analyse.py
```diff
@@ -131,24 +167,55 @@ def analyse_results(results_filepath):
...
+
+    if output_filepath:
+        f = open(output_filepath, "w")
+        f.write(content)
+    else:
+        print(content)
```
feluelle, line 182:
```suggestion
    if output_filepath:
        with open(output_filepath, "w") as file:
            file.write(content)
    else:
        print(content)
```

python-sdk/tests/benchmark/analyse.py
utkarsharma2: @tatiana Make sense.

python-sdk/tests/benchmark/analyse.py
utkarsharma2: @feluelle Thanks, I have added it :)

Issue #887, opened 2022-09-21
Use git sha to pick docker image for k8 pods


utkarsharma2:
**Please describe the feature you'd like to see**
Every new docker image is labeled as `latest` and if we two parallel branches that runs benchmark one might conflict with another branch. We should have new labels per commit basis.

**Describe the solution you'd like**
We should modify the terraform script to use git sha for instead of `latest` tag

**Acceptance Criteria**

- [ ] Stop using `latest` tag and use git sha for this.

PR #927, opened 2022-09-27
Make docker image tag dynamic

utkarsharma2:
# Description
Every new docker image is labeled as latest and if we have two parallel branches that run benchmark one might conflict with another branch. We should have new labels per commit basis.

closes: #887

## What is the new behavior?
Modified Makefile to use git sha instead of the latest tag.

Nope

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/Makefile
+++ b/python-sdk/tests/benchmark/Makefile
@@ -36,7 +36,7 @@ run: clean
 container: clean check_google_credentials
 	@echo "Building and pushing container $(CONTAINER_REGISTRY)"
 	@gcloud auth configure-docker gcr.io
-	@docker build --platform=$(PLATFORM) --build-arg=GIT_HASH=$(GIT_HASH) --tag=$(CONTAINER_REGISTRY):$(GIT_HASH) --tag=$(CONTAINER_REGISTRY):latest -f ./Dockerfile  ../../
+	@docker build --platform=$(PLATFORM) --build-arg=GIT_HASH=$(GIT_HASH) --tag=$(CONTAINER_REGISTRY):$(GIT_HASH) --tag=$(CONTAINER_REGISTRY):$(GIT_HASH) -f ./Dockerfile  ../../
 	@gcloud auth activate-service-account --key-file $(GOOGLE_APPLICATION_CREDENTIALS)
 	@docker push $(CONTAINER_REGISTRY)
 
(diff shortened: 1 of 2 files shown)


pankajkoti (commented):
This would mean every time we make local changes and want to run benchmark, we would have to commit right? How about instead generating a hash so that although 2 developers have the same git commit hash on local (before they commit), so that they have different tags?

tatiana (approved):

pankajastro (approved):

pankajastro:
> This would mean every time we make local changes and want to run benchmark, we would have to commit right? How about instead generating a hash so that although 2 developers have the same git commit hash on local (before they commit), so that they have different tags?

Benchmark we don't run so frequently so I'm ok with this change also otherwise maybe we can take the current time in milliseconds + gittag WDYT?

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: benchmark
-        image: gcr.io/astronomer-dag-authoring/benchmark
+        image: gcr.io/astronomer-dag-authoring/benchmark:%VERSION%
```
feluelle, line 13: can you not access the env var here?

python-sdk/tests/benchmark/Makefile
```diff
@@ -59,7 +59,8 @@ run_job:
 	@kubectl apply -f infrastructure/kubernetes/namespace.yaml
 	@kubectl apply -f infrastructure/kubernetes/postgres.yaml
 	@kubectl apply -f infrastructure/kubernetes/postgres_svc.yaml
-	@kubectl create -f infrastructure/kubernetes/job.yaml
+	@sed s/%VERSION%/${GIT_HASH}/g infrastructure/kubernetes/job.yaml > infrastructure/kubernetes/job_final.yaml
+	@kubectl create -f infrastructure/kubernetes/job_final.yaml
```
feluelle, line 63: This looks like a hack to me. There has to be another way of passing values from outside into the yaml file.

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
utkarsharma2: @feluelle As per my understanding. We can't. But do let me know if you think there is a way

python-sdk/tests/benchmark/Makefile
utkarsharma2: @feluelle This is not the ideal solution but is there an easier alternative? 

I considered `yq` to replace the yaml file but then again you have to create a new file and you need to install `yq`.

kaxil: There is another hack - https://stackoverflow.com/a/56240712 :)

kaxil: But it is just a hack -- you can't template k8s yaml without Helm natively I think

kaxil (approved):

python-sdk/tests/benchmark/Makefile
feluelle: Right - you would have to use helm.

PR #928, opened 2022-09-28
Run benchmark job on label

pankajastro:
# Description
Run the benchmark if the label "benchmark" added to the PR

### Checklist

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -12,10 +12,11 @@ on:
       - 'python-sdk/**'
       - '.github/workflows/ci-python-sdk.yaml'
       - '*'
+    types: ['labeled', 'opened', 'synchronize']
   # Run on PRs from forks
-  pull_request_target:
-    branches: [ 'main' ]
-    types: ['labeled']
+#  pull_request_target:
+#    branches: [ 'main' ]
+#    types: ['labeled']
   release:
     types: [ 'created' ]
 defaults:
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -12,11 +12,8 @@ on:
       - 'python-sdk/**'
       - '.github/workflows/ci-python-sdk.yaml'
       - '*'
-    types:
-      - labeled
-      - opened
-      - synchronize
-      # Run on PRs from forks
+    types: ['labeled', 'opened', 'synchronize']
+  # Run on PRs from forks
   pull_request_target:
     branches: [ 'main' ]
     types: ['labeled']
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -80,8 +80,7 @@ jobs:
       ) ||
       (
         github.event_name == 'pull_request_target' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-python-sdk.yaml
```diff
@@ -384,3 +385,41 @@ jobs:
     env:
       TWINE_USERNAME: __token__
       TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+
+  Run-Benchmark:
+    if: github.event.label.name == 'benchmark'
```
utkarsharma2, line 390: @pankajastro Can we change this to run-benchmark? We might use the benchmark to categorize other tickets down the line.

.github/workflows/ci-python-sdk.yaml
```diff
@@ -12,6 +12,7 @@ on:
       - 'python-sdk/**'
```
tatiana, line 12: @pankajastro please, share your findings about why this PR has
`28 successful and 6 skipped checks`

When compared to #924
`24 successful and 1 skipped checks` 

Why did the checks increase so much?

pankajastro: Looking..., yes ideally it should have 25 successful and 1 skipped

pankajastro: Look like some job from the event `pull_request_target` is also getting executed and if the condition is not matching then getting skipped but still it strange since we have added branch filter

pankajastro: Also, I can see some jobs are double for example 

`Build and test astro / Type-Check (pull_request_target) ` and 
`Build and test astro / Type-Check (pull_request_target`

pankajastro: @tatiana this look like a pre-existing issue in our CI we add any label in our PR it ran couple job twice only by `pull_request` and other by `pull_request_target`

pankajastro: we fixed some issues of running twice here https://github.com/astronomer/astro-sdk/pull/950 but I feel we still need to re-investigate how we should run CI on the fork

utkarsharma2 (approved):
LGTM

.github/workflows/ci-python-sdk.yaml
```diff
@@ -387,3 +388,41 @@ jobs:
...
+      SNOWFLAKE_WAREHOUSE: DEMO
+      SNOWFLAKE_HOST: https://gp21411.us-east-1.snowflakecomputing.com
+      SNOWFLAKE_ACCOUNT: gp21411
+      SNOWFLAKE_REGION: us-east-1
+      SNOWFLAKE_ROLE: AIRFLOW_TEST_USER
+      AIRFLOW__CORE__LOAD_DEFAULT_CONNECTIONS: True
+      AIRFLOW__CORE__ENABLE_XCOM_PICKLING: True
+      AIRFLOW_VAR_FOO: templated_file_name
+      AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE: True
+      FORCE_COLOR: "true"
```
utkarsharma2, line 422:
```suggestion
      FORCE_COLOR: "true"
      ASTRO_PUBLISH_BENCHMARK_DATA: True
```

PR #933, opened 2022-09-28
Add temporary ls step in benchmark job

pankajastro:
# Description
Add temporary ls step in benchmark job. I'll revert it once testing is done

## What is the new behavior?

-
-
-

### Checklist


sunank200 (approved):
This is just for test right @pankajastro So approving it. We should rollback to default after this issue is fixed.

utkarsharma2 (approved):

PR #935, opened 2022-09-28
[Test] Temp fix conn config path

pankajastro:
# Description
Temp fix conn config path. This is just for testing the benchmark script at a scheduled time once testing is done I'll revert the changes

## What is the new behavior?

-
-
-

### Checklist


sunank200 (approved):
This is for test @pankajastro . We should change the cronjob schedule after this test is successful

Issue #884, opened 2022-09-21
Modify config to contain values for load_file kwargs operator.


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we need to modify the dag if we want to modify the behavior of `load_file` operator. This is not scalable since we need to run the same dag or all paths - native and non-natives.

**Describe the solution you'd like**
We can keep the kwargs that will be passed to load_file() operator in config.json in a new key "load_file_kwargs" as follows:
```
"databases": [{
    "name": "postgres",
    "params": {
      "conn_id": "postgres_conn_benchmark",
      "metadata": {
        "database": "postgres"
      }
    },
    "load_file_kwargs": {
      "use_native_support": True,
      "enable_native_fallback": False
    }
  }]
```
and pass this new value to `load_file()` operator.

**Are there any alternatives to this feature?**
Is there another way we could solve this problem or enable this use-case?
Maybe we can change the config a bit. Open to suggestions.

**Acceptance Criteria**

- [ ] We should be able to run benchmarking scripts for both native and non-native paths without modifying the dag.

PR #939, opened 2022-09-28
Moved kwargs to config

utkarsharma2:
# Description
Currently, we need to modify the dag if we want to modify the behavior of load_file operator. This is not scalable since we need to run the same dag or all paths - native and non-natives.

closes: #884

## What is the new behavior?
We can keep the kwargs that will be passed to load_file() operator in config.json in a new key "load_file_kwargs" as follows:
```
"databases": [{
    "name": "postgres",
    "params": {
      "conn_id": "postgres_conn_benchmark",
      "metadata": {
        "database": "postgres"
      }
    },
    "kwargs": {
      "use_native_support": True,
      "enable_native_fallback": False
    }
  }]
```
and pass this new value to `load_file()` operator.

### Checklist
- [ ] We should be able to run benchmarking scripts for both native and non-native paths without modifying the dag.

### DIFF
--- a/python-sdk/tests/benchmark/config.json
+++ b/python-sdk/tests/benchmark/config.json
@@ -1,14 +1,20 @@
 {
   "databases": [
     {
+      "kwargs": {
+        "enable_native_fallback": false
+      },
       "name": "snowflake",
-      "params": {
+      "output_table": {
         "conn_id": "snowflake_conn"
       }
     },
     {
+      "kwargs": {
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Approving at the assumption that this change was tested and the new version of the benchmark was run successfully :)

utkarsharma2:
@tatiana Yes I tested it locally.

Issue #827, opened 2022-09-12
Verify "logging" when falling back to pandas-path


kaxil:
No logs seen when native loads fails and we take the pandas path course. Add logs when falling back to pandas path.

@utkarsharma2 to add details

PR #940, opened 2022-09-28
Updated the log level to logging.WARNING

utkarsharma2:
# Description
Currently, we are not showing the warning message when the native path fails and we default to pandas path to load the file in DB. 

closes: #827

## What is the new behavior?
Found the default log level to ERROR and changed it to WARNING.

Nope

### DIFF
--- a/python-sdk/src/astro/__init__.py
+++ b/python-sdk/src/astro/__init__.py
@@ -15,9 +15,8 @@ __version__ = "1.2.0.dev1"
 
 import logging
 
-from airflow.configuration import conf
-
 import astro.sql  # noqa: F401
+from airflow.configuration import conf
 
 # Override the default log level from logging.ERROR to logging.WARNING
 logging.getLogger().setLevel(level=logging.WARNING)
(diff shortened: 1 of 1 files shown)


python-sdk/src/astro/__init__.py
```diff
@@ -13,9 +13,14 @@
 # imported beforehand, it will also load astro.sql. In astro.sql we import lots of operators which depend on
 # astro.database, and this is what leads to the circular dependency.
 
+import logging
+
 import astro.sql  # noqa: F401
 from airflow.configuration import conf
 
+# Override the default log level from logging.ERROR to logging.WARNING
+logging.getLogger().setLevel(level=logging.WARNING)
```
kaxil, line 22: Wait, where is the logging level of `ERROR` being set?

utkarsharma2:
Found the Error level was set in the benchmark/run.sh script. We no longer need this PR. Since when user uses Astro-sdk the warning will be printed as intended.

Issue #883, opened 2022-09-21
Create and Use docker volume to save all the datasets


utkarsharma2:
**Please describe the feature you'd like to see**
We would like to run benchmarking script for loading local files. For that, the datasets need to be locally available. 

**Describe the solution you'd like**
We can use a docker volume to keep the datasets and use it in benchmarking script.

**Are there any alternatives to this feature?**
Copy data from GCS to Docker container, but this will be very slow.

**Acceptance Criteria**

- [ ] Volume should contain all the datasets
- [ ] Should be read-only

PR #945, opened 2022-09-28
Mount NFS persistent volumes for local to database benchmark

sunank200:
# Description

We would like to run benchmarking script for loading local files. For that, the datasets need to be locally available.

closes: #883 

## What is the new behavior?

- Create Persistent Disk in Google Compute Engine 
- Create NFS Server in GKE
- Create NFS Service
- Create Persistent Volume and Persistent Volume Claims
- Configuring Pods with PVC

No

### Checklist


python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
```diff
@@ -17,7 +17,16 @@ spec:
             memory: 6Gi
           limits:
             memory: 10Gi
+        volumeMounts:
+          # mount the nfs volume for local files
+          - name: benchmark-pvc
+            mountPath: "/mnt"
```
utkarsharma2, line 23: @sunank200 can you specify the path to local datasets?

sunank200: Yes its changed to `/data`

pankajastro (approved):

python-sdk/tests/benchmark/infrastructure/kubernetes/create-nfs-server-gke.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: apps/v1
```
tatiana, line 1: @sunank200 What do you think of having all the NFS-related specs in the same file `nfs.yml`?
Since this is a declaration of infrastructure, I believe the file's name should be a resource, as opposed to action (eg. create). And it could simplify the maintenance t keep them all together.

dimberman: +1. Kubernetes is declarative, so you think in terms of nouns instead of verbs. Think "this is what I expect to exist" instead of "let me create this"

sunank200: Added it here: https://github.com/astronomer/astro-sdk/pull/945/commits/efff09b2e7d690f25968e15a1db2466f79c60521

PR #950, opened 2022-09-28
Fix CI job running twice

pankajastro:
# Description
## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -7,7 +7,7 @@ on:
       - '.github/workflows/ci-python-sdk.yaml'
       - '*'
   pull_request:
-    branches: [ 'main', 'release-**' ]
+    branches: [ 'main', 'release-**', 'labeled' ]
     paths:
       - 'python-sdk/**'
       - '.github/workflows/ci-python-sdk.yaml'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-sql-cli.yaml
```diff
@@ -21,15 +21,8 @@ defaults:
   run:
     working-directory: sql-cli
 jobs:
-  Markdown-link-check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: gaurav-nelson/github-action-markdown-link-check@v1
-        with:
-          config-file: '.github/workflows/mlc_config.json'
```
kaxil, line 30: We didn't need it here, we already run it project wide in Python SDK workflow

python-sdk/src/astro/settings.py
```diff
@@ -37,6 +37,7 @@
     section="astro_sdk", key="run_raw_sql_response_size", fallback=-1
 )
 
+# Temp changes
```
kaxil, line 40: We should remove this line :(

.github/workflows/ci-python-sdk.yaml
```diff
@@ -23,6 +23,7 @@ defaults:
     working-directory: python-sdk
 jobs:
   Markdown-link-check:
+    if: github.event.action != 'labeled'
```
kaxil, line 26: We don't want to run this jobs when a label is added as it is redundant with when PR is created

pankajastro:
Thank you so much @kaxil for looking into it

Issue #917, opened 2022-09-26
Disable the SDK from changing capitalization of output columns by default


pgzmnk:
**Describe the bug**

Users expressed confusion on why the column names for output tables are lowercased. It's due to this [default setting](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/dataframe.py#L110). 

**Expected behavior**

The default behavior should be to not change casing and it should be configureable with a global variable.

That way we can avoid hardcoding this setting into the Astro Build dagfile generator script.

**Version**
1.1.0

**To Reproduce**
Run [this Pipeline](https://cloud.astronomer.io/cku7t3fvx59046554xr4g0siv7r/build/17982df9-c45f-43ec-b2dd-bfd6f42d1c8c/fd5d862c-bcad-4be2-a3a4-f7b3b9920bca).

**Screenshots**
<img width="1311" alt="image" src="https://user-images.githubusercontent.com/27398253/192357157-e5c7b9d4-3517-4f5c-8521-0098375bc052.png">

**Additional context**
[Slack thread from September 26 2022](https://astronomer.slack.com/archives/C02B8SPT93K/p1664208204100329).

kaxil:
Some related snowflake issues:
- https://github.com/snowflakedb/snowflake-sqlalchemy/issues/157
- https://github.com/apache/superset/issues/18085

PR #951, opened 2022-09-28
Add a global config for col capitalization

utkarsharma2:
# Description
Users expressed confusion on why the column names for output tables are lowercased. It's due to this [default setting](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/dataframe.py#L110).

closes: #917

## What is the new behavior?
The default behavior should be to not change casing and it should be configureable with a global variable.

That way we can avoid hardcoding this setting into the Astro Build dagfile generator script.

Nope

### Checklist

### DIFF
--- a/python-sdk/src/astro/constants.py
+++ b/python-sdk/src/astro/constants.py
@@ -65,4 +65,3 @@ ExportExistsStrategy = Literal["replace", "exception"]
 MergeConflictStrategy = Literal["ignore", "update", "exception"]
 
 ColumnCapitalization = Literal["upper", "lower", "original"]
-DEFAULT_COLUMN_CAPITALIZATION = "original"
--- a/python-sdk/src/astro/constants.py
+++ b/python-sdk/src/astro/constants.py
@@ -65,3 +65,4 @@ ExportExistsStrategy = Literal["replace", "exception"]
 MergeConflictStrategy = Literal["ignore", "update", "exception"]
 
 ColumnCapitalization = Literal["upper", "lower", "original"]
+DEFAULT_COLUMN_CAPITALIZATION = "original"
--- a/python-sdk/src/astro/files/types/csv.py
+++ b/python-sdk/src/astro/files/types/csv.py
@@ -7,12 +7,17 @@ from astro.constants import FileType as FileTypeConstants
 from astro.files.types.base import FileType
 from astro.utils.dataframe import convert_columns_names_capitalization
 
+from astro import settings
+
 
 class CSVFileType(FileType):
     """Concrete implementation to handle CSV file type"""
 
     def export_to_dataframe(
-        self, stream, columns_names_capitalization="original", **kwargs
+        self,
+        stream,
+        columns_names_capitalization=settings.COLUMN_CAPITALIZATION,
+        **kwargs,
     ) -> pd.DataFrame:
         """read csv file from one of the supported locations and return dataframe
 
--- a/python-sdk/src/astro/files/types/csv.py
+++ b/python-sdk/src/astro/files/types/csv.py
@@ -3,11 +3,10 @@ from __future__ import annotations
 import io
 
 import pandas as pd
+from astro import settings
 from astro.constants import FileType as FileTypeConstants
 from astro.files.types.base import FileType
 from astro.utils.dataframe import convert_columns_names_capitalization
-
-from astro import settings
 
 
 class CSVFileType(FileType):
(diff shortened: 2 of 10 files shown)


python-sdk/src/astro/constants.py
```diff
@@ -65,3 +65,4 @@ def __str__(self) -> str:
 MergeConflictStrategy = Literal["ignore", "update", "exception"]
 
 ColumnCapitalization = Literal["upper", "lower", "original"]
+DEFAULT_COLUMN_CAPITALIZATION = "original"
```
kaxil, line 68:
```suggestion
```

kaxil: We can just add this as fallback in settings.py

python-sdk/src/astro/settings.py
```diff
@@ -41,3 +41,8 @@
 AUTO_ADD_INLETS_OUTLETS = conf.getboolean(
     "astro_sdk", "auto_add_inlets_outlets", fallback=True
 )
+
+
+COLUMN_CAPITALIZATION = conf.get(
+    "astro_sdk", "column_capitalization", fallback=DEFAULT_COLUMN_CAPITALIZATION
+)
```
kaxil, line 48:
```suggestion
COLUMN_CAPITALIZATION = conf.get("astro_sdk", "column_capitalization", fallback="original")
```

python-sdk/src/astro/settings.py
```diff
@@ -1,5 +1,5 @@
 from airflow.configuration import conf
-from astro.constants import DEFAULT_SCHEMA
+from astro.constants import DEFAULT_COLUMN_CAPITALIZATION, DEFAULT_SCHEMA
```
kaxil, line 2:
```suggestion
from astro.constants import DEFAULT_SCHEMA
```

python-sdk/src/astro/settings.py
```diff
@@ -41,3 +41,8 @@
 AUTO_ADD_INLETS_OUTLETS = conf.getboolean(
     "astro_sdk", "auto_add_inlets_outlets", fallback=True
 )
+
+
+COLUMN_CAPITALIZATION = conf.get(
+    "astro_sdk", "column_capitalization", fallback="original"
+)
```
kaxil, line 48: We should add docs about it in https://github.com/astronomer/astro-sdk/blob/main/python-sdk/docs/configurations.rst

kaxil: Also, you have the following in the PR description

>Does this introduce a breaking change?
> Nope

But this does introduce a breaking change. When writing the release notes, we should ask users to change the config if needed if they relied on this behavior and mention that we consider it a "bug" hence the change

feluelle: Why is that a breaking change? The default does not change as we are setting `COLUMN_CAPITALIZATION` default/fallback to `original`.

python-sdk/src/astro/files/types/csv.py
```diff
@@ -12,7 +13,10 @@ class CSVFileType(FileType):
     """Concrete implementation to handle CSV file type"""
 
     def export_to_dataframe(
-        self, stream, columns_names_capitalization="original", **kwargs
+        self,
+        stream,
+        columns_names_capitalization=settings.COLUMN_CAPITALIZATION,
+        **kwargs,
```
dimberman, line 19: Can you add type hints here

python-sdk/src/astro/files/types/ndjson.py
```diff
@@ -14,7 +15,10 @@ class NDJSONFileType(FileType):
     """Concrete implementation to handle NDJSON file type"""
 
     def export_to_dataframe(
-        self, stream, columns_names_capitalization="original", **kwargs
+        self,
+        stream,
+        columns_names_capitalization=settings.COLUMN_CAPITALIZATION,
+        **kwargs,
```
dimberman, line 21: type hints here as well

python-sdk/src/astro/settings.py
dimberman: +1

python-sdk/tests/sql/operators/test_dataframe.py
```diff
@@ -334,3 +335,24 @@ def sample_df_1(**kwargs):
 
     task = sample_df_1(output_table=output_table)
     assert task.operator.outlets == []
+
+
+def test_col_capitalization_default_dataframe(sample_dag):
+    """Test that capitalization of col remains the same"""
+
+    assert settings.COLUMN_CAPITALIZATION == "original"
```
dimberman, line 343: You can use pytest parameters to make this test apply to all three known settings (and maybe even an invalid setting?)

dimberman (changes requested):

python-sdk/src/astro/settings.py
feluelle: Thanks @kaxil for pointing this out 👍

utkarsharma2:
We no longer need this PR, since we have added a fix in a separate PR.

Issue #885, opened 2022-09-21
Move benchmarking config out of codebase to GCS


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, we have config kept in the Codebase due to which whenever we want to modify it we have to generate a new docker image.

**Describe the solution you'd like**
We can move the config to GCS and load it from there. This way any mods to config won't affect the docker image.

**Are there any alternatives to this feature?**
Open to suggestions.

**Acceptance Criteria**

- [ ] While changing the config we should not generate a new docker image

kaxil:
When would we ever want to change the "benchmark config"?

sunank200:
It doesn't make sense to have the config both on github and GCS.

phanikumv:
Closing this as discussed with the team that it is not required anymore

PR #954, opened 2022-09-28
Moved config out of codebase to GCS

utkarsharma2:
# Description
Currently, we have config kept in the Codebase due to which whenever we want to modify it we have to generate a new docker image.

closes: #885 

## What is the new behavior?
We can move the config to GCS and load it from there. This way any mods to config won't affect the docker image.

Nope


python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -1,21 +1,27 @@
...
 
 START_DATE = datetime(2000, 1, 1)
 
 
 def load_config():
-    config_path = Path(Path(__file__).parent.parent, "config.json").resolve()
-    with open(config_path) as fp:
-        return json.load(fp)
+    tmp_config = "/tmp/config.json"
+    storage_client = storage.Client()
```
utkarsharma2, line 18: @kaxil @tatiana @dimberman 
1.  Is there a way we can do without this intermediate step of creating a file?
2. is there a better way to explicitly specify credentials?

kaxil: The following is not clear to me:

- When would we ever want to change the "benchmark config" file? Why?
- Why do changes in config need a new dockerfile? Can you link the code for that please?

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -1,21 +1,26 @@
...
+    storage_client.download_blob_to_file(
+        file_obj=open(tmp_config, "wb"),
+        blob_or_uri=benchmark_settings.publish_benchmarks_config_path,
+    )
+    return json.load(open(tmp_config))
```
feluelle, line 23: Please use the `open` context manager. Otherwise in case of errors it might not close the file.

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -1,21 +1,26 @@
...
 
 
 def load_config():
-    config_path = Path(Path(__file__).parent.parent, "config.json").resolve()
-    with open(config_path) as fp:
-        return json.load(fp)
+    tmp_config = "/tmp/config.json"
+    storage_client = storage.Client()
+    storage_client.download_blob_to_file(
+        file_obj=open(tmp_config, "wb"),
```
feluelle, line 20: Same here. You are also opening the file twice.

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -1,21 +1,26 @@
...
+from tests.benchmark import settings as benchmark_settings
 
 START_DATE = datetime(2000, 1, 1)
 
 
 def load_config():
-    config_path = Path(Path(__file__).parent.parent, "config.json").resolve()
-    with open(config_path) as fp:
-        return json.load(fp)
+    tmp_config = "/tmp/config.json"
```
feluelle, line 17: Please use [gettempdir](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir).

python-sdk/tests/benchmark/dags/evaluate_load_file.py
feluelle: It seems you are actually not interested about the file. In this case you could use [download_as_text](https://googleapis.dev/python/storage/latest/storage/blobs.html?highlight=download_as_string#google.cloud.storage.blob.Blob.download_as_text) and just pass the result to `json.loads`.

python-sdk/tests/benchmark/dags/evaluate_load_file.py
tatiana: The `tempfile.NamedTemporaryFile` may be more suitable, if we use something like:
```
from tempfile import NamedTemporaryFile

with tempfile.NamedTemporaryFile(mode="w") as temp_file:
    print(temp_file.name)
```
https://docs.python.org/3/library/tempfile.html

tatiana (approved):

feluelle (changes requested):
I see you already got an approval, but please resolve my comments first. I think the `open` calls can be an issue.

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -1,21 +1,26 @@
...
+    storage_client = storage.Client()
+    storage_client.download_blob_to_file(
```
tatiana, line 19: It may be worth having it in a versioned GCS bucket and printing the object's version.
I think it is important that we print the content of this file for troubleshooting.

python-sdk/tests/benchmark/dags/evaluate_load_file.py
feluelle: Yes, or https://github.com/astronomer/astro-sdk/pull/954#discussion_r983117319

python-sdk/tests/benchmark/settings.py
```diff
@@ -11,3 +11,6 @@
 publish_benchmarks_table_grouping_col = os.getenv(
     "ASTRO_PUBLISH_BENCHMARK_GROUPING_COL", "revision"
 )
+publish_benchmarks_config_path = os.getenv(
```
tatiana, line 14: Great improvement!

python-sdk/tests/benchmark/settings.py
```diff
@@ -11,3 +11,6 @@
...
+    "ASTRO_PUBLISH_BENCHMARK_CONFIG_PATH", "gs://astro-sdk-config/config.json"
```
tatiana, line 15: Perhaps we should name this `ASTRO_BENCHMARK_CONFIG_PATH`

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -1,21 +1,26 @@
 import json
 import os
 from datetime import datetime
-from pathlib import Path
```
tatiana, line 4: @utkarsharma2 I think this is a nice feature to debug locally (e.g. run the benchmark for a subset of files), but I believe the CI should be running the tests for a more stable version of the config.

My suggestion is:
* by default, we keep the behaviour of retrieving the configuration from the config.json file. It is available in the Docker image. In this case, we'd print saying we are using the built-in version of the config.
* If `ASTRO_BENCHMARK_CONFIG_PATH` is defined, we can retrieve the config from either the local or remote file path referenced by it, and we'd be printing the content of that configuration file / its location within the logs.

This would give us the flexibility to troubleshoot while also having some reproducibility and trackability.

PR #960, opened 2022-09-29
Generate benchmark config file

pankajastro:
# Description
A util script to generate the benchmark configuration file.

gen_config.py will generate configuration json that will have datasets of sizes (10kb, 100kb, 10mb, 100mb, 1gb, 2gb, 5gb, 10gb) at locations(gcs, s3) of types (ndjson, csv, parquet) and will have databases (bigquery, postgres, redshift, snowflake, sqlite). 
Added some new connections in `ci-test-connections.yaml` that would be used only in benchmarking


python-sdk/tests/benchmark/wait_for_completion.sh
```diff
@@ -10,7 +10,7 @@ while [[ $complete_status -ne 0 ]] && [[ $failed_status -ne 0 ]]; do
   complete_status=$?
 done
 
-kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
+# kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
```
tatiana, line 13: It's probably better to do this as part of a separate PR

PR #970, opened 2022-09-30
Create a custom xcom backend that removes XCOM_PICKLING argument

dimberman:
# Description
## What is the new behavior?

-
-
-

### Checklist


python-sdk/src/astro/files/base.py
```diff
@@ -163,6 +169,33 @@ def _path_to_dataset_uri(self) -> str:
...
 
+    def to_json(self):
+        self.log.debug("converting file %s into json", self.path)
+        filetype = None if not self.filetype else self.filetype.value
+        return {
+            "class": "File",
+            "conn_id": self.conn_id,
+            "path": self.path,
+            "filetype": filetype,
+            "normalize_config": self.normalize_config,
```
tatiana, line 180: @dimberman bringing the discussion here, that we may want to use `attrs.asdict`, as previously suggested by @kaxil , doing some filtering - also mentioned by @feluelle :
https://github.com/apache/airflow/blob/029ebacd9cbbb5e307a03530bdaf111c2c3d4f51/airflow/lineage/__init__.py#L79-L81
https://www.attrs.org/en/stable/api.html#attrs.asdict

python-sdk/src/astro/__init__.py
```diff
@@ -13,14 +13,20 @@
...
 # multiple places, but at this point it's the only workaround if you'd like
 # your custom conn type to show up in the Airflow UI.
+
+# To prevent this from being a breaking change, we have to add our backend to taskinstance explicitly
+# The alternative would be to force users to set our custom xcom backend in their airflow configs, which would
+# be a breaking change which would require a 2.0 release
+from airflow.models import taskinstance
+from astro.custom_backend.astro_custom_backend import AstroCustomXcomBackend
+
+taskinstance.XCom = AstroCustomXcomBackend
```
tatiana, line 27: I'm curious if Airflow has any alternative, cleaner way for a library to redefine the default `XCom` backend. I would love to have feedback from other Airflow committers @kaxil @ashb @feluelle

python-sdk/src/astro/files/base.py
dimberman: @tatiana I mentioned this elsewhere but the asdict function ended up adding a bunch of fields that didn't work with deserialization and made the serializer more complex. There are also certain fields we have to add to make the desert work that wouldn't be added by asdict. I don't think it buys us much.

python-sdk/src/astro/__init__.py
```diff
@@ -2,25 +2,20 @@
...
-
-import astro.sql  # noqa: F401
-from airflow.configuration import conf
+# To prevent this from being a breaking change, we have to add our backend to taskinstance explicitly
+# The alternative would be to force users to set our custom xcom backend in their airflow configs, which would
+# be a breaking change which would require a 2.0 release
+from airflow.models import taskinstance
+from astro.custom_backend.astro_custom_backend import AstroCustomXcomBackend
 
+taskinstance.XCom = AstroCustomXcomBackend
```
kaxil, line 11: Are you sure this works?

python-sdk/src/astro/utils/dataframe.py
```diff
@@ -17,3 +27,32 @@ def convert_columns_names_capitalization(
...
+    """
+    Passes a dataframe into a File using parquet as an efficient storage format. This allows us to use
+    Json as a storage method without filling the metadata database. the values for conn_id and bucket path can
+    be found in the airflow.cfg as follows:
+
+    [astro]
+    dataframe_storage_conn_id=...
+    dataframe_storage_url=///
+    :param df: Dataframe to convert to file
+    :return:
```
feluelle, line 42: missing description for `:return:`

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -164,12 +191,15 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
...
-            return pandas_dataframe
+            if isinstance(function_output, pd.DataFrame):
+                return convert_to_file(pandas_dataframe)
+            elif isinstance(function_output, list):
+                return [
+                    convert_to_file(obj) if isinstance(obj, pd.DataFrame) else obj
+                    for obj in function_output
+                ]
+            else:
+                return function_output
```
feluelle, line 202:
```suggestion
            if isinstance(function_output, list):
                return [
                    convert_to_file(obj) if isinstance(obj, pd.DataFrame) else obj
                    for obj in function_output
                ]
            return function_output
```

python-sdk/src/astro/files/base.py
```diff
@@ -163,6 +169,33 @@ def _path_to_dataset_uri(self) -> str:
...
+            "is_dataframe": self.is_dataframe,
+        }
+
+    @classmethod
+    def from_json(cls, serialized_object: dict):
+        filetype = (
+            constants.FileType(serialized_object["filetype"])
+            if serialized_object.get("filetype")
```
feluelle, line 188:
```suggestion
            if "filetype" in serialized_object
```

python-sdk/src/astro/files/base.py
```diff
@@ -163,6 +169,33 @@ def _path_to_dataset_uri(self) -> str:
         )
         return new_parsed_url.geturl()
 
+    def to_json(self):
+        self.log.debug("converting file %s into json", self.path)
+        filetype = None if not self.filetype else self.filetype.value
```
feluelle, line 174:
```suggestion
        filetype = self.filetype.value if self.filetype else None
```

python-sdk/src/astro/utils/dataframe.py
```diff
@@ -17,3 +27,32 @@ def convert_columns_names_capitalization(
...
+    """
+    unique_id = random.choice(string.ascii_lowercase) + "".join(
+        random.choice(string.ascii_lowercase + string.digits) for _ in range(64)
+    )
+
+    # importing here to prevent circular imports
+    from astro.files import File
+
+    file = File(
+        path=DATAFRAME_STORAGE_URL + "/" + unique_id + ".parquet",
```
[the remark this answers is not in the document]
dimberman: @feluelle if you have two slashes it resolves to a single slash. 

foo//bar=foo////bar=foo/bar

python-sdk/src/astro/utils/serializer.py
```diff
@@ -0,0 +1,96 @@
...
+        elif obj["class"] == "File":
+            log.debug("Found file dictionary %s, will attempt to deserialize", obj)
+            return _deserialize_file(obj)
+        else:
+            return obj["value"]
+    elif isinstance(obj, str):
+        log.debug("Found string, will attempt to deserialize")
+        return _attempt_to_deser_unknown_object(obj)
+    else:
+        return obj
```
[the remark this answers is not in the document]
dimberman: I don't understand what you're trying to do here?

python-sdk/tests/sql/operators/transform/test_transform.py
```diff
@@ -109,7 +109,12 @@ def raw_sql_query(my_input_table: Table, created_table: Table, num_rows: int):
         return "SELECT * FROM {{my_input_table}} LIMIT {{num_rows}}"
 
     @task
-    def validate_raw_sql(cur):
+    def validate_raw_sql(cur: pd.DataFrame):
+        from sqlalchemy.engine.row import LegacyRow
+
+        # Note: It's a broken feature on th main branch that this is return in a list of lists. Problem reported here:
```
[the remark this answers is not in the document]
dimberman: Naw I just hadn't written the ticket yet :)

python-sdk/src/astro/utils/serializer.py
```diff
@@ -0,0 +1,96 @@
...
+        file = convert_to_file(obj)
+        return serialize(file)
+    elif isinstance(obj, File):
+        return obj.to_json()
+    elif isinstance(obj, list):
+        return [serialize(o) for o in obj]
+    elif isinstance(obj, str):
+        return {"class": "string", "value": obj}
+    else:
+        return _attempt_to_serialize_unknown_object(obj)
```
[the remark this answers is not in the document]
dimberman: I think it's safer to keep these as elif statements as it guarantees we won't run multiple in case one doesn't have a return statement. WDYT @kaxil

python-sdk/src/astro/utils/dataframe.py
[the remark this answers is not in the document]
feluelle: Okay, but this might in 0,00000001% not be what the user want. For example, in aws s3 you can have `/` as part of the prefix. You can also have `//` and it seems `File` wouldn't allow it. 😅

python-sdk/src/astro/utils/serializer.py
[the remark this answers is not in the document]
feluelle: Removing the `elif` and `else`'s as if you have _if-return_'s you don't need them.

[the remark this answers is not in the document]
feluelle: With multiple returns what you want to do is essentially a row of shortcuts. A shortcut will be taken if the condition is met. In the end in case of no shortcuts taken you just return what you would return without shortcuts. (I don't know if this was helpful or not 😅)

It is my personal preference, but feel free to keep it as it is :)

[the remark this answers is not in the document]
dimberman: I'm gonna keep it as is to prevent any unwanted actions if a return statement is removed :)

python-sdk/tests/utils/test_dataframe.py
```diff
@@ -0,0 +1,12 @@
+import pandas as pd
+from astro.utils.dataframe import convert_to_file
+
+
+def test_convert_to_file():
+    """Ensures that there is parity when converting to file and exporting back to dataframe"""
+    data_for_df = {"col1": [1, 2], "col2": [3, 4]}
+    # Calling DataFrame constructor on list
+    df = pd.DataFrame(data_for_df)
```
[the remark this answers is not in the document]
dimberman: Please forgive me, as an American I don't speak English very well :)

python-sdk/src/astro/utils/dataframe.py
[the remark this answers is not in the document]
dimberman: @feluelle I think this is a case of "don't let the perfect be the enemy of the good". If we see this causing problem for users it can be fixed in a patch, but I believe this helps a lot more people than it hurts.

.github/workflows/ci-benchmark.yaml
```diff
@@ -44,6 +44,7 @@ jobs:
       AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE: True
       FORCE_COLOR: "true"
       ASTRO_PUBLISH_BENCHMARK_DATA: True
+      AIRFLOW__CORE__CUSTOM_XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
```
kaxil, line 47:
```suggestion
      AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
```

feluelle (approved):
LGTM in general, but please resolve these small comments first. :)

python-sdk/src/astro/custom_backend/astro_custom_backend.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @staticmethod
+    def deserialize_value(result: "XCom") -> Any:
+        """
+        Deserializing the result of a xcom_pull before passing th result to the next task.
+        :param result:
+        :return:
+        """
+        result = BaseXCom.deserialize_value(result)
+        return deserialize(result)
```
uranusjr, line 49: Since `serialize_value` only calls `serialize` when the storage variables are set, why does this always call `deserialize`? It feels to me that one of these may be wrong.

python-sdk/src/astro/custom_backend/serializer.py
```diff
@@ -0,0 +1,97 @@
...
+        file = convert_to_file(obj)
+        return serialize(file)
+    elif isinstance(obj, File):
+        return obj.to_json()
+    elif isinstance(obj, list):
+        return [serialize(o) for o in obj]
+    elif isinstance(obj, str):
+        return {"class": "string", "value": obj}
+    else:
+        return _attempt_to_serialize_unknown_object(obj)
```
uranusjr, line 38: Should this handle tuple?

PR #971, opened 2022-10-01
Users can now pass multiple dataframes through aql.dataframe

dimberman:
# Description
Currently, there is no way to output multiple dataframes from an aql.dataframe function and expect both outputted dataframes to be used in a future task. This limits our flows to single dataframe workflows which could go against many common use-cases (e.g. separating test data and and run data for model testing).

closes: #909 

## What is the new behavior?
Now users can return lists of items from the aql.dataframe, including both dataframes and non-dataframe items.

### Checklist


python-sdk/src/astro/sql/operators/base_operator.py
```diff
@@ -1,8 +1,9 @@
 from abc import ABC
 
 from airflow.models.baseoperator import BaseOperator
+from airflow.utils.log.logging_mixin import LoggingMixin
 from astro.sql.operators.upstream_task_mixin import UpstreamTaskMixin
 
 
-class AstroSQLBaseOperator(UpstreamTaskMixin, BaseOperator, ABC):
+class AstroSQLBaseOperator(UpstreamTaskMixin, BaseOperator, LoggingMixin, ABC):
```
kaxil, line 8: The `BaseOperator` inherits `LoggingMixin` via `AbstractOperator` class so this is not needed

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -148,17 +148,21 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
...
+        function_output = self._convert_column_capitlization_for_output()
         if self.output_table:
+            self.log.debug(
+                "Output table found, converting function output to SQL table"
+            )
+            if not isinstance(function_output, pd.DataFrame):
+                raise ValueError(
+                    "Astro can only turn a single dataframe into a table. Please change your"
+                    "function output."
+                )
```
kaxil, line 160: Can we add a test case for this code block

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -130,7 +130,7 @@ def __init__(
             **kwargs_with_datasets(kwargs=kwargs, output_datasets=self.output_table),
         )
 
-    def execute(self, context: Context) -> Table | pd.DataFrame:
+    def execute(self, context: Context) -> BaseTable | pd.DataFrame | list:
```
kaxil, line 133: Will the list always be a `list[pd.DataFrame]`?

If so we should change it to:

```suggestion
    def execute(self, context: Context) -> BaseTable | pd.DataFrame | list[pd.Dataframe]:
```

python-sdk/tests/sql/operators/test_dataframe.py
```diff
@@ -334,3 +334,43 @@ def sample_df_1(**kwargs):
 
     task = sample_df_1(output_table=output_table)
     assert task.operator.outlets == []
+
```
kaxil, line 337: Deletion of `tests/airflow/__init__.py` file looks like a mistake ?

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -169,7 +173,31 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
...
+                    df=f, columns_names_capitalization=self.columns_names_capitalization
+                )
+                for f in function_output
+            ]
+        elif isinstance(function_output, dict):
+            function_output = {
+                k: convert_columns_names_capitalization(
+                    df=v, columns_names_capitalization=self.columns_names_capitalization
+                )
+                for k, v in function_output.items()
```
kaxil, line 193: This needs test too

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -169,7 +173,31 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
                 and not settings.ALLOW_UNSAFE_DF_STORAGE
             ):
                 raise IllegalLoadToDatabaseException()
-            return pandas_dataframe
+            return function_output
+
+    def _convert_column_capitlization_for_output(self):
+        """Handles column capitalization for single outputs, lists, and dictionaries"""
+        function_output = self.python_callable(*self.op_args, **self.op_kwargs)
+        if isinstance(function_output, (list, tuple)):
```
kaxil, line 181: This needs test

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -169,7 +173,31 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
...
+                )
+                for f in function_output
+            ]
+        elif isinstance(function_output, dict):
+            function_output = {
+                k: convert_columns_names_capitalization(
+                    df=v, columns_names_capitalization=self.columns_names_capitalization
+                )
+                for k, v in function_output.items()
+            }
```
kaxil, line 194: This needs test

python-sdk/tests/sql/operators/test_dataframe.py
```diff
@@ -334,3 +334,43 @@ def sample_df_1(**kwargs):
...
+    df2 = pandas.DataFrame({"numbers": [4, 5, 6], "colors": ["red", "white", "blue"]})
+
+    @aql.dataframe
+    def create_multiple_dataframes():
+        return [df1, df2]
+
+    @aql.dataframe()
+    def validate_dfs(input_dfs):
+        assert df1.equals(input_dfs[0])
+        assert df2.equals(input_dfs[1])
```
kaxil, line 350: Should we add/modify one of the example DAGs or docs - https://github.com/astronomer/astro-sdk/blob/main/python-sdk/docs/astro/sql/operators/dataframe.rst to tell that we support multiple dataframes

python-sdk/src/astro/sql/operators/dataframe.py
dimberman: Not necessarily. It could be a list of anything, we're just doing our special storage method on dataframes. So [1, "foo", DataFrame, 4] would be legal here

python-sdk/tests/sql/operators/test_dataframe.py
dimberman: @kaxil we need to figure out how to handle that, because it's causing all kinds of issues in pycharm. Basically airflow imports don't work because the ide thinks we're importing the tests airflow package instead of the apache airflow package.

kaxil: I would like to see that when you are back. I can't replicate it on my Pycharm

python-sdk/src/astro/sql/operators/base_operator.py
kaxil: Try upgrading `apache-airflow` in the environment that your IDE use -- nevertheless, it is something in your local setup

https://github.com/apache/airflow/blob/2.4.1/airflow/models/baseoperator.py#L438-L439
https://github.com/apache/airflow/blob/2.4.1/airflow/models/abstractoperator.py#L67

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -169,7 +175,33 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
                 and not settings.ALLOW_UNSAFE_DF_STORAGE
             ):
                 raise IllegalLoadToDatabaseException()
-            return pandas_dataframe
+            return function_output
+
+    @staticmethod
+    def _convert_column_capitlization_for_output(
```
feluelle, line 181:
```suggestion
    def _convert_column_capitalization_for_output(
```

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -147,18 +147,24 @@ def execute(self, context: Context) -> Table | pd.DataFrame:
         self.op_kwargs = load_op_kwarg_table_into_dataframe(
             self.op_kwargs, self.python_callable, self.columns_names_capitalization
         )
-
-        pandas_dataframe = self.python_callable(*self.op_args, **self.op_kwargs)
-        pandas_dataframe = convert_columns_names_capitalization(
-            df=pandas_dataframe,
-            columns_names_capitalization=self.columns_names_capitalization,
+        function_output = self.python_callable(*self.op_args, **self.op_kwargs)
+        function_output = self._convert_column_capitlization_for_output(
```
feluelle, line 151:
```suggestion
        function_output = self._convert_column_capitalization_for_output(
```

python-sdk/src/astro/sql/operators/dataframe.py
dimberman: added

python-sdk/tests/sql/operators/test_dataframe.py
dimberman: @kaxil added

dimberman:
@kaxil I added an example in the docs

.gitignore
```diff
@@ -2,6 +2,7 @@
 coverage.xml
 .env
 .coverage
+.python-version
```
feluelle, line 5:
```suggestion
**/.python-version
```

feluelle (approved):

Issue #932, opened 2022-09-28
Create basic project directory structure and version command using typer


tatiana:
**Please describe the feature you'd like to see**

Create a project structure, so we implement the SQL CLI using  https://typer.tiangolo.com/typer-cli/

**Describe the solution you'd like**

Replace existing Python argparse by a library which will give more features to end-users (including auto-complete).

Setup Python modules so different developers can contribute to different parts of the project concurrently 

By the end of this task, users should be able to run:
```
flow version
```

And see the SQL CLI version.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [x] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [x] Integration tests (if the feature relates to a new database or external service)
- [x] Example DAG
- [x] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [x] Exception handling in case of errors
- [x] Logging (are we exposing useful information to the user? e.g. source and destination)
- [x] Improve the documentation (README, Sphinx, and any other relevant)
- [x] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

tatiana:
Created a PR:
https://github.com/astronomer/astro-sdk/pull/984

Still fixing things so the CI checks pass

PR #984, opened 2022-10-03
Implement SQL CLI flow version

tatiana:
# Description

This PR implements the command `flow version`. Closes: #932

The package was made available at: https://pypi.org/project/astro-sql-cli/

There are a few things which were done as part of this PR, and I could separate them if the team prefers different PRs (there are separate commits), including:
- the adoption of Poetry
- addition of the Makefile
- changes to Github actions workflows

No

### Checklist


.github/workflows/ci-python-sdk.yaml
```diff
@@ -1,21 +1,22 @@
...
     paths:
       - 'python-sdk/**'
       - '.github/workflows/ci-python-sdk.yaml'
-      - '*'
   pull_request:
     branches: [ 'main', 'release-**' ]
     paths:
       - 'python-sdk/**'
       - '.github/workflows/ci-python-sdk.yaml'
-      - '*'
```
tatiana, line 14: We don't want to trigger `python-sdk` tests if only the `sql-cli` changes

.deepsource.toml
```diff
@@ -2,7 +2,8 @@ version = 1
 
 test_patterns = [
   "python-sdk/tests/**",
-  "python-sdk/test-docker/**"
+  "python-sdk/test-docker/**",
```
tatiana, line 5: We need to add these otherwise Deepsource complains about the asserts

.pre-commit-config.yaml
```diff
@@ -40,9 +40,9 @@ repos:
       - id: forbid-crlf
       - id: remove-crlf
       - id: forbid-tabs
-        exclude: ^python-sdk/mk/|^python-sdk/docs/Makefile|^python-sdk/Makefile$|^python-sdk/tests/benchmark/Makefile$
+        exclude: ^python-sdk/mk/|^python-sdk/docs/Makefile|^python-sdk/Makefile$|^python-sdk/tests/benchmark/Makefile|^sql-cli/Makefile$
```
tatiana, line 43: Make sure we don't convert Makefile tabs to spaces

sql-cli/noxfile.py
```diff
@@ -16,61 +16,35 @@ def dev(session: nox.Session) -> None:
     development environment to ``.nox/dev``.
     """
     session.install("nox")
-    session.install("-e", "../python-sdk/.[all]")
-    session.install("-e", "../python-sdk/.[tests]")
-    session.install("-e", ".[all]")
-    session.install("-e", ".[tests]")
+    session.install("poetry")
```
tatiana, line 19: Trying to unify to install things with Poetry, where possible.

sql-cli/noxfile.py
```diff
@@ -16,61 +16,35 @@ def dev(session: nox.Session) -> None:
...
+    session.run("poetry", "install")
 
 
 @nox.session(python=["3.7", "3.8", "3.9"])
-@nox.parametrize("airflow", ["2.2.5", "2.3"])
+@nox.parametrize("airflow", ["2.3"])
```
tatiana, line 24: I feel that if the Python SDK is tested with both versions of Airflow, that is probably good enough - unless we find errors. I'm happy to rollback this change.

sql-cli/noxfile.py
```diff
@@ -16,61 +16,35 @@ def dev(session: nox.Session) -> None:
...
-@nox.session()
-@nox.parametrize(
-    "extras",
-    [
-        ("postgres-amazon", {"include": ["postgres", "amazon"]}),
-        ("snowflake-amazon", {"include": ["snowflake", "amazon"]}),
-        ("sqlite", {"include": ["sqlite"]}),
-    ],
-)
-def test_examples_by_dependency(session: nox.Session, extras):
```
tatiana, line 60: We may want to have something similar for the DAGs generated by the SQL CLI - but that's not part of the scope or the code ATM.

sql-cli/pyproject.toml
```diff
@@ -1,31 +1,14 @@
...
-    "smart-open",
-    "SQLAlchemy>=1.3.18"
-]
-
-keywords = ["airflow", "provider", "astronomer", "sql", "decorator", "task flow", "elt", "etl", "dag"]
+license = "Apache-2.0"
+keywords = ["sql", "transform", "analyst", "workflow", "airflow", "astronomer"]
+packages = [{include = "sql_cli"}]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
```
pankajkoti, line 13: Not your change, but curious what does this classifier mean? and if it is relevant/suitable for us?

pankajkoti (commented):
Looks great to me Tatiana, just a minor question inline.

sql-cli/pyproject.toml
kaxil: Yeah that should be changed to Alpha. It is used for PyPI classifiers: https://pypi.org/classifiers/

sql-cli/Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+setup: system install  ## Setup a local development environment
+
+pre-commit:  # Run pre-commit
+	@git ls-files -- . | xargs pre-commit run --files
+
+flow:  # Build and run the latest version of flow
+
+test:  # Run tests
+	@poetry run pytest -s --cov --cov-branch --cov-report=term-missing
```
feluelle, line 34: Are we not using nox to run pytest?

tatiana: Nox is using this same command line at the moment..!

For local dev, if we can avoid yet another tool, I'd prefer - but if you feel strongly about having nox here, I can add it, @feluelle!

sql-cli/pyproject.toml
```diff
@@ -1,33 +1,16 @@
-[build-system]
-requires = ["flit_core ~=3.2"]
-build-backend = "flit_core.buildapi"
-
-[project]
+[tool.poetry]
```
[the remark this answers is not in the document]
tatiana: We don't need - @kaxil - and, yes, I'll make a separate PR!

I like the fact it materialises the dependencies in a lock file - it is better for reproducibility. And it is much quicker than `pipenv`, for instance. I believe it is a middle ground between using Docker for local development and pure pip.

sql-cli/Makefile
feluelle: I would say we should be consistent and use the same things in local dev and ci. If we use nox in CI, let's use it locally as well. If we can remove nox from ci, I am fine with this too.

.github/workflows/ci-python-sdk.yaml
feluelle: The `*` here only finds files in the root directory e.g. `.pre-commit-config.yaml`

.pre-commit-config.yaml
```diff
@@ -94,13 +94,8 @@ repos:
       - id: mypy
         additional_dependencies: [types-PyYAML, types-attrs, attrs]
         pass_filenames: false
-        files: "^python-sdk/"
+        files: "^python-sdk/|sql-cli/$"
```
feluelle, line 97:
```suggestion
        files: "^python-sdk/|^sql-cli/"
```
Why did you change this? Woudn't `sql-cli/$` only find the folder `sql-cli/` - but mypy has to run on files or?

sql-cli/Makefile
feluelle: `sql-sdk` I have never heard of it. What is it? 😆

sql-cli/noxfile.py
tatiana: Removed!

.github/workflows/ci-python-sdk.yaml
```diff
@@ -1,21 +1,22 @@
-name: Build and test astro
+name: Build and test astro Python SDK
 on:
   push:
     branches: [ 'main', 'release-**' ]
     paths:
       - 'python-sdk/**'
       - '.github/workflows/ci-python-sdk.yaml'
-      - '*'
```
tatiana, line 8: To revert this change - since we want the tests to run if we change pre-commit hook etc

sql-cli/pyproject.toml
```diff
@@ -1,30 +1,16 @@
...
+description = "Empower analysts to build workflows to transform data using SQL"
 authors = [
     {name = "Astronomer", email = "humans@astronomer.io"},
 ]
 readme = "README.md"
 license = {file = "LICENSE"}
-
 requires-python = ">=3.7"
-dependencies = [
-    "astro-sdk-python[all]>=1.0.0",
```
tatiana, line 20: We'll need this to validate the DAGs we've generated are valid. But this is out of the scope of the current ticket.
@feluelle, can we add a ticket for us to test the generated DAGs using a similar strategy to the Python SDK?

tatiana: This probably makes sense to be a test dependency

pankajkoti (approved):
Looks great to me. Thanks a lot for the all the work!

PR #987, opened 2022-10-03
Track the airflow operator error in benchmark

pankajastro:
# Description
Keep track of the GCS result file if an error occurs on any run and at the end of execution print these error result file paths and exit with error code 1. Also changed while loop to for loop since pipe run the commands in the new process so it was difficult to update the global variable error_code to track the error.

The Benchmark job is green even if the operator has failed since we silently ignore the error and run of the next dataset and database. https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/benchmark/run.sh#L79

## What is the new behavior?
CI will still run for all the databases and datasets but it will fail if the operator will fail and print the GCS log path in CI

No

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/run.sh
+++ b/python-sdk/tests/benchmark/run.sh
@@ -4,7 +4,6 @@ set -x  # print the shell command before being executed
 set -v  # run bash in verbose mode (echos each command before running it)
 set -e  # stop the execution instantly if a command returns a non-zero status
 
-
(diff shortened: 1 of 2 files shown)


tatiana (commented):
@pankajastro are you still working on this ticket? It seems the benchmark job failed in the CI

pankajastro:
> @pankajastro are you still working on this ticket? It seems the benchmark job failed in the CI

yes, @tatiana will check and fix it today

PR #990, opened 2022-10-04
speed up CI using new dag running system

dimberman:
# Description
## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/python-sdk/tests/sql/operators/utils.py
+++ b/python-sdk/tests/sql/operators/utils.py
@@ -1,16 +1,35 @@
-import copy
-import os
+from __future__ import annotations
 import time
 
 import pandas as pd
 from airflow.exceptions import BackfillUnfinished
-from airflow.executors.debug_executor import DebugExecutor
 from airflow.models.dag import DAG
-from airflow.models.taskinstance import TaskInstance
-from airflow.utils import timezone
-from airflow.utils.state import State
 from astro.sql.table import Metadata
 from pandas.testing import assert_frame_equal
+
+import copy
+import logging
+import os
+import sys
+from datetime import datetime
+from typing import (
+    Any,
+)
+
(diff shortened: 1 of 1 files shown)


python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -427,14 +427,11 @@ def test_load_file_using_file_connection_fails_nonexistent_conn(
...
         "output_table": Table(name=OUTPUT_TABLE_NAME, **sql_server_params),
     }
-    with pytest.raises(BackfillUnfinished):
+    with pytest.raises(AirflowNotFoundException):
         with sample_dag:
             load_file(**task_params)
         test_utils.run_dag(sample_dag)
 
-    expected_error = "Failed to execute task: The conn_id `fake_conn` isn't defined."
-    assert expected_error in caplog.text
```
kaxil, line 436: Should we keep this error since `AirflowNotFoundException` can occur for few more cases?

image

kaxil (approved):
Minor comment added, lgtm. <3 it

python-sdk/tests/sql/operators/test_load_file.py
dimberman: @kaxil turns out you can actually do this with pytest.raises to check error message

```
    with pytest.raises(AirflowNotFoundException, match=r"The conn_id `fake_conn` isn't defined"):
        with sample_dag:
            load_file(**task_params)
```

kaxil: Yeah, go for it 🎉

dimberman: @kaxil added that. The other exceptions are simple things like "filenotfound" so we don't need to check error logs for those.

kaxil (approved):

PR #992, opened 2022-10-04
Run all example dags

feluelle:
# Description

Currently, we have to manually select dag_ids from examples to be tested.

## What is the new behavior?

We automatically test all example dags.

No.

### Checklist

### DIFF
--- a/python-sdk/tests/test_example_dags.py
+++ b/python-sdk/tests/test_example_dags.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+import airflow
 import pytest
 from airflow.models.dagbag import DagBag
 from airflow.utils import timezone
@@ -47,9 +48,20 @@ def session():
 
 
 dag_bag = DagBag(Path.cwd().parent / "example_dags")
+airflow_2_3 = pytest.mark.skipif(
+    airflow.__version__ < "2.3.0", reason="Require Airflow version >= 2.3.0"
+)
+airflow_2_3_dag_ids = [
+    "example_dynamic_map_task",
+    "example_dynamic_task_template",
+]
+dag_ids = [
+    pytest.param(dag_id, marks=airflow_2_3) if dag_id in airflow_2_3_dag_ids else dag_id
+    for dag_id in dag_bag.dag_ids
+]
 
 
-@pytest.mark.parametrize("dag_id", dag_bag.dag_ids)
+@pytest.mark.parametrize("dag_id", dag_ids)
 def test_example_dag(session, dag_id):
     dag = dag_bag.get_dag(dag_id)
 
--- a/python-sdk/tests/test_example_dags.py
+++ b/python-sdk/tests/test_example_dags.py
@@ -1,6 +1,5 @@
-import os
+from pathlib import Path
 
-import airflow
 import pytest
 from airflow.models.dagbag import DagBag
 from airflow.utils import timezone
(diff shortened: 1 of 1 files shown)


python-sdk/tests/test_example_dags.py
```diff
@@ -47,47 +47,23 @@ def session():
...
-        "example_snowflake_partial_table_with_append",
-        "example_sqlite_load_transform",
-        "example_append",
-        "example_load_file",
-        "example_transform",
-        "example_merge_bigquery",
-        "example_transform_file",
-        "calculate_popular_movies",
-    ],
+dag_bag = DagBag(Path.cwd().parent / "example_dags")
```
kaxil, line 50: Can we assert that this (`dag_bag.dags`) is not empty, please?

kaxil: https://github.com/astronomer/astro-sdk/actions/runs/3183804425/jobs/5191462176 is failing -- I haven't checked if it is related or not

feluelle: It was related. Fixed in c99e549.

feluelle: The issue was that `cwd` points to the `python-sdk` - not the tests folder. We always run from the project directory e.g. `python-sdk` or `sql-cli`.

And actually `assert dag_bag.dags` wouldn't have helped because it was using the airflow default example dags.

feluelle: WDYT of af23cb3 instead of `.cwd`?

python-sdk/tests/test_example_dags.py
```diff
@@ -47,47 +47,23 @@ def session():
...
 )
-def test_example_dag(session, dag_id):
-    dir_path = os.path.dirname(os.path.realpath(__file__))
-    db = DagBag(dir_path + "/../example_dags")
-    dag = db.get_dag(dag_id)
-
-    if dag is None:
-        raise NameError(f"The DAG with dag_id: {dag_id} was not found")
-    wrapper_run_dag(dag)
+airflow_2_3_dag_ids = {
```
tatiana, line 54: @feluelle do you know if there is a clean way for us to set the Airflow version a DAG depends on the DAG itself? It would be great if this information was there

python-sdk/tests/test_example_dags.py
```diff
@@ -47,47 +47,23 @@ def session():
...
+    "example_dynamic_map_task",
+    "example_dynamic_task_template",
+}
+dag_ids = [
+    pytest.param(dag_id, marks=airflow_2_3) if dag_id in airflow_2_3_dag_ids else dag_id
+    for dag_id in dag_bag.dag_ids
+]
```
tatiana, line 61: Following @kaxil 's feedback, perhaps we could have a simple test asserting `dag_ids` is not empty

python-sdk/tests/test_example_dags.py
feluelle: We could use DAG `tags` 🤔

feluelle: Something like this bd6add4

python-sdk/tests/test_example_dags.py
feluelle: Done 0a6f104. :)

python-sdk/tests/test_example_dags.py
```diff
@@ -47,48 +51,43 @@ def session():
...
-        "example_merge_bigquery",
-        "example_transform_file",
-        "calculate_popular_movies",
-    ],
-)
-def test_example_dag(session, dag_id):
-    dir_path = os.path.dirname(os.path.realpath(__file__))
-    db = DagBag(dir_path + "/../example_dags")
-    dag = db.get_dag(dag_id)
+DAG_BAG = DagBag(Path(__file__).parent.parent / "example_dags", include_examples=False)
```
kaxil, line 54: This looks good, last thought, how can we protect against dag parsing errors in one of the dags?

For example, if `example_amazon_s3_postgres` dag has a parsing error, currently it will be skipped from running the test.

python-sdk/tests/test_example_dags.py
kaxil: Sounds good, yeah. Similar point -> https://github.com/astronomer/astro-sdk/pull/992#discussion_r987802528

How can we protect if one of our example dags isn't a valid DAG, should we validate the # of files equal to # of dags? or is that too brittle, any other thoughts?

python-sdk/tests/test_example_dags.py
feluelle: Added a test for import errors 8f0f067

python-sdk/tests/test_example_dags.py
feluelle: I would not check for `# of files equal to # of dags` as some files could have more than one dag. one file = one dag is more of a best practice than a rule.

feluelle: I have added a suggestion to the comment you linked.

tatiana (approved):
Looks great, thanks, @feluelle !

kaxil (approved):
Love it, definitely a better solution

PR #995, opened 2022-10-04
Add changelog for 1.1.1

kaxil:
Sync's changelog from release to main

### DIFF
--- a/python-sdk/docs/CHANGELOG.md
+++ b/python-sdk/docs/CHANGELOG.md
@@ -1,4 +1,13 @@
 # Changelog
+## 1.1.1
+
+### Improvements
+* Add filetype when resolving path in case of loading into dataframe [#881](https://github.com/astronomer/astro-sdk/issues/881)
+
+### Bug fixes
+* Fix postgres performance regression (example from one_gb file - 5.56min to 1.84min) [#876](https://github.com/astronomer/astro-sdk/issues/876)
+
+
 ## 1.1.0
 
 ### Features


python-sdk/docs/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
 # Changelog
+## 1.1.1
+
+### Improvements
+* Add filetype when resolving path in case of loading into dataframe [#881](https://github.com/astronomer/astro-sdk/issues/881)
+
+### Bug fixes
+* Fix postgres performance regression (example from one_gb file - 5.56min to 1.84min) [#876](https://github.com/astronomer/astro-sdk/issues/876)
+
+
```
tatiana, line 10: It  may be worth adding a section summarising the improvements to the benchmark, including running in the CI

kaxil: That's internal details, not user facing

feluelle: I would be in favor of having a "Misc/Internal" section.

kaxil: 👍 Are you both ok with us doing that in 1.2 when we publish some of it publicly?

Since our 1.1.1 release does not contain any of those things

tatiana: BTW: my suggestion wasn't for it to be within "Bug fixes", but in one of the sections we used in the past - we already had `Misc` and `Enhancements` sections in previous releases - and they were used for this type of thing.

I don't consider the benchmark changes only internal - since they affect users (the results changed, how we run the benchmarks changed) and developers (they can see the benchmark running by adding labels to PRs, and the benchmarks are running regularly in an automated way).

kaxil: But they are not in 1.1.1 release - they are in main branch. Check https://github.com/astronomer/astro-sdk/tree/1.1.1

tatiana (approved):

Issue #795, opened 2022-09-07
Remove dependency on AIRFLOW__CORE__ENABLE_XCOM_PICKLING


dimberman:
**Please describe the feature you'd like to see**
A clear and concise description of what problem feature solves or what use-case it would enable.

To ensure maximum adoption by enterprise organizations, we should ensure that we follow best security practices. Enabling xcom pickling globally just to handle dataframe storage can be seen as a fairly global solution to what is ultimately a local problem. We should find a way to handle large dataframe storage that is actually independent of Airflow's xcom s.t. users can use whatever xcom backend they like.

**Describe the solution you'd like**
A solution that would work in this case would be to have an env variable `AIRFLOW__ASTRO_SDK__DATAFRAME_STORAGE_CONN_ID` which points to an s3 or GCS (and eventually azure storage) conn_id. This conn_id can be used by `aql.dataframe` to store dataframes in parquet or arrow binary format and store the resulting URL in xcom. We can still have `AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE` for Astro developers who do not want to store dataframes in external blob storage.

**Are there any alternatives to this feature?**
Another alternative would be to create an astro-sdk custom backend, which would prevent using any other custom backend with said astro-sdk backend.

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] Create a way to store dataframes that does not involve setting the `AIRFLOW__CORE__ENABLE_XCOM_PICKLING` env var]
- [ ] Should use a binary format like parquet for storing dataframes
- [ ] should not degrade aql.dataframe performance
- [ ] Should not break any existing user flows
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

[...]

PR #997, opened 2022-10-04
Allow installing Python SDK without enabling pickling

kaxil:
# Description

Users can not even run Airflow if Astro SDK is installed but if they don't want to run it. For example, we can't install Astro SDK in the runtime image.

## What is the new behavior?

This PR reverts https://github.com/astronomer/astro-sdk/pull/446 . Without this, users can not even run Airflow if Astro SDK is installed.

This will even allow users to use `File` and `Table` objects in `inlets` and `outlets` without using any operators or decorators.

This will also allow us to install SDK in Runtime image without impacting users who don't want to use SDK

The only caveat if users don't enable XCom pickling, the following error is raised but that is OK. We would like to get this in Astro SDK 1.2 (or 1.1.2) as we plan to tackle removal of pickling completely in https://github.com/astronomer/astro-sdk/issues/795:

```python
[2022-10-04, 23:09:10 UTC] {xcom.py:599} ERROR - Could not serialize the XCom value into JSON. If you are using pickle instead of JSON for XCom, then you need to enable pickle support for XCom in your *** config.
[2022-10-04, 23:09:10 UTC] {taskinstance.py:1851} ERROR - Task failed with exception
  File "/usr/local/lib/python3.9/site-packages/airflow/utils/session.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/airflow/models/taskinstance.py", line 2378, in xcom_push
    XCom.set(
  File "/usr/local/lib/python3.9/site-packages/airflow/utils/session.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/airflow/models/xcom.py", line 206, in set
    value = cls.serialize_value(
  File "/usr/local/lib/python3.9/site-packages/airflow/models/xcom.py", line 597, in serialize_value
    return json.dumps(value).encode('UTF-8')
  File "/usr/local/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Table is not JSON serializable
```

This will allow users to use the following without pickling:

```python
import pendulum
from airflow import DAG, Dataset
from airflow.operators.bash import BashOperator

from astro.files.base import File
dag1_dataset = File(path="s3://dag1/output_1.txt")

with DAG(
    dag_id='dataset_produces_1',
    catchup=False,
    start_date=pendulum.datetime(2021, 1, 1, tz="UTC"),
    schedule='@daily',
    tags=['produces', 'dataset-scheduled'],
) as dag1:
    BashOperator(outlets=[dag1_dataset], task_id='producing_task_1', bash_command="sleep 5")

with DAG(
    dag_id='dataset_consumes_1',
    catchup=False,
    start_date=pendulum.datetime(2021, 1, 1, tz="UTC"),
    schedule=[dag1_dataset],
    tags=['consumes', 'dataset-scheduled'],
) as dag3:
    # [END dag_dep]
    BashOperator(
        outlets=[Dataset('s3://consuming_1_task/dataset_other.txt')],
        task_id='consuming_1',
        bash_command="sleep 5",
    )
```

This works because of https://github.com/apache/airflow/commit/25a1a1b597ab20fa601a6b701f83ada02b6334a5 which we added in 2.4.0

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/__init__.py
+++ b/python-sdk/src/astro/__init__.py
@@ -14,7 +14,6 @@ __version__ = "1.2.0.dev1"
 # astro.database, and this is what leads to the circular dependency.
 
 import astro.sql  # noqa: F401
-from airflow.configuration import conf
 
 
 # This is needed to allow Airflow to pick up specific metadata fields it needs
@@ -32,10 +31,3 @@ def get_provider_info() -> dict:
         "hook-class-names": [],
         "extra-links": [],
     }
-
-
-if not conf.getboolean(section="core", key="enable_xcom_pickling"):
-    raise OSError(
-        "AIRFLOW__CORE__ENABLE_XCOM_PICKLING environment variable needs to be set to True or enable_xcom_pickling=true "
-        "in airflow.cfg before importing astro-sdk-python."
-    )
--- a/python-sdk/tests/test_constants.py
+++ b/python-sdk/tests/test_constants.py
@@ -1,8 +1,3 @@
-import importlib
-import os
-from unittest import mock
-
-import pytest
 from astro.constants import (
     SUPPORTED_DATABASES,
     SUPPORTED_FILE_LOCATIONS,
@@ -23,19 +18,3 @@ def test_supported_file_types():
 def test_supported_databases():
     expected = ["bigquery", "postgres", "redshift", "snowflake", "sqlite"]
     assert sorted(SUPPORTED_DATABASES) == expected
-
-
-@mock.patch.dict(
-    os.environ, {"AIRFLOW__CORE__ENABLE_XCOM_PICKLING": "False"}, clear=True
-)
-def test_enable_xcom_pickling_set_false_in_env():
-    import astro
-
-    with pytest.raises(OSError) as exe_info:
-        importlib.reload(astro)
-
-    assert (
-        exe_info.value.args[0]
-        == "AIRFLOW__CORE__ENABLE_XCOM_PICKLING environment variable needs to be set to True or "
-        "enable_xcom_pickling=true in airflow.cfg before importing astro-sdk-python."
-    )


python-sdk/src/astro/__init__.py
```diff
@@ -32,10 +31,3 @@ def get_provider_info() -> dict:
         "hook-class-names": [],
         "extra-links": [],
     }
-
-
-if not conf.getboolean(section="core", key="enable_xcom_pickling"):
-    raise OSError(
```
tatiana, line 38: @kaxil, what if we didn't raise an `OSError` but had a warning alerting users that some features may not work if `enable_xcom_pickling` is False? (e.g. `aql.dataframe`)? It can be misleading to not have any warning, and several features do not work.

I understand the motivation in allowing users to install and use the parts worth it without pickling - but, still, it's better to have a warning (at the very list on the tasks/decorators which will not work, such as `aql.dataframe`).

kaxil: `aql.dataframe` has a check before pushing to XCom though:

https://github.com/astronomer/astro-sdk/blob/b9c974ed105a4b0a95c09a17a055410ee0ad29c5/python-sdk/src/astro/sql/operators/dataframe.py#L166-L171

kaxil: where would you like me to add the warning

feluelle (approved):
The current exception looks sufficient to me:

> Could not serialize the XCom value into JSON. If you are using pickle instead of JSON for XCom, then you need to enable pickle support for XCom in your *** config.

PR #998, opened 2022-10-05
Fix tasks template to use transform_file

feluelle:
# Description

Currently, the tasks template creates `transform_sql` tasks, but we only have `transform_file`.

related: #836 

## What is the new behavior?

Fix references to `transform_sql` by replacing it with `transform_file`

No.

### Checklist


feluelle:
This was not covered by tests since we don't have integration tests yet. I would suggest we add it when we have the optimized local runner. WDYT @tatiana @dimberman? Isn't it easier to add it with the new runner?

kaxil (approved):

Issue #892, opened 2022-09-23
Make transform_file not depend on the transform decorator


dimberman:
**Please describe the feature you'd like to see**

The current `transform_file` operator is a hack around the `transform` decorator. It calls a function that returns the SQL file path. This hack was reasonable for a 1.0 release. Still, since we plan to make the transform_file operator a basis for our SQL CLI, we should invest in improving it to make it return an actual operator instead of a taskflow decorator. This solution will add the following benefits:

1. Looking into our benchmarking, operators work better than taskflow decorators at scale.
2. Having a straightforward operator will make debugging simpler. Users will not need to understand any taskflow magic to see why their SQL did no run.
3. We can offer a rendered "SQL" object in the Airflow UI, which would be much harder with a taskflow object that only creates a SQL object at runtime.

**Describe the solution you'd like**

We should create a TransformFileOperator that a) parses a SQL file with templating, b) applies all needed parameters, and c) runs the SQL to return a table object. To keep the `aql.transform_file` function, the function can return `TransformFileOperator().output` in a similar fashion to the merge operator.

**Are there any alternatives to this feature?**
None that I can think of

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1004, opened 2022-10-06
Fix transform_file to not depend on transform decorator

feluelle:
# Description

Currently, the `transform_file` is using the `transform` decorator underneath which has a few limitations. For more information check out the issue below.

closes: #892  

## What is the new behavior?

- use TransformOperator in transform_file
- set default task_id for TransformOperator to "transform"
- fix example for transform_file
- resolve XComArgs when finding first table
- add parameters to template_fields

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/transform.py
+++ b/python-sdk/src/astro/sql/operators/transform.py
@@ -9,7 +9,9 @@ except ImportError:
     from airflow.decorators.base import task_decorator_factory
     from airflow.decorators import _TaskDecorator as TaskDecorator
 
+from airflow.decorators.base import get_unique_task_id
 from airflow.models.xcom_arg import XComArg
+from sqlalchemy.sql.functions import Function
 
 from astro.sql.operators.base_decorator import BaseSQLDecoratedOperator
 from astro.utils.typing_compat import Context
@@ -20,6 +22,33 @@ class TransformOperator(BaseSQLDecoratedOperator):
     Given a SQL statement and (optional) tables, execute the SQL statement and output
     the result into a SQL table.
     """
+
+    def __init__(
+        self,
+        conn_id: str | None = None,
+        parameters: dict | None = None,
+        handler: Function | None = None,
+        database: str | None = None,
+        schema: str | None = None,
+        response_limit: int = -1,
+        response_size: int = -1,
+        sql: str = "",
+        task_id: str = "",
+        **kwargs: Any,
+    ):
+        task_id = task_id or get_unique_task_id("transform")
+        super().__init__(
+            conn_id=conn_id,
+            parameters=parameters,
+            handler=handler,
+            database=database,
+            schema=schema,
+            response_limit=response_limit,
+            response_size=response_size,
+            sql=sql,
+            task_id=task_id,
+            **kwargs,
+        )
 
     def execute(self, context: Context):
         super().execute(context)
@@ -104,7 +133,7 @@ def transform(
 def transform_file(
     file_path: str,
     conn_id: str = "",
-    parameters: Mapping | Iterable | None = None,
+    parameters: dict | None = None,
     database: str | None = None,
     schema: str | None = None,
     **kwargs: Any,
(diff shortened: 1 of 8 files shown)


python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -103,7 +132,7 @@ def my_sql_statement(table1: Table, table2: Table, execution_date) -> Table:
 def transform_file(
```
kaxil, line 132: Do we even need a separate `transform_file` decorator? or can we have the following too instead?

```
        table_from_query = aql.transform(
            file_path="tests/sql/operators/transform/test.sql",
            parameters={"input_table": imdb_table},
            op_kwargs={"output_table": target_table},
        )
```

dimberman: I'd honestly love this solution if we can get it to work. Keeps the API simple. Eventually we could even have the transform function to point at directories to do the live rendering :).

python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -128,14 +157,13 @@ def transform_file(
...
 
-    @transform(
+    return TransformOperator(
         conn_id=conn_id,
         parameters=parameters,
+        handler=None,
         database=database,
         schema=schema,
+        sql=file_path,
+        python_callable=lambda: (file_path, parameters),
```
dimberman, line 167: So would this not run into the same scaling issues that the decorated tasks run into? Have you tested this at scale?

feluelle: I am 100% sure that it won't run into the scaling issues, because I tested `transform` decorator vs `TransformOperator`.

python-sdk/src/astro/sql/operators/transform.py
feluelle: > Do we even need a separate transform_file decorator?

The `transform_file` is a function - not a decorator.

> or can we have the following too instead?

No, because `aql.transform` already exists as a decorator in the module and we want to call a function. That is why I previously had created a `transform_sql` which you can use to pass either sql or a file_path, but I reverted it because of discussion in https://github.com/astronomer/astro-sdk/pull/897#discussion_r984080333.

If you only want to have the same name, we can have it in a separate module, but `aql.transform` won't work.

python-sdk/src/astro/sql/operators/transform.py
tatiana: Out of curiosity, what are the numbers for both approaches?

python-sdk/src/astro/utils/table.py
```diff
@@ -1,97 +1,122 @@
...
+    :return: first valid table found in op_kwargs.
     """
-    first_table = None
-    kwargs_of_table_type = [
-        op_kwargs[kwarg.name]
+    kwargs = [
+        op_kwargs[kwarg.name].resolve(context)
+        if isinstance(op_kwargs[kwarg.name], XComArg)
+        else op_kwargs[kwarg.name]
         for kwarg in inspect.signature(python_callable).parameters.values()
```
tatiana, line 57: I have the feeling we're trying to accomplish a lot in a single list comprehension - would it be worth to break this logic in (a) separate function(s)?

python-sdk/src/astro/utils/table.py
```diff
@@ -1,97 +1,122 @@
...
-    return first_table
+    params = [
+        param.resolve(context) if isinstance(param, XComArg) else param
+        for param in parameters.values()
+    ]
+    tables = [param for param in params if isinstance(param, BaseTable)]
+
+    if _have_same_conn_id(tables):
+        return tables[0]
+    return None
```
tatiana, line 84: It would be great if we could have tests covering the new scenarios introduced as part of this PR.

python-sdk/src/astro/sql/operators/transform.py
feluelle: See https://github.com/astronomer/astro-sdk/pull/897#discussion_r986817532

python-sdk/src/astro/utils/table.py
feluelle: 🤔 not sure.. It collects all kwargs from the python callable signature and if they are XcomArg they will be resolved first.

python-sdk/src/astro/utils/table.py
feluelle: This is not a new scenario. We had the same scenarios already before that, maybe you could not see it that well. But we were check for conn_id and either return the first table or none.

feluelle: Added tests for the xcom arg feature in find_first_table which you probably meant. See 504c9d1

dimberman (approved):

PR #1006, opened 2022-10-06
Add docs requirements to venv via make

feluelle:
# Description

Currently, in our local setup we don't have docs requirement installed.

## What is the new behavior?

We add docs requirements to the venv (via make).

No.

### Checklist


python-sdk/mk/local.mk
```diff
@@ -37,6 +37,7 @@ install: virtualenv  ## Install python dependencies in existing virtualenv
 	@$(PIP) install pre-commit
 	@cd .. && $(PIP) install -e .[all]
 	@cd .. && $(PIP) install .[tests]
+	@cd .. && $(PIP) install .[doc]
```
kaxil, line 40: Should we just change this to the following:

```suggestion
	@cd .. && $(PIP) install -e .[all, tests, doc]
```

Issue #955, opened 2022-09-28
Test End to End CI benchmarking


utkarsharma2:
We need to verify the end-to-end CI/Benchmarking is working.
1. Adding a label
2. Cron

sunank200:
@utkarsharma2 did the cron job run successfully? If so we should update this ticket

sunank200:
@utkarsharma2 to sync with @uranusjr to debug this

utkarsharma2:
Test end to end with 10kb to 1gb files.

PR #1012, opened 2022-10-06
Add benchmarking config

utkarsharma2:
# Description

1. We don't have the complete benchmarking config added to the main branch.
2. As per the earlier discussion we need to run benchmarks in parallel
3. Because of a memory leak we need to increase the config for instance. - https://github.com/astronomer/astro-sdk/issues/1058

related: #955 

## What is the new behavior?
1. Added complete grid config
2. Added multiple configs to be used in parallel jobs
3. Increased the machine size to n2-standard-16

Nope

### Checklist
- [ ] Created tests that fail without the change (if possible)


python-sdk/tests/benchmark/infrastructure/terraform/vars.tf
```diff
@@ -25,5 +25,5 @@ variable "tfstate_bucket" {
 variable "gke_node_pool_machine_type" {
   type        = string
   description = "GCP machine type used for the GKE Node Pool"
-  default     = "n2-standard-8"  # 8 vCPU and 32GB RAM, more options at https://cloud.google.com/compute/vm-instance-pricing
+  default     = "n2-standard-16"  # 8 vCPU and 32GB RAM, more options at https://cloud.google.com/compute/vm-instance-pricing
```
utkarsharma2, line 28:
```suggestion
  default     = "n2-standard-16"  # 16 vCPU and 64GB RAM, more options at https://cloud.google.com/compute/vm-instance-pricing
```

python-sdk/tests/benchmark/run.sh
```diff
@@ -9,6 +9,7 @@ repeat=${1:-3}  # how many times we want to repeat each DAG run (default: 3)
 
 benchmark_dir="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 config_path="${benchmark_dir}/config.json"
+#config_path="${benchmark_dir}/$1"
```
utkarsharma2, line 12:
```suggestion
```

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
```diff
@@ -14,9 +14,9 @@ spec:
         command: ["./run.sh"]
         resources:
           requests:
-            memory: 16Gi
+            memory: 50Gi
```
sunank200, line 17: Can we add a comment about why is this high?

sunank200 (approved):

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
pankajastro: We can't expect a single worker to have 50Gi in a real-world scenario, this is too much memory IMO

phanikumv: why are postgres and snowflake tests failing here?

kaxil: 50 GI: this is just wrong... 

I 100% agree with @pankajastro -- we can't keep increasing the worker memory. We should instead fix the issue which is causing us to need that memory

kaxil (dismissed):
50 GI: this is just wrong...

I 100% agree with @pankajastro -- we can't keep increasing the worker memory. We should instead fix the issue which is causing us to need that memory

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
kaxil: We have talked about not changing the benchmarking environment (Datasets, worker size etc) multiple times now

python-sdk/tests/benchmark/infrastructure/terraform/vars.tf
```diff
@@ -25,5 +25,5 @@ variable "tfstate_bucket" {
 variable "gke_node_pool_machine_type" {
   type        = string
   description = "GCP machine type used for the GKE Node Pool"
-  default     = "n2-standard-8"  # 8 vCPU and 32GB RAM, more options at https://cloud.google.com/compute/vm-instance-pricing
+  default     = "n2-standard-16"  # 16 vCPU and 64GB RAM, more options at https://cloud.google.com/compute/vm-instance-pricing
```
kaxil, line 28: Why!

kaxil:
The PR description also just says that "we are doing something" and "not why do we need this"

python-sdk/tests/benchmark/infrastructure/terraform/vars.tf
utkarsharma2: Hey Kaxil, I have created the ticket for memory leak, which is the reason why we require higher config - https://github.com/astronomer/astro-sdk/issues/1058

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
utkarsharma2: @kaxil - You can still run all the benchmarks one at a time in "n2-standard-8" and 16Gi memory. This is a separate issue of memory leak which needs to be investigated. We are only changing this because we are running 150+ benchmarks on the same machine which leads to consuming the memory and not freeing it. WDYT?

python-sdk/tests/benchmark/infrastructure/terraform/vars.tf
kaxil: Let's solve that first

python-sdk/tests/benchmark/infrastructure/kubernetes/job.yaml
kaxil: >We are only changing this because we are running 150+ benchmarks on the same machine which leads to consuming the memory and not freeing it. WDYT?

We should investigate and fix it rather than working around it

utkarsharma2:
When running end to end, observing the pods are killed due to - `OOMKilled` issue 

```
kubectl describe pod -n benchmark benchmark-207ebe0-txn9h
W1031 18:55:22.893163   28770 gcp.go:119] WARNING: the gcp auth plugin is deprecated in v1.22+, unavailable in v1.26+; use gcloud instead.
To learn more, consult [https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke)
Name:             benchmark-207ebe0-txn9h
Namespace:        benchmark
Priority:         0
Service Account:  default
Node:             gke-astro-sdk-benchmark-4d102245-cmwm/10.128.0.30
Start Time:       Mon, 31 Oct 2022 17:51:12 +0530
Labels:           controller-uid=83ca59cd-5d4e-4ad8-9317-db6cb7c68b1c
job-name=benchmark-207ebe0
Annotations:      <none>
Status:           Failed
Reason:           Evicted
Message:          The node was low on resource: memory. Container benchmark was using 27698652Ki, which exceeds its request of 16Gi.
IP:               10.84.5.10
IPs:
IP:           10.84.5.10
Controlled By:  Job/benchmark-207ebe0
Containers:
benchmark:
Container ID:  containerd://aaabe8ad80555104ca37b724961068c2fd9889e069dc5619d0f5a55b175ce73d
Image:         [[gcr.io/astronomer-dag-authoring/benchmark:207ebe0](http://gcr.io/astronomer-dag-authoring/benchmark:207ebe0)](http://gcr.io/astronomer-dag-authoring/benchmark:207ebe0)
Image ID:      [[gcr.io/astronomer-dag-authoring/benchmark@sha256:381642d3f6051beeac015b7f3e81422d37f7177f5b2fc2071c2723a04e0c46ca](http://gcr.io/astronomer-dag-authoring/benchmark@sha256:381642d3f6051beeac015b7f3e81422d37f7177f5b2fc2071c2723a04e0c46ca)](http://gcr.io/astronomer-dag-authoring/benchmark@sha256:381642d3f6051beeac015b7f3e81422d37f7177f5b2fc2071c2723a04e0c46ca)
Port:          <none>
Host Port:     <none>
Command:
./run.sh
State:          Terminated
Reason:       OOMKilled
Exit Code:    137
Started:      Mon, 31 Oct 2022 17:51:52 +0530
Finished:     Mon, 31 Oct 2022 18:45:35 +0530
Ready:          False
Restart Count:  0
Limits:
memory:  32Gi
Requests:
memory:     16Gi
Environment:  <none>
Mounts:
/var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-fjkff (ro)
Conditions:
Type              Status
Initialized       True
Ready             False
ContainersReady   False
PodScheduled      True
Volumes:
kube-api-access-fjkff:
Type:                    Projected (a volume that contains injected data from multiple sources)
TokenExpirationSeconds:  3607
ConfigMapName:           kube-root-ca.crt
ConfigMapOptional:       <nil>
DownwardAPI:             true
QoS Class:                   Burstable
Node-Selectors:              <none>
Tolerations:                 [[node.kubernetes.io/not-ready:NoExecute](http://node.kubernetes.io/not-ready:NoExecute)](http://node.kubernetes.io/not-ready:NoExecute) op=Exists for 300s
[[node.kubernetes.io/unreachable:NoExecute](http://node.kubernetes.io/unreachable:NoExecute)](http://node.kubernetes.io/unreachable:NoExecute) op=Exists for 300s
Events:
Type     Reason               Age   From     Message

---

Warning  Evicted              10m   kubelet  The node was low on resource: memory. Container benchmark was using 27698652Ki, which exceeds its request of 16Gi.
Normal   Killing              10m   kubelet  Stopping container benchmark
Warning  ExceededGracePeriod  10m   kubelet  Container runtime did not kill the pod within specified grace period.
```

utkarsharma2:
On further investigation - not able to reproduce the memory leaks locally. All the processes are freeing up the memory as expected.
1. 400MB parquet file - Postgres 
Screenshot 2022-10-31 at 7 28 34 PM
2. 1GB parquet file - Postgres
Screenshot 2022-10-31 at 7 29 54 PM
3. 5GB parquet file - Postgres
Screenshot 2022-10-31 at 7 30 57 PM

python-sdk/tests/benchmark/infrastructure/terraform/vars.tf
utkarsharma2: As per my discussion with @kaxil We agreed to run the benchmark on 1KB to 1GB grid, which runs fine on the existing machine size of `n2-standard-8`

.github/workflows/ci-benchmark.yaml
```diff
@@ -4,7 +4,7 @@ on:
     types:
       - labeled
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '0 0 * * MON'
```
pankajastro, line 7: Not sure if MON work in GH action cron can you change this to `cron: '0 0 * * 1` or should we run twice in a week then `cron: '0 0 * * 1,3` (for Mon and Wed) @phanikumv @utkarsharma2 WDYT?

utkarsharma2: I'm okay with `0 0 * * 1`

utkarsharma2, line 7:
```suggestion
    - cron: '0 0 * * 1'
```

utkarsharma2:
Merging into the main, since the failing tests are the fixed in main/flaky

PR #1015, opened 2022-10-07
Fix nox' airflow version

feluelle:
# Description

nox, did not use the airflow version we wanted it to use as we forgot to install it into the session.

## What is the new behavior?

Add session install command to install specific apache airflow version.

No.

### Checklist

### DIFF
--- a/python-sdk/noxfile.py
+++ b/python-sdk/noxfile.py
@@ -24,6 +24,7 @@ def dev(session: nox.Session) -> None:
 @nox.parametrize("airflow", ["2.2.5", "2.3"])
 def test(session: nox.Session, airflow) -> None:
     """Run both unit and integration tests."""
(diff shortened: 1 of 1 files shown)


kaxil (approved):

pankajastro (commented):
I think we need similar changes in sql-cli nox.py file. WDYT?

feluelle:
> I think we need similar changes in sql-cli nox.py file. WDYT?

Currently, sql-cli does not explicitly depend on airflow. We removed it from the dependencies and the nox test function. But will probably add it back as soon as we run end-to-end tests there.

pankajastro (approved):

kaxil:
There was one test failure due to which 2.2.5 test did not run -- but that is ok, it will run on main and we can check there

PR #1016, opened 2022-10-07
Use directory structure for running version specific example DAGs

feluelle:
# Description

Currently, we use Airflow tags for determining which Airflow DAGs to run. The problem with that is that the DAG needs to be in the DagBag to get the tag. Therefore import_errors test would fail always. By using the directory structure we only load the DagBag with files which are supported for the current installed airflow version.

## What is the new behavior?

It now uses the directory structure within example_dags folder to determine which DAGs to run for the installed version.

No.

### Checklist


python-sdk/tests/test_example_dags.py
```diff
@@ -51,38 +49,31 @@ def session():
     return get_session()
 
 
-DAG_BAG = DagBag(Path(__file__).parent.parent / "example_dags", include_examples=False)
-AIRFLOW_VERSION_INDICATOR = "airflow_version:"
-MINIMUM_AIRFLOW_VERSION = "2.2.5"
+def get_dag_folder() -> Path:
```
kaxil, line 52: hmm I don't like the Airflow versions in folder tbh :( 

I would rather prefer a list of dags that needs to be skipped for certain airflow versions in tests

feluelle: But that is not possible if we check for import errors, I believe. We cannot exclude files from the dag_bag dag folder.

feluelle: Or we dynamically create an `.airflowignore` :/

feluelle: but this would still mean we need to add the version information somehow/somewhere outside of the DAG.

python-sdk/tests/test_example_dags.py
```diff
@@ -91,4 +86,5 @@ def test_example_dags_loaded():
 
 
 def test_example_dags_no_import_errors():
+    print(DAG_BAG.dag_folder)
```
feluelle, line 89:
```suggestion
```

python-sdk/docs/astro/sql/operators/append.rst
```diff
@@ -7,7 +7,7 @@ When to use the ``append`` operator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The ``append`` operator allows you to append data from a source table to a target table.
 
-.. literalinclude:: ../../../../example_dags/example_append.py
```
dimberman, line 10: @feluelle why are we nesting these DAGs instead of just having them as separate folders? We are already filling a DAGBag manually so it wouldn't be too hard to just point based on an env variable https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/test_example_dags.py#L54

We could even have it take a list so if we expect 2.4 to run the 2.3 and 2.2.5 DAGs we just have the 2.4 run take the argument `TEST_AIRFLOW_VERSIONS=2.4,2.3,2.2.5`. 

This structure would be confusing to users.

python-sdk/tests/test_example_dags.py
```diff
@@ -51,38 +50,34 @@ def session():
...
+        if path.is_dir():
+            _version = version.parse(path.name)
+            if (
+                isinstance(_version, version.Version)
+                and version.parse(airflow.__version__) >= _version
+            ):
+                return path
 
+    raise ValueError(
+        "Could not find a dag folder with the same name as the current airflow version!"
```
dimberman, line 71: Yeah I think with some very minor changes here we could have separate 2.4, 2.3, and 2.2.5 folders.

dimberman (changes requested):

kaxil:
@feluelle  I have merged https://github.com/astronomer/astro-sdk/pull/1024 -- but do let me know if you don't like that approach

python-sdk/tests/test_example_dags.py
feluelle: Correct, see https://github.com/astronomer/astro-sdk/pull/1016#discussion_r991217537

feluelle:
Closing in favor of https://github.com/astronomer/astro-sdk/pull/1024

PR #1023, opened 2022-10-07
Run CI tests for Airflow 2.4 instead of 2.3

kaxil:
We should run our CI test on Airflow 2.2.5 and 2.4. I don't think we should run tests against 2.3 anymore


kaxil:
I am merging this without approval as currently no-one is available to review PR and this needs to go in sooner than later

PR #1024, opened 2022-10-07
Skip certain example DAGs by using `.airflowignore`

kaxil:
This is an alternative to https://github.com/astronomer/astro-sdk/pull/1016

The main benefit is it is clear which dags will be skipped, is completely automated without any user-facing changes

### DIFF
--- a/python-sdk/tests/test_example_dags.py
+++ b/python-sdk/tests/test_example_dags.py
@@ -16,7 +16,6 @@ from tenacity import (
     stop_after_attempt,
     wait_exponential,
 )
-
 from tests.sql.operators import utils as test_utils
 
 RETRY_ON_EXCEPTIONS = []
--- a/python-sdk/tests/test_example_dags.py
+++ b/python-sdk/tests/test_example_dags.py
@@ -73,9 +73,6 @@ def get_dag_bag() -> DagBag:
             if Version(airflow.__version__) < min_version:
                 print(f"Adding {files} to .airflowignore")
                 file.writelines(files)
-                # We want to find a version in the map that is less than the installed version
-                # but greater than other versions
-                break
 
     dag_bag = DagBag(example_dags_dir, include_examples=False)
     return dag_bag
--- a/python-sdk/tests/test_example_dags.py
+++ b/python-sdk/tests/test_example_dags.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Iterator
 
 import airflow
 import pytest
@@ -10,12 +9,14 @@ from airflow.models.dagbag import DagBag
 from airflow.utils import timezone
 from airflow.utils.db import create_default_connections
 from airflow.utils.session import provide_session
+from packaging.version import Version
 from tenacity import (
     retry,
     retry_if_exception_type,
     stop_after_attempt,
     wait_exponential,
 )
+
 from tests.sql.operators import utils as test_utils
 
 RETRY_ON_EXCEPTIONS = []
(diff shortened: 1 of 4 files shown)


dimberman (approved):
@kaxil I like this approach much better. One thing to consider is instead of saying which "files to skip" we can actually set a "minimum version" for certain files. So `example_datasets.py` would be minimum "2.4". That way the list doesn't become cumbersome as it grows.

python-sdk/tests/test_example_dags.py
```diff
@@ -51,44 +52,45 @@ def session():
...
-            dag,
-            id=dag_id,
-            marks=pytest.mark.skipif(
-                airflow.__version__ < dag_airflow_version,
-                reason=f"Require Airflow version >= {dag_airflow_version}",
-            ),
-        )
-        for dag_id, dag, dag_airflow_version in get_airflow_dags()
-    ],
+    [pytest.param(dag, id=dag.dag_id) for dag in get_dag_bag().dags.values()],
```
kaxil, line 87: This can also be:

```suggestion
    [pytest.param(dag, id=dag_id) for dag_id, dag in get_dag_bag().dags.items()],
```

kaxil:
> @kaxil I like this approach much better. One thing to consider is instead of saying which "files to skip" we can actually set a "minimum version" for certain files. So `example_datasets.py` would be minimum "2.4". That way the list doesn't become cumbersome as it grows.

Updated, based on your suggestion

python-sdk/tests/test_example_dags.py
```diff
@@ -51,44 +51,44 @@ def session():
...
-            dag,
-            id=dag_id,
-            marks=pytest.mark.skipif(
-                airflow.__version__ < dag_airflow_version,
-                reason=f"Require Airflow version >= {dag_airflow_version}",
-            ),
-        )
-        for dag_id, dag, dag_airflow_version in get_airflow_dags()
-    ],
+    [pytest.param(dag, id=dag_id) for dag_id, dag in get_dag_bag().dags.items()],
```
feluelle, line 85: It is not necessary that you pull the dags. You can use `@pytest.mark.parametrize("dag_id", dag_bag.dag_ids)` and then pull the dag inside the function. This might be a bit faster. WDYT?

python-sdk/tests/test_example_dags.py
```diff
@@ -51,44 +51,44 @@ def session():
...
 
 
-def test_example_dags_loaded():
-    assert DAG_BAG.dags
-
-
-def test_example_dags_no_import_errors():
-    assert not DAG_BAG.import_errors
+def test_example_dags_loaded_with_no_errors():
+    dag_bag = get_dag_bag()
```
feluelle, line 92: Why not access a global dag_bag? `get_dag_bag()` is being called globally anyway. You currently retrieve the dag bag two times.

python-sdk/tests/test_example_dags.py
```diff
@@ -51,44 +51,44 @@ def session():
...
+    airflow_ignore_file = example_dags_dir / ".airflowignore"
 
-def get_airflow_dags() -> Iterator[tuple[str, str]]:
-    for dag_id, dag in DAG_BAG.dags.items():
-        yield dag_id, dag, get_airflow_version(dag)
+    with open(airflow_ignore_file, "w+") as file:
+        for min_version, files in MIN_VER_DAG_FILE_VER.items():
+            if Version(airflow.__version__) < min_version:
+                print(f"Adding {files} to .airflowignore")
+                file.writelines([f"{file}\n" for file in files])
```
feluelle, line 75: This line looks a bit strange because you "redefine" `file`. 

Btw you could also just do this, or?
```diff
- file.writelines([f"{file}\n" for file in files])
+ file.write("\n".join(files)
```

feluelle (commented):
I really like the `.airflowignore` approach. But we could (imo) improve it even further.. we could create a dir for 2.3 and 2.4 within example_dags - not nested - and then you don't have to think about adding the file manually here.

I personally don't think version folders harms the users' experience. In the docs we can directly link the example_dags anyway. So it will only be a different browsing experience.

PR #1029, opened 2022-10-07
Benchmarking Script Fixes

utkarsharma2:
# Description
1. We are not adding revision for when dag fails
2. We create dags even for skipped datasets

## What is the new behavior?
1. Add revision in case of an error in bq
3. Skip dag creation when the dataset is skipped for a DB.

Nope

### DIFF
--- a/python-sdk/tests/benchmark/dags/evaluate_load_file.py
+++ b/python-sdk/tests/benchmark/dags/evaluate_load_file.py
@@ -7,11 +7,12 @@ from urllib.parse import urlparse
 
 import settings as benchmark_settings
 from airflow import DAG
-from astro import sql as aql
 from astro.constants import DEFAULT_CHUNK_SIZE, FileType
-from astro.files import File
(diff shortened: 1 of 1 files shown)


kaxil (approved):

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -195,13 +203,28 @@ def handle_failure(context):
...
+    """
+    Check if the dataset is skipped for database. Default value is False.
+    :param database_name: name of db expected in database_kwargs
+    :param dataset: dataset config
+    """
+    database_kwargs = dataset["database_kwargs"]
+    if database_name not in database_kwargs:
+        return False
+    elif "skip" not in database_kwargs[database_name]:
+        return False
```
pankajkoti, line 216: We might not need this check. The below return bool statement would return False if the 'skip' is absent, because bool(None) is False.

Issue #941, opened 2022-09-28
Implement connection validation (flow validate)


tatiana:
**Please describe the feature you'd like to see**

Using the previously CLI code structure using click, implement the init command to validate all the connections:
```
flow validate
```
or to only validate connections within a specific environment:

```
flow validate --env dev
```

It should output:
```
Checking the <project-name> connections

Validating connection sample_imdb_conn PASSED 
Validating connection output_conn PASSED 

======================= 2 checks passed in 0.1s ==========================
```

**Dependencies**
* Depends on #828 
* Depends on #932

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more)
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)

tatiana:
Implemented as part of #1031

Issue #930, opened 2022-09-28
Create or replace existing local Airflow connections using YAML


tatiana:
**Please describe the feature you'd like to see**

Given a previously generated project structure:
```
sql-test/
    config/
        local/
            configuration.yml
        dev/
            configuration.yml
    workflows/
        workflow_example1/
            example_1a.sql
            example_1b.sql
    .dags/
```
In which a configuration.yml contains:
```
connections:
  - conn_id: postgres_conn
    conn_type: postgres
    host: localhost
    schema:
    login: postgres
    password: postgres
    port: 5433
    extra:
```

Implement a Python function which receives the environment (e.g. local) and creates or replaces all the connections in Airflow.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

Issue #829, opened 2022-09-12
flow init: Implement the SQL project initialization


tatiana:
**Please describe the feature you'd like to see**

Using the previously CLI code structure using click, implement the init command:
```
flow init .
```
or

```
flow init <new project dir>
```

It should output:
```
Initialized Astro SQL project at <dir-path>
Added example workflow, database and connection
```

This command creates a new folder `workflows`, sample configuration/connection files and an example workflow. Examples of addition can be switched off by setting the config `[airflow][astro][sql_examples]` to false.

Example of directory structure:
```
sql-test/
    config/
        local/
            configuration.yml
        dev/
            configuration.yml
    workflows/
        workflow_example1/
            example_1a.sql
            example_1b.sql
    .dags/
```

Example of connection file:
```
config/dev/configuration.yml 
---
connections:
  - conn_id: postgres_conn
    conn_type: postgres
    host: localhost
    schema:
    login: postgres
    password: postgres
    port: 5433
    extra:
```

Possible examples:
```
$ more workflows/workflow_example1/example1.sql 
SELECT * FROM orders_table WHERE amount > 150;

 $ more workflows/workflow_example1/example2.sql
CREATE OR REPLACE TABLE transformed_orders as 
SELECT c.customer_id, customer_name, order_id, purchase_date, amount, type 
FROM {{example1}} fo JOIN customers_table c ON fo.customer_id = c.customer_id
```

**Dependencies**
* Depends on #828 
* Depends on #932

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more)
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)

PR #1031, opened 2022-10-10
Add implementation for validate connections command

pankajkoti:
# Description

No support for validate connections command

closes: #930 
closes: #941 

## What is the new behavior?

Adds a validate command which accepts two optional arguments environment
and connection. Based on the given environment, appropriate configuration.yaml
is utilised to create new or replace existing airflow connections.

No

### Checklist


sql-cli/tests/test___main__.py
```diff
@@ -15,3 +15,8 @@ def test_version():
     result = runner.invoke(app, ["version"])
     assert result.exit_code == 0
     assert "Astro SQL CLI 0.0.0" in result.stdout
+
+
+def test_validate():
+    result = runner.invoke(app, ["validate"])
+    assert result.exit_code == 0
```
tatiana, line 22: It would be great if we could test other aspects of the feature, e.g.:
- did it create the connection?
- did it log the expected message?
- did it not log unwanted messages?

We may want to have separate tests to cover:
- if the connection existed with a different value, was it replaced? It would be nice to not print (deleted, valid, created in those cases - only replaced)
- if the connection existed with the same value, could we just report to the end-user that it is valid?
- if the connection is not valid

sql-cli/sql_cli/settings.py
```diff
@@ -0,0 +1,8 @@
+from pathlib import Path
+
+from airflow.configuration import conf
+
+# This needs to be updated with the `astro dev init` command to set the appropriate project directory.
+SQL_CLI_PROJECT_DIRECTORY = Path(
+    conf.get("astro_sdk", "sql_cli_project_directory", fallback="../tests/")
```
tatiana, line 7: Since we are trying to abstract Airflow, other options to achieve the same could be:
-  pass a parameter on the command line
- default to the "current directory" from where the user is running the command
Without the need to set an Airflow variable.

sql-cli/sql_cli/connections.py
```diff
@@ -0,0 +1,99 @@
...
+    connections.
+    """
+    config_file_contains_connection = False
+    connections = _load_yaml_connections(environment)
+    for conn in connections:
+        conn_id = conn["conn_id"]
+        conn["connection_id"] = conn_id
+        conn.pop("conn_id")
+        data = connection_schema.load(conn)
+        if connection and conn_id != connection:
```
feluelle, line 87: I think `connection` is the wrong word here. Wouldn't `connection_id`/`conn_id` make more sense? It begins already in the `validate` app command.

sql-cli/sql_cli/connections.py
```diff
@@ -0,0 +1,99 @@
...
+    """Gets the configuration yaml for the given environment and loads the connections from it into a dictionary"""
+    config_file = (
+        SQL_CLI_PROJECT_DIRECTORY / "config" / environment / "configuration.yaml"
+    )
+    if not config_file.exists():
+        raise FileNotFoundError(
+            f"Config file configuration.yaml does not exist for environment {environment}"
+        )
+
+    connections = yaml.safe_load(open(config_file))["connections"]
```
feluelle, line 26: Please don't use `open()` like this - instead please use the context manager to open and close the file reliably. In case of a crash the file might not be closed. [This](https://betterprogramming.pub/context-managers-in-python-go-beyond-with-open-as-file-85a27e392114) describes the difference quite well.

sql-cli/sql_cli/connections.py
pankajkoti: I referred for this name for this exposed flag in the validate command from the PRD, where it says connection: https://docs.google.com/document/d/1_iUp0Mci3ZTdcRUUHTvx8F_lffxUv601uBkpJBMhiXs/edit#

pankajkoti: @mikeshwe Is it okay if we call it `connection_id` instead of `connection` as the exposed flag for the `validate` command?

sql-cli/sql_cli/settings.py
pankajkoti: This might need some more thought. Can I work on this later in another PR once the scaffolding/init command and astro-cli integration progresses a bit more how to mount/use the project directory in the docker container? I am a bit unclear at the moment.

sql-cli/sql_cli/connections.py
```diff
@@ -0,0 +1,84 @@
...
+def _test_connection(conn_obj: Connection) -> bool:
+    """Tests whether connection is established successfully with the given data."""
+    try:
+        status, _ = conn_obj.test_connection()
+        if not status:
+            return False
+
+        return True
+    except ValidationError as err:
+        raise BadRequest(detail=str(err.messages))
```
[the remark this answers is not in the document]
pankajkoti: Actually, no. You're right. I had copied the code from REST API endpoint for test_connection which could raise such an error when the POST body is a BAD request. We don't need that check. I have refactored the code.

sql-cli/sql_cli/__main__.py
```diff
@@ -1,13 +1,20 @@
...
 
 from sql_cli import __version__
+from sql_cli.connections import validate_connections
 from sql_cli.dag_generator import generate_dag
 
 app = Typer(add_completion=False)
 
+NON_SQL_CLI_LOGGERS = [logging.getLogger(name) for name in logging.root.manager.loggerDict]
+for logger in NON_SQL_CLI_LOGGERS:
+    logger.setLevel(logging.ERROR)
```
feluelle, line 16:
```suggestion
for logger in logging.root.manager.loggerDict.values():
    logger.setLevel(logging.ERROR)
```
Wondering if this would also work 🤔

feluelle: If not, I would use this so that you don't have to iterate twice of the loggers. :)
```suggestion
for name in logging.root.manager.loggerDict:
    logging.getLogger(name).setLevel(logging.ERROR)
```

feluelle (approved):

sql-cli/sql_cli/settings.py
```diff
@@ -0,0 +1,10 @@
+from pathlib import Path
+
+from airflow.configuration import conf
+
+PROJECT_BASE_PATH = Path(__file__).parent.parent.resolve()
+
+# This needs to be updated with the `astro dev init` command to set the appropriate project directory.
```
tatiana, line 7: I believe we should not depend on `astro dev init`, only on `flow init` to set up a SQL project. I also believe we should not rely on an Airflow configuration to define the project root directory.

I can refactor and remove this as part of the flow init PR #829

sql-cli/sql_cli/settings.py
```diff
@@ -0,0 +1,10 @@
...
+SQL_CLI_PROJECT_DIRECTORY = Path(
+    conf.get("astro_sdk", "sql_cli_project_directory", fallback=PROJECT_BASE_PATH / "tests")
+)
```
tatiana, line 10: I believe we should not depend on `astro dev init`, only on `flow init` to set up a SQL project. I also believe we should not rely on an Airflow configuration to define the project root directory.

I can refactor and remove this as part of the flow init PR #829

sql-cli/tests/test___main__.py
tatiana: That looks very nice, thanks, @pankajkoti ! We can have a separate ticket to add colours if things pass or fail.

sql-cli/sql_cli/connections.py
```diff
@@ -0,0 +1,71 @@
...
+    with create_session() as session:
+        db_connection = session.query(Connection).filter_by(conn_id=conn_id).one_or_none()
+        if db_connection:
+            session.delete(db_connection)
+            session.commit()
+        session.add(conn_obj)
+        session.commit()
+
+
+def validate_connections(environment: str = "default", connection: str | None = None) -> None:
```
tatiana, line 40:
```suggestion
def validate_connections(environment: str = "default", conn_id: str | None = None) -> None:
```

sql-cli/sql_cli/connections.py
```diff
@@ -0,0 +1,71 @@
...
+def _load_yaml_connections(environment: str) -> list[dict[str, Any]]:
+    """Gets the configuration yaml for the given environment and loads the connections from it into a dictionary"""
+    config_file = SQL_CLI_PROJECT_DIRECTORY / "config" / environment / "configuration.yaml"
+    if not config_file.exists():
+        raise FileNotFoundError(
+            f"Config file configuration.yaml does not exist for environment {environment}"
+        )
+
+    with open(config_file) as connections_file:
+        connections: list[dict[str, Any]] = yaml.safe_load(connections_file)["connections"]
```
tatiana, line 23: @pankajkoti we also should expand environment variables, similar to:
```
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/conftest.py#L48
```
If the users don't want to hard-code the password and other sensitive data inside the YAML file itself.

It would be great if you could also have a test for this.

sql-cli/sql_cli/settings.py
pankajkoti: okay Tatiana, will appreciate your help on this in the init PR.

sql-cli/sql_cli/__main__.py
pankajkoti: The former did not work because it does not give a logger instance. The latter works. Thanks, nice suggestion .

sql-cli/sql_cli/connections.py
pankajkoti: Renamed it to connection_id for internal functions.

sql-cli/sql_cli/connections.py
pankajkoti: Thanks, this was needed. Using a redshift connection now with secrets loaded for environment variable. Altered the test to check that Redshift connections passes locally with .env file environment vars and in CI using CI yaml induced env variables for the Redshift in the Tests job.

sql-cli/tests/test___main__.py
```diff
@@ -43,3 +45,9 @@ def test_generate(root_directory, target_directory, dags_directory):
     result_stdout = get_stdout(result)
     assert result_stdout.startswith("The DAG file ")
     assert result_stdout.endswith(" has been successfully generated. 🎉")
+
+
+def test_validate():
+    os.system("airflow db init")
```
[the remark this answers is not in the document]
pankajkoti: @tatiana irrespective of the directory we're calling `airflow db init`, I am observing that it creates `airflow.db` in the directory set in the environment variable AIRFLOW_HOME. 
Also, I believe calling `airflow db init` multiple times should not be a problem. It's like a no-operation, isn't it?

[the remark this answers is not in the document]
pankajkoti: Just wondering and trying to figure out how're initialising the Airflow DB for `python-sdk`.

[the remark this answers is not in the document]
pankajkoti: I see `python-sdk` is initialising Airflow DB in the nox test session. Will try to do the same here.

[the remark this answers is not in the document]
pankajkoti: It works!

sql-cli/noxfile.py
```diff
@@ -23,16 +23,27 @@ def dev(session: nox.Session) -> None:
...
 @nox.session(python=["3.7", "3.8", "3.9"])
-def test(session: nox.Session) -> None:
+@nox.parametrize("airflow", ["2.2.5", "2.4"])
+def test(session: nox.Session, airflow: str) -> None:
     """Run both unit and integration tests."""
     session.install("-e", "../python-sdk/.[all]")
     session.install("-e", "../python-sdk/.[tests]")
     session.install("poetry")
-    session.run("poetry", "install")
+    session.run("poetry", "install", "--with", "apache_airflow_providers_amazon")
```
feluelle, line 32: Why only this provider? We should add all here.

Btw I think this is confusing that we have nox. Nox is currently not in our dependencies of sql-cli. We only use it in CI.

sql-cli/pyproject.toml
```diff
@@ -26,11 +26,51 @@ Home = "https://astronomer.io/"
...
+[tool.poetry.group.apache_airflow_providers_amazon.dependencies]
+apache-airflow-providers-amazon = "*"
+s3fs = "*"
+smart-open = {extras = ["s3"], version = ">=5.2.1"}
+
+[tool.poetry.extras]
+google = ["protobuf", "apache-airflow-providers-google", "sqlalchemy-bigquery", "smart-open[gcs]"]
+snowflake = ["apache-airflow-providers-snowflake","snowflake-sqlalchemy", "snowflake-connector-python[pandas]"]
+postgres = ["apache-airflow-providers-postgres"]
+amazon = ["apache-airflow-providers-amazon", "s3fs", "smart-open[s3]"]
```
feluelle, line 73: This will be replaced by `astro-sdk[all]` at some point or? Do we have explicit need for these requirements in the connection validation?

sql-cli/sql_cli/__main__.py
```diff
@@ -38,5 +46,10 @@ def generate(
     rprint("The DAG file", dag_file.resolve(), "has been successfully generated. 🎉")
 
 
+@app.command()
+def validate(environment: str = "default", connection: Optional[str] = None) -> None:
```
feluelle, line 50: Could you add here a bit of information in the docstring, please?

sql-cli/tests/test_connections.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+
+from sql_cli.connections import _load_yaml_connections, validate_connections
+
+
+def test_validate_connections(caplog):
+    validate_connections()
+
+    postgres_conn_id = "postgres_conn"
+    postgres_conn_id_formatted_string = postgres_conn_id + " " * (25 - len(postgres_conn_id))
```
feluelle, line 10: This looks a bit strange tbh. Why does the length of the conn_id matter? Where does the magical `25` come from? :D

.github/workflows/ci-sql-cli.yaml
```diff
@@ -26,6 +26,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+# This list should only have non-sensitive environment variables.
+# Environment variables with secrets should be in the specific jobs.
+env:
```
tatiana, line 31: Do we need to define these environment variables in the context of all tests - or are they only required for a specific test related to rendering `REDSHIFT` a connection file which has environment variables? If the latter, we can consider patching them only within the test of interest, using something like:
```
@mock.patch.dict(os.environ, {"REDSHIFT_DATABASE": "fake.redshift.amazonaws.com", ...}, clear=True)
```

sql-cli/noxfile.py
tatiana: :+1: it would be great if we could centralize the dependencies management using Poetry

sql-cli/tests/config/default/configuration.yaml
```diff
@@ -0,0 +1,48 @@
+connections:
```
tatiana, line 1: Do we need such a long file? Could we have a simpler one?

sql-cli/tests/test_connections.py
```diff
@@ -0,0 +1,50 @@
...
+    failed_connection_log = f"Validating connection {postgres_conn_id_formatted_string} FAILED"
+    assert failed_connection_log in caplog.text
+
+    sqlite_conn_id = "sqlite_conn"
+    sqlite_conn_id_formatted_string = sqlite_conn_id + " " * (25 - len(sqlite_conn_id))
+    added_connection_log = f"Validating connection {sqlite_conn_id_formatted_string} PASSED"
+    assert added_connection_log in caplog.text
+
+
+def test_specific_connection_with_env_vars_in_config(caplog):
```
tatiana, line 20: Is this an integration test? If yes, could we mock the communication with Redshift and change it to a unittest?

sql-cli/pyproject.toml
```diff
@@ -26,11 +26,14 @@ Home = "https://astronomer.io/"
 Source = "https://github.com/astronomer/astro-sdk/sql-cli"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<4.0.0"
+python = ">=3.8,<3.11"
```
pankajkoti, line 29:
```suggestion
python = ">=3.8,<4.0.0"
```

.github/workflows/ci-sql-cli.yaml
```diff
@@ -92,7 +92,6 @@ jobs:
           name: coverage
           path: ./sql-cli/.coverage
```
pankajkoti, line 94:
```suggestion


```

sql-cli/noxfile.py
pankajkoti: Removed this, not needed anymore since we're mocking the connection with Redshift.

sql-cli/pyproject.toml
pankajkoti: Removed all of it.

sql-cli/tests/test_connections.py
pankajkoti: So, for a prettier output I'm setting the connection ID string width to 25. So that the results appears in a columnar fashion. I have moved this to a constant now and reusing the constant here to fill up spaced for connection IDs having smaller lengths.

.github/workflows/ci-sql-cli.yaml
pankajkoti: Right. Nice suggestion, thanks. Removed it from here.

sql-cli/tests/config/default/configuration.yaml
pankajkoti: Trimmed the file to have only 2 connections (including one with vars)

sql-cli/tests/test_connections.py
pankajkoti: Refactored the tests. Communications are mocked now.

tatiana (approved):

Issue #1008, opened 2022-10-06
Implement the flow generate command


tatiana:
**Please describe the feature you'd like to see**

Create the command `flow generate` which can be used to materialise a folder of SQL files as an Airflow DAG.

**Describe the solution you'd like**

Users should be able to run:
```
flow generate some_workflow ./dags
```

And the command will convert the SQL workflow, defined within `some_workflow` folder, into a dag `some_workflow.py` inside the `./dags` directory, using the existing DAG generation logic.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1013, opened 2022-10-06
Fix cyclic dependency and improve import time

kaxil:
This PR/commit improves the import time and removes cyclic dependency.

Reduces the import time for `astro/databases/__init__.py` from **23.254** seconds to **0.062** seconds

**Before**:

```
Thu Oct  6 23:12:29 2022    /Users/kaxilnaik/Documents/GitHub/astronomer/astro-sdk/python-sdk/dev/dags/program.prof

         3726676 function calls (3646580 primitive calls) in 23.254 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   4194/1    0.399    0.000   23.281   23.281 {built-in method builtins.exec}
3899/1    0.109    0.000   23.280   23.280 <frozen importlib._bootstrap>:1002([_find_and_load](http://127.0.0.1:4000/?func_name=_find_and_load))
   3849/1    0.074    0.000   23.280   23.280 <frozen importlib._bootstrap>:967([_find_and_load_unlocked](http://127.0.0.1:4000/?func_name=_find_and_load_unlocked))
   4500/2    0.023    0.000   23.276   11.638 <frozen importlib._bootstrap>:220([_call_with_frames_removed](http://127.0.0.1:4000/?func_name=_call_with_frames_removed))
   3672/2    0.084    0.000   23.274   11.637 <frozen importlib._bootstrap>:659([_load_unlocked](http://127.0.0.1:4000/?func_name=_load_unlocked))
   3522/2    0.042    0.000   23.273   11.637 <frozen importlib._bootstrap_external>:844([exec_module](http://127.0.0.1:4000/?func_name=exec_module))
        1    0.000    0.000   23.271   23.271 /usr/local/airflow/astro_sdk/src/astro/databases/__init__.py:1([<module>](http://127.0.0.1:4000/?func_name=%3Cmodule%3E))
        1    0.000    0.000   23.224   23.224 /usr/local/airflow/astro_sdk/src/astro/databases/base.py:1([<module>](http://127.0.0.1:4000/?func_name=%3Cmodule%3E))
  1079/49    0.018    0.000   21.545    0.440 {built-in method builtins.__import__}
        1    0.000    0.000   17.856   17.856 /usr/local/lib/python3.9/site-packages/airflow/__init__.py:18([<module>](http://127.0.0.1:4000/?func_name=%3Cmodule%3E))
[...]

PR #1032, opened 2022-10-10
Add typer app command for generate

feluelle:
# Description

Currently, we don't have a CLI command for generating a DAG - only through the `generate_dag.py` script.

closes: #1008

## What is the new behavior?

We are adding a typer command for generating a DAG.

- refactor Makefile
- add missing deps
- add airflow and astro-sdk to optional deps; we install them via makefile and in dev/test/type_check nox sessions
- return dag file path in generate_dag
- use rich for printing
- improve tests
- remove unused project init and cli entrypoint
- add pytest-freezegun as dev deps
- set static start_date
- fix templates

No.

### Checklist


sql-cli/noxfile.py
```diff
@@ -16,19 +16,13 @@ def dev(session: nox.Session) -> None:
     development environment to ``.nox/dev``.
     """
     session.install("nox")
-    session.install("-e", "../python-sdk/.[all]")
-    session.install("-e", "../python-sdk/.[tests]")
-    session.install("-e", ".[all]")
-    session.install("-e", ".[tests]")
+    session.install("-e", ".[all,tests]")
```
feluelle, line 19: The `python-sdk` will be installed through `all`.

sql-cli/pyproject.toml
```diff
@@ -34,6 +36,10 @@ tests = [
     "pytest-cov",
     "mypy"
 ]
+all = [
+    "apache-airflow>=2.0",
+    "astro-sdk-python[all]>=1.0.0",
+]
```
feluelle, line 42: I have added these two as optional dependencies. For example, they are already useful to validate the syntax of the generated DAG.

I would suggest we make them required dependencies when we have `flow run`.

sql-cli/sql_cli/__main__.py
```diff
@@ -1,24 +1,41 @@
...
+app = Typer(add_completion=False)
 
 
 @app.command()
 def version() -> None:
     """
     Print the SQL CLI version.
     """
-    typer.echo(f"Astro SQL CLI {sql_cli.__version__}")
+    rprint("Astro SQL CLI", __version__)
```
feluelle, line 17: using rich is [recommended](https://typer.tiangolo.com/tutorial/printing/).

sql-cli/noxfile.py
tatiana: The advantage of the previous way is that we'd be testing the latest version of the Python SDK, as opposed to the last released version. Perhaps we should have both, WDYT?

feluelle: Not sure if this is an advantage tbh. We should, in my opinion, treat `python-sdk` as any other package. We should only test against released stable versions.

feluelle: In my opinion testing against non-stable versions does not make sense as they could quickly change.

tatiana: So if the `python-sdk`  introduces a breaking change, the SQL CLI would only know after it is released?

feluelle: If we use proper semver in python-sdk, this is not a problem in my opinion. We could restrict the version also the same we do for other dependencies.

tatiana (approved):

sql-cli/Makefile
```diff
@@ -32,21 +31,14 @@ virtualenv:  ## Create Python virtualenv
...
 install: virtualenv  ## Install python dependencies in existing virtualenv
 	@echo "Installing Python dependencies using $(PIP)"
 	@$(PIP) install --upgrade pip
 	@$(PIP) install nox
 	@$(PIP) install pre-commit
-	@$(PIP) install -e .[all]
-	@$(PIP) install .[tests]
+	@$(PIP) install -e .[all,tests]
 
-setup: system install  ## Setup a local development environment
```
dimberman, line 49: Is this relevent to the current PR or should it be a different PR?

sql-cli/examples/_dags/advanced.py
```diff
@@ -2,11 +2,11 @@
 from airflow import DAG
 from airflow.utils import timezone
 from astro import sql as aql
-from astro.table import Table
+from astro.sql.table import Table
 
 with DAG(
     dag_id="advanced",
-    start_date=timezone.parse("2022-10-05 06:47:15.790544"),
+    start_date=timezone.parse("2022-10-10 09:11:57.308687"),
```
dimberman, line 9: Why this change?

sql-cli/examples/_dags/simple.py
```diff
@@ -2,11 +2,11 @@
 from airflow import DAG
 from airflow.utils import timezone
 from astro import sql as aql
-from astro.table import Table
+from astro.sql.table import Table
```
dimberman, line 5: Why this change?

sql-cli/noxfile.py
dimberman: I'm a -1 on this. There will be situations where we are both changing the SQL SDK and the python SDK and we will need to test the changes together.

dimberman (changes requested):

sql-cli/noxfile.py
feluelle: I agree that this may need its own PR.

sql-cli/examples/_dags/advanced.py
feluelle: This is the time of the last auto-generated example. I ran it because of the change you see below. Let's continue the discussion in https://github.com/astronomer/astro-sdk/pull/1032#discussion_r991511811.

sql-cli/examples/_dags/simple.py
feluelle: Because, in my opinion we should use the released python-sdk version - not the current one. Otherwise it will become very confusing for users if they see the example dag of an unreleased version. The import path change was introduced in #1013

PR #1033, opened 2022-10-10
Use poetry to improve the local SQL CLI development

tatiana:
# Description
We manage local development using pip and our own managed virtualenv.

## What is the new behaviour?
Now, by default, we use poetry to add/remove dependencies, test and release the SQL CLI.

No

### Checklist

### DIFF
--- a/sql-cli/pyproject.toml
+++ b/sql-cli/pyproject.toml
@@ -29,7 +29,7 @@ Source = "https://github.com/astronomer/astro-sdk/sql-cli"
 python = ">=3.8,<4.0.0"
 typer = {extras = ["all"], version = "^0.6.1"}
 networkx = "^2.8.7"
-Jinja2 = "^3.1.2"
+Jinja2 = ">=2.10.1, <3.1"
 python-frontmatter = "^1.0.0"
 
 [tool.poetry.scripts]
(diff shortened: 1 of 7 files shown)


sql-cli/pyproject.toml
```diff
@@ -20,20 +20,29 @@ classifiers = [
...
-tests = [
-    "pytest",
-    "pytest-cov",
-    "mypy"
-]
+[tool.poetry.dependencies]
+python = ">=3.8,<4.0.0"
+typer = {extras = ["all"], version = "^0.6.1"}
+networkx = "^2.8.7"
+Jinja2 = "^3.1.2"
```
feluelle, line 32: This version won't work with Airflow. For example, airflow 2.2.5 uses `jinja2>=2.10.1,<3.1` 

https://github.com/astronomer/airflow/blob/2.2.5/setup.cfg#L131

airflow 2.0.0 has even `jinja2>=2.10.1, <2.12.0`

https://github.com/astronomer/airflow/blob/2.0.0/setup.cfg#L111

feluelle: We support airflow >2.0.0 right? So we should set it in my opinion to `jinja2>=2.10.1`

tatiana: It may be worth to have an upper bound similar to Airflow, so we don't start using versions which will cause dependency conflicts to end-users

Issue #942, opened 2022-09-28
 flow run: Run a SQL workflow


tatiana:
**Please describe the feature you'd like to see**

Using the previously CLI code structure using click, implement the init command to run a SQL workflow.
```
flow run <workflow_name>
```

It should output:
```
Running the workflow example_imdb…

Processing popular_movies.sql to output_db.popular_movies SUCCESS 
Processing top_directors.sql to output_db.top_directors SUCCESS 
Completed successfully running the pipeline example_imdb

Summary
–--------------|-------------|---------------|-----------------|---------|
Model          | Total rows  | Total columns | Processing time | Status  |
–--------------|-------------|---------------|-----------------|---------|
example_imdb   | 1,000,000   |          20   |            -    |         |
popular_movies |       100   |           6   |            1s   | Success |
top_directors  |        10   |           3   |            1s   | Success |
–--------------|-------------|---------------|-----------------|---------|

Total time:        3.218s

====================== 2 models processed in 3s ========================

```

**Dependencies**
* Depends on #828 
* Depends on #932

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more)
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)

PR #1037, opened 2022-10-11
Add run command to SQL CLI

dimberman:
# Description
Currently there is no way for SQL CLI users to run a DAG locally. They would need to either deploy their DAG to an airflow using `astro dev start` (which can take multiple minutes and involves an entire airflow UI) or use `airflow dags test` which only exists in airflow 2.5+. 

closes: #942

## What is the new behavior?
Now sql-cli users can run `flow run <dag id> <subdir>` to locally run their DAGs. they can also include connections files and variables files to override their airflow settings.

The help documentation looks like this:
Screen Shot 2022-10-12 at 9 14 34 AM

No

### Checklist


sql-cli/sql_cli/__main__.py
```diff
@@ -38,5 +42,50 @@ def generate(
...
+    ),
+    conn_file_path: Optional[str] = typer.Argument(
+        help="path to connections yaml or json file", default=None
+    ),
+    variable_file_path: Optional[str] = typer.Argument(
+        help="path to variables yaml or json file", default=None
+    ),
+) -> None:
+    """
+    Run a DAG locally. This task assumes that there is a local airflow DB (can be a SQLite file), that has been
```
tatiana, line 62: I think we should be avoiding the word DAG where possible in the SQL CLI. If we could focus on workflow, it would be great.

It's also important that the run command is compatible to receive the name of a SQL directory, within `workflows`, and both generates & runs it

feluelle: 🤔 but all DAGs are workflows - not all workflows are DAGs - and we only allow workflows which are DAGs to run. So I think DAG is more precisely what we run.

I think we should have in general discussion about terms/names we use in sql-cli. I also thought about renaming some classes and functions to flow/workflow instead of dag and then only the dag_generator is allowed to be called **dag**_generator.

In my opinion this is something we should clean up / make clear before 1.0 definitely.

sql-cli/tests/test_run_dag.py
```diff
@@ -0,0 +1,89 @@
...
+    run_dag(sample_dag)
+    assert "witch_wins ran successfully!" not in caplog.text
+    assert "snow_white_wins ran successfully!" in caplog.text
+    assert "movie_ends ran successfully!" in caplog.text
+
+
+def test_run_dag(caplog):
+    dag = get_dag(dag_id="example_dataframe", subdir=f"{CWD}/test_dag")
+    run_dag(dag)
+    assert "The worst month was 2020-05" in caplog.text
```
tatiana, line 83: I can see the value of these tests from a Python SDK perspective. @dimberman how meaningful and representative do ou find them from a SQL CLI perspective?

Would it make sense to run DAGs auto-generated by the SQL CLI as part of the run?

sql-cli/sql_cli/run_dag.py
```diff
@@ -0,0 +1,161 @@
+from __future__ import annotations
```
tatiana, line 1: ATM, we have a module named `dag_generator.py`. What do you think about renaming it to `dag.py` and adding all the dag-related logic to it?

sql-cli/sql_cli/__main__.py
feluelle: Ah sorry that is not what I meant. It was only about naming.

feluelle: We can only run DAGs right now, because Airflow only runs DAGs. A cyclic dependency for example is not allowed.

sunkickr:
Where can I see an example the variable and connections yaml files?

sql-cli/tests/test___main__.py
```diff
@@ -53,27 +53,15 @@ def test_validate():
...
 
 
-def test_run_dag(root_directory_dags):
-    result = runner.invoke(
-        app,
-        ["run", "example_dataframe", root_directory_dags.as_posix(), "--no-is-sql"],
-    )
-    assert result.exit_code == 0
-    result_stdout = get_stdout(result)
-    assert "The worst month was 2020-05" in result_stdout
```
feluelle: @dimberman The goal for sql-cli was to run on a directory of sql files, hence we don't pass a dag_id. We instead generate the dag first and then get the dag id internally from the generated file which is why I removed it this test. We have another test below for testing the run statement.

@tatiana is this correct?

sql-cli/tests/test_run_dag.py
```diff
@@ -0,0 +1,89 @@
...
+    @task
+    def witch_wins():
+        return "the witch is happy"
+
+    @task(trigger_rule=TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS)
+    def movie_ends():
+        return "movie ends"
+
+    with sample_dag:
+        who_is_prettiest() >> [snow_white_wins(), witch_wins()] >> movie_ends()
```
feluelle, line 73:
```suggestion
        who_is_prettiest() >> [snow_white_wins(), witch_wins()] >> movie_ends()  # skipcq: PYL-W0106
```

sql-cli/sql_cli/run_dag.py
```diff
@@ -0,0 +1,167 @@
...
+    :param ti: TaskInstance to run
+    """
+    pprint("[bold green]*****************************************************[/bold green]")
+    if hasattr(ti, "map_index") and ti.map_index > 0:
+        pprint("Running task %s index %d", ti.task_id, ti.map_index)
+    else:
+        pprint(f"Running task [bold red]{ti.task_id}[/bold red]")
+    try:
+        warnings.filterwarnings(action="ignore")
+        ti._run_raw_task(session=session)
```
feluelle, line 121:
```suggestion
        ti._run_raw_task(session=session)  # skipcq: PYL-W0212
```

sql-cli/tests/test_dag_generator.py
```diff
@@ -19,3 +19,18 @@ def test_generate_dag(root_directory, dags_directory):
...
+    import tempfile
+    from pathlib import Path
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        shutil.copytree(root_directory, tmp_dir + "/root")
+        os.mkdir(Path(tmp_dir) / "target")
+        os.mkdir(Path(tmp_dir) / "dag")
+        print(tmp_dir)
+        generate_dag(Path(tmp_dir) / "root", dags_directory=Path(tmp_dir) / "dag")
+        print(tmp_dir)
```
feluelle, line 36:
```suggestion
```

feluelle:
@dimberman @tatiana any idea how to fix the tests?
```
FAILED tests/test_run_dag.py::test_run_task_cleanup_log - AssertionError: assert 'aql.cleanup async, continuing' in ''
FAILED tests/test_run_dag.py::test_run_dag_dynamic_task - AssertionError: assert 'Running task print_val index 1' in ''
FAILED tests/test_run_dag.py::test_run_dag_with_skip - AssertionError: assert 'snow_white_wins ran successfully!' in ''
```

My guess is that this is due to:
```python
for name in logging.root.manager.loggerDict:
    logging.getLogger(name).setLevel(logging.ERROR)
```

sql-cli/tests/test___main__.py
dimberman: @feluelle testing this mode helps ensure that nothing breaks in the run dag regarding DAG running. Only having integration tests for the generate + run flow will make it more difficult to debug in the future.

feluelle (approved):

PR #1038, opened 2022-10-11
Open lineage append op

pankajastro:
# Description
Added methods to collect the lineage for the Append operator.
Expose `get_openlineage_facets` on the operator which will be used by PythonExtractor once task execution is complete. Exposed `openlineage_dataset_name` to get dataset qualified name and `openlineage_dataset_namespace` to get dataset namespace on BaseTable object. Since Table is an input/output for astro-sdk operator so I feel this is the right place to expose it. The BaseTable `openlineage_dataset_name` and `openlineage_dataset_namespace` depend on table type construct the right name and namespace of dataset by calling astro-sdk databases class `openlineage_dataset_name` and `openlineage_dataset_namespace` method at runtime.

We do not collect lineage for the Append operator

closes: #899 

## What is the new behavior?
Add lineage support for the Append operator

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/aws/redshift.py
+++ b/python-sdk/src/astro/databases/aws/redshift.py
@@ -384,3 +384,11 @@ class RedshiftDatabase(BaseDatabase):
         for Redshift.
         """
         return "SELECT 1 + 1"
+
+    def openlineage_dataset_name(self, table: BaseTable = None):
+        conn = self.hook.get_connection(self.conn_id)
+        return f"{conn.schema}.{table.name}"
+
+    def openlineage_dataset_namespace(self):
+        conn = self.hook.conn
+        return f"{self.sql_type}://{conn.host}:{conn.port}"
--- a/python-sdk/src/astro/databases/aws/redshift.py
+++ b/python-sdk/src/astro/databases/aws/redshift.py
@@ -385,10 +385,20 @@ class RedshiftDatabase(BaseDatabase):
         """
         return "SELECT 1 + 1"
 
-    def openlineage_dataset_name(self, table: BaseTable = None):
+    def openlineage_dataset_name(self, table: BaseTable = None) -> str:
+        """
+        Returns the open lineage dataset name as per
+        https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
+        Example: schema_name.table_name
+        """
         conn = self.hook.get_connection(self.conn_id)
         return f"{conn.schema}.{table.name}"
 
-    def openlineage_dataset_namespace(self):
+    def openlineage_dataset_namespace(self) -> str:
+        """
+        Returns the open lineage dataset namespace as per
+        https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
+        Example: redshift://cluster:5439
+        """
         conn = self.hook.conn
         return f"{self.sql_type}://{conn.host}:{conn.port}"
--- a/python-sdk/src/astro/databases/aws/redshift.py
+++ b/python-sdk/src/astro/databases/aws/redshift.py
@@ -385,7 +385,7 @@ class RedshiftDatabase(BaseDatabase):
         """
         return "SELECT 1 + 1"
 
-    def openlineage_dataset_name(self, table: BaseTable = None) -> str:
+    def openlineage_dataset_name(self, table: BaseTable) -> str:
         """
         Returns the open lineage dataset name as per
         https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import warnings
 from abc import ABC
-from typing import Any, Callable, Mapping
+from typing import Any, Callable, Mapping, Union
 
 import pandas as pd
 import sqlalchemy
@@ -63,7 +63,7 @@ class BaseDatabase(ABC):
 
     def __init__(self, conn_id: str):
         self.conn_id = conn_id
-        self.sql = None
+        self.sql: Union[str, ClauseElement] = ""
 
     def __repr__(self):
         return f'{self.__class__.__name__}(conn_id="{self.conn_id})'
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -63,7 +63,7 @@ class BaseDatabase(ABC):
 
     def __init__(self, conn_id: str):
         self.conn_id = conn_id
-        self.sql = ""
+        self.sql = None
 
     def __repr__(self):
         return f'{self.__class__.__name__}(conn_id="{self.conn_id})'
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import warnings
 from abc import ABC
-from typing import Any, Callable, Mapping, Union
+from typing import Any, Callable, Mapping
 
 import pandas as pd
 import sqlalchemy
@@ -63,7 +63,7 @@ class BaseDatabase(ABC):
 
     def __init__(self, conn_id: str):
         self.conn_id = conn_id
-        self.sql: Union[str, ClauseElement] = ""
+        self.sql: str | ClauseElement = ""
 
     def __repr__(self):
         return f'{self.__class__.__name__}(conn_id="{self.conn_id})'
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -789,3 +789,9 @@ class BaseDatabase(ABC):
             source_file.location.location_type in self.FILE_PATTERN_BASED_AUTODETECT_SCHEMA_SUPPORTED
         )
         return is_file_pattern_based_schema_autodetection_supported
+
+    def openlineage_dataset_name(self, table: BaseTable):
+        raise NotImplementedError
+
+    def openlineage_dataset_namespace(self):
+        raise NotImplementedError
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -183,21 +183,6 @@ class BaseDatabase(ABC):
         """
         raise NotImplementedError
 
-    def openlineage_dataset_namespace(self) -> str:
-        """
-        Returns the open lineage dataset namespace as per
-        https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
-        """
-        airflow_conn = self.hook.get_connection(self.conn_id)
-        return f"{self.sql_type}://{airflow_conn.host}"
-
-    def openlineage_dataset_name(self, table: BaseTable) -> str:
-        """
-        Returns the open lineage dataset name as per
-        https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
-        """
-        return self.get_table_qualified_name(table)
-
     def populate_table_metadata(self, table: BaseTable) -> BaseTable:
         """
         Given a table, check if the table has metadata.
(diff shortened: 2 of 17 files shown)


python-sdk/src/astro/databases/base.py
```diff
@@ -774,3 +774,9 @@ def check_file_pattern_based_schema_autodetection_is_supported(self, source_file
             source_file.location.location_type in self.FILE_PATTERN_BASED_AUTODETECT_SCHEMA_SUPPORTED
         )
         return is_file_pattern_based_schema_autodetection_supported
+
+    def openlineage_dataset_name(self, table: BaseTable):
+        raise NotImplementedError
```
pankajastro, line 779: added `table` param to get the dataset and table name in case of bigquery. Airflow biquery hook method `hook.get_uri()` return string which has only project id and scheme f"bigquery://{self.project_id}".

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +62,43 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
         )
+
         return self.target_table
 
+    def get_openlineage_facets(self) -> OpenLineageFacets:
+        input_dataset: list[OpenlineageDataset] = [
+            OpenlineageDataset(
+                namespace=self.source_table.openlineage_dataset_namespace(),
+                name=self.source_table.openlineage_dataset_name(),
+                facets={
```
sunank200, line 73: @pankajastro This has to be an object from class Basefacet. You might need to add a custom facet class like https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/load_file.py#L205

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +62,43 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
+                    "metadata": self.target_table.metadata,
+                },
+            )
+        ]
+
+        output_dataset: list[OpenlineageDataset] = [
+            OpenlineageDataset(
+                namespace=self.target_table.openlineage_dataset_namespace(),
+                name=self.target_table.openlineage_dataset_name(),
+                facets={
```
sunank200, line 86: @pankajastro This has to be an object from class Basefacet. You might need to add a custom facet class like https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/load_file.py#L205

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +62,43 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
+                    "table_name": self.target_table.name,
+                    "row_affected": 0,  # FixMe
+                    "columns": self.columns,
+                    "metadata": self.target_table.metadata,
+                },
+            )
+        ]
+
+        run_facets: dict[str, BaseFacet] = {}
```
sunank200, line 95: Do you think this will always be empty in append operator. Job facet can contain sql as mentioned in https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#run-facets

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +62,43 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
         return self.target_table
 
+    def get_openlineage_facets(self) -> OpenLineageFacets:
+        input_dataset: list[OpenlineageDataset] = [
+            OpenlineageDataset(
+                namespace=self.source_table.openlineage_dataset_namespace(),
+                name=self.source_table.openlineage_dataset_name(),
+                facets={
+                    "table_name": self.source_table.name,
+                    "row_affected": 0,  # FixMe
```
sunank200, line 75: Is this always 0?

python-sdk/tests/sql/test_table.py
```diff
@@ -151,3 +152,76 @@ def test_if_table_object_can_be_pickled():
     """Verify if we can pickle Table object"""
     table = Table()
     assert pickle.loads(pickle.dumps(table)) == table
+
+
+@pytest.mark.parametrize(
```
sunank200, line 157: I think it would be good to add an integration test as well to see if extractor is getting called when operator is called

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -429,6 +429,13 @@ def load_local_file_to_table(
             )
         job.result()
 
+    def openlineage_dataset_name(self, table: BaseTable = None):
+        dataset = table.metadata.database or table.metadata.schema
+        return f"{self.hook.project_id}.{dataset}.{table.name}"
```
sunank200, line 439: Isn't get qualified table giving you the same thing?

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -384,3 +384,11 @@ def get_merge_initialization_query(parameters: tuple) -> str:
         for Redshift.
         """
         return "SELECT 1 + 1"
+
+    def openlineage_dataset_name(self, table: BaseTable = None):
+        conn = self.hook.get_connection(self.conn_id)
+        return f"{conn.schema}.{table.name}"
```
sunank200, line 395: wouldn't get qualified table method give you the same thing?

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +61,36 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
+
         return self.target_table
 
+    def get_openlineage_facets(self):
+        input_dataset = [
+            OpenlineageDataset(
+                namespace=self.source_table.openlineage_dataset_namespace(),
+                name=self.source_table.openlineage_dataset_name(),
+                facets={
+                    "table_name": self.source_table.name,
```
sunank200, line 74: This should be an object of BaseFacet class like in https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/load_file.py#L205

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +61,36 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
+                    "row_affected": 0,  # FixMe
```
sunank200, line 75: Is this always 0?

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -59,8 +61,36 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
...
+                    "columns": self.columns,
+                },
+            )
+        ]
+
+        output_dataset = [
+            OpenlineageDataset(
+                namespace=self.target_table.openlineage_dataset_namespace(),
+                name=self.target_table.openlineage_dataset_name(),
+                facets={
```
sunank200, line 86: This should be object of BaseFacet class like in https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/load_file.py#L227

python-sdk/tests/sql/test_table.py
```diff
@@ -151,3 +152,74 @@ def test_if_table_object_can_be_pickled():
     """Verify if we can pickle Table object"""
     table = Table()
     assert pickle.loads(pickle.dumps(table)) == table
+
```
sunank200, line 155: We should have a unit or integration test to check the if extractor is called when operator is called

sunank200 (changes requested):
Added few comments. As per open lineage documentation - facet should be always a dict which contains object of `BaseFacet` class.

python-sdk/tests/sql/test_table.py
sunank200: For example: https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/extractors/test_extractor.py#L59

python-sdk/src/astro/sql/operators/append.py
pankajastro: any thought on how we can access sql stmt it get form in base class https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/base.py#L594 ?

python-sdk/src/astro/sql/operators/append.py
pankajastro: Added let me know how it looks now

python-sdk/src/astro/sql/operators/append.py
pankajastro: look like our Table object does not have the row count data. do you think it make sense to expose the `get_row_count` on Table object for some database query like `select count(*) from xyz` time taking query

python-sdk/tests/sql/test_table.py
pankajastro: added, please let me know how it is looking now

python-sdk/src/astro/databases/aws/redshift.py
pankajastro: qualified table method, consider schema name Table object schema field only so if I have schema in connection not Table then it won't be same

python-sdk/src/astro/sql/operators/append.py
sunank200: Yes I was thinking something like following in BaseTable class:
```
    def get_rows(self):
        db_connection = create_engine(self.conn_id)
        db_metadata = MetaData(bind=db_connection, reflect=True)
        sqlalchemy_table = db_metadata.tables[self.name]
        with Session(db_connection) as session:
            return session.query(sqlalchemy_table).count()
```

python-sdk/src/astro/databases/aws/redshift.py
sunank200: Sure. Makes sense

python-sdk/src/astro/sql/operators/append.py
sunank200: LGTM

python-sdk/src/astro/sql/operators/append.py
sunank200: You could add `sql` to a class attribute and use it as part of `self.<attribute-name>`

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -220,11 +220,10 @@ def get_openlineage_facets(self) -> OpenLineageFacets:
 
         output_dataset: list[OpenlineageDataset] = [OpenlineageDataset(namespace=None, name=None, facets={})]
         if self.output_table:
-            output_database = create_database(self.output_table.conn_id)
             output_dataset = [
                 OpenlineageDataset(
-                    namespace=output_database.openlineage_dataset_namespace(),
-                    name=output_database.openlineage_dataset_name(table=self.output_table),
+                    namespace=self.output_table.openlineage_dataset_namespace(),
```
sunank200, line 225: Does the test for load file need to change after this change?

sunank200 (commented):
Added few comments

python-sdk/src/astro/sql/operators/load_file.py
pankajastro: No

python-sdk/src/astro/sql/operators/append.py
pankajastro: As discussed offline we are using xcom for this

python-sdk/src/astro/sql/operators/append.py
pankajastro: expose row count on BaseTable

python-sdk/tests/sql/test_table.py
```diff
@@ -151,3 +152,76 @@ def test_if_table_object_can_be_pickled():
...
+        (
+            Connection(conn_id="test_conn", conn_type="sqlite", host="/tmp/sqlite.db"),
+            "/tmp/sqlite.db.test_tb",
+            "/tmp/sqlite.db",
+        ),
+    ],
+)
+@mock.patch("airflow.providers.google.cloud.utils.credentials_provider.get_credentials_and_project_id")
+@mock.patch("airflow.hooks.base.BaseHook.get_connection")
+def test_openlineage_dataset(mock_get_connection, gcp_cred, connection, name, namespace):
```
sunank200, line 221: We should add a doc string mentioning what we are testing here.

sunank200 (commented):
Overall it LGTM me now. I have added few comments. Also deepsource and pre-commit is throwing errors.

sunank200 (approved):
LGTM. Other test is failing because of flaky test on snowflake

pankajastro:
This requires for other PR so merging it. If there will any further suggestions will take that in a separate PR.

PR #1039, opened 2022-10-11
Set different AIRFLOW_HOME per test instance

feluelle:
# Description

Currently, running nox is always using the same (default) `AIRFLOW_HOME` which can result in errors when running multiple different nox test sessions as they use different airflow versions which have different database tables for example.

## What is the new behavior?

We set the AIRFLOW_HOME per test session - means per airflow and python version.

No.

### Checklist

### DIFF
--- a/python-sdk/noxfile.py
+++ b/python-sdk/noxfile.py
@@ -69,6 +69,7 @@ def test_examples_by_dependency(session: nox.Session, extras):
     session.install("-e", ".[tests]")
     session.run("airflow", "db", "init")
 
+    # Since pytest is not installed in the nox session directly, we need to set `external=true`.
     session.run("pytest", "tests/test_example_dags.py", *pytest_args, *session.posargs, external=True)
 
 
--- a/python-sdk/noxfile.py
+++ b/python-sdk/noxfile.py
@@ -38,6 +38,7 @@ def test(session: nox.Session, airflow) -> None:
     session.run("pip3", "freeze")
     AIRFLOW_HOME = f"~/airflow-{airflow}-{session.python}"
     session.run("airflow", "db", "init", env={"AIRFLOW_HOME": AIRFLOW_HOME})
+    # Since pytest is not installed in the nox session directly, we need to set `external=true`.
     session.run("pytest", *session.posargs, env={"AIRFLOW_HOME": AIRFLOW_HOME}, external=True)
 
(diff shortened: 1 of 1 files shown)


python-sdk/noxfile.py
```diff
@@ -68,7 +69,7 @@ def test_examples_by_dependency(session: nox.Session, extras):
     session.install("-e", ".[tests]")
     session.run("airflow", "db", "init")
 
-    session.run("pytest", "tests/test_example_dags.py", *pytest_args, *session.posargs)
+    session.run("pytest", "tests/test_example_dags.py", *pytest_args, *session.posargs, external=True)
```
feluelle, line 72: I added `external=True` because nox was complaining (with a warning) that this is not in the (I think) required dependencies. This fix simply suppresses the warning as suggested by the warning.

sunank200: Can we add this as a comment here, it will be useful.

sunank200 (approved):

python-sdk/noxfile.py
feluelle:
```suggestion
    # Since pytest is not installed in the nox session directly, we need to set `external=true`.
    session.run("pytest", "tests/test_example_dags.py", *pytest_args, *session.posargs, external=True)
```

tatiana (approved):

python-sdk/noxfile.py
```diff
@@ -36,8 +36,9 @@ def test(session: nox.Session, airflow) -> None:
     # Log all the installed dependencies
     session.log("Installed Dependencies:")
     session.run("pip3", "freeze")
-    session.run("airflow", "db", "init")
-    session.run("pytest", *session.posargs)
+    AIRFLOW_HOME = f"~/airflow-{airflow}-{session.python}"
+    session.run("airflow", "db", "init", env={"AIRFLOW_HOME": AIRFLOW_HOME})
+    session.run("pytest", *session.posargs, env={"AIRFLOW_HOME": AIRFLOW_HOME}, external=True)
```
feluelle, line 41:
```suggestion
    # Since pytest is not installed in the nox session directly, we need to set `external=true`.
    session.run("pytest", *session.posargs, env={"AIRFLOW_HOME": AIRFLOW_HOME}, external=True)
```

PR #1047, opened 2022-10-12
Remove pytest-freezegun dev deps

feluelle:
# Description

Currently, we have pytest-freezegun as a dev dependency even-though we don't need.

## What is the new behavior?

We remove pytest-freezegun from dev dependencies as we are using a static datetime now. We don't need it anymore.

- improve test for dag_generator
- fix get_sql_files is_file check
- refactor utils

No.

### Checklist


pankajkoti (approved):

sql-cli/sql_cli/utils.py
```diff
@@ -9,39 +9,44 @@
 from jinja2.runtime import StrictUndefined
```
tatiana, line 9: @feluelle, what if utils was a folder/package and within it, we had the `jinja.py` module? This would avoid `utils` becoming a wild place for other things

feluelle: Sure! Works for me. 👍

feluelle: Done in 71d91c9

tatiana (approved):
:rocket:

PR #1050, opened 2022-10-12
Add quotes on col names to preserve the casing

utkarsharma2:
# Description
Currently, we were changing the case of col names to lowercase for snowflakes. Which trips user. 

closes: #917 

## What is the new behavior?
We can keep the col names intact by passing them in quotes when creating tables added quotes. 
We no longer need the ColCapitalization feature. 

Yes

### Checklist


python-sdk/src/astro/databases/snowflake.py
```diff
@@ -480,30 +481,16 @@ def create_table_using_schema_autodetection(
         table: BaseTable,
         file: File | None = None,
         dataframe: pd.DataFrame | None = None,
-        columns_names_capitalization: ColumnCapitalization = "lower",
+        columns_names_capitalization: ColumnCapitalization = "original",
     ) -> None:
         """
-        Create a SQL table, automatically inferring the schema using the given file.
+        Dummy function. We are inferring schema and creating table in `load_pandas_dataframe_to_table`.
```
utkarsharma2, line 487:
```suggestion
        A dummy function is required to override the default behavior. We are inferring schema and creating a table in `load_pandas_dataframe_to_table`.
```

tatiana: Why are we defining this function at all? Perhaps we shouldn't define and just use the parent class implementation.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -579,17 +566,26 @@ def load_pandas_dataframe_to_table(
...
+
+        auto_create_table = False
+        if if_exists == "replace" or not self.table_exists(target_table):
+            auto_create_table = True
+
+        if target_table.metadata.schema:
+            target_table.metadata.schema = target_table.metadata.schema.upper()
+
+        if target_table.metadata.database:
+            target_table.metadata.database = target_table.metadata.database.upper()
```
tatiana, line 578: Are we happy changing the schema and database capitalization to upper?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -579,17 +566,26 @@ def load_pandas_dataframe_to_table(
...
 
         pandas_tools.write_pandas(
             conn=self.hook.get_conn(),
             df=source_dataframe,
-            table_name=target_table.name,
+            table_name=target_table.name.upper(),
```
tatiana, line 583: What are the implications of making the table name only uppercase here?

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: We want to override the default implementation. Since prep_table() is not allowing us to add quotes. we want to skip it and use a single method(pandas_tool.write_write_pandas()) to create and populate a snowflake table.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +762,49 @@ def _build_merge_sql(
...
+
+    @staticmethod
+    def get_merge_initialization_query(parameters: tuple) -> str:
+        """
+        Handles database-specific logic to handle constraints, keeping
+        it agnostic to database.
+        """
+        constraints = ",".join([f'"{p}"' for p in parameters])
+        sql = "ALTER TABLE {{table}} ADD CONSTRAINT airflow UNIQUE (%s)" % constraints
+        return sql
```
tatiana, line 806: Why are we introducing these methods as part of the capitalization PR? Is it because of the quoting? Doesn't the table identifier need quotes? It may be worth adding comments explaining why we have this and highlighting the underlining issue.

python-sdk/src/astro/databases/base.py
```diff
@@ -248,7 +248,7 @@ def create_table_using_schema_autodetection(
         table: BaseTable,
         file: File | None = None,
         dataframe: pd.DataFrame | None = None,
-        columns_names_capitalization: ColumnCapitalization = "lower",  # skipcq
+        columns_names_capitalization: ColumnCapitalization = "original",  # skipcq
```
tatiana, line 251: This is a breaking change. Will this be part of 1.2? Are we happy to introduce breaking changes as part of a minor release? Would it make sense to keep the old default, but support original - and encourage users to use the "original" capitalization?

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: 1. Yes, I want to keep these methods isolated from other databases.
2. No table identifiers already preserve the case.
3. Good idea, I'll add comments.

utkarsharma2: Added explanation in docstring.

python-sdk/src/astro/databases/base.py
utkarsharma2: @tatiana This was one of the requirements of - https://github.com/astronomer/astro-sdk/issues/917 to change the default to the original. 

cc: @kaxil

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: 1. This is the current behavior.
2.  Also, pandas_tools.write_pandas() is not able to create a table when a lowercase name is passed. However, it is possible in Snowflake to create a table in lowercase. Which will break if someone tries to point to an existing table which is in lowercase.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Was not needed, removed it.

python-sdk/src/astro/databases/base.py
dimberman: @tatiana I think we can consider the original functionality as a bug. It makes no sense to modify data by default.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -579,17 +567,20 @@ def load_pandas_dataframe_to_table(
...
+
+        auto_create_table = False
+        if if_exists == "replace" or not self.table_exists(target_table):
+            auto_create_table = True
 
         pandas_tools.write_pandas(
             conn=self.hook.get_conn(),
             df=source_dataframe,
-            table_name=target_table.name,
+            table_name=target_table.name.upper(),
```
dimberman, line 578: Why is this set to upper now?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -717,11 +708,11 @@ def _build_merge_sql(
         )
 
         merge_target_dict = {
-            f"merge_clause_target_{i}": f"{target_table_name}.{x}"
+            f"merge_clause_target_{i}": f'{target_table_name}."{x}"'
             for i, x in enumerate(target_conflict_columns)
         }
         merge_source_dict = {
-            f"merge_clause_source_{i}": f"{source_table_name}.{x}"
+            f"merge_clause_source_{i}": f'{source_table_name}."{x}"'
```
dimberman, line 715: Does quoting fully upper or fully lower values affect those values in snowflake? Like if I go to snowflake do I need to use quotes going forward?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +757,52 @@ def _build_merge_sql(
         }
         return statement, params
 
+    def append_table(
```
dimberman, line 760: Why are we adding this function?

dimberman (changes requested):

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Added a response here - https://github.com/astronomer/astro-sdk/pull/1050#discussion_r995468099

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: @dimberman You can run the query without quotes if we have the cols name in all caps. 
For example - 
If you have below table:
X, Y
1, 2
3,4

case 1:
SELECT x, y From <table_nma> 

case 2:
SELECT X, Y From <table_nma> 

case 3:
SELECT "X", "Y" From <table_nma> 

All of the listed queries will work. 

With the current approach, we are passing every col name in quotes. So that if someone has created cols in upper/lower/mixed cases it will work regardless.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: This is just isolation from other databases. Since adding quoted was failing for bigquery.

dimberman (approved):

python-sdk/src/astro/databases/base.py
```diff
@@ -115,7 +115,9 @@ def run_sql(
             sql = kwargs.get("sql_statement")  # type: ignore
 
         if isinstance(sql, str):
-            result = self.connection.execute(sqlalchemy.text(sql), parameters)
+            result = self.connection.execute(
+                sqlalchemy.text(sql).execution_options(autocommit=True), parameters
```
kaxil, line 119: why `.execution_options(autocommit=True)`?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -480,30 +482,36 @@ def create_table_using_schema_autodetection(
...
         """
         Create a SQL table, automatically inferring the schema using the given file.
+        Overriding default behaviour and not using the `prep_table` since it doesn't allow the adding quotes.
 
         :param table: The table to be created.
         :param file: File used to infer the new table columns.
         :param dataframe: Dataframe used to infer the new table columns if there is no file
         """
+        if file is None:
+            if dataframe is None:
```
kaxil, line 496:
```suggestion
        if file is None and dataframe is None:
```

python-sdk/src/astro/databases/snowflake.py
kaxil: Can you add code comments about this please?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -717,11 +728,11 @@ def _build_merge_sql(
         )
 
         merge_target_dict = {
-            f"merge_clause_target_{i}": f"{target_table_name}.{x}"
+            f"merge_clause_target_{i}": f'{target_table_name}."{x}"'
```
kaxil, line 731: Why this change?

python-sdk/src/astro/databases/snowflake.py
kaxil: but it shouldn't matter, the column names in snowflake are case in-sensitive, no?

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -196,14 +199,33 @@ def test_if_exist_param_of__load_pandas_dataframe_to_table(database_table_fixtur
...
 
-    with mock.patch("snowflake.connector.pandas_tools.write_pandas"):
-        with mock.patch("astro.databases.base.BaseDatabase.create_table") as method:
-            database.load_pandas_dataframe_to_table(pandas_dataframe, table, if_exists="replace")
-            method.assert_called_with(table, dataframe=pandas_dataframe)
+    with mock.patch("snowflake.connector.pandas_tools.write_pandas") as method:
+        database.load_pandas_dataframe_to_table(pandas_dataframe, table, if_exists="replace")
+        method.assert_called_with(
+            conn=ANY,
+            df=ANY,
```
kaxil, line 206: why is `df` ANY?

shouldn't it be `df=pandas_dataframe`?

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -155,8 +159,7 @@ def test_snowflake_create_table_using_native_schema_autodetection(
     [
         {
             "database": Database.SNOWFLAKE,
-            "table": Table(metadata=Metadata(schema=SCHEMA)),
-            "file": File(str(pathlib.Path(CWD.parent, "data/sample.csv"))),
```
kaxil, line 159: Why are this file objects no longer required?

python-sdk/src/astro/databases/base.py
utkarsharma2: @kaxil I'm using run_sql() to run the truncate query, which needs to be executed immediately.  I can create a separate function to execute this query or move this logic to the tuncate_table() function.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Nope, doesn't seem like it.
If i have a table with lowercase cols:
x, y
1, 2
3,4

and you execute query 
case 1: 
SELECT x, y From <table_name>
Screenshot 2022-10-14 at 11 16 48 PM


case 2:
SELECT "x", "y" From <table_name>

Screenshot 2022-10-14 at 11 17 53 PM

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: We need to keep identifiers in double quotes.

python-sdk/tests/databases/test_snowflake.py
utkarsharma2: With this test case, we are just concerned if the `auto_create_table` is true or false base on the append/replace value. I can add it, no strong opinion.

python-sdk/src/astro/databases/snowflake.py
kaxil: The are only case sensitive if you create the table with quotes.

in your screenshot, how are you creating the table? Can you paste the query

utkarsharma2: Exactly, If someone create a table with lower case/Mixed case cols we will be able to handle that.

utkarsharma2: Table creation query - 
CREATE TABLE persons (
    "id" int,
    "name" varchar(255)
);

kaxil: But now we are requiring users to add quotes to all the columns names in their SQL, this is a big con in my opinion

utkarsharma2: So we are keeping them consistent - Whatever user is passing, we are keeping them as it is in the database, and when they are using it in another operator, like data frame they will get the same case for cols.

utkarsharma2: @kaxil They only have to add that if the cols are in lower/mixed case. If they are in uppercase they don't have to add it.

kaxil: Isn't column being in lower case more common?

And I am also talking about the usage in run_raw_sql for example.

as a user now I am forced to add quotes to all my column names, no?

utkarsharma2: yes that's true, it's a tradeoff

## With changes:

### Case 1:
if we insert the below table in snowflake:
X,Y
1, 2,
3, 4

**run_raw_sql():**
we can query it like this:
Select x,y from <table_name>
or 
Select X,Y from <table_name>

**dataframe():**
will return:
X,Y  <-- same as table 
1, 2,
3, 4

### Case 2:
if we insert the below table in snowflake:
x,y
1, 2,
3, 4

**run_raw_sql():**
we can query it like this:
Select "x", "y" from <table_name>

**dataframe():**
will return:
x,y  <-- same as table 
1, 2,
3, 4


## Without Changes

### Case 1:
if we insert the below table in snowflake:
X, Y
1, 2,
3, 4

**run_raw_sql():**
we can query it like this:
Select x,y from <table_name>
or 
Select X,Y from <table_name>

**dataframe():**
will return:
X, Y  <-- same as table 
1, 2,
3, 4

### Case 2:
if we insert the below table in snowflake:
x,y
1, 2,
3, 4

**run_raw_sql():**
we can query it like this:
Select x, y from <table_name>

**dataframe():**
will return:
X,Y  <-- not same as table 
1, 2,
3, 4

python-sdk/src/astro/databases/base.py
pgzmnk: The original functionality is confusing for Astro Build users directly working with output data in subsequent cells. Unaware of the SDK internals, users expect to e.g. work in a Python task with columns in the same casing as they were in an upstream SQL task.

We are looking for a way to disable this without an explicit flag in the DAG code. We initially thought an option could be to define casing with an env variable we could set for all projects; but this would cause an inconsistency for Pipelines exported to deployments.

Astro Build aside, it makes sense that the default behavior is to not change casing. Especially if it's not clearly obvious to the user.

python-sdk/src/astro/databases/snowflake.py
kaxil: I did some investigation and I think we should follow `snowflake-sqlalchemy` to resolve this (https://docs.snowflake.com/en/user-guide/sqlalchemy.html#object-name-case-handling) — which is also mentioned in their official docs

My recommendation is we should have the same path as well instead of the current approach  where we use identifiers for everything:

- If data from load_file has “non lower-case” (upper or mixed case), we use “identifiers”
- If data from load_file has “lower_case”, we use it “as-is”

That way we are consistent with snowflake-sqlalchemy  and the most common case of “lower-case” will be catered — i.e. those users remain completely unaffected.

Slack Thread 🧵: https://astronomer.slack.com/archives/C03868KGF2Q/p1666046525734929?thread_ts=1665999175.728989&cid=C03868KGF2Q

.github/workflows/ci-benchmark.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       AIRFLOW__ASTRO_SDK__SQL_SCHEMA: astroflow_ci
+      AIRFLOW__ASTRO_SDK__SNOWFLAKE_DEFAULT_SCHEMA: ASTROFLOW_CI
```
kaxil, line 32: Why is this needed?

python-sdk/src/astro/databases/snowflake.py
kaxil: @utkarsharma2 Can you address this please

python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -1068,3 +1069,77 @@ def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
...
+        {
+            "database": Database.SNOWFLAKE,
+        }
+    ],
+    indirect=True,
+    ids=["snowflake"],
+)
+def test_load_file_col_cap(sample_dag, database_table_fixture, text_cases):
+    db, test_table = database_table_fixture
+    # path = str(CWD) + "/../../data/homes_upper.csv"
```
kaxil, line 1101: ?

python-sdk/src/astro/databases/base.py
kaxil: Why was it not required earlier? and why is it needed now?

Code comments on it would be good if we still need it

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -8,6 +8,7 @@
 from dataclasses import dataclass, field
 from typing import Any
 
+import pandas
```
kaxil, line 11: pandas is already imported on L12

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +478,62 @@ def create_table_using_native_schema_autodetection(
...
+    def use_quotes(cols: tuple[str]):
+        return SnowflakeDatabase.get_cols_case(cols) in ["upper", "mixed"]
+
     def create_table_using_schema_autodetection(
         self,
         table: BaseTable,
         file: File | None = None,
         dataframe: pd.DataFrame | None = None,
-        columns_names_capitalization: ColumnCapitalization = "lower",
+        columns_names_capitalization: ColumnCapitalization = "original",
```
kaxil, line 507: This parameter isn't used anymore, right?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +478,62 @@ def create_table_using_native_schema_autodetection(
...
-                columns_names_capitalization=columns_names_capitalization,
-            )
-
-        # Snowflake doesn't handle well mixed capitalisation of column name chars
-        # we are handling this more gracefully in a separate PR
-        super().create_table_using_schema_autodetection(table, dataframe=dataframe)
+        pandas_tools.write_pandas(
+            conn=self.hook.get_conn(),
+            df=source_dataframe,
+            table_name=table.name.upper(),
```
kaxil, line 529: Do we need `.upper()` here?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +478,62 @@ def create_table_using_native_schema_autodetection(
...
+
+        total = sum(result)
+        if total == 0:
+            return "lower"
+        elif total == len(cols):
+            return "upper"
+
+    @staticmethod
+    def use_quotes(cols: tuple[str]):
+        return SnowflakeDatabase.get_cols_case(cols) in ["upper", "mixed"]
```
kaxil, line 500: You could simplify this as below:

```suggestion
    @staticmethod
    def use_quotes(cols: tuple[str]):
        return any(col for col in cols if not col.islower())
```


Example tests:

```python
In [5]: use_quotes(["list", "sell"])
Out[5]: False

In [6]: use_quotes(["List", "sell"])
Out[6]: True

In [7]: use_quotes(["List", "Sell"])
Out[7]: True

In [8]: use_quotes(["List", "sEll"])
Out[8]: True

In [9]: use_quotes(["list", "sEll"])
Out[9]: True

In [10]: use_quotes(["LIST", "sEll"])
Out[10]: True

In [11]: use_quotes(["LIST", "sell"])
Out[11]: True

```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +478,62 @@ def create_table_using_native_schema_autodetection(
...
-        # we are handling this more gracefully in a separate PR
-        super().create_table_using_schema_autodetection(table, dataframe=dataframe)
+        pandas_tools.write_pandas(
+            conn=self.hook.get_conn(),
+            df=source_dataframe,
+            table_name=table.name.upper(),
+            schema=table.metadata.schema,
+            database=table.metadata.database,
+            chunk_size=DEFAULT_CHUNK_SIZE,
+            quote_identifiers=SnowflakeDatabase.use_quotes(source_dataframe),
```
kaxil, line 533:
```suggestion
            quote_identifiers=self.use_quotes(source_dataframe),
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +478,62 @@ def create_table_using_native_schema_autodetection(
...
+            auto_create_table=True,
+        )
+        self.truncate_table(table)
```
kaxil, line 536: Why do we truncate table after loading the data?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -702,6 +735,16 @@ def _build_merge_sql(
         source_cols = source_to_target_columns_map.keys()
         target_cols = source_to_target_columns_map.values()
 
+        target_table_sqla = self.get_sqla_table(target_table)
+        target_identifier_enclosure = ""
+        if SnowflakeDatabase.use_quotes(target_table_sqla.columns.keys()):
```
kaxil, line 740:
```suggestion
        if self.use_quotes(target_table_sqla.columns.keys()):
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -702,6 +735,16 @@ def _build_merge_sql(
...
+            target_identifier_enclosure = '"'
+
+        source_table_sqla = self.get_sqla_table(source_table)
+        source_identifier_enclosure = ""
+        if SnowflakeDatabase.use_quotes(source_table_sqla.columns.keys()):
```
kaxil, line 745:
```suggestion
        if self.use_quotes(source_table_sqla.columns.keys()):
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +818,68 @@ def _build_merge_sql(
...
 
+    def append_table(
+        self,
+        source_table: BaseTable,
+        target_table: BaseTable,
+        source_to_target_columns_map: dict[str, str],
+    ) -> None:
+        """
+        Append the source table rows into a destination table.
+        The argument `if_conflicts` allows the user to define how to handle conflicts.
```
kaxil, line 829: Where is the `if_conflicts` argument?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +818,68 @@ def _build_merge_sql(
...
+
+        Overriding the base method since we need to add quotes around the identifiers for
+         snowflake to preserver case of cols - Column(name=col, quote=True)
```
kaxil, line 832:
```suggestion
         snowflake to preserve case of cols - Column(name=col, quote=True)
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +818,68 @@ def _build_merge_sql(
...
+
+        :param source_table: Contains the rows to be appended to the target_table
+        :param target_table: Contains the destination table in which the rows will be appended
+        :param source_to_target_columns_map: Dict of source_table columns names to target_table columns names
+        """
+        target_table_sqla = self.get_sqla_table(target_table)
+        source_table_sqla = self.get_sqla_table(source_table)
+        use_quotes_target_table = SnowflakeDatabase.use_quotes(target_table_sqla.columns.keys())
+        use_quotes_source_table = SnowflakeDatabase.use_quotes(source_table_sqla.columns.keys())
```
kaxil, line 841:
```suggestion
        use_quotes_target_table = self.use_quotes(target_table_sqla.columns.keys())
        use_quotes_source_table = self.use_quotes(source_table_sqla.columns.keys())
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +818,68 @@ def _build_merge_sql(
...
+                for col in source_to_target_columns_map.keys()
+            ]
+
+        sel = select(source_columns).select_from(source_table_sqla)
+        # TODO: We should fix the following Type Error
+        # incompatible type List[ColumnClause[<nothing>]]; expected List[Column[Any]]
+        sql = insert(target_table_sqla).from_select(target_columns, sel)  # type: ignore[arg-type]
+        self.run_sql(sql=sql)
+
+    @staticmethod
```
kaxil, line 869: you can convert this to class method so you can do `cls.use_quotes`

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -766,6 +818,68 @@ def _build_merge_sql(
...
+    def get_merge_initialization_query(parameters: tuple[str]) -> str:
+        """
+        Handles database-specific logic to handle constraints, keeping
+        it agnostic to database.
+        """
+        identifier_enclosure = ""
+        if SnowflakeDatabase.use_quotes(parameters):
```
kaxil, line 876:
```suggestion
        if cls.use_quotes(parameters):
```

if you make this https://github.com/astronomer/astro-sdk/pull/1050/files#r998794279 change

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -51,7 +55,7 @@ def test_snowflake_run_sql():
 def test_table_exists_raises_exception():
     """Test if table exists in snowflake database"""
     database = SnowflakeDatabase(conn_id=CUSTOM_CONN_ID)
-    table = Table(name="inexistent-table", metadata=Metadata(schema=SCHEMA))
+    table = Table(name="inexistent-table", metadata=Metadata(schema=SNOWFLAKE_SCHEMA))
```
kaxil, line 58: Why did we need to change `SCHEMA` to `SNOWFLAKE_SCHEMA` (I see only difference is lower-case / upper-case) -- why though?

python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -1068,3 +1069,77 @@ def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
...
+    "database_table_fixture",
+    [
+        {
+            "database": Database.SNOWFLAKE,
+        }
+    ],
+    indirect=True,
+    ids=["snowflake"],
+)
+def test_load_file_col_cap(sample_dag, database_table_fixture, text_cases):
```
kaxil, line 1099: docstring on what are we testing here please

python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -1068,3 +1069,77 @@ def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
...
+    "database_table_fixture",
+    [
+        {
+            "database": Database.SNOWFLAKE,
+        }
+    ],
+    indirect=True,
+    ids=["snowflake"],
+)
+def test_load_file_col_cap_native_path(sample_dag, database_table_fixture):
```
kaxil, line 1127: docstring on what are we testing here please

python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -1068,3 +1069,77 @@ def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
...
+    path = str(CWD) + "/../../data/homes_upper.csv"
+    with sample_dag:
+        load_file(
+            input_file=File(path),
+            output_table=test_table,
+            native_support_kwargs={
+                "snowflake_storage_integration_google": "gcs_int_python_sdk",
+                "storage_integration": "gcs_int_python_sdk",
+            },
+            enable_native_fallback=False,
```
kaxil, line 1138: Do we still fallback by default?

Have we not disabled it by default and created a config to control that? If not we should absolutely create a ticket for us to tackle it.

cc @phanikumv

python-sdk/tests/sql/operators/test_snowflake_merge_func.py
```diff
@@ -44,12 +44,12 @@ def test_merge_func(self):
         assert (
             sql == "merge into IDENTIFIER(:target_table) using IDENTIFIER(:source_table) "
             "on Identifier(:merge_clause_target_0)=Identifier(:merge_clause_source_0) "
-            f"when matched then UPDATE SET {self.target_table_name}.sell={self.source_table_name}.sell "
-            f"when not matched then insert({self.target_table_name}.sell) values ({self.source_table_name}.sell)"
+            f'when matched then UPDATE SET {self.target_table_name}."sell"={self.source_table_name}."sell" '
+            f'when not matched then insert({self.target_table_name}."sell") values ({self.source_table_name}."sell")'
```
kaxil, line 48: Weren't this lower case already, do we need quotes for it?

python-sdk/tests/sql/operators/test_load_file.py
phanikumv: Created https://github.com/astronomer/astro-sdk/issues/1089 to tackle this

phanikumv:
@pankajastro @sunank200 need your feedback on this PR

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Nice!

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Yes, we don't need the `columns_names_capitalization` feature itself. We can remove it altogether.

python-sdk/src/astro/databases/base.py
pankajastro: what would happen if we don't do we need to call commit manually?

python-sdk/src/astro/databases/snowflake.py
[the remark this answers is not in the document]
utkarsharma2: Removed `get_cols_case` and added doc string to use_quotes.

pankajastro:
> @pankajastro @sunank200 need your feedback on this PR

I still think python-sdk does not change the column, the snowflake DB store it in capital unless we pass it in quotes. So I'm not sure about preserving the col name by putting quotes from our side is the right idea or not.

python-sdk/tests/sql/operators/test_snowflake_merge_func.py
utkarsharma2: Removed it.

python-sdk/tests/sql/operators/test_load_file.py
utkarsharma2: Added.

python-sdk/tests/sql/operators/test_load_file.py
utkarsharma2: Added.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: This metho was copied from base.py. Removing it both. Thanks for pointing out.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Done.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Done.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Done.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Done.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Done

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: So earlier we were using prep_table() which just takes a dataframe and then comes up with an SQL query to create the table and we executed the query. But that function doesn't give us the quotes=True/False. So this was the next best option I had to create a table. Since we are just expecting an empty table to be created, we are truncating. Loading of actual data happens in two separate flows - 1. Pandas 2. Native.

python-sdk/tests/databases/test_snowflake.py
utkarsharma2: I think we still need it, since we are `write_pandas()` doesn't give us an option of treating schema and table name differently, everything is an identifier. So for mixed/upper case, we preserver the case of all the identifiers, which will lead the query to form with "astroflow_ci" and that scheme won't be found since it's not there.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +477,66 @@ def create_table_using_native_schema_autodetection(
             },
         )
 
+    @classmethod
+    def use_quotes(cls, cols: tuple[str, ...]) -> bool:
+        """
+        With snowflake identifier we have two cases,
+
+        1. When Upper/Mixed case col names are used
+            We are required to preserver the text casing of the col names. By adding the quotes around identifier.
```
kaxil, line 486:
```suggestion
            We are required to preserverthe text casing of the col names. By adding the quotes around identifier.
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +477,66 @@ def create_table_using_native_schema_autodetection(
             },
         )
 
+    @classmethod
+    def use_quotes(cls, cols: tuple[str, ...]) -> bool:
```
kaxil, line 481:
```suggestion
    def use_quotes(cls, cols: tuple[str]) -> bool:
```
probably?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -787,6 +901,10 @@ def openlineage_dataset_namespace(self) -> str:
         account = self.hook.get_connection(self.conn_id).extra_dejson.get("account")
         return f"{self.sql_type}://{account}"
 
+    def truncate_table(self, table):
```
kaxil, line 904:
```suggestion
    def truncate_table(self, table: BaseTable):
```

python-sdk/src/astro/databases/snowflake.py
kaxil: So we are double loading the file (although the first time we do -- we probably only load 1000 rows for schema detection)

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -475,35 +477,68 @@ def create_table_using_native_schema_autodetection(
...
+        lowercase object names to be case-insensitive. Snowflake SQLAlchemy converts the object name case during
+        schema-level communication (i.e. during table and index reflection). If you use uppercase object names,
+        SQLAlchemy assumes they are case-sensitive and encloses the names with quotes. This behavior will cause
+        mismatches against data dictionary data received from Snowflake, so unless identifier names have been truly
+        created as case sensitive using quotes (e.g. "TestDb"), all lowercase names should be used on the SQLAlchemy
+        side.
+
+        :param cols: tuple of columns
+        """
+        return any(col for col in cols if not col.islower())
```
kaxil, line 503: We should add a unit test with parameterize for this method

kaxil (approved):
We should add a user-facing doc about the new "casing" behaviour !!

This was a toughy

python-sdk/src/astro/databases/snowflake.py
kaxil: Separately, a TODO for you is that you are going to create a Github issue on the double-loading for non-native path

utkarsharma2: Added the testcases for use_quotes() function.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Reverted

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: @kaxil Added the ticket - https://github.com/astronomer/astro-sdk/issues/1097

python-sdk/src/astro/databases/snowflake.py
sunank200: We should call it out as comment as TODO

python-sdk/src/astro/databases/snowflake.py
sunank200: We should add it as comment

python-sdk/src/astro/databases/snowflake.py
sunank200: This needs to be added as comment

sunank200 (approved):
Overall LGTM now but this PR comment resolution @utkarsharma2

python-sdk/src/astro/databases/base.py
utkarsharma2: @pankajastro We don't have to call it manually, but sqlalchemy calls it based on the optimization logic they have, we don't have the control over it.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Added a ticket - https://github.com/astronomer/astro-sdk/issues/1099

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: Done.

kaxil (approved):

kaxil (commented):
oh, user-facing docs are missing

utkarsharma2:
Screenshot 2022-10-20 at 12 13 28 PM
Screenshot 2022-10-20 at 12 13 41 PM

Issue #963, opened 2022-09-29
Mapped Tasks Not `cleanup`'d?


fritz-astronomer:
**Describe the bug**
It appears that mapped tasks aren't cleaned up via `cleanup`.
Relevant logs (note: there were `TempTable` objects in the XCOM view for mapped tasks in this DAG)
```
[2022-09-29, 16:01:35 UTC] {cleanup.py:183} INFO - No tables provided, will delete all temporary tables
[2022-09-29, 16:01:35 UTC] {cleanup.py:78} INFO - Tables found for cleanup: 
```

**Version**
* astro-sdk-python[google,snowflake]==1.1.0
* astro-runtime: 6.0.0

**To Reproduce**
Load was a Parquet File from GCS into Snowflake. There was another `MergeOperator` after this, but that shouldn't be relevant

```python
from airflow.models.dag import DAG
from astro.sql import LoadFileOperator, Table, cleanup
from astro.files import File
from datetime import datetime

with DAG("error", start_date=datetime(1970,1,1), schedule=None) as dag:
  LoadFileOperator.partial(
    task_id="foo",
    ...
  ).expand_kwargs([{
    "input_file": File(path=..., conn_id=...),
    "output_table": Table(temp=True, conn_id=..., metadata=..., columns=...)
  }])

  cleanup()
```

**Expected behavior**
Temp tables should be cleaned up, even with mapped tasks

kaxil:
@feluelle This would be a good one for you to take on if you have some bandwidth this week.

feluelle:
Yes, let me take it :)

phanikumv:
Moving to 1.3.0 as we are not sure that this will be completed by the deadline.

phanikumv:
@feluelle is this ready for review yet?

feluelle:
The PR is still in draft, because the issue is not yet fixed.

feluelle:
Short update: Good news. I can reproduce this.

The following gets cleaned up on Snowflake by using `aql.cleanup()`:
```
LoadFileOperator(
    task_id="foo",
    input_file=File(path=(CWD.parent.parent / "data/sample.csv").as_posix()),
    output_table=temp_table,
)
```
where as this not:
```
LoadFileOperator.partial(
    task_id="foo",
).expand_kwargs([{
    "input_file": File(path=(CWD.parent.parent / "data/sample.csv").as_posix()),
    "output_table": temp_table,
}])
```
[...]

PR #1053, opened 2022-10-13
Fix `CleanupOperator` for mapped tasks

feluelle:
# Description

Currently, `aql.cleanup()` does not cleanup/remove all temp tables.

We know this for sure for Snowflake:

The following [CI logs](https://github.com/astronomer/astro-sdk/actions/runs/3236060888/jobs/5309951211#step:11:967) were giving us the hint that the tables are not being deleted after running the tests.
```
E       sqlalchemy.exc.ProgrammingError: (snowflake.connector.errors.ProgrammingError) 090153 (22000): 01a7981b-0605-9c81-0000-68211381e012: The result set size exceeded the max number of rows(10000) supported for SHOW statements. Use LIMIT option to limit result set to a smaller number.
E       [SQL: SHOW /* sqlalchemy:get_table_names */ TABLES IN astroflow_ci]
E       (Background on this error at: https://sqlalche.me/e/14/f405)
```
The schema where we write the tables to exceeded 10000 tables.

closes: #963 

## What is the new behavior?

Cleanup mapped tasks

- run cleanup tests on all databases

No, but we will remove temp tables when the issue is fixed.

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/cleanup.py
+++ b/python-sdk/src/astro/sql/operators/cleanup.py
@@ -8,7 +8,15 @@ from airflow.decorators.base import get_unique_task_id
 from airflow.exceptions import AirflowException
 from airflow.models.baseoperator import BaseOperator
 from airflow.models.dagrun import DagRun
-from airflow.models.mappedoperator import MappedOperator
+
+try:
+    # Airflow >= 2.3
+    from airflow.models.mappedoperator import MappedOperator
+except ImportError:
+    # Airflow < 2.3
+    MappedOperator = None
+
+
 from airflow.models.taskinstance import TaskInstance
 from airflow.utils.state import State
 
@@ -213,7 +221,11 @@ class CleanupOperator(AstroSQLBaseOperator):
                     t = task.output.resolve(context)
                     if isinstance(t, BaseTable):
                         res.append(t)
-                elif isinstance(task, MappedOperator) and task.operator_class is LoadFileOperator:
+                elif (
+                    MappedOperator
+                    and isinstance(task, MappedOperator)
+                    and task.operator_class is LoadFileOperator
+                ):
                     for t in task.output.resolve(context):
                         if isinstance(t, BaseTable):
                             res.append(t)
--- a/python-sdk/src/astro/sql/operators/cleanup.py
+++ b/python-sdk/src/astro/sql/operators/cleanup.py
@@ -8,6 +8,7 @@ from airflow.decorators.base import get_unique_task_id
 from airflow.exceptions import AirflowException
 from airflow.models.baseoperator import BaseOperator
 from airflow.models.dagrun import DagRun
+from airflow.models.mappedoperator import MappedOperator
 from airflow.models.taskinstance import TaskInstance
 from airflow.utils.state import State
 
@@ -55,8 +56,8 @@ class CleanupOperator(AstroSQLBaseOperator):
         Very relevant if using a synchronous executor. The default is 3.
     :param retry_delay: Delay between running retries. Very relevant if using a synchronous executor.
         The default is 10s.
-    :param run_sync_mode: Whether to wait for the DAG to finish or not. Set to False if you want
-        to immediately clean all DAGs. Note that if you supply anything to `tables_to_cleanup`
+    :param run_sync_mode: Whether to wait for the DAG to finish or not. Set to True if you want
+        to immediately clean all tables. Note that if you supply anything to `tables_to_cleanup`
         this argument is ignored.
     """
 
@@ -191,7 +192,9 @@ class CleanupOperator(AstroSQLBaseOperator):
         task_outputs = self.resolve_tables_from_tasks(tasks=tasks, context=context)
         return task_outputs
 
-    def resolve_tables_from_tasks(self, tasks: list[BaseOperator], context: Context) -> list[BaseTable]:
+    def resolve_tables_from_tasks(  # noqa: C901
+        self, tasks: list[BaseOperator], context: Context
+    ) -> list[BaseTable]:
         """
         For the moment, these are the only two classes that create temporary tables.
         This function allows us to only resolve xcom for those objects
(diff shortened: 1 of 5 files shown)


python-sdk/tests/sql/operators/test_cleanup.py
```diff
@@ -186,16 +186,26 @@ def test_cleanup_multiple_table(database_table_fixture, multiple_tables_fixture)
...
+        assert not db.table_exists(table_2)
+
     with sample_dag:
         foo(table_1)
         foo(table_2)
 
-        aql.cleanup([])
+        # FIXME: on_success_callback is being ignored
+        # see https://airflow.apache.org/docs/apache-airflow/stable/logging-monitoring/callbacks.html#callbacks
+        aql.cleanup(on_success_callback=assert_tables_do_not_exist)
```
feluelle, line 208: @dimberman any idea how we can test this? I want to assert the tables have been removed after running `aql.cleanup`. The solution I would have loved to work is by using `on_success_callback`. But unfortunately callbacks are being ignored by our (& Airflow's) runner.

kaxil: Probably one solution until `on_*_callback` is supported might be a follow-up python task that gets the table names from cleanup (as I think `aql.cleanup` pushes table names to XCom -- do verify it though) and checks if they exists?

feluelle: Good point, but I already tested similar ways, but none of them worked:
* create a task downstream to run `assert` -> this does not work as cleanup waits until the dagrun is completed
* create another dag that waits until the dag finishes and then runs `assert` -> this did not work as we delete the dag after running it, it seems (but not sure)

python-sdk/tests/sql/operators/test_cleanup.py
```diff
@@ -42,16 +39,19 @@
...
 
 
 @pytest.mark.integration
 @pytest.mark.parametrize(
     "database_table_fixture",
-    SQLITE_ONLY,
+    SUPPORTED_DATABASES_WITH_FILE,
     indirect=True,
-    ids=["sqlite"],
+    ids=["sqlite", "postgres", "bigquery", "snowflake", "redshift"],
```
tatiana, line 54: Any reason why we're not using `astro_sdk.constants.SUPPORTED_DATABASES` here, and in other similar test cases?

feluelle: Because I didn't know that we have this :) Thanks :)

python-sdk/tests/sql/operators/test_cleanup.py
kaxil: How about a non-Airflow command after `run_dag` to assert this manually by calling the DB?

feluelle: So you mean kind of a sensoring operation but not as an airflow task?

kaxil: No I just meant the following on L210:

```
assert not db.table_exists(table_1)
assert not db.table_exists(table_2)
```

feluelle: That was the first thing I tried 😂 This does not work as the cleanup waits until the dagrun has finished. These asserts would check before cleanup would have a chance to remove the tables.

kaxil: But on L210, the DAG will have completed the execution. It will only be running in case of a "failed test" to prevent which you can add a "timeout" of few seconds.

dimberman: @feluelle do you want to pair on this tomorrow?

feluelle: Yes, let's pair on this :)

python-sdk/src/astro/sql/operators/cleanup.py
```diff
@@ -55,7 +55,7 @@ class CleanupOperator(AstroSQLBaseOperator):
         Very relevant if using a synchronous executor. The default is 3.
     :param retry_delay: Delay between running retries. Very relevant if using a synchronous executor.
         The default is 10s.
-    :param run_sync_mode: Whether to wait for the DAG to finish or not. Set to False if you want
+    :param run_sync_mode: Whether to wait for the DAG to finish or not. Set to True if you want
         to immediately clean all DAGs. Note that if you supply anything to `tables_to_cleanup`
```
utkarsharma2, line 59: is `Set to True if you want to immediately clean all DAGs.` right? or it should be `tables`?

feluelle: Makes sense

feluelle:
```suggestion
        to immediately clean all tables. Note that if you supply anything to `tables_to_cleanup`
```

python-sdk/tests/sql/operators/test_cleanup.py
feluelle: Thanks @kaxil, @dimberman has found the solution. I was testing for the wrong tables to not exist. I have pushed the changes. All tests are passing.

feluelle:
FYI There are some weird issues going on with the `test_cleanup_non_temp_table` test. Running them one by one works as expected but running all together fails for some. I will continue to investigate..

python-sdk/src/astro/sql/operators/cleanup.py
```diff
@@ -205,16 +216,24 @@ def resolve_tables_from_tasks(self, tasks: list[BaseOperator], context: Context)
...
                         res.append(t)
-                except AirflowException:
-                    self.log.info(
-                        "xcom output for %s not found. Will not clean up this task",
-                        task.task_id,
-                    )
+                elif (
+                    MappedOperator
+                    and isinstance(task, MappedOperator)
+                    and task.operator_class is LoadFileOperator
```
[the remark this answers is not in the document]
feluelle: Yes, thanks for the reminder.

feluelle:
Fixed the issues related to cleanup. All current failing issues are unrelated. They are also failing in main.

kaxil (approved):

Issue #902, opened 2022-09-23
Openlineage support - Add Extractor for `TransformOperator`


kaxil:
**Please describe the feature you'd like to see**
We should be able to extract open lineage info from the `TransformOperator`.

**Describe the solution you'd like**
- Add a method `get_openlineage_facets` on the Operator (based on [this doc](https://docs.google.com/document/d/1vPsvHejQ24xTbzpz_LYSf0_ixk9oUuBiEUHVEaF9J2U/edit?usp=sharing))
- The "PythonSDKExtractor" built-in https://github.com/astronomer/astro-sdk/issues/898 should be able to work with `TransformOperator.get_openlineage_facets`. Test it to make sure it works

**Acceptance Criteria**

- [ ] Post the screenshot of how it looks in the Openlineage/Marquez UI
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)

PR #1056, opened 2022-10-13
Add open lineage integration for BaseSQLDecoratedOperator and TransformOperator

sunank200:
# Description

Open Lineage support for `BaseSQLDecoratedOperator` and  `TransformOperator` is missing.

closes: #901
closes: #902 

## What is the new behavior?

- Add open Lineage support for `BaseSQLDecoratedOperator` and  `TransformOperator` is missing.
- Add the test for the same

No

### Checklist


utkarsharma2:
@sunank200 
1. Do we have to add anything specifically for the Transform operator which is not covered by BaseSQLDecoratedOperator?

2. How will this look on UI? will we be able to distinguish? b/w TransformOperator and another operator that inherits from BaseSQLDecoratedOperator?

sunank200:
> @sunank200
> 
> 1. Do we have to add anything specifically for the Transform operator which is not covered by BaseSQLDecoratedOperator?
> 2. How will this look on UI? will we be able to distinguish? b/w TransformOperator and another operator that inherits from BaseSQLDecoratedOperator?

@utkarsharma2 
1. No all the facets are the same in `BaseSQLDecoratedOperator` and in `TransformOperator` as both have same constructor.
2. This is how the facets are reflected on the json file from OL:

pankajastro (approved):
LGTM

PR #1057, opened 2022-10-13
Remove the target_directory from generate_dag

feluelle:
# Description

Currently, the user has to input a target directory where the generated sql files land. However we don't need a separate folder for the sql files without headers we can just put them in a relative path inside the dags folder.

## What is the new behavior?

The sql files without headers will be written to a `.sql/<sql-files-dir>` instead of a directory passed by the user.

We remove the target_directory from generate_dag function.

- restructure directories for sql files and dags

N/A

### Checklist


kaxil (approved):

sql-cli/sql_cli/dag_generator.py
```diff
@@ -80,17 +80,16 @@ def sorted_sql_files(self) -> list[SqlFile]:
...
     Generate a DAG from SQL files.
 
     :params directory: The directory containing the raw sql files.
-    :params target_directory: The directory containing the executable sql files.
     :params dags_directory: The directory containing the generated DAG.
 
     :returns: the path to the DAG file.
     """
-    sql_files = sorted(get_sql_files(directory, target_directory))
+    sql_files = sorted(get_sql_files(directory, target_directory=dags_directory / ".sql"))
```
dimberman, line 92: Why do you sort these?

sql-cli/examples/.airflow/dags/simple.py
```diff
@@ -6,15 +6,15 @@
 from astro import sql as aql
 from astro.table import Table
 
-DAGS_DIRECTORY = Path(__file__).parent.as_posix()
+PROJECT_DIRECTORY = Path(__file__).parent.parent.parent.as_posix()
```
dimberman, line 9: I'm fine with this for an 0.1, but it feels weird to have this level of directory structure dependence. Like this would ONLY work if we push the entire project as is to the airflow deployment.

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -99,15 +99,17 @@ def get_relative_target_path(self) -> str:
...
+        )
         target_full_directory.mkdir(parents=True, exist_ok=True)
 
         target_path = target_full_directory / self.path.name
 
         with open(target_path, "w") as file:
             file.write(f"{self.content}\n")
 
-        return target_path.relative_to(self.root_directory.parent).as_posix()
+        return target_path.relative_to(self.root_directory.parent.parent).as_posix()
```
dimberman, line 112: Having `parent.parent` or `parent.parent.parent` feels like technical debt. It's ok for now but I don't want this to be how this is released 1.0.

dimberman (changes requested):

dimberman (approved):
I'm approving this for the sake of us meeting the 0.1 deadline, but I think we need to discuss if this can be done in a more sustainable manner after the release.

sql-cli/sql_cli/sql_directory_parser.py
feluelle: I agree. But I could not get it to work quickly without all these parents :D

sql-cli/examples/.airflow/dags/simple.py
feluelle: No, from the generator perspective it should be sufficient to deploy `.airflow` directory which contains `dags` and `.sql`.

feluelle: This path resolution would still work like this, because the first parent of the dag file is `dags`, the 2nd is `.airflow` and the third is whatever you copy the directory to.

sql-cli/sql_cli/dag_generator.py
feluelle: Because it is a `set` of sql_files. A SqlFile has a unique path. You cannot have the same file twice. I found it more logical to return a set. But happy to change it to a list. However I am not sure if the `glob` (used underneath) is deterministic (see its [documentation](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob)). In the example they actually also sort the files after.

feluelle: If we use a non-deterministic structure like set for generating the tasks in the DAG, we will have a non-deterministic order of tasks if they have no dependencies for example.

PR #1060, opened 2022-10-14
return error for improper xcom

dimberman:
…com-pickling-with-exception# Description
Currently the XCom custom backend has incorrect logs that mispell the env variables that need to be set for proper function. Also we have found that giving a warning is not loud enough and people end up with errors that are hard to understand.
## What is the new behavior?

- To remedy this situation we now throw an exception if the custom backend is set without proper companion settings.

No

### Checklist


python-sdk/src/astro/custom_backend/serializer.py
```diff
@@ -43,7 +43,7 @@ def _attempt_to_serialize_unknown_object(obj: object):
     try:
         return json.dumps(obj)
     except TypeError:
-        return pickle.dumps(obj).hex()
+        return obj
```
sunank200, line 46: Why are we returning the `obj` when `TypeError`?

python-sdk/src/astro/custom_backend/serializer.py
```diff
@@ -87,12 +87,5 @@ def _attempt_to_deser_unknown_object(obj: str):
...
-            log.debug(
-                "Json debugging failed for object %s, attempting to pickle deserialize",
-                obj,
-            )
-            return pickle.loads(bytes.fromhex(obj))
-        except UnpicklingError:
-            log.debug("unpickling failed for object %s, returning raw object", obj)
-            return obj
+        log.debug("Json deserializing failed for object %s, returning raw object", obj)
+        return obj
```
sunank200, line 91: Why are we returning the `obj` when `JSONDecodeError`?

sunank200 (commented):
Added few comments

python-sdk/src/astro/custom_backend/serializer.py
dimberman: @sunank200, if we don't know how to debug an object, we have to assume that the user has an object that is not a json object and let them decide how to handle it.

python-sdk/src/astro/custom_backend/serializer.py
dimberman: Because if we can't serialize the object, we have to leave it to the user to decide whether they want to turn on pickling or not. The entire point of this backend is to remove the pickling dependency and I think people wouldn't be stoked if we were pickling objects and not telling them.

sunank200 (approved):

sunank200 (commented):
Tests are failing and that needs to be fixed

python-sdk/src/astro/custom_backend/astro_custom_backend.py
```diff
@@ -31,11 +32,12 @@ def serialize_value(value: Any, **kwargs) -> Any:  # type: ignore[override]
...
-                "Warning: Since you have not provided a remote object storage conn_id, we are defaulting to"
-                "storing your data in the Metadata DB. This does not scale well and can cause degradation to"
-                "your airflow DB. Please set the AIRFLOW__ASTRO__XCOM_STORAGE_CONN_ID and "
-                "AIRFLOW__ASTRO__XCOM_STORAGE_URL variables"
+            raise AirflowException(
+                "Since you have not provided a remote object storage conn_id, we are defaulting to "
+                "storing your data in the Metadata DB. This does not scale well and can cause degradation to "
+                "your airflow DB. Please set the AIRFLOW__ASTRO_SDK__XCOM_STORAGE_CONN_ID and "
+                "AIRFLOW__ASTRO_SDK__XCOM_STORAGE_URL variables "
+                "(or you can set AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV for local development)"
```
feluelle, line 40: Where do we catch this exception? Because it is saying that we default to sth...

python-sdk/src/astro/custom_backend/serializer.py
kaxil: in that case, should we just do `json.dumps(obj)` in `_attempt_to_serialize_unknown_object`? so that it will blow up for unknown type and they know that it is because of the serialization error?

phanikumv: where are we uploading to a given storage url here?

python-sdk/src/astro/custom_backend/astro_custom_backend.py
dimberman: fixed

PR #1062, opened 2022-10-14
Use include/base as official examples

feluelle:
# Description

We have old examples outside of the project structure. It makes more sense to have them in the project.

## What is the new behavior?

Remove old examples in `examples` directory. Instead, we use the examples in `include/base`.

No.

### Checklist


sql-cli/include/base/.airflow/.dags/.sql/example_basic_transform/top_animations.sql
```diff
@@ -0,0 +1,5 @@
+SELECT Title, Rating
```
tatiana, line 1: @feluelle shouldn't these SQL example workflows be within `include/base/workflows` for now, until we make examples optional (0.2)?

I'd expect only the materialised SQL files to be in hidden .airflow folder

feluelle: These are the materialised ones.

sql-cli/include/base/.airflow/.dags/.sql/example_templating/joint_orders_customers.sql
```diff
@@ -0,0 +1,4 @@
+SELECT c.customer_id, customer_name, order_id, purchase_date, amount, type
```
tatiana, line 1: Same here; shouldn't this be within `include/base/workflows`?

I'd expect for the folder `.airflow/dags/` to be empty by default

feluelle: I would suggest we have some example dags in the repo.

.gitignore
```diff
@@ -46,9 +46,14 @@ python-sdk/docs/autoapi/
 **/dev/plugins
 
 # Generated by tests
-sql-cli/tests/.airflow/dags/*
-!sql-cli/tests/.airflow/dags/.sql/.gitkeep
-!sql-cli/tests/.airflow/dags/.gitkeep
+sql-cli/tests/.airflow/.dags/*
```
tatiana, line 49: @feluelle, since we're adding this here and to pre-commit, could you already name things to what we agreed? I'm doing that in my PR, but it may be easier if we do as part of this one:
* `sql-cli/tests/.airflow/.dags/*` -> `sql-cli/tests/.airflow/dags/*`
* `sql-cli/tests/.airflow/.dags/.sql*` -> `sql-cli/tests/.airflow/dags/sql*`

tatiana (approved):

PR #1063, opened 2022-10-14
Add astro-sdk-python as a dependency for sql-cli project

pankajkoti:
We also are facing dependency resolution issue for certain providers like the Google providers where it is failing to resolve its internal dependencies for python versions >3.10. Hence, we are restricting the python version for the project to <3.11.

### DIFF
--- a/sql-cli/noxfile.py
+++ b/sql-cli/noxfile.py
@@ -16,8 +16,6 @@ def dev(session: nox.Session) -> None:
     development environment to ``.nox/dev``.
     """
     session.install("nox")
-    session.install("-e", "../python-sdk/.[all]")
-    session.install("-e", "../python-sdk/.[tests]")
     session.install("poetry")
     session.run("poetry", "install")
 
@@ -25,8 +23,6 @@ def dev(session: nox.Session) -> None:
 @nox.session(python=["3.7", "3.8", "3.9"])
 def test(session: nox.Session) -> None:
     """Run both unit and integration tests."""
-    session.install("-e", "../python-sdk/.[all]")
-    session.install("-e", "../python-sdk/.[tests]")
     session.install("poetry")
     session.run("poetry", "install")
     # Log all the installed dependencies
--- a/sql-cli/noxfile.py
+++ b/sql-cli/noxfile.py
@@ -24,7 +24,7 @@ def dev(session: nox.Session) -> None:
 def test(session: nox.Session) -> None:
     """Run both unit and integration tests."""
     session.install("poetry")
-    session.run("poetry", "install")
+    session.run("poetry", "install", "--with", "dev")
     # Log all the installed dependencies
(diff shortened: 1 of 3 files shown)


tatiana:
Thanks for looking into this so quickly, @pankajkoti !

It is probably worth it as part of this PR to remove the lines in Nox which install the Python SDK, e.g.:
https://github.com/astronomer/astro-sdk/blob/main/sql-cli/noxfile.py#L28-L29
(there may be more)

This way, we have a consistent way of setting the dependencies - both as a package and for the tests

pankajkoti:
> Thanks for looking into this so quickly, @pankajkoti !
> 
> It is probably worth it as part of this PR to remove the lines in Nox which install the Python SDK, e.g.: https://github.com/astronomer/astro-sdk/blob/main/sql-cli/noxfile.py#L28-L29 (there may be more)
> 
> This way, we have a consistent way of setting the dependencies - both as a package and for the tests

doing now, I will also mark the group.dev as optional and add `astro-sdk-python = {path = "../python-sdk", extras = ["tests"]} ` to its dependencies. We will explicitly need to install it for local development. Like this, those dependencies won't be packaged in the build, otherwise all the group dependencies get installed too by default I believe.

feluelle (approved):
This is great! Thank you @pankajkoti! 🚀

sql-cli/noxfile.py
```diff
@@ -49,7 +45,7 @@ def test(session: nox.Session) -> None:
 def type_check(session: nox.Session) -> None:
     """Run MyPy checks."""
     session.install("poetry")
-    session.run("poetry", "install")
+    session.run("poetry", "install", "mypy")
```
feluelle, line 48:
```suggestion
    session.run("poetry", "install")
    session.run("poetry", "add", "mypy")
```

sql-cli/pyproject.toml
```diff
@@ -26,23 +26,35 @@ Home = "https://astronomer.io/"
...
-python = ">=3.8,<4.0.0"
+python = ">=3.8,<3.11"
 typer = {extras = ["all"], version = "^0.6.1"}
 networkx = "^2.8.7"
 Jinja2 = ">=2.10.1, <=3.1"
 python-frontmatter = "^1.0.0"
 PyYAML = ">5.4.1"
 apache-airflow = ">2.0"
 python-dotenv = "^0.21.0"
+astro-sdk-python = {path = "../python-sdk", extras = ["all"]}
```
kaxil, line 37: Remember to change this to "non-dev" before the release though

pankajkoti: yes, @kaxil . Created a quick ticket for this so that we don't forget: https://app.zenhub.com/workspaces/sql-cli-631f43da09a3a67f2c795913/issues/astronomer/astro-sdk/1067

Issue #1064, opened 2022-10-14
Append Operator xcom push is non-json serializable


sunank200:
**Describe the bug**
Make the Database SQL query as string object to make it serializable

**Expected behavior**
Make the Database SQL query as string object to make it serializable

PR #1065, opened 2022-10-14
Cast the SQLAlchemy db object to string

sunank200:
# Description

Make the Database SQL query as string object to make it serializable

closes: #1064 

## What is the new behavior?

- Make the Database SQL query as string object to make it serializable

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/append.py
+++ b/python-sdk/src/astro/sql/operators/append.py
@@ -68,7 +68,7 @@ class AppendOperator(AstroSQLBaseOperator):
             target_table=self.target_table,
             source_to_target_columns_map=self.columns,
         )
-        context["ti"].xcom_push(key="append_query", value=db.sql)
+        context["ti"].xcom_push(key="append_query", value=str(db.sql))
         return self.target_table
 
     def get_openlineage_facets(self, task_instance) -> OpenLineageFacets:


python-sdk/src/astro/sql/operators/append.py
```diff
@@ -68,7 +68,7 @@ def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
             target_table=self.target_table,
             source_to_target_columns_map=self.columns,
         )
-        context["ti"].xcom_push(key="append_query", value=db.sql)
+        context["ti"].xcom_push(key="append_query", value=str(db.sql))
```
kaxil, line 71: Have you tested this?

Can we add a test to check the type / contents of what is pushed

dimberman (approved):

python-sdk/src/astro/sql/operators/append.py
sunank200: Yes we tested it on the DAG

pankajastro: yes, 
Screenshot 2022-10-14 at 8 03 07 PM

sunank200: We also checked the type and it is `str`

pankajastro (approved):

python-sdk/src/astro/sql/operators/append.py
kaxil: sounds good

kaxil: We should also a add a unit-test in a follow-up PR next week

PR #1068, opened 2022-10-14
Add open lineage support for MergeOperator

sunank200:
# Description

Open Lineage support for `MergeOperator` is missing.

closes: #900 

## What is the new behavior?

- Add open Lineage support for `MergeOperator`
- Add the test for the same

No

### Checklist


python-sdk/src/astro/sql/operators/merge.py
```diff
@@ -73,8 +83,62 @@ def execute(self, context: Context) -> BaseTable:
...
+                    ),
+                    "output_stats": OutputStatisticsOutputDatasetFacet(rowCount=self.target_table.row_count),
+                },
+            )
+        ]
+
+        run_facets: dict[str, BaseFacet] = {}
+
+        merge_query = task_instance.xcom_pull(task_ids=task_instance.task_id, key="merge_query")
+        job_facets: dict[str, BaseFacet] = {"sql": SqlJobFacet(query=str(merge_query))}
```
pankajastro, line 136: @sunank200 do we need typecast `str` since we are pushing str in xcom?

pankajastro (approved):

pankajastro:
LGTM

python-sdk/src/astro/sql/operators/merge.py
sunank200: I can remove this

pankajastro:
@sunank200 to fix the test look like you will have to pass context having ti here https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/sql/operators/test_merge.py#L297 something like https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/sql/operators/test_append.py#L81

sunank200:
> Add context to the task in test

yup https://github.com/astronomer/astro-sdk/pull/1068/commits/224e4de4d8fed82f8b3f9f38426c205d9c5efa57

python-sdk/src/astro/sql/operators/merge.py
```diff
@@ -73,8 +83,62 @@ def execute(self, context: Context) -> BaseTable:
             target_conflict_columns=self.target_conflict_columns,
             source_to_target_columns_map=self.columns,
         )
+        context["ti"].xcom_push(key="merge_query", value=str(db.sql))
```
uranusjr, line 86: Why is this additional `str` needed?

pankajastro: I think because sqlalchemy insert statement is not serializable

uranusjr: Ah OK that makes sense, I thought `sql` is the raw string.

sunank200: @uranusjr we wanted to typecast this to string before pushing as per discussion: https://astronomer.slack.com/archives/C03868KGF2Q/p1665754738297099

PR #1070, opened 2022-10-14
Set SQL project initialisation to also initialises Airflow db

tatiana:
Improve the project initialisation, so it also initialises an Airflow DB
Improve the project's interface so commands used after init can reference the project directories & files at easy. Example of how we can benefit from this:

```
def generate(directory: Path, dags_directory: Path) -> None:
    """
    Generate the Airflow DAG from a directory of SQL files.

    :params directory: The directory containing the raw sql files.
    :params dags_directory: The directory containing the generated DAG.
    """
    project_dir = project_dir or Path.cwd()
    project = Project.load_config()

    dag_file = generate_dag(project.workflows_directory, project.dags_directory)
    rprint("The DAG file", dag_file.resolve(), "has been successfully generated. :tada:") retrieve the configuration, by doing:
```


sql-cli/sql_cli/constants.py
```diff
@@ -0,0 +1,6 @@
+CONFIG_DIR = "config"
+CONFIG_FILENAME = "configuration.yml"
+
+DEFAULT_ENVIRONMENT = "default"
+DEFAULT_AIRFLOW_HOME = f"./.airflow/{DEFAULT_ENVIRONMENT}"
+DEFAULT_DAGS_FOLDER = "./.airflow/.dags"
```
feluelle, line 6:
```suggestion
DEFAULT_AIRFLOW_HOME = f".airflow/{DEFAULT_ENVIRONMENT}"
DEFAULT_DAGS_FOLDER = ".airflow/dags"
```

sql-cli/sql_cli/project.py
```diff
@@ -1,27 +1,44 @@
...
+
+    @property
+    def airflow_home(self):
+        """
+        Where the Airflow DAGs are placed. Can be either user-defined, during initialisation, or the default one.
+        """
+        return self._airflow_home or Path(self.directory, DEFAULT_AIRFLOW_HOME)
+
+    @property
+    def airflow_dags_folder(self):
```
feluelle, line 40: Could you add a docstring here, please? :)

sql-cli/sql_cli/project.py
```diff
@@ -41,10 +58,15 @@ def _update_config(self) -> None:
...
         """
         config = Config(environment=DEFAULT_ENVIRONMENT, project_dir=self.directory)
-        if self.airflow_home is not None:
-            config.write_value_to_yaml("airflow", "home", str(self.airflow_home))
-        if self.airflow_dags_folder is not None:
-            config.write_value_to_yaml("airflow", "dags_folder", str(self.airflow_dags_folder))
+        if self._airflow_home is not None:
+            config.write_value_to_yaml("airflow", "home", str(self._airflow_home))
+        if self._airflow_dags_folder is not None:
+            config.write_value_to_yaml("airflow", "dags_folder", str(self._airflow_dags_folder))
```
feluelle, line 64:
```suggestion
        if self._airflow_home:
            config.write_value_to_yaml("airflow", "home", self._airflow_home.as_posix())
        if self._airflow_dags_folder:
            config.write_value_to_yaml("airflow", "dags_folder", self._airflow_dags_folder.as_posix())
```

sql-cli/sql_cli/project.py
```diff
@@ -41,10 +58,15 @@ def _update_config(self) -> None:
...
+
+    def _initialise_airflow(self) -> None:
+        cmd = f'PYTHONWARNINGS="ignore" AIRFLOW__CORE__LOAD_EXAMPLES=False AIRFLOW_HOME={self.airflow_home} airflow db init'  # noqa: E501
+        os.system(cmd)
+        # replace by subprocess.run
```
feluelle, line 69: Because we have airflow as a python dependency we could also just call the [python interface](https://github.com/apache/airflow/blob/main/airflow/cli/commands/db_command.py#L34-L38) for initialising the db. I think this is cleaner.

sql-cli/sql_cli/project.py
```diff
@@ -56,3 +78,15 @@ def initialise(self) -> None:
...
         self._update_config()
+        self._initialise_airflow()
+
+    def is_valid_project(self) -> bool:
+        existing_paths = {path.relative_to(self.directory) for path in Path(self.directory).rglob("*")}
+        return MANDATORY_PATHS.issubset(existing_paths)
+
+    def load_config(self, environment: Optional[str] = DEFAULT_ENVIRONMENT):
+        if not self.is_valid_project():
+            logging.error("This is not a valid SQL project. Please, use `flow init`")
```
feluelle, line 89: Shouldn't we raise an exception instead?

sql-cli/sql_cli/project.py
uranusjr: Note that Airflow’s Python interface has a ton of side effects (we’re working on reducing them but it’s still not all the way there). It’s probably still a good idea to use it here for now to see how far we can go.

feluelle (approved):

PR #1072, opened 2022-10-14
Fix sql-cli examples

feluelle:
# Description

The current examples didn't work end-to-end e.g. via `flow run`, because the `conn_id` was missing.

## What is the new behavior?

- add sqlite_conn
- specify exact paths for test_project tests
- ignore .gitkeep files in copytree
- use tmp_path fixture instead of tempfile.TemporaryDirectory

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/project.py
+++ b/sql-cli/sql_cli/project.py
@@ -23,16 +23,6 @@ class Project:
         self.airflow_home = airflow_home
         self.airflow_dags_folder = airflow_dags_folder
 
-    def _delete_temporary_files(self) -> None:
-        """
-        Delete files which should not be part of the user project directory structure.
-        An example are the .gitkeep files, which were added just so the default folder structure
-        can be added to Git (by default empty directories are not versioned).
-        """
-        gitkeep_files = Path(self.directory).rglob(".gitkeep")
-        for file_ in gitkeep_files:
-            file_.unlink()
-
     def _update_config(self) -> None:
         """
         Sets custom Airflow configuration in case the user is not using the default values.
@@ -53,6 +43,10 @@ class Project:
         :param airflow_home: Custom user-defined Airflow Home directory
         :param airflow_dags_folder: Custom user-defined Airflow DAGs folder
         """
-        shutil.copytree(src=BASE_SOURCE_DIR, dst=self.directory, dirs_exist_ok=True)
-        self._delete_temporary_files()
+        shutil.copytree(
+            src=BASE_SOURCE_DIR,
+            dst=self.directory,
+            ignore=shutil.ignore_patterns(".gitkeep"),
+            dirs_exist_ok=True,
+        )
         self._update_config()
--- a/sql-cli/tests/test_project.py
+++ b/sql-cli/tests/test_project.py
@@ -1,29 +1,45 @@
-import tempfile
+from pathlib import Path
+from tempfile import TemporaryDirectory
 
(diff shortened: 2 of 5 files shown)


sql-cli/include/base/config/default/configuration.yml
```diff
@@ -3,7 +3,7 @@ connections:
 ## Example of SQLite connection
   - conn_id: sqlite_conn
     conn_type: sqlite
-    host: ../../data/imdb.db
+    host: ../../data/movies.db
```
tatiana, line 6: Since a Sqlite file represents a database and can contain multiple tables, I don't think the database file should match the table name used. An alternative could be to have the SQL query referencing `imdb.movies` (SQL Lite database . SQLite tale)

sql-cli/include/base/config/dev/configuration.yml
```diff
@@ -3,7 +3,7 @@ connections:
 ## Example of SQLite connection
   - conn_id: sqlite_conn
     conn_type: sqlite
-    host: $PWD/data/retail.db
+    host: $PWD/data/dev_movies.db
```
tatiana, line 6: :+1: I'm happy to drop the retail example for now - we can always use it in future if we need it

sql-cli/sql_cli/project.py
```diff
@@ -53,6 +43,10 @@ def initialise(self) -> None:
...
         :param airflow_dags_folder: Custom user-defined Airflow DAGs folder
         """
-        shutil.copytree(src=BASE_SOURCE_DIR, dst=self.directory, dirs_exist_ok=True)
-        self._delete_temporary_files()
+        shutil.copytree(
+            src=BASE_SOURCE_DIR,
+            dst=self.directory,
+            ignore=shutil.ignore_patterns(".gitkeep"),
+            dirs_exist_ok=True,
+        )
```
tatiana, line 51: Looks much better!

sql-cli/tests/test_project.py
```diff
@@ -1,29 +1,22 @@
-import tempfile
+from tempfile import TemporaryDirectory
```
tatiana, line 1: This feels like a matter of style

sql-cli/tests/test_project.py
```diff
@@ -1,29 +1,22 @@
...
 
 
 def test_initialise_project_with_dirname():
-    with tempfile.TemporaryDirectory() as dir_name:
-        assert not list_dir(dir_name)
+    with TemporaryDirectory() as dir_name:
         Project(dir_name).initialise()
-        new_dir_files_list = list_dir(dir_name)
-        assert new_dir_files_list == BASE_PATHS
-        assert len(new_dir_files_list) == BASE_PATHS_COUNT
```
tatiana, line 16: I would rather we continue checking the number of files as well - since there is a risk someone may update the `BASE_SOURCE_DIR` to an empty directory

sql-cli/tests/test_project.py
feluelle: Yes. Are you okay with that change?

tatiana: It feels unnecessary, but I'm okay :)

tatiana (approved):

sql-cli/tests/test_project.py
feluelle: I found an even better way then using the `tempfile.TemporaryDirectory` see 5ee08b7 :)

Issue #1073, opened 2022-10-17
Row count implementation for table missing


sunank200:
**Describe the bug**
Row count for `BaseTable` class is missing.

**Expected behavior**
Add the implementation for row count for `BaseTable`.

PR #1074, opened 2022-10-17
Add `row_count` on Table

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1073

Add row_count property in BaseTable class. 
Fixed the extractor tests accordingly 

We do not have a way to get table row count

## What is the new behavior?
Add row_count property in BaseTable class

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -129,10 +129,15 @@ class BaseTable:
     @property
     def row_count(self) -> Any:
         """
-        Return the row count of table
+        Return the row count of table.
         """
-        # TODO: Implement this property
-        return 0
+        try:
+            db = create_database(self.conn_id)
+            return db.run_sql(
+                f"SELECT COUNT(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
+            ).scalar()
+        except Exception:
+            return 0
 
     def to_json(self):
         return {
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -5,7 +5,7 @@ import string
 from typing import Any
 
 from attr import define, field, fields_dict
-from sqlalchemy import Column, MetaData
+from sqlalchemy import Column, MetaData, select, func
 
 from astro.airflow.datasets import Dataset
 from astro.databases import create_database
@@ -131,8 +131,6 @@ class BaseTable:
         """
         Return the row count of table.
         """
-        from sqlalchemy import select, func
-
         db = create_database(self.conn_id)
         tb = db.get_sqla_table(table=self)
         query = select(func.count("*")).select_from(tb)
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -131,13 +131,12 @@ class BaseTable:
         """
         Return the row count of table.
         """
-        try:
-            db = create_database(self.conn_id)
-            return db.run_sql(
-                f"SELECT COUNT(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
-            ).scalar()
-        except Exception:
-            return 0
+        from sqlalchemy import select, func
+
+        db = create_database(self.conn_id)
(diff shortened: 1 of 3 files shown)


python-sdk/src/astro/table.py
```diff
@@ -129,10 +129,15 @@ def name(self, value: str) -> None:
...
         """
-        # TODO: Implement this property
-        return 0
+        try:
+            db = create_database(self.conn_id)
+            return db.run_sql(
+                f"SELECT COUNT(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
+            ).scalar()
+        except Exception:
+            return 0
```
uranusjr, line 140: Not sure if it’s a good idea to blanket-catch and return 0 here; why not let the exception bubble up?

pankajastro: @uranusjr If I do Table(...) then it just creates metadata and not the actual table in the source database because of that if I do Table(...).row_count it would complain table does not exist but yeah agree it is not a good idea to ignore the exception.

uranusjr: It should be possible to do some reflection to check whether the table exists. It is not unreasonable to simply raise an exception in this case either.

python-sdk/src/astro/table.py
```diff
@@ -129,10 +129,14 @@ def name(self, value: str) -> None:
...
         """
-        Return the row count of table
+        Return the row count of table.
         """
-        # TODO: Implement this property
-        return 0
+        from sqlalchemy import func, select
+
+        db = create_database(self.conn_id)
+        tb = db.get_sqla_table(table=self)
```
utkarsharma2, line 137: You can avoid this step. Directly create the query as "SELECT count(*) from table".

pankajastro: query strings are a bit vulnerable to SQL injection

python-sdk/src/astro/table.py
```diff
@@ -129,10 +129,14 @@ def name(self, value: str) -> None:
     @property
     def row_count(self) -> Any:
         """
```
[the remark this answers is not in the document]
pankajastro: Added a integration test 2f226812e9802fac07b10bf751d5df58812fcabd

python-sdk/src/astro/table.py
utkarsharma2: But there won't be any input from the user that can manipulate the query string, right? This will be a hard-coded string.

pankajastro: right, but still deepsource will complain and will have to using `# skipcq: BAN-B608`

python-sdk/src/astro/table.py
pankajastro: removed try-catch block

python-sdk/src/astro/table.py
```diff
@@ -129,10 +129,14 @@ def name(self, value: str) -> None:
...
-        Return the row count of table
+        Return the row count of table.
         """
-        # TODO: Implement this property
-        return 0
+        from sqlalchemy import func, select
```
sunank200, line 134: Should this import statement be on the top of file?

python-sdk/tests/utils/test_table.py
```diff
@@ -89,3 +91,16 @@ def test_find_first_table(kwargs, return_type):
 @mock.patch("airflow.models.xcom_arg.PlainXComArg.resolve", return_value=Table())
 def test_find_first_table_with_xcom_arg(xcom_arg_resolve, kwargs, return_type):
     assert isinstance(find_first_table(context={}, **kwargs), return_type)
+
+
+@pytest.mark.integration
+def test_row_count():
+    imdb_table = LoadFileOperator(
+        task_id="load_file",
```
sunank200, line 99: we should add a doc-string

sunank200 (commented):
Overall LGTM. Can we resolve the comments, please?

sunank200 (approved):

PR #1075, opened 2022-10-17
Integrate project with generate and run

feluelle:
# Description

We need to integrate the project init with the cli generate & run commands.

## What is the new behavior?

We integrate the project with the cli generate & run commands by passing all the necessary paths to the commands.

- fixes some issues in project init (see comments below)

No.

### Checklist


sql-cli/conftest.py
```diff
@@ -21,6 +21,16 @@
...
 
+@pytest.fixture()
+def project_directory():
+    return CWD / "tests" / "example_project"
+
+
+@pytest.fixture()
+def workflow_directory():
+    return CWD / "tests" / "example_project" / "workflows" / "example_basic_transform"
+
```
tatiana, line 32: If we used temporary directories, it would be easier to parallelise tests

sql-cli/sql_cli/__main__.py
```diff
@@ -37,20 +38,19 @@ def about() -> None:
 
 @app.command(help="Generate the Airflow DAG from a directory of SQL files.")
 def generate(
-    directory: Path = typer.Argument(
+    workflow_directory: Path = typer.Argument(
```
tatiana, line 41: The idea was that this would be `workflow_name` (only the name of the directory), and that, together with the `project_dir` we can build the path to it.

sql-cli/sql_cli/__main__.py
```diff
@@ -37,20 +38,19 @@ def about() -> None:
...
         default=...,
         show_default=False,
         exists=True,
         help="directory containing the sql files.",
     ),
-    dags_directory: Path = typer.Argument(
```
tatiana, line 46: I believe, from the generate perspective, it is better to have the `dags_directory` - since Airflow users can use it as a stand-alone command and generate the DAG anywhere, they'd like - without the need to be respecting the SQL CLI project structure for output.

sql-cli/sql_cli/__main__.py
```diff
@@ -91,18 +91,12 @@ def validate(
     """
 )
 def run(
-    directory: Path = typer.Argument(
+    workflow_directory: Path = typer.Argument(
```
tatiana, line 94: Same here. I'd expect this to be `workflow_name` - and the name would match the directory of interest.

sql-cli/sql_cli/__main__.py
```diff
@@ -91,18 +91,12 @@ def validate(
...
         exists=True,
         help="directory containing the sql files.",
     ),
-    dags_directory: Path = typer.Argument(
-        default=...,
-        show_default=False,
-        exists=True,
-        help="directory for the generated DAGs.",
-    ),
     connection_file: Path = typer.Option(
```
tatiana, line 100: We should be consistent with the other commands, which use `environment`, and we retrieve the connection based on the environment path containing configurations.

sql-cli/sql_cli/__main__.py
```diff
@@ -113,9 +107,14 @@ def run(
         exists=True,
         help="path to variables yaml or json file",
     ),
```
tatiana, line 109: I believe we shouldn't expose this in 0.1 - it doesn't seem to fit our 0.1 user journey.

sql-cli/sql_cli/project.py
```diff
@@ -124,7 +127,9 @@ def load_config(self, environment: str = DEFAULT_ENVIRONMENT) -> None:
         :param environment: string referencing the desired environment, uses "default" unless specified
         """
         if not self.is_valid_project():
-            logging.exception("This is not a valid SQL project. Please, use `flow init`")
+            raise Exception("This is not a valid SQL project. Please, use `flow init`")
```
tatiana, line 130: We should probably raise a custom exception

sql-cli/sql_cli/__main__.py
feluelle: I have removed it for now. I think we can re-add it later. I would not add to many arguments in the beginning especially because this as you said would only be useful to airflow users - not necessarily sql-cli users.

tatiana:
The work developed here will be merged as part of #1082

PR #1077, opened 2022-10-17
Integrate SQL CLI project with validate

tatiana:
Aligns validate with the init project structure and arguments structure.
Fixes initialisation issues (e.g. check if a project is valid, the path of the default datasets)

This is in the draft because I'm fixing tests.


sql-cli/sql_cli/__main__.py
```diff
@@ -55,19 +55,29 @@ def generate(
...
 
 @app.command(
-    help="Validate Airflow connection(s) provided in the configuration file for the given environment."
+    help="""
+    Validate Airflow connection(s) provided in the configuration file for the given environment.
+    """
 )
 def validate(
-    environment: str = typer.Argument(
+    project_dir: Optional[Path] = typer.Argument(
```
dimberman, line 63:
```suggestion
    project_dir: Optional[Path] = typer.Option(
```

I think if we're going to make this optional it should be a named option instead of a an Argument that is dependent on order

sql-cli/sql_cli/connections.py
```diff
@@ -40,15 +26,25 @@ def _create_or_replace_connection(conn_obj: Connection) -> None:
...
     config_file_contains_connection = False
-    connections = _load_yaml_connections(environment)
+
+    airflow_meta_conn = retrieve_airflow_meta_database_conn(project.directory / project.airflow_home)
+    os.environ["AIRFLOW__DATABASE__SQL_ALCHEMY_CONN"] = airflow_meta_conn
+
+    importlib.reload(airflow)
+    importlib.reload(airflow.configuration)
+    importlib.reload(airflow.models.base)
+    importlib.reload(airflow.models.connection)
```
dimberman, line 44: We should add documentation around this as it seems somewhat hacky. Also do we need to put this in its own function? Will this be required for other functions like run?

dimberman (changes requested):

sql-cli/sql_cli/__main__.py
tatiana: I have mixed feelings about this one - as it is, we are consistent with `flow init`.

sql-cli/sql_cli/connections.py
tatiana: Done, @dimberman ! I'm not sure if run will need this, but I refactored and isolated this, trying to add some comments.

tatiana:
The work developed here will be merged as part of #1082

Issue #1080, opened 2022-10-18
File size is returning -1 instead of file size


sunank200:
**Describe the bug**
Currently the file size is returning -1 in `File` class instead of actual file size.

**Expected behavior**
Return actual file size

PR #1081, opened 2022-10-18
Fix file size property

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1080
Fixed `size` property on location objects.

The `size` property on location objects return -1 except for the local location type

## What is the new behavior?
Now, the `size` property on location objects will return the actual file size

### Checklist

### DIFF
--- a/python-sdk/src/astro/files/locations/amazon/s3.py
+++ b/python-sdk/src/astro/files/locations/amazon/s3.py
@@ -43,7 +43,9 @@ class S3Location(BaseFileLocation):
 
     @property
     def size(self) -> int:
-        return -1
+        url = urlparse(self.path)
+        bucket_name = url.netloc
+        return self.hook.head_object(key=self.path, bucket_name=bucket_name).get("ContentLength") or -1
 
     @property
     def openlineage_dataset_namespace(self) -> str:
--- a/python-sdk/src/astro/files/locations/amazon/s3.py
+++ b/python-sdk/src/astro/files/locations/amazon/s3.py
@@ -43,6 +43,7 @@ class S3Location(BaseFileLocation):
 
     @property
     def size(self) -> int:
+        """Return file size for S3 location"""
         url = urlparse(self.path)
         bucket_name = url.netloc
         return self.hook.head_object(key=self.path, bucket_name=bucket_name).get("ContentLength") or -1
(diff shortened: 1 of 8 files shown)


python-sdk/src/astro/files/locations/google/gcs.py
```diff
@@ -35,7 +35,12 @@ def paths(self) -> list[str]:
 
     @property
     def size(self) -> int:
-        return -1
+        url = urlparse(self.path)
+        bucket_name = url.netloc
+        object_name = url.path
+        if object_name[0] == "/":
```
sunank200: What would happen when path is like `gs://bucket/*` ? We should add a test for this

python-sdk/src/astro/files/locations/http.py
```diff
@@ -18,7 +19,10 @@ def paths(self) -> list[str]:
 
     @property
     def size(self) -> int:
-        return -1
+        print("self.pathself.pathself.pathself.path", self.path)
```
sunank200: Print statement should be removed

python-sdk/src/astro/files/locations/amazon/s3.py
```diff
@@ -43,7 +43,9 @@ def paths(self) -> list[str]:
 
     @property
     def size(self) -> int:
-        return -1
+        url = urlparse(self.path)
+        bucket_name = url.netloc
```
sunank200, line 48: We should add a test for `s3://bucket/*`

sunank200 (commented):
Added a few comments

python-sdk/src/astro/files/locations/http.py
pankajastro: Look like you looking at an old commit

python-sdk/src/astro/files/locations/google/gcs.py
pankajastro: if you pass `gs://bucket/*` it would probably not work. 
because GCS hook `get_size` uses `get_blob` which expects blob name i.e object name and `*` is not the correct object name here
https://github.com/apache/airflow/blob/main/airflow/providers/google/cloud/hooks/gcs.py#L776
https://googleapis.dev/python/storage/latest/buckets.html

pankajastro: infact even our paths property on location  will return an empty list

PR #1091, opened 2022-10-19
Use Intersphinx to avoid direct link to Airflow docs

uranusjr:
Currently the documentation links to Airflow documentation pages directly; this uses [Intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) instead as a more robust way to manage the references.

There are probably other links that can be converted; they can be converted incrementally when encountered—I only changed the most obvious ones.

### DIFF
--- a/python-sdk/docs/astro/files/operators/get_file_list.rst
+++ b/python-sdk/docs/astro/files/operators/get_file_list.rst
@@ -23,5 +23,5 @@ The following example retrieves a file list from the GCS bucket and dynamically 
 Related references
 ~~~~~~~~~~~~~~~~~~
 
-- `Dynamic task mapping - Apache Airflow <https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html>`_
+- :external+airflow:doc:`Dynamic task mapping - Apache Airflow <concepts/dynamic-task-mapping>`
 - `Dynamic tasks - Astronomer <https://www.astronomer.io/guides/dynamic-tasks/>`_
--- a/python-sdk/docs/conf.py
+++ b/python-sdk/docs/conf.py
@@ -37,7 +37,7 @@ release = __version__
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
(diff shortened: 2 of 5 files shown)


python-sdk/docs/astro/files/operators/get_file_list.rst
```diff
@@ -23,5 +23,5 @@ The following example retrieves a file list from the GCS bucket and dynamically
 Related references
 ~~~~~~~~~~~~~~~~~~
 
-- `Dynamic task mapping - Apache Airflow <https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html>`_
+- :external+airflow:doc:`Dynamic task mapping - Apache Airflow <concepts/dynamic-task-mapping>`
```
kaxil, line 26: TIL :)

python-sdk/docs/conf.py
```diff
@@ -63,6 +63,12 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", "*.txt"]
 
+# Mapping to inter-sphinx linking.
+intersphinx_mapping = {
+    "airflow": ("https://airflow.apache.org/docs/apache-airflow/stable", None),
+    "airflow-postgres": ("https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable", None),
+}
```
kaxil, line 70: On Providers I had added the following to register intersphinx for all deps:

https://github.com/astronomer/astronomer-providers/blob/5ab58aa67f0e1549e14c802e0b2ad7ac04cc9ffb/docs/conf.py#L141-L155


```python
config = configparser.ConfigParser(strict=False)
config.read(_REPO_DIR / "setup.cfg")

_PROV_DEPS = [
    re.match(r"([a-zA-Z-]+)", dep).groups()[0]
    for dep in config["options.extras_require"]["all"].split()
    if dep.startswith("apache-airflow-providers-")
]
intersphinx_mapping = {
    "airflow": ("https://airflow.apache.org/docs/apache-airflow/stable/", None),
    **{provider: (f"https://airflow.apache.org/docs/{provider}/stable", None) for provider in _PROV_DEPS},
}
```

kaxil (approved):

Issue #1107, opened 2022-10-19
Run should have a clean and direct message if the table the SQL statement is trying to access isn't available


tatiana:
**Describe the bug**
At the moment, if the user tries to run a DAG, but the database they are trying to connect to does not have the desired table, they generate almost 100 lines of traceback. We should print a more graceful message to guide the user on the problem.

**Version**
* SQL CLI: 0.0.9

**To Reproduce**
```
$ mkdir buggy

$ cd buggy

$ flow init

$ flow run example_templating

Running the workflow example_templating

Processing filtered_orders... ╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/tati/.virtualenvs/sql-cli-0.0.9/lib/python3.8/site-packages/sqlalchemy/engine/base.py:1802 │
│ in _execute_context                                                                              │
│                                                                                                  │
│   1799 │   │   │   │   │   │   │   evt_handled = True                                            │
│   1800 │   │   │   │   │   │   │   break                                                         │
│   1801 │   │   │   │   if not evt_handled:                                                       │
│ ❱ 1802 │   │   │   │   │   self.dialect.do_execute(                                              │
│   1803 │   │   │   │   │   │   cursor, statement, parameters, context                            │
│   1804 │   │   │   │   │   )                                                                     │
│   1805                                                                                           │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /home/tati/.virtualenvs/sql-cli-0.0.9/lib/python3.8/site-packages/astro/databases/base.py:118 in │
│ run_sql                                                                                          │
│                                                                                                  │
│   115 │   │   │   sql = kwargs.get("sql_statement")  # type: ignore                              │
│   116 │   │                                                                                      │
│   117 │   │   if isinstance(sql, str):                                                           │
│ ❱ 118 │   │   │   result = self.connection.execute(sqlalchemy.text(sql), parameters)             │
│   119 │   │   else:                                                                              │
│   120 │   │   │   result = self.connection.execute(sql, parameters)                              │
│   121 │   │   return result                                                                      │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │     kwargs = {}                                                                              │ │
│ │ parameters = {}                                                                              │ │
[...]

PR #1111, opened 2022-10-20
Print custom error message for failed runs caused by Operational errors

feluelle:
# Description

Currently, we don't capture Operational database errors. For more details see the issue below.

closes: #1107

## What is the new behavior?

We print a custom error message for failed runs caused by Operational errors.

- change logger level to CRITICAL
- move workflow print statements to cli
- use tempdir instead of /tmp
- add more tests for failed runs

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/run_dag.py
+++ b/sql-cli/sql_cli/run_dag.py
@@ -16,6 +16,7 @@ from airflow.utils.session import NEW_SESSION, provide_session
 from airflow.utils.state import DagRunState, State
 from airflow.utils.types import DagRunType
 from rich import print as pprint
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm.session import Session
 
 from astro.sql.operators.cleanup import AstroCleanupException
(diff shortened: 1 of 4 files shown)


sql-cli/sql_cli/run_dag.py
```diff
@@ -149,6 +149,10 @@ def _run_task(ti: TaskInstance, session: Session) -> None:
         session.flush()
         session.commit()
         pprint("[bold green]SUCCESS[/bold green]")
+    except OperationalError as operational_exception:
+        pprint("[bold red]FAILED[/bold red]")
+        original_message = operational_exception.orig.args[0]
+        pprint(f"  Error [red]{original_message}[/red] using connection [bold]{ti.task.conn_id}[/bold].")
```
feluelle, line 155: @tatiana I think it is better to print the original message and adding a bit more information to it - rather than us reading the message (string) and parsing the information out of it. If we want to do that that could cause a massive overhead we would have to maintain.

feluelle: Screenshot 2022-10-20 at 11 17 20

WDYT?

tatiana: Sounds great, @feluelle , as long as we avoid massive Python tracebacks and keep these messages small and straightforward, it is great!

tatiana (approved):

PR #1113, opened 2022-10-20
Change version to 1.3.0dev

sunank200:
# Description

Version in main is 1.2.0

## What is the new behavior?

- Change version to 1.3.0dev

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/__init__.py
+++ b/python-sdk/src/astro/__init__.py
@@ -1,6 +1,6 @@
 """A decorator that allows users to run SQL queries natively in Airflow."""
 
-__version__ = "1.3.0dev"
+__version__ = "1.3.0.dev1"
 
 
 # This is needed to allow Airflow to pick up specific metadata fields it needs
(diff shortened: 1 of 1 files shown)


pankajastro (approved):

python-sdk/src/astro/__init__.py
```diff
@@ -1,6 +1,6 @@
 """A decorator that allows users to run SQL queries natively in Airflow."""
 
-__version__ = "1.2.0"
+__version__ = "1.3.0dev"
```
pankajastro, line 3: it should be 1.3.0 or 1.2.1?

kaxil: 1.3.0 -- Main should always point to the next minor

kaxil, line 3:
```suggestion
__version__ = "1.3.0.dev1"
```

kaxil, line 3:
```suggestion
__version__ = "1.3.0.dev1"
```

kaxil (approved):

Issue #1034, opened 2022-10-10
Transform op - database is not picked from metadata


utkarsharma2:
**Describe the bug**

```
with DAG(dag_id="test", schedule=None, start_date=datetime(1970, 1, 1)) as dag:
    @aql.transform
    def select(input_table: Table):
        return "SELECT * FROM my_db.my_schema.my_table LIMIT 4;"  # {{ input_table }} was also failing
    select(
        input_table=Table(metadata={"schema": "my_schema", "database": "db"}, name="my_table", conn_id="my_conn_id"),
        output_table=Table(name="fritz_test_1234")
    )
```
`snowflake.connector.errors.ProgrammingError: 090105 (22000): 01a77af0-0605-96b0-0000-682112f6e8ae: Cannot perform SELECT. This session does not have a current database. Call 'USE DATABASE', or use a qualified name.
`
ref: https://astronomer.slack.com/archives/C02B8SPT93K/p1665414029917169

**Version**
* Astro: 1.1.1

**To Reproduce**
Try the above code. 

**Expected behavior**
If the DB is not specified in the connection we should be able to pick from metadata.

pankajastro:
This looks like a problem when using snowflake. I quickly tested the  bigquery worked fine

pankajastro:
For the hook creation in astro-sdk we tightly depends the airflow connection 
for example
```
 @property
    def hook(self) -> BigQueryHook:
        """Retrieve Airflow hook to interface with the BigQuery database."""
        return BigQueryHook(gcp_conn_id=self.conn_id, use_legacy_sql=False)
```
Snowflake hook expects the database, schema, etc., so the ideal place to set these values is a connection only. Even in our case we do
```
 @property
    def hook(self) -> SnowflakeHook:
        """Retrieve Airflow hook to interface with the Snowflake database."""
        return SnowflakeHook(snowflake_conn_id=self.conn_id)
```
If a user is setting up the schema, and database in the input and output table in the transform operator which value should we use from input_table or output_table or will we create multiple hooks? it looks a little bit complicated to me in the current design. 

For now, I have modified the `create_database` method to accept conn_id as well as a table and I'm passing the output table from the base database class and setting the schema and database in the snowflake hook if the table metadata has these values.
I'll if fix the CI if people will agree with the approach. PR https://github.com/astronomer/astro-sdk/pull/1117

PR #1117, opened 2022-10-20
Transform op - database is not picked from metadata

pankajastro:
# Description
closes: #1034 
Add table param in ```create_database``` function so that can access and use table metadata while creating a hook

Currently, we don't use Table metadata param schema and database while creating so if these values is not set in the airflow connection we are getting an error in the case of snowflake

## What is the new behavior?
change 

```
def create_database(conn_id: str) -> BaseDatabase:
    ...
```
to
```
def create_database(
    conn_id: str,
    table: object | None = None,
) -> BaseDatabase: 
    ...
```
and use table metadata if available while creating snowflake hook

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/__init__.py
+++ b/python-sdk/src/astro/databases/__init__.py
@@ -8,6 +8,7 @@ from astro.utils.path import get_class_name, get_dict_with_module_names_to_dot_n
 
 if TYPE_CHECKING:
     from astro.databases.base import BaseDatabase
+    from astro.table import BaseTable
 
 DEFAULT_CONN_TYPE_TO_MODULE_PATH = get_dict_with_module_names_to_dot_notations(Path(__file__))
 CUSTOM_CONN_TYPE_TO_MODULE_PATH = {
@@ -21,7 +22,7 @@ CONN_TYPE_TO_MODULE_PATH = {
 SUPPORTED_DATABASES = set(DEFAULT_CONN_TYPE_TO_MODULE_PATH.keys())
 
 
-def create_database(conn_id: str, table: object | None = None, ) -> "BaseDatabase":
+def create_database(conn_id: str, table: BaseTable | None = None, ) -> "BaseDatabase":
     """
     Given a conn_id, return the associated Database class.
 
--- a/python-sdk/src/astro/databases/__init__.py
+++ b/python-sdk/src/astro/databases/__init__.py
@@ -22,7 +22,10 @@ CONN_TYPE_TO_MODULE_PATH = {
 SUPPORTED_DATABASES = set(DEFAULT_CONN_TYPE_TO_MODULE_PATH.keys())
 
 
-def create_database(conn_id: str, table: BaseTable | None = None, ) -> "BaseDatabase":
+def create_database(
+    conn_id: str,
+    table: BaseTable | None = None,
+) -> BaseDatabase:
     """
     Given a conn_id, return the associated Database class.
 
@@ -35,5 +38,5 @@ def create_database(conn_id: str, table: BaseTable | None = None, ) -> "BaseData
     module_path = CONN_TYPE_TO_MODULE_PATH[conn_type]
     module = importlib.import_module(module_path)
     class_name = get_class_name(module_ref=module, suffix="Database")
-    database: "BaseDatabase" = getattr(module, class_name)(conn_id, table)
+    database: BaseDatabase = getattr(module, class_name)(conn_id, table)
     return database
--- a/python-sdk/src/astro/databases/__init__.py
+++ b/python-sdk/src/astro/databases/__init__.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING
 
 from astro.utils.path import get_class_name, get_dict_with_module_names_to_dot_notations
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     from astro.databases.base import BaseDatabase
     from astro.table import BaseTable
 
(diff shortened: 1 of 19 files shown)


python-sdk/src/astro/databases/__init__.py
```diff
@@ -19,17 +21,21 @@
 SUPPORTED_DATABASES = set(DEFAULT_CONN_TYPE_TO_MODULE_PATH.keys())
 
 
-def create_database(conn_id: str) -> "BaseDatabase":
+def create_database(
+    conn_id: str,
+    table: object | None = None,
+) -> BaseDatabase:
```
kaxil, line 27: Can you verify that this does not introduce cyclic dependency or increase import time that we fixed in https://github.com/astronomer/astro-sdk/pull/1013

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -248,13 +248,17 @@ class SnowflakeDatabase(BaseDatabase):
...
+        self.table = table
 
     @property
     def hook(self) -> SnowflakeHook:
         """Retrieve Airflow hook to interface with the snowflake database."""
-        return SnowflakeHook(snowflake_conn_id=self.conn_id)
+        kwargs = {}
+        if self.table and self.table.metadata:
+            kwargs = {"schema": self.table.metadata.schema, "database": self.table.metadata.database}
+        return SnowflakeHook(snowflake_conn_id=self.conn_id, **kwargs)
```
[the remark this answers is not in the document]
pankajastro: I believe It would require if metadata depend on conn too so I feel we need similar change for Redshift and Postgres, I'll do those once we feel this approach is right

pankajastro:
I have posted some findings on https://github.com/astronomer/astro-sdk/issues/1034 also posting here for better visibility

python-sdk/src/astro/databases/__init__.py
kaxil: you can use one of the following for it:

- https://docs.python.org/3/library/profile.html
- https://github.com/nschloe/tuna
- https://github.com/ymichael/cprofilev

pankajastro: I tried to get the import time for `astro-sdk/python-sdk/src/astro/databases/__init__.py`

Screenshot 2022-10-26 at 2 26 30 PM

dimberman (approved):

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -68,12 +70,13 @@ class RedshiftDatabase(BaseDatabase):
         FileLocation.S3: "load_s3_file_to_table",
     }
 
-    illegal_column_name_chars: List[str] = ["."]
-    illegal_column_name_chars_replacement: List[str] = ["_"]
+    illegal_column_name_chars: list[str] = ["."]
+    illegal_column_name_chars_replacement: list[str] = ["_"]
 
-    def __init__(self, conn_id: str = DEFAULT_CONN_ID):
+    def __init__(self, conn_id: str = DEFAULT_CONN_ID, table: BaseTable | None = None):
```
utkarsharma2, line 76: @pankajastro Just a thought, I think the database is a superset and the table is a subset. For example - One database can be associated with multiple tables, but one table can only be associated with one DB. Based on this train of thought, it feels odd that we are adding tables as part of database creation. I feel this is tight coupling. 

If we need a table for hook creation can we pass the table as a param to self.hook instead? WDYT?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -102,8 +102,9 @@ class BigqueryDatabase(BaseDatabase):
         DatabaseCustomError,
     )
 
-    def __init__(self, conn_id: str = DEFAULT_CONN_ID):
+    def __init__(self, conn_id: str = DEFAULT_CONN_ID, table: BaseTable | None = None):
         super().__init__(conn_id)
+        self.table = table
```
utkarsharma2, line 107: @pankajastro  Don't we need a similar self.hook changes for bigquery?

utkarsharma2:
@pankajastro I think we should also define a priority when picking up values for metadata. 
P0 - Table object
P1 - Connection id
P2 - Settings(Airflow config ex - schema postgres_default_schema)  
P3 - Database defaults

pankajastro:
Postgres-related tests are failing because ```database_table_fixture``` create below object which has schema ```astroflow_ci``` but in our test env it does not exist.

```database_table_fixture = (PostgresDatabase(conn_id="postgres_conn), TempTable(_name='', conn_id='postgres_conn', metadata=Metadata(schema='astroflow_ci', database=None), columns=[], temp=True))```

python-sdk/src/astro/databases/aws/redshift.py
pankajastro: Yeah, I agree, but it is a bit difficult to handle it in the current design, I  posted https://github.com/astronomer/astro-sdk/issues/1034#issuecomment-1285647029 some insight

python-sdk/src/astro/databases/google/bigquery.py
pankajastro: No, since bigquery hook does not expect database as kwargs

pankajastro:
> @pankajastro I think we should also define a priority when picking up values for metadata. P0 - Table object P1 - Connection id P2 - Settings(Airflow config ex - schema postgres_default_schema) P3 - Database defaults
 
currently, since this depends on airflow hook too and that is not consistent so not sure how can control this

phanikumv (approved):

PR #1119, opened 2022-10-20
Update CODEOWNERS to reflect SQL CLI work

kaxil:
This will ensure correct folks are tagged for reviews for SQL CLI


feluelle:
What about the root folder? @kaxil It now does not have any code owners I think.

kaxil:
> What about the root folder? @kaxil It now does not have any code owners I think.

Created https://github.com/astronomer/astro-sdk/pull/1132

PR #1122, opened 2022-10-20
Reduce command run times for SQL CLI

kaxil:
This PR moves the import lines to the functions that use them to delay imports. It also changes the "logging" code that sets loggers to critical. Previously since "airflow" was loaded from imports, it sets the level for it as Critical, now we explicitly set the level just for Airflow

**`flow version`**:

Before: 2.156s
Now: 0.175s

image
image

**`flow init`**:
Before: 2.108s
Now: 0.409s

image
image

**`flow validate`**:
Before: 2.342s
Now: 1.121s

image
image

**`flow run`**:
No improvement here as it needs most of the libraries

Before: 1.953s
Now: 2.036s
image
image

### DIFF
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -6,28 +6,13 @@ from dotenv import load_dotenv
 from rich import print as rprint
 
 import sql_cli
-from sql_cli.connections import validate_connections
 from sql_cli.constants import DEFAULT_AIRFLOW_HOME, DEFAULT_DAGS_FOLDER
-from sql_cli.dag_generator import generate_dag
-from sql_cli.exceptions import EmptyDag, SqlFilesDirectoryNotFound
-from sql_cli.project import Project
-from sql_cli.run_dag import run_dag
-from sql_cli.utils.airflow import (
-    get_dag,
-    retrieve_airflow_database_conn_from_config,
-    set_airflow_database_conn,
-)
 
 load_dotenv()
 app = typer.Typer(add_completion=False, context_settings={"help_option_names": ["-h", "--help"]})
 
 
-def set_logger_level(log_level: int) -> None:
-    for name in logging.root.manager.loggerDict:
-        logging.getLogger(name).setLevel(log_level)
-
-
-set_logger_level(logging.CRITICAL)
+logging.getLogger("airflow").setLevel(logging.CRITICAL)
 
 
 @app.command(help="Print the SQL CLI version.")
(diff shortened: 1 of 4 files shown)


kaxil:
Feel to update, remove, adapt any changes from the PR.

feluelle:
Hmm 🤔 not sure If I like it or not. Is it really worth it moving the imports for 1-2 seconds speed improvement? I am wondering if this a best practice for cli apps?

feluelle:
@uranusjr WDYT of this change?

kaxil:
> If we decide to go this way, we could use something like I created in airflint to [automatically move the imports](https://github.com/feluelle/airflint/blob/main/airflint/rules/use_function_level_imports.py) into functions via a pre-commit hook. This would make it as fast as we can 😬.
> 
> But It also depends on which commands and how often you call the commands, because python will cache the imports right?

Yup, if a module is already imported, it will be cached. 


>Hmm 🤔 not sure If I like it or not. Is it really worth it moving the imports for 1-2 seconds speed improvement? I am wondering if this a best practice for cli apps?

Yeah I was not sure as well but generally it is annoying if it takes a long time to load the version 😄  -- but it is more of "annoyance" rather than "problem". In Airflow, we load the CLI commands lazily to avoid it:

https://github.com/apache/airflow/blob/7be777254192f5a7cdcdd9066917217980916f0e/airflow/cli/cli_parser.py#L46-L56
https://github.com/apache/airflow/blob/7be777254192f5a7cdcdd9066917217980916f0e/airflow/cli/cli_parser.py#L1002-L1008

kaxil:
>If we decide to go this way, we could use something like I created in airflint to [automatically move the imports](https://github.com/feluelle/airflint/blob/main/airflint/rules/use_function_level_imports.py) into functions via a pre-commit hook. This would make it as fast as we can 😬.

Oohhh nice, personally I would be in favor of it

feluelle:
@tatiana WDYT? I think we can give this a try in the next release after 0.1? Together with a pre-commit hook which would move the imports automatically.

sql-cli/sql_cli/__main__.py
```diff
@@ -6,28 +6,13 @@
 from rich import print as rprint
 
 import sql_cli
-from sql_cli.connections import validate_connections
 from sql_cli.constants import DEFAULT_AIRFLOW_HOME, DEFAULT_DAGS_FOLDER
```
kaxil, line 9: I didn't move this as this module only has constants and no imported modules

dimberman (approved):
LGTM!

dimberman:
@feluelle is there such a pre-commit hook? I almost wonder if we can do this in a cleaner manner by moving the functions out of the main class? Either way I think this is fine for now and we can decide later if we want a more permanent solution.

feluelle:
@dimberman please see https://github.com/astronomer/astro-sdk/pull/1122#issuecomment-1286696737

PR #1125, opened 2022-10-21
Use env vars to create connections

kaxil:
Instead of using session and storing connections in the database, we can simply store the connections as Environment Variables.


feluelle (commented):
@kaxil good point! I am not sure how safe this is. I mean can't we run into race conditions with other sql-cli projects?

I am wondering if we need that at all. Isn't the [astro filesystem backend](https://github.com/astronomer/astro-sdk/blob/main/sql-cli/sql_cli/run_dag.py#L28) enough?

kaxil:
> @kaxil good point! I am not sure how safe this is. I mean can't we run into race conditions with other sql-cli projects?
> 
> I am wondering if we need that at all. Isn't the [astro filesystem backend](https://github.com/astronomer/astro-sdk/blob/main/sql-cli/sql_cli/run_dag.py#L28) enough?

Yeah I think [astro filesystem backend](https://github.com/astronomer/astro-sdk/blob/main/sql-cli/sql_cli/run_dag.py#L28) should be enough too. But I don't think it will conflict with any other SQL CLI project since the scope of this env vars would be the "current python session".

But will leave the final call to you :) feel free to close the PR if not needed

feluelle:
@tatiana I would consider the env approach but for 0.2 or so. WDYT?

feluelle:
Or even checking if we can directly validate the backend being used.

dimberman:
We're gonna need the filesystem for variables anyways, and it would be more consistent with how we're doing the run function. No super strong convictions on this one though I think it would be better to keep things consistent.

feluelle:
> We're gonna need the filesystem for variables anyways

We could also [use env vars](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html#storing-variables-in-environment-variables).

sql-cli/sql_cli/connections.py
```diff
@@ -31,7 +19,8 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
     for connection in connections:
         if connection.id == connection_id:
             config_file_contains_connection = True
-        _create_or_replace_connection(connection)
+        # Create connection using Environment Variable
+        os.environ[f"AIRFLOW_CONN_{connection.conn_id.upper()}"] = connection.get_uri()
```
dimberman, line 23: @kaxil maybe we can have a way to convert the yaml in the test-connections.yaml into these URIs? the test-connections.yaml is very useful it would be a bummer to lose it.

dimberman (changes requested):

feluelle:
Why did you close the PR?

kaxil:
Fixed the merge conflict :)

kaxil:
> Why did you close the PR?

Mistake :)

sql-cli/sql_cli/connections.py
kaxil: hmm, how does this impact test-connections.yaml?

feluelle:
@tatiana I think this can be a really useful feature. WDYT?

tatiana:
> @tatiana I think this can be a really useful feature. WDYT?

@feluelle, you've been investigating further how we set the Airflow configs as part of #1219
It would be great if we had a consistent way of setting our Airflow configuration - and if environment variables (internally) are the best way to be consistent, I'll for it!

sql-cli/sql_cli/connections.py
```diff
@@ -30,9 +18,10 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
     config_file_contains_connection = False
 
     for connection in connections:
-        if connection.id == connection_id:
+        if connection.conn_id == connection_id:
```
feluelle, line 21: 👀 @tatiana @pankajkoti 

..found this while testing the new env vars

sql-cli/sql_cli/connections.py
```diff
@@ -30,9 +18,10 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
...
             config_file_contains_connection = True
-        _create_or_replace_connection(connection)
+        # Create connection using Environment Variable
```
feluelle, line 23:
```suggestion
```
This is not helping in my opinion.

sql-cli/sql_cli/connections.py
tatiana: I understand that we're only using env vars internally - the file continues to be used by the end users.

tatiana (approved):
Happy for this to be merged after the tests are passed, and the branch is rebased! :)

feluelle:
Thanks @kaxil for the great idea! ❤️

Issue #1067, opened 2022-10-14
Change python-sdk dependency to non-dev before release


pankajkoti:
Currently, we have a local folder path dependency in `pyproject.toml `for `python-sdk`. We need to update it to non-dev version before the `sql-cli` release.

tatiana:
Already been addressed and merged to main - it is unpinned at the moment.

PR #1131, opened 2022-10-21
Update astro-sdk-python version requirement

feluelle:
# Description

Currently, the astro-sdk-python minimum version is a beta version which is not optimal, instead we should rely only on stable versions.

related: #1067

## What is the new behavior?

Update astro-sdk-python version requirement

- fix run_dag for airflow version prior to 2.3

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/run_dag.py
+++ b/sql-cli/sql_cli/run_dag.py
@@ -138,7 +138,7 @@ def _run_task(ti: TaskInstance, session: Session) -> None:
 
     :param ti: TaskInstance to run
     """
-    if ti.map_index >= 0:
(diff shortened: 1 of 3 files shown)


sql-cli/sql_cli/run_dag.py
```diff
@@ -138,7 +138,7 @@ def _run_task(ti: TaskInstance, session: Session) -> None:
 
     :param ti: TaskInstance to run
     """
-    if ti.map_index >= 0:
+    if hasattr(ti, "map_index") and ti.map_index >= 0:
```
feluelle, line 141: This is required for Airflow version prior to 2.3 which we currently allow i.e. `>2.0`.

pankajkoti (approved):

PR #1132, opened 2022-10-21
Add CodeOwners for root

kaxil:
The root contains pre-commit hooks, isort configs etc.

I am just adding a few folks I know would be interested in reviewing those files, we can add others if they are interested in it.

The last matching pattern takes the most precedence

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,6 @@
+# Order is important; the last matching pattern takes the most
+# precedence.
+
+*               @pankajastro @feluelle @tatiana @kaxil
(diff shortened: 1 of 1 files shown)


feluelle (approved):

CODEOWNERS
```diff
@@ -1,2 +1,6 @@
+# Order is important; the last matching pattern takes the most
+# precedence.
+
+*               @pankajastro @feluelle @tatiana @kaxil
```
utkarsharma2, line 4:
```suggestion
*               @pankajastro @feluelle @tatiana @kaxil @utkarsharma2
```

utkarsharma2: @kaxil interested :)

utkarsharma2 (approved):

Issue #946, opened 2022-09-28
Support dynamic DAG generation in SQL CLI


tatiana:
placeholder

PR #1140, opened 2022-10-26
Add DAG Renderer for run command

dimberman:
# Description
Currently, if a user wants to run a DAG, we first must generate a DAG that generates each task, it's dependencies, and it's variables. This approach leads a DAG that is more explicit, but also can be hard to reach and involves a lot of code maintenance on our part/very little flexibility (e.g. if a user later wants to add a task, they either need to manually add it into the DAG or re-generate from scratch).

[#ISSUE](https://github.com/astronomer/astro-sdk/issues/946)

## What is the new behavior?

To explore multiple options, we will also create a DAG "render" option that does not generate any files, but instead creates a DAG object and runs that object. In the coming weeks we will compare the two solutions.

No

### Checklist


sql-cli/conftest.py
```diff
@@ -161,3 +164,15 @@ def initialised_project_with_empty_workflow(initialised_project: Project):
...
+
+
+@pytest.fixture()
+def test_connections():
+
+    merge_conn(
+        Connection(
+            conn_id="my_test_sqlite",
+            conn_type="sqlite",
+            host=os.path.join(gettempdir(), "sqlite_default.db"),
```
feluelle, line 176: Do we need this path to be "static" ?

If not, I would suggest we use pytests `tmp_path` fixture:
```suggestion
            host=(tmp_path / "sqlite_default.db").as_posix(),
```

sql-cli/sql_cli/dag_generator.py
```diff
@@ -83,6 +88,45 @@ def sorted_sql_files(self) -> list[SqlFile]:
...
+    """
+    if not directory.exists():
+        raise SqlFilesDirectoryNotFound("The directory does not exist!")
+    sql_files: list[SqlFile] = sorted(get_sql_files(directory, target_directory=None))
+    sql_files_dag = SqlFilesDAG(
+        dag_id=directory.name,
+        start_date=datetime(2020, 1, 1),
+        sql_files=sql_files,
+    )
+    dag = DAG(dag_id=workflow_name, start_date=start_date)
```
feluelle, line 125: Let's use the information from the SqlFilesDAG. WDYT?
```suggestion
    dag = DAG(dag_id=sql_files_dag.dag_id, start_date=sql_files_dag.start_date)
```

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -110,8 +111,25 @@ def get_relative_target_path(self) -> Path:
...
+    def to_transform_operator(self) -> TransformOperator:
+        """
+        Converts SQLFile into a TransformOperator that can be added to a DAG.
+        Any relevant metadata from the file frontmatter will be passed to the TransformOperator,
+        though we do not pass parameter dependencies at this stage.
+        :return: a TransformOperator
+        """
+        return TransformOperator(
+            conn_id=self.metadata.get("conn_id"),
+            parameters=None,
```
feluelle, line 123: Why are we not adding parameters here, like how you did it above? 🤔 Is it too complex?
```python
param_dict[s.path.stem].parameters[p] = param_dict[p].output
```

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -110,8 +111,25 @@ def get_relative_target_path(self) -> Path:
...
+            handler=None,
+            database=self.metadata.get("database"),
+            schema=self.metadata.get("schema"),
+            python_callable=lambda: (str(self.path), None),
```
feluelle, line 127: The `None` here are the `parameters` right? Don't we need to set this here?

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -121,7 +139,7 @@ def get_sql_files(directory: Path, target_directory: Path) -> set[SqlFile]:
     :returns: the sql files found in the directory.
     """
     return {
-        SqlFile(root_directory=directory, path=child, target_directory=target_directory)
+        SqlFile(root_directory=directory, path=child, target_directory=target_directory)  # type: ignore
```
feluelle, line 142: I probably know why you added `# type: ignore`, I guess it is to coupled with `target_directory` being set. But before merging we should decouple this in `SqlFile`, or?

sql-cli/tests/test_dag_generator.py
```diff
@@ -31,6 +32,11 @@ def test_generate_dag(root_directory, dags_directory):
     assert dag_file
 
 
+def test_generate_dag_render(test_connections, root_directory, sample_dag):
+    dag = render_dag(directory=root_directory, workflow_name=sample_dag.dag_id)
+    run_dag(dag)
```
feluelle, line 37: Ideally, we should not run the dag here. Instead we should `assert` on the dag object itself and its tasks.

We can have an end-to-end test in `test___main__.py`.

feluelle (changes requested):
This looks super cool @dimberman! and very cleanly implemented!

Left a few comments for you..

One bigger concern I have with that is the deployment part. How would you deploy this? The advantage we have with generating a dag file is that we can ship the dag file to any airflow installation.

For local development we can let the user decide if they want to have the file or not. But it has the risk of having a deviation between local dag and deployed dag (exported using `generate`).

sql-cli/sql_cli/sql_directory_parser.py
dimberman: Because the parameters don't exist yet. We have to create all of the transform operators and THEN add them as parameters/set dependencies.

feluelle: Ah makes sense 👍

sql-cli/sql_cli/__main__.py
```diff
@@ -125,6 +125,7 @@ def run(
     project_dir: Path = typer.Option(
         None, dir_okay=True, metavar="PATH", help="(Optional) Default: current directory.", show_default=False
     ),
+    gen_dag: bool = typer.Option(default=False, help="whether to generate a DAG file", show_default=True),
```
tatiana, line 128: I find the name `gen_dag` misleading. I suggest we name the `Option` `--export-dag`. I believe, by default, we should generate the DAG - so we have a consistent way of representing what we are running locally and remotely.

sql-cli/sql_cli/__main__.py
```diff
@@ -139,17 +140,20 @@ def run(
...
-            dags_directory=project.airflow_dags_folder,
-        )
+        if gen_dag:
+            dag_file = generate_dag(
+                directory=project.directory / project.workflows_directory / workflow_name,
+                dags_directory=project.airflow_dags_folder,
+            )
+            dag = get_dag(dag_id=workflow_name, subdir=dag_file.parent.as_posix(), include_examples=False)
+        else:
+            dag = render_dag(directory=project_dir / "workflows" / workflow_name, workflow_name=workflow_name)
```
tatiana, line 150:
```suggestion
            dag = render_dag(directory=project.directory / project.workflows_directory / workflow_name, workflow_name=workflow_name)
```

tatiana, line 150: I find the name `render_dag` confusing. What if we called this function `get_virtual_dag`?

sql-cli/sql_cli/dag_generator.py
```diff
@@ -83,6 +88,45 @@ def sorted_sql_files(self) -> list[SqlFile]:
 
         return list(depth_first_search.dfs_postorder_nodes(graph))
 
+    def to_transform_task_list(self) -> list[TransformOperator]:
```
tatiana, line 91:
```suggestion
    def to_task_list(self) -> list[TransformOperator]:
```

sql-cli/sql_cli/dag_generator.py
```diff
@@ -83,6 +88,45 @@ def sorted_sql_files(self) -> list[SqlFile]:
...
+        """
+        param_dict = {s.path.stem: s.to_transform_operator() for s in self.sql_files}
+        for s in self.sql_files:
+            for p in s.get_parameters():
+                param_dict[s.path.stem].parameters[p] = param_dict[p].output
+                param_dict[s.path.stem].set_upstream(param_dict[p])
+        return list(param_dict.values())
+
+
+def render_dag(directory: Path, workflow_name: str, start_date: datetime = datetime(2020, 1, 1)) -> DAG:
```
tatiana, line 105:
```suggestion
def get_virtual_dag(directory: Path, workflow_name: str, start_date: datetime = DEFAULT_DATETIME) -> DAG:
```
Suggestions:
* Rename `render_dag` to `get_virtual_dag`
* Have a constant to define the DEFAULT_DATETIME

dimberman:
Hi @feluelle that's a really good point. I've modified the PR. Now both options will "generate" a DAG file that users can easily deploy, one just has the tasks pre-generated and one generates the tasks at runtime based on the directory. Please take a look when you have a chance

sql-cli/sql_cli/cli.py
```diff
@@ -34,6 +34,25 @@ def generate_dag(project: Project, env: str, workflow_name: str) -> Path:
         rprint(f"[bold red]The workflow {workflow_name} contains a cycle! {dag_cycle}[/bold red]")
         raise Exit(code=1)
     rprint("The DAG file", dag_file.resolve(), "has been successfully generated. 🎉")
+    import_errors = DagBag(process_subdir(str(dag_file))).import_errors
+    if import_errors:
+        all_errors = "\n\n".join(list(import_errors.values()))
```
feluelle, line 39: This is an additional great validation step. I am wondering if this alone could replace some of the exception handling like checking for dag cycles for example. But wondering how cryptic the output looks for (sql analyst) users.

Could you move that into a function, please? It looks very complex.

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -21,6 +22,7 @@ def __init__(self, root_directory: Path, path: Path, target_directory: Path) ->
         self.root_directory = root_directory
         self.path = path
         self.target_directory = target_directory
+        self.raw_content = open(self.path).read()
```
feluelle, line 25:
```suggestion
        self.raw_content = self.path.read_text()
```
[Path.read_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text)

sql-cli/sql_cli/sql_directory_parser.py
feluelle: Could you revert this please since you know pass a target_directory.

sql-cli/sql_cli/templates/render_dag.py.jinja2
```diff
@@ -0,0 +1,20 @@
+# NOTE: This is an auto-generated file. Please do not edit this file manually.
+from pathlib import Path
+
+from airflow import DAG
+from airflow.utils import timezone
+from astro import sql as aql
+from astro.table import Table
+from sql_cli.dag_render import render_tasks
+DAGS_FOLDER = Path(__file__).parent.as_posix()
```
feluelle, line 9:
```suggestion
from sql_cli.dag_render import render_tasks

DAGS_FOLDER = Path(__file__).parent.as_posix()
```

sql-cli/sql_cli/dag_generator.py
```diff
@@ -108,3 +108,34 @@ def generate_dag(directory: Path, dags_directory: Path) -> Path:
...
+        sql_files=sql_files,
+    )
+    render(
+        template_file=Path("templates/render_dag.py.jinja2"),
+        context={
+            "dag": sql_files_dag,
+            "dag_id": directory.name,
+            "start_date": datetime(2020, 1, 1),
+            "project_path": directory,
+        },
```
feluelle, line 138:
```suggestion
        context={
            "dag": sql_files_dag,
        },
```
Since you now still need the `sql_files_dag`, I would suggest you only pass the dag as context.

feluelle: And after removing this, the only part that is different is the `template_file`. Maybe it makes sense to pass the template_file as argument to the function OR merging the template files into one. WDYT?

feluelle (commented):
Nice work @dimberman! Left some comments for further improvements.

sql-cli/sql_cli/cli.py
dimberman: @feluelle I think for some common use-cases it might be better for us to own the error handling so we can send analyst friendly error messages (e.g. an analyst shouldn't need an airflow stacktrace). that said I'll move the import error checking into function.

dimberman:
@feluelle that was a really great call! Really reduces the code footprint. Now it's as simple as picking one of two rendering options.

sql-cli/sql_cli/dag_generator.py
dimberman: @feluelle done!

feluelle (approved):
Looking great @dimberman! 🚀

feluelle:
Please only add some more tests to improve coverage, when done, feel free to merge.

sql-cli/sql_cli/__main__.py
```diff
@@ -127,6 +127,13 @@ def run(
     project_dir: Path = typer.Option(
         None, dir_okay=True, metavar="PATH", help="(Optional) Default: current directory.", show_default=False
     ),
+    generate_tasks: bool = typer.Option(
+        default=False,
+        help="whether to explicitly generate the tasks in your SQL CLI "
+        "DAG. By default we will keep the DAGs smaller and read SQL"
+        "files at runtime",
```
kaxil, line 134:
```suggestion
        "DAG. By default we will keep the DAGs smaller and read SQL "
        "files at runtime",
```

sql-cli/sql_cli/dag_render.py
```diff
@@ -0,0 +1,44 @@
...
+
+if TYPE_CHECKING:
+    from astro.sql.operators.transform import TransformOperator
+
+
+def to_task_list(sql_files: list[SqlFile]) -> list[TransformOperator]:
+    """
+    Converts the list of SQL Files into a list of TranformOperator tasks
+    that all have proper dependencies set.
+    :return:
```
kaxil, line 17: missing docstring for `sql_files` and either add docstring for `:return:` or remove it

sql-cli/sql_cli/dag_render.py
```diff
@@ -0,0 +1,44 @@
...
+            param_dict[s.path.stem].set_upstream(param_dict[p])
+    return list(param_dict.values())
+
+
+def render_tasks(directory: Path) -> list[TransformOperator]:
+    """
+    render_dag allows a user to take any directory and turn it into a runnable
+    Airflow DAG. This function will read all SQL files, and set dependencies based
+    on jinja template-based variables.
+    :param directory: Base directory for SQL files. We will recursively parse
```
kaxil, line 34:
```suggestion
    on jinja template-based variables.

    :param directory: Base directory for SQL files. We will recursively parse
```

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -110,8 +112,41 @@ def get_relative_target_path(self) -> Path:
 
         return target_path.relative_to(self.target_directory)
 
+    def write_raw_content_to_target_path(self) -> None:
+        """
+        Writes both content and headers to the target directory.
+        This is because with the "render" function, we will still need
+        the headers for creating proper TransformOperators
+        :return:
```
kaxil, line 120: same as https://github.com/astronomer/astro-sdk/pull/1140/files#r1015793376

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -110,8 +112,41 @@ def get_relative_target_path(self) -> Path:
...
+        target_path = target_full_directory / self.path.name
+
+        target_path.write_text(self.raw_content)
+
+    def to_transform_operator(self) -> TransformOperator:
+        """
+        Converts SQLFile into a TransformOperator that can be added to a DAG.
+        Any relevant metadata from the file frontmatter will be passed to the TransformOperator,
+        though we do not pass parameter dependencies at this stage.
+        :return: a TransformOperator
```
kaxil, line 136:
```suggestion
        though we do not pass parameter dependencies at this stage.

        :return: a TransformOperator
```

sql-cli/sql_cli/templates/render_dag.py.jinja2
```diff
@@ -0,0 +1,17 @@
+# NOTE: This is an auto-generated file. Please do not edit this file manually.
+from pathlib import Path
+
+from airflow import DAG
+from airflow.utils import timezone
+from astro import sql as aql
+from astro.table import Table
+from sql_cli.dag_render import render_tasks
```
kaxil, line 8: In a future version, we can move this render to python SDK, and use `from astro` import? so that their is a clear separation between CLI (SQL CLI) vs library (Python SDK)?

dimberman: @kaxil that is absolutely the plan. I started making a POC of it, but I realized it would be better to wait until I go approval to do so.

sql-cli/sql_cli/cli.py
```diff
@@ -14,11 +17,16 @@
 from sql_cli.project import Project
 
 
-def generate_dag(project: Project, workflow_name: str) -> Path:
+def generate_dag(project: Project, env: str, workflow_name: str, generate_tasks: bool = False) -> Path:
+    rprint(
+        f"\nGenerating the DAG file from workflow [bold blue]{workflow_name}[/bold blue]"
+        f" for [bold]{env}[/bold] environment..\n"
+    )
```
feluelle, line 24: @dimberman I think there went something wrong on merging main into your branch. With that change we now have duplicate print statements in generate.

sql-cli/sql_cli/cli.py
```diff
@@ -29,11 +37,25 @@ def generate_dag(project: Project, workflow_name: str) -> Path:
     except DagCycle as dag_cycle:
         rprint(f"[bold red]The workflow {workflow_name} contains a cycle! {dag_cycle}[/bold red]")
         raise Exit(code=1)
+    rprint("The DAG file", dag_file.resolve(), "has been successfully generated. 🎉")
```
feluelle, line 40: Same here.

sql-cli/sql_cli/cli.py
```diff
@@ -29,11 +37,25 @@ def generate_dag(project: Project, workflow_name: str) -> Path:
...
+    if import_errors:
+        all_errors = "\n\n".join(list(import_errors.values()))
+        rprint(f"[bold red]Workflow failed to render[/bold red]\n errors found:\n\n {all_errors}")
+        raise Exit(code=1)
+
+
+def run_dag(project: Project, env: str, dag: DAG, verbose: bool) -> None:
+    rprint(
+        f"\nRunning the workflow [bold blue]{dag.dag_id}[/bold blue] for [bold]{env}[/bold] environment..\n"
+    )
```
feluelle, line 56: And here.

sql-cli/sql_cli/cli.py
```diff
@@ -47,3 +69,14 @@ def run_dag(project: Project, dag: DAG, verbose: bool) -> DagRun:
...
+    final_state = None
+    if dr.state == State.SUCCESS:
+        final_state = "[bold green]SUCCESS[/bold green]"
+    elif final_state == State.FAILED:
+        final_state = "[bold red]FAILED[/bold red]"
+    else:
+        final_state = dr.state
+    rprint(f"Final state: {final_state}")
+    elapsed_seconds = (dr.end_date - dr.start_date).microseconds / 10**6
+    rprint(f"Total elapsed time: [bold blue]{elapsed_seconds:.2}s[/bold blue]")
```
feluelle, line 82: Could you move this please to the `__main__` call. This way we ensure that the normal print statements only get printed once i.e. per command call.

FYI we moved the print in PR https://github.com/astronomer/astro-sdk/pull/1199

sql-cli/sql_cli/dag_generator.py
```diff
@@ -101,9 +103,15 @@ def generate_dag(directory: Path, dags_directory: Path) -> Path:
...
+            sql_file.write_raw_content_to_target_path()
+        template_dag_name = "render_dag"
+
     output_file = dags_directory / f"{sql_files_dag.dag_id}.py"
     render(
-        template_file=Path("templates/dag.py.jinja2"),
+        template_file=Path(f"templates/{template_dag_name}.py.jinja2"),
         context={"dag": sql_files_dag},
         output_file=output_file,
     )
```
feluelle, line 117: I think this would have looked nicer 😬 WDYT?
```python
    if generate_tasks:
        for sql_file in sql_files_dag.sorted_sql_files():
            sql_file.write_raw_content_to_target_path()
        template_file = Path("templates/render_dag.py.jinja2")
    else:
        template_file = Path("templates/gen_tasks_dag.py.jinja2")

    output_file = dags_directory / f"{sql_files_dag.dag_id}.py"
    render(
        template_file=template_file,
        context={"dag": sql_files_dag},
        output_file=output_file,
    )
```

feluelle (commented):
Also codecov did not pass. 😕

dimberman:
@feluelle that's weird codecov was passing when I merged, not sure what happened there. regarding the other issues I've created a PR here https://github.com/astronomer/astro-sdk/pull/1231/files

Issue #1141, opened 2022-10-27
Fix namespace for OL for sqlite


sunank200:
**Describe the bug**
The sqlite namespace is missing in [OpenLineage naming convention](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md). So we created the namespace by default. OL team has requested to change this namespace.
The Namespaces extracted from AstroSDK do not conform to the [OpenLineage naming convention](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md) for sqlite. For example, /tmp/sqlite_default.db instead of e.g. sqllite://

PR #1142, opened 2022-10-27
Fix open lineage namespace for Sqlite as per OL team request

sunank200:
# Description

The sqlite namespace is missing in [OpenLineage naming convention](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md). So we created the namespace by default. OL team has requested to change this namespace.
The Namespaces extracted from AstroSDK do not conform to the [OpenLineage naming convention](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md) for sqlite. For example, /tmp/sqlite_default.db instead of e.g. sqllite://

closes: #1141

## What is the new behavior?

- Fix the namespace for sqlite
- Fix the test for this

No

### Checklist

### DIFF
--- a/python-sdk/tests/databases/test_base_database.py
+++ b/python-sdk/tests/databases/test_base_database.py
@@ -9,13 +9,31 @@ from astro.databases import create_database
 from astro.databases.base import BaseDatabase
 from astro.files import File
 from astro.settings import SCHEMA
-from astro.table import Metadata, Table
+from astro.table import BaseTable, Metadata, Table
 
 CWD = pathlib.Path(__file__).parent
 
 
 class DatabaseSubclass(BaseDatabase):
     pass
+
+
+def test_openlineage_database_dataset_namespace():
+    """
+    Test the open lineage dataset namespace for base class
+    """
+    db = DatabaseSubclass(conn_id="fake_conn_id")
+    with pytest.raises(NotImplementedError):
+        db.openlineage_dataset_namespace()
+
+
+def test_openlineage_database_dataset_name():
+    """
+    Test the open lineage dataset names for the base class
+    """
+    db = DatabaseSubclass(conn_id="fake_conn_id")
+    with pytest.raises(NotImplementedError):
+        db.openlineage_dataset_name(table=BaseTable)
 
 
 def test_subclass_missing_not_implemented_methods_raise_exception():
--- a/python-sdk/tests/sql/test_table.py
+++ b/python-sdk/tests/sql/test_table.py
@@ -211,8 +211,8 @@ def test_if_table_object_can_be_pickled():
         ),
         (
             Connection(conn_id="test_conn", conn_type="sqlite", host="/tmp/sqlite.db"),
-            "/tmp/sqlite.db.test_tb",
-            "/tmp/sqlite.db",
+            "tmp/sqlite.db.test_tb",
+            "sqlite://tmp/sqlite.db",
         ),
     ],
 )
--- a/python-sdk/tests/sql/test_table.py
+++ b/python-sdk/tests/sql/test_table.py
@@ -210,7 +210,7 @@ def test_if_table_object_can_be_pickled():
             "snowflake://astro-sdk",
         ),
         (
-            Connection(conn_id="test_conn", conn_type="sqlite", host="/tmp/sqlite.db"),
+            Connection(conn_id="test_conn", conn_type="sqlite", host="tmp/sqlite.db"),
             "tmp/sqlite.db.test_tb",
             "sqlite://tmp/sqlite.db",
         ),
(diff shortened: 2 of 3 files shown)


pankajastro (approved):
LGTM. Thanks, @sunank200 for fixing it.

python-sdk/tests/databases/test_base_database.py
```diff
@@ -18,6 +18,24 @@ class DatabaseSubclass(BaseDatabase):
     pass
 
 
+def test_openlineage_database_dataset_namespace():
```
pankajastro, line 21: just wondering does this test add some value?

sunank200: It is just for code coverage. It was failing for coverage

julienledem: I think this is a good point from @pankajastro, the test should check that the method returns what you expect. 
Test coverage reports tell you where you're missing tests, coverage is not a goal in itself.
In this case, the test should check that the method returns a namespace that conforms with the spec.

python-sdk/tests/databases/test_base_database.py
```diff
@@ -18,6 +18,24 @@ class DatabaseSubclass(BaseDatabase):
...
+        db.openlineage_dataset_namespace()
+
+
+def test_openlineage_database_dataset_name():
+    """
+    Test the open lineage dataset names for the base class
+    """
+    db = DatabaseSubclass(conn_id="fake_conn_id")
+    with pytest.raises(NotImplementedError):
+        db.openlineage_dataset_name(table=BaseTable)
```
julienledem, line 36: Similarly, here you want to mock the method being called to check that openlineage_dataset_name returns the expected value.

python-sdk/tests/databases/test_base_database.py
```diff
@@ -18,6 +18,24 @@ class DatabaseSubclass(BaseDatabase):
...
+
+
 def test_subclass_missing_not_implemented_methods_raise_exception():
     db = DatabaseSubclass(conn_id="fake_conn_id")
     with pytest.raises(NotImplementedError):
```
julienledem, line 41: ditto

python-sdk/tests/sql/test_table.py
```diff
@@ -210,9 +210,9 @@ def test_if_table_object_can_be_pickled():
             "snowflake://astro-sdk",
         ),
         (
-            Connection(conn_id="test_conn", conn_type="sqlite", host="/tmp/sqlite.db"),
-            "/tmp/sqlite.db.test_tb",
-            "/tmp/sqlite.db",
+            Connection(conn_id="test_conn", conn_type="sqlite", host="tmp/sqlite.db"),
+            "tmp/sqlite.db.test_tb",
+            "sqlite://tmp/sqlite.db",
```
julienledem, line 215: Two comments here: 
- The namespace should not repeat the dataset name ("tmp/sqlite.db.test_tb").
- The tricky aspect of those local datasets that are private to the machine they are running on is we want the dataset ns+name to be unique. Otherwise if you run this on two different machines the lineage graph will assume that you are writing and reading the same dataset, which is not true. You need to make the namespace more unique by (for example) adding the machine name/IP in it.

julienledem (commented):
Thank you @sunank200 for addressing the request.
Sorry for not giving my review before this got merged.
Please look at the few comments bellow.
I think you still need to tweak it before we can consider it done.
Best, Julien

python-sdk/tests/databases/test_base_database.py
sunank200: Sure. We already have a test which covers the scenarios: https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/sql/test_table.py#L221

python-sdk/tests/sql/test_table.py
sunank200: I have added the PR for the change https://github.com/astronomer/astro-sdk/pull/1178. Please review

JDarDagran:
Hey, there have been new conventions introduced: [Naming.md#local-file-system](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#local-file-system)
I believe sqlite counts as local file, therefore you could adjust naming convention in astro as well.

A word of explanation: both IP and port are mandatory. It means that if port is unknown (e. g. if file exists in Airflow's local file system only) you should default to some value. In one of OL extractors it is defaulted to `paramiko.config.SSH_PORT`, in most cases it's equal to 22.

sunank200:
@JDarDagran created the issue for this: https://github.com/astronomer/astro-sdk/issues/1281

sunank200:
> Hey, there have been new conventions introduced: [Naming.md#local-file-system](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#local-file-system) I believe sqlite counts as local file, therefore you could adjust naming convention in astro as well.
> 
> A word of explanation: both IP and port are mandatory. It means that if port is unknown (e. g. if file exists in Airflow's local file system only) you should default to some value. In one of OL extractors it is defaulted to `paramiko.config.SSH_PORT`, in most cases it's equal to 22.

@JDarDagran created the issue for this: https://github.com/astronomer/astro-sdk/issues/1281

Issue #750, opened 2022-08-25
Support for s3 compatible object storage 


mohittalele:
**Describe the bug**
I am trying to use the astr sdk with airflow 2.3.3, postgresql as database and minio as an s3 compatible object storage
**Version**
* Astro: [1.0.2]
* OS: [ Debian]
* Airflow : 2.3.3

**To Reproduce**
1. Write the task to load the csv file stored in pandas dataframe  or database 
 ```
 test_s3_load = aql.load_file(
        input_file=File(path=f"{s3_bucket}/data/falcon_heartbeat.csv", conn_id="s3_conn"),
    )
my_homes_table = aql.load_file(
    input_file=File(path=f"{s3_bucket}/data/falcon_heartbeat.csv", conn_id="s3_conn"),
    # [START temp_table_example]  skipcq: PY-W0069
    output_table=Table(
        conn_id="falcon_postgresql_conn",
        name="astro_example_test",
    )
```
3. Create connection using env variables as :
```
AIRFLOW_CONN_S3_CONN: 's3://MINIO_USERNAME:MINIO_PASSWORD@?host=http%3A%2F%2Fminio.airflow.svc.cluster.local%3A9000'
AIRFLOW_CONN_FALCON_POSTGRESQL_CONN: 'postgresql://falcon:falcon@falcon-postgres-postgresql.airflow.svc.cluster.local:5432/falcon_db'
```

4. Run using  the dag 
5. See error

```
  File "/home/airflow/.local/lib/python3.7/site-packages/botocore/retryhandler.py", line 374, in _check_caught_exception
    raise caught_exception
  File "/home/airflow/.local/lib/python3.7/site-packages/botocore/endpoint.py", line 249, in _do_get_response
    http_response = self._send(request)
  File "/home/airflow/.local/lib/python3.7/site-packages/botocore/endpoint.py", line 321, in _send
    return self.http_session.send(request)
  File "/home/airflow/.local/lib/python3.7/site-packages/botocore/httpsession.py", line 434, in send
    raise EndpointConnectionError(endpoint_url=request.url, error=e)
botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "https://falcon.s3.amazonaws.com/data/falcon_heartbeat.csv"
[...]

PR #1146, opened 2022-10-27
Fix minio connection issue

utkarsharma2:
# Description
Currently, we were not able to connect to minio(https://min.io/) as a replacement for S3.
Impact: load_file(), export_file() and minio as custom xcom backend

related: #750
and slack thread - https://astronomer.slack.com/archives/C02B8SPT93K/p1666289425280609

## What is the new behavior?
Now we are passing all the required parameters to open_smart() to be able to communicate to minio server.

Nope


phanikumv (commented):
Wondering if we could have an example DAG that (integration) tests this scenario?

.github/ci-test-connections.yaml
```diff
@@ -73,3 +73,17 @@ connections:
     extra:
       aws_access_key_id: $AWS_ACCESS_KEY_ID
       aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
+  - conn_id: minio_conn_1
+    conn_type: aws
+    description: null
+    extra:
+      aws_access_key_id: "ROOTNAME"
+      aws_secret_access_key: "CHANGEME123"
+      host: "http://127.0.0.1:9000"
```
pankajastro, line 82: what about using ```endpoint_url``` instead of host since ```host``` in extra has deprecated in Airflow aws conn?

utkarsharma2: @pankajastro We are testing it with both actually. `minio_conn_2` has `endpoint_url`

tatiana (approved):
LGTM, thanks, @utkarsharma2 ! Do we need to spin up Minio as a service in the CI? I'm happy for you to merge this PR once the tests are passing

.github/ci-test-connections.yaml
pankajastro: ok, just curious why testing for both i.e host and endpoint_url

utkarsharma2: @pankajastro  We needed that earlier since we were passing the values like endpoint_url while creating a client, and the filed name has changed in apache-airflow-providers-amazon from `host` to `endpoint_url` in the 5.0.0 version and we wanted to handle both cases. Now we don't need that since we are using provide a method to do that.

utkarsharma2:
@phanikumv @tatiana 
1. Minio works very similarly to S3, so I think we can do away without adding an integration test
2. Also since Minio is locally hosted we need to have a server of it running in our CI env for us to have an integration test.

I think just having a UT which verifies that we are creating the transport params correctly should be enough.

PR #1148, opened 2022-10-27
Add  global config `enable_native_fallback` for load file op

pankajastro:
# Description
closes: #1089 
Add `LOAD_FILE_ENABLE_NATIVE_FALLBACK` config to globally disable native fallback

currently, we do not have a way to disable the native fallback globally

## What is the new behavior?
you can set LOAD_FILE_ENABLE_NATIVE_FALLBACK to disable/enable native fallback still local config will overwrite the global config in case it is provided

No (But, by default the ENABLE_NATIVE_FALLBACK was true now its false)

### Checklist

### DIFF
--- a/python-sdk/docs/configurations.rst
+++ b/python-sdk/docs/configurations.rst
@@ -143,3 +143,21 @@ or by updating Airflow's configuration
 
    [astro_sdk]
    openlineage_emit_temp_table_event = True
+
+
+Configuring the native fallback mechanism
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Astro SDK ``LoadFileOperator`` have a fallback mechanism to load data in database using dataframe.
+By default, We have disabled this feature. This can override at task level using ``enable_native_fallback`` param.
+This might be inconvenient for some users who do want to leverage it. Such users can set the following config to ``True`` to enable it at global level.
+
+.. code:: ini
+
+   LOAD_FILE_ENABLE_NATIVE_FALLBACK = False
+
+or by updating Airflow's configuration
+
+.. code:: ini
+
+   [astro_sdk]
+   load_file_enable_native_fallback = False
--- a/python-sdk/docs/configurations.rst
+++ b/python-sdk/docs/configurations.rst
@@ -124,6 +124,7 @@ or by updating Airflow's configuration
    [astro_sdk]
    auto_add_inlets_outlets = True
 
+.. _configure_native_fallback:
 
 Configuring to emit temp table event in openlineage
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -147,17 +148,19 @@ or by updating Airflow's configuration
 
 Configuring the native fallback mechanism
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Astro SDK ``LoadFileOperator`` have a fallback mechanism to load data in database using dataframe.
-By default, We have disabled this feature. This can override at task level using ``enable_native_fallback`` param.
-This might be inconvenient for some users who do want to leverage it. Such users can set the following config to ``True`` to enable it at global level.
+The ``LoadFileOperator`` has a fallback mechanism when loading data to the database from file storage as explained in :ref:`load_file_working`.
 
-.. code:: ini
+This fallback can be configured at the task level using ``enable_native_fallback`` param.
 
-   LOAD_FILE_ENABLE_NATIVE_FALLBACK = False
+Users can also control this setting and override the default at a global level (for all tasks) by setting the following config. Set it to ``True`` to allow falling back to "pandas" path.
+
+.. code-block:: shell
+
+   AIRFLOW__ASTRO_SDK__LOAD_FILE_ENABLE_NATIVE_FALLBACK = False
 
 or by updating Airflow's configuration
 
-.. code:: ini
+.. code-block:: ini
 
    [astro_sdk]
    load_file_enable_native_fallback = False
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -118,7 +118,6 @@ with dag:
             "allow_jagged_rows": True,
             "skip_leading_rows": "1",
         },
-        enable_native_fallback=False,
     )
     # [END load_file_example_9]
 
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -165,7 +165,7 @@ with dag:
     # [START load_file_example_14]
     aql.load_file(
         input_file=File(
-            "gs://astro-sdk/workspace/sample_pattern.csv",
+            "gs://astro-sdk/workspace/sample_pattern",
             conn_id="bigquery",
             filetype=FileType.CSV,
         ),
@@ -199,7 +199,7 @@ with dag:
     # [START load_file_example_17]
     aql.load_file(
         input_file=File(
-            "gs://astro-sdk/workspace/sample_pattern.csv",
+            "gs://astro-sdk/workspace/sample_pattern",
             conn_id="bigquery",
             filetype=FileType.CSV,
         ),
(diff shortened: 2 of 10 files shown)


python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -33,7 +35,7 @@ class LoadFileOperator(AstroSQLBaseOperator):
     :param native_support_kwargs: kwargs to be used by method involved in native support flow
     :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
             in the resulting dataframe
-    :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
+    :param enable_native_fallback: (deprecated) Use enable_native_fallback=True to fall back to default transfer
```
kaxil, line 38: I would say instead of deprecating, we allow both configs (global and local)

So users can disable it globally but selectively enable it based on a certain transfer

pankajastro: @kaxil I believe global config means airflow config here but can you please elaborate a bit on what you mean by a local config (the existing one i.e based on the value of the `enable_native_fallback ` param)? Also when you say
> selectively enable it based on a certain transfer

do you mean using `enable_native_fallback` in the operator?

python-sdk/example_dags/example_load_file.py
```diff
@@ -142,7 +141,7 @@
     # [START load_file_example_12]
     aql.load_file(
         input_file=File(
-            "gs://astro-sdk/workspace/sample_pattern",
+            "gs://astro-sdk/workspace/sample_pattern.csv",
```
pankajastro, line 144: Look like old resource does not exist I'm getting 404 in CI for old path

python-sdk/src/astro/sql/operators/load_file.py
pankajastro: Removed deprecation, now we have global and local config. the local config will overwrite the global one

python-sdk/example_dags/example_load_file.py
kaxil: hmm, it should have worked no? Or else the tests in main would have failed?

python-sdk/tests/extractors/test_extractor.py
```diff
@@ -52,7 +52,7 @@
                 columns=[],
                 schema="astro",
                 used_native_path=False,
-                enabled_native_fallback=True,
+                enabled_native_fallback=False,
```
utkarsharma2, line 55: @pankajastro Should we keep this as default?

python-sdk/example_dags/example_load_file.py
kaxil: actually I was not able to find the file too, how is this working? or do we not run this tests on main?

utkarsharma2 (approved):
LGTM.

utkarsharma2:
@pankajastro Changes in `python-sdk/example_dags/example_load_file.py` looks odd. Let's investigate it before merging the PR.

python-sdk/src/astro/settings.py
```diff
@@ -10,6 +10,8 @@
 SNOWFLAKE_SCHEMA = conf.get("astro_sdk", "snowflake_default_schema", fallback=SCHEMA)
 REDSHIFT_SCHEMA = conf.get("astro_sdk", "redshift_default_schema", fallback=SCHEMA)
 
+LOAD_FILE_ENABLE_NATIVE_FALLBACK = conf.get("astro_sdk", "load_file_enable_native_fallback", fallback=False)
```
kaxil, line 13: We should add this to https://github.com/astronomer/astro-sdk/blob/main/python-sdk/docs/configurations.rst

python-sdk/example_dags/example_load_file.py
phanikumv: created https://github.com/astronomer/astro-sdk/issues/1163 to address this issue

sunank200:
PR wise it LGTM but we should investigate the `sample_pattern` working on example_load_file before merging it. It shouldn't have passed the CI.

python-sdk/src/astro/settings.py
pankajastro: Added config docs 5b3f7e44e973799fe9508b4812aa695b9e8f91b8

python-sdk/docs/configurations.rst
```diff
@@ -123,3 +123,21 @@ or by updating Airflow's configuration
 
    [astro_sdk]
    auto_add_inlets_outlets = True
+
+
+Configuring the native fallback mechanism
```
kaxil, line 128: Should we update point 3 in https://astro-sdk-python.readthedocs.io/en/stable/astro/sql/operators/load_file.html#parameters-for-native-transfer too @pankajastro ?

python-sdk/docs/configurations.rst
```diff
@@ -123,3 +123,21 @@ or by updating Airflow's configuration
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Astro SDK ``LoadFileOperator`` have a fallback mechanism to load data in database using dataframe.
+By default, We have disabled this feature. This can override at task level using ``enable_native_fallback`` param.
+This might be inconvenient for some users who do want to leverage it. Such users can set the following config to ``True`` to enable it at global level.
+
+.. code:: ini
+
+   LOAD_FILE_ENABLE_NATIVE_FALLBACK = False
```
kaxil, line 136:
```suggestion
   AIRFLOW__ASTRO_SDK__LOAD_FILE_ENABLE_NATIVE_FALLBACK = False
```

python-sdk/docs/configurations.rst
```diff
@@ -123,3 +123,21 @@ or by updating Airflow's configuration
...
    auto_add_inlets_outlets = True
+
+
+Configuring the native fallback mechanism
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Astro SDK ``LoadFileOperator`` have a fallback mechanism to load data in database using dataframe.
+By default, We have disabled this feature. This can override at task level using ``enable_native_fallback`` param.
+This might be inconvenient for some users who do want to leverage it. Such users can set the following config to ``True`` to enable it at global level.
+
+.. code:: ini
```
kaxil, line 134:
```suggestion
.. code-block:: shell
```

python-sdk/docs/configurations.rst
```diff
@@ -123,3 +123,21 @@ or by updating Airflow's configuration
...
+
+   LOAD_FILE_ENABLE_NATIVE_FALLBACK = False
+
+or by updating Airflow's configuration
+
+.. code:: ini
```
kaxil, line 140:
```suggestion
.. code-block:: ini
```

python-sdk/docs/configurations.rst
```diff
@@ -123,3 +123,21 @@ or by updating Airflow's configuration
 
    [astro_sdk]
    auto_add_inlets_outlets = True
+
+
+Configuring the native fallback mechanism
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Astro SDK ``LoadFileOperator`` have a fallback mechanism to load data in database using dataframe.
+By default, We have disabled this feature. This can override at task level using ``enable_native_fallback`` param.
+This might be inconvenient for some users who do want to leverage it. Such users can set the following config to ``True`` to enable it at global level.
```
kaxil, line 132: We should rephrase this:

```suggestion
The ``LoadFileOperator`` has a fallback mechanism when loading data to the database from file storage as explained in :ref:`load_file_working`.
This fallback can be configured at the task level using ``enable_native_fallback`` param.
Users can also control this setting and override the default at a global level (for all tasks) by setting the following config. Set it to ``True`` to allow falling back to "pandas" path.
```

python-sdk/docs/configurations.rst
pankajastro: updated 6dccda0ad378c40578138421e5c21692fde27395

pankajastro:
> PR wise it LGTM but we should investigate the `sample_pattern` working on example_load_file before merging it. It shouldn't have passed the CI.

Look like `example_load_file.py` does not run on main

pankajastro:
> > PR wise it LGTM but we should investigate the `sample_pattern` working on example_load_file before merging it. It shouldn't have passed the CI.
> 
> Look like `example_load_file.py` does not run on main

We investigated this in https://github.com/astronomer/astro-sdk/issues/1163 and as a conclusion for now at least will have to add the full path to pass Airflow-2-2-5 tests

Issue #1149, opened 2022-10-27
SQL CLI 0.1 in Astro CLI 1.7: Verbose run command


tatiana:
**Describe the bug**
When trying out the Astro CLI 1.7 alpha, we noticed that the run command (both on @mikeshwe & on my laptop), is more verbose than we'd expect.

**Version**
* Astro CLI: 1.7 alpha
* SQL CLI: 0.1.0

**To Reproduce**

```
$ astro flow run example_basic_transform

Running the workflow example_basic_transform for default environment

Processing top_animations... [2022-10-27 22:08:31,482] {taskinstance.py:1590} INFO - Exporting the following env vars:
AIRFLOW_CTX_DAG_OWNER=airflow
AIRFLOW_CTX_DAG_ID=example_basic_transform
AIRFLOW_CTX_TASK_ID=top_animations
AIRFLOW_CTX_EXECUTION_DATE=2022-10-27T22:08:31.173150+00:00
AIRFLOW_CTX_TRY_NUMBER=1
AIRFLOW_CTX_DAG_RUN_ID=manual__2022-10-27T22:08:31.173150+00:00
[2022-10-27 22:08:31,484] {base_decorator.py:125} INFO - Returning table Table(_name='top_animations', conn_id='sqlite_conn', metadata=Metadata(schema=None, database=None), columns=[], temp=False, uri='astro://@?table=top_animations', extra={})
[2022-10-27 22:08:31,681] {taskinstance.py:1401} INFO - Marking task as SUCCESS. dag_id=example_basic_transform, task_id=top_animations, execution_date=20221027T220831, start_date=, end_date=20221027T220831
SUCCESS
Completed running the workflow example_basic_transform: [SUCCESS]
Total elapsed time: 0.6s
```

**Expected behaviour**

```
$ astro flow run example_basic_transform

Running the workflow example_basic_transform for default environment

Processing top_animations...  SUCCESS
Completed running the workflow example_basic_transform: [SUCCESS]
Total elapsed time: 0.6s
```

tatiana:
Initially, only @mikeshwe and I were able to reproduce this issue. While troubleshooting with @feluelle , we were able to reproduce the issue in his laptop after running deleting the SQL CLI docker image, by running:  `docker rmi sql_cli -f`

PR #1157, opened 2022-10-31
Disable propagation for airflow logs in sql-cli

feluelle:
# Description

In Airflow version `2.4.2` the airflow logs propagation has changed from `False` to `True`.

closes: #1149
related: https://github.com/apache/airflow/issues/27345

## What is the new behavior?

Since we don't want to show any airflow logs to the user by default we explicitly disable it now.

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -12,7 +12,9 @@ load_dotenv()
 app = typer.Typer(add_completion=False, context_settings={"help_option_names": ["-h", "--help"]})
 
 
-logging.getLogger("airflow").setLevel(logging.CRITICAL)
+airflow_logger = logging.getLogger("airflow")
+airflow_logger.setLevel(logging.CRITICAL)
+airflow_logger.propagate = False
 
 
 @app.command(help="Print the SQL CLI version.")
--- a/sql-cli/tests/test_run_dag.py
+++ b/sql-cli/tests/test_run_dag.py
@@ -121,7 +121,7 @@ def test_run_dag(capsys, caplog):
     assert not any(record.name == "airflow.task" for record in caplog.records)
 
 
-def test_run_dag_verbose(capsys, caplog):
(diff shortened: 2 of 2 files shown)


sql-cli/sql_cli/__main__.py
```diff
@@ -12,7 +12,9 @@
 app = typer.Typer(add_completion=False, context_settings={"help_option_names": ["-h", "--help"]})
 
 
-logging.getLogger("airflow").setLevel(logging.CRITICAL)
+airflow_logger = logging.getLogger("airflow")
```
tatiana, line 15: @feluelle did you check out if there would be a way for us to test this behaviour?

feluelle: Good point. Yes, we can.

kaxil: pytest's caplog fixture would be useful

kaxil: https://docs.pytest.org/en/7.1.x/how-to/logging.html

feluelle: Added in ee3da30

sql-cli/tests/test_run_dag.py
```diff
@@ -113,11 +113,19 @@ def movie_ends():
...
-def test_run_dag(capsys):
+def test_run_dag(capsys, caplog):
     dag = get_dag(dag_id="example_dataframe", subdir=f"{CWD}/test_dag")
     run_dag(dag)
     captured = capsys.readouterr()
     assert "The worst month was" in captured.out
+    assert not any(record.name == "airflow.task" for record in caplog.records)
+
+
+def test_run_dag_verbose(capsys, caplog):
```
feluelle, line 124:
```suggestion
def test_run_dag_verbose(capsys):
```

tatiana (approved):
LGTM, thanks a lot, @feluelle !

PR #1160, opened 2022-10-31
Row count not working for snowflake while calling from Openlineage

### LINKED ISSUE #1145: Row count not working for snowflake
**Describe the bug**
Currently row count fails for snowflake 

**Expected behavior**
Row count should work snowflake

rajaths010494:
# Description
## What is the current behaviour?
Currently, row count fails for snowflake

## What is the new behaviour?
Row count should work snowflake and other databases as well

closes #1145

### DIFF
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -133,7 +133,7 @@ class BaseTable:
         """
         db = create_database(self.conn_id)
         table_name = db.get_table_qualified_name(self)
-        result = db.run_sql(f"select count(*) from {table_name};").scalar()
+        result = db.run_sql("select count(*) from %s;" % table_name).scalar()
         return result
 
     def to_json(self):
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -133,8 +133,8 @@ class BaseTable:
         """
         db = create_database(self.conn_id)
         result = db.run_sql(
-            f"select count(*) from {db.get_table_qualified_name(self)}"
-        ).scalar()  # skipcq: BAN-B608
+            f"select count(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
+        ).scalar()
         return result
 
     def to_json(self):
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -5,7 +5,7 @@ import string
 from typing import Any
 
 from attr import define, field, fields_dict
-from sqlalchemy import Column, MetaData, func, select
+from sqlalchemy import Column, MetaData
 
 from astro.airflow.datasets import Dataset
 from astro.databases import create_database
@@ -132,9 +132,8 @@ class BaseTable:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
-        tb = db.get_sqla_table(table=self)
-        query = select(func.count("*")).select_from(tb)
-        return db.run_sql(query).scalar()
+        result = db.run_sql(f"select count(*) from {db.get_table_qualified_name(self)}").scalar()
+        return result
 
     def to_json(self):
         return {
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -132,8 +132,9 @@ class BaseTable:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
-        table_name = db.get_table_qualified_name(self)
-        result = db.run_sql("select count(*) from %s;" % table_name).scalar()
+        result = db.run_sql(
+            f"select count(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
+        ).scalar()
         return result
 
     def to_json(self):
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -132,9 +132,8 @@ class BaseTable:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
-        result = db.run_sql(
-            f"select count(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
-        ).scalar()
+        table_name = db.get_table_qualified_name(self)
+        result = db.run_sql(f"select count(*) from {table_name};").scalar()
         return result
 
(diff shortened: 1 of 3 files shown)


python-sdk/tests/utils/test_table.py
```diff
@@ -94,16 +95,39 @@ def test_find_first_table_with_xcom_arg(xcom_arg_resolve, kwargs, return_type):
...
         task_id="load_file",
         input_file=File(
             path="https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb_v2.csv"
         ),
-        output_table=Table(conn_id="gcp_conn", metadata=Metadata(schema="astro")),
+        output_table=test_table,
     ).execute({})
 
-    assert imdb_table.row_count > 0
+    assert imdb_table.row_count == 117
```
pankajastro, line 133: I still feel `>` check would be more stable

python-sdk/src/astro/table.py
```diff
@@ -132,9 +132,8 @@ def row_count(self) -> Any:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
-        tb = db.get_sqla_table(table=self)
-        query = select(func.count("*")).select_from(tb)
-        return db.run_sql(query).scalar()
+        result = db.run_sql(f"select count(*) from {db.get_table_qualified_name(self)}").scalar()
```
pankajastro, line 135: looking fine to me. But just wondering why the old query didn't work maybe adding error logs on the ticket would be helpful for me to understand the issue.

pankajastro: concatenating a string to build a query is unsafe since it is easy to inject the wrong query though it is not that important here.
you will have to add `# skipcq: BAN-B608` to pass CI

pankajastro (approved):

python-sdk/tests/utils/test_table.py
rajaths010494: changed

python-sdk/src/astro/table.py
rajaths010494: image
it was failing silently in Open lineage but not all the queries getting were generated for the sqla.

python-sdk/src/astro/table.py
```diff
@@ -132,9 +132,10 @@ def row_count(self) -> Any:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
-        tb = db.get_sqla_table(table=self)
-        query = select(func.count("*")).select_from(tb)
-        return db.run_sql(query).scalar()
+        result = db.run_sql(
+            f"select count(*) from {db.get_table_qualified_name(self)}"  # skipcq: BAN-B608
```
utkarsharma2, line 136: @rajaths010494 can we not skip this?

python-sdk/tests/utils/test_table.py
```diff
@@ -94,16 +95,39 @@ def test_find_first_table_with_xcom_arg(xcom_arg_resolve, kwargs, return_type):
...
     imdb_table = LoadFileOperator(
         task_id="load_file",
         input_file=File(
             path="https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb_v2.csv"
         ),
-        output_table=Table(conn_id="gcp_conn", metadata=Metadata(schema="astro")),
+        output_table=test_table,
     ).execute({})
 
     assert imdb_table.row_count > 0
```
utkarsharma2, line 133: @rajaths010494 can we assert it to a specific number?

python-sdk/src/astro/table.py
rajaths010494: ok will fix this

python-sdk/tests/utils/test_table.py
rajaths010494: https://github.com/astronomer/astro-sdk/pull/1160#discussion_r1010132118 i had asserted to specific number then i got a comment so I changed.
cc @pankajastro

pankajastro: @rajaths010494 @utkarsharma2 I'm ok with `==` also but I was just thinking in case of equal if someone will update `imdb_v2.csv` file test will start failing but not a big concern.

python-sdk/src/astro/table.py
rajaths010494: As discussed i am currently skipping this.

pankajastro:
for some reason, I can see only 6 jobs have run on the latest commit. @rajaths010494 do we again need to add a label?

python-sdk/src/astro/table.py
```diff
@@ -132,9 +132,10 @@ def row_count(self) -> Any:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
```
sunank200, line 134: Is the `conn_id` always available? For example this is the error i getting as `conn_id` is not available for transform operator
```
  File "/usr/local/airflow/astro_sdk/src/astro/table.py", line 173, in openlineage_dataset_namespace
    database = create_database(self.conn_id)
  File "/usr/local/airflow/astro_sdk/src/astro/databases/__init__.py", line 30, in create_database
    conn_type = BaseHook.get_connection(conn_id).conn_type
  File "/usr/local/lib/python3.9/site-packages/airflow/hooks/base.py", line 70, in get_connection
    conn = Connection.get_connection_from_secrets(conn_id)
  File "/usr/local/lib/python3.9/site-packages/airflow/models/connection.py", line 432, in get_connection_from_secrets
    raise AirflowNotFoundException(f"The conn_id `{conn_id}` isn't defined")
airflow.exceptions.AirflowNotFoundException: The conn_id `` isn't defined
```

pankajastro: @sunank200  does it make sense to have a table object without conn_id? if not then we should check this in the `openlineage_dataset_namespace` method before calling `row_count`

pankajastro:
why code coverage not running here?

python-sdk/src/astro/table.py
rajaths010494: https://github.com/astronomer/astro-sdk/pull/1181

rajaths010494:
> why code coverage not running here?

How do i rerun code coverage?

PR #1161, opened 2022-10-31
Make the root README consistent on Pickling

kaxil:
There was inconsistency between the README in the root of the repo and the one in the python-sdk folder


tatiana (approved):
@kaxil Would it be better if we had a symbolic link to avoid this type of issue in future?

kaxil:
> @kaxil Would it be better if we had a symbolic link to avoid this type of issue in future?

I like your proposal, made the change

Issue #1173, opened 2022-11-01
Mosts of the tests in example DAGs are not running


sunank200:
**Describe the bug**
Currently the DEFAULT_DATE  is set to `timezone.datetime(2016, 1, 1)` in `python-sdk/tests/sql/operators/utils.py` So all the DAGs in example DAGs have datetime set to after 2016 are not running and they are marked as success.

**Expected behavior**
Change to  `DEFAULT_DATE = timezone.utcnow()`

PR #1174, opened 2022-11-01
Change the default date to utcnow()

sunank200:
# Description

Currently the DEFAULT_DATE is set to `timezone.datetime(2016, 1, 1)` in `python-sdk/tests/sql/operators/utils.py` So all the DAGs in example DAGs have datetime set to after 2016 are not running and they are marked as success.

closes: #1173 

## What is the new behavior?

- Change to DEFAULT_DATE = timezone.utcnow()

Note: We should merge this change to main ASAP as other tests are not affected.

No

### Checklist


kaxil (approved):
We expect a lot of failures but that is expected and we would merge this PR with failures and fix it in the second one.

PR #1175, opened 2022-11-01
Fix some of the failing tests

kaxil:
Uses a different approach for pulling XCom


utkarsharma2 (approved):

pankajastro (approved):
Let's merge it, since we are fixing other fail cases in different PR's

PR #1176, opened 2022-11-01
Change the date to past date for example dag

utkarsharma2:
# Description
Fixing group 1 - https://github.com/astronomer/astro-sdk/actions/runs/3368325263/jobs/5586745425#step:11:834

### DIFF
--- a/python-sdk/example_dags/example_datasets.py
+++ b/python-sdk/example_dags/example_datasets.py
@@ -34,7 +34,7 @@ imdb_movies_table = Table(name="imdb_movies", conn_id="sqlite_default")
 # [END dataset_table]
 
 top_animations_table = Table(name="top_animation", conn_id="sqlite_default")
-START_DATE = datetime(2022, 9, 1)
+START_DATE = datetime(2016, 9, 1)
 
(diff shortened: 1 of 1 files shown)


phanikumv (approved):

python-sdk/example_dags/example_datasets.py
```diff
@@ -34,7 +34,7 @@
 # [END dataset_table]
 
 top_animations_table = Table(name="top_animation", conn_id="sqlite_default")
-START_DATE = datetime(2022, 9, 1)
+START_DATE = datetime(2016, 9, 1)
```
kaxil, line 37: Why was this needed? The date was already in past, Sep 9, 2022?

utkarsharma2: Nope, START_DATE was used to initialize dag.

Screenshot 2022-11-01 at 7 28 54 PM

utkarsharma2: ahh yes, but weird that fixed the issue in CI. Maybe an intermittent issue

Issue #1040, opened 2022-10-11
Allow users to enable SQL CLI verbose output


tatiana:
**Please describe the feature you'd like to see**

By default, we want the SQL CLI commands to return very clean and tidy outputs so the end-users don't feel overwhelmed. However, we also understand more advanced users would like to see the full logs.

**Describe the solution you'd like**

We'd like to have a flag `--debug` which will set the log level globally for all external modules. This helps especially for debugging internal airflow issues.

Additionally we'd like to have more `--verbose` flags to log more information about airflow init, for example.

**Are there any alternatives to this feature?**

Using environment variables/airflow configuration.

**Additional context**

This issue popped up during the implementation of ticket #941

**Acceptance Criteria**

- [ ] By default, debug is enabled, and we only see the clean output
- [ ] If debug is enabled, we should see INFO logs 
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1177, opened 2022-11-01
Add global debug option and more verbose options to sql-cli

feluelle:
# Description

Currently, we don't have an option for the user to show verbose output to debug.

closes: #1040

## What is the new behavior?

We add a global `--debug` option and more local `--verbose` flags.

- show airflow and other deps debug logs when debug is set
- show raw exceptions when verbose is set
- add verbose log to validate_connections
- use sql-cli logger instead of new logger in run_dag
- add RichHandler instead of StreamHandler to ti logger
- add test for log level
- set log level for ti.log
- show airflow db init stdout/stderr when debug is set
- catch exception in validate connections to continue validating

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -12,8 +12,15 @@ from sql_cli.astro.command import AstroCommand
 from sql_cli.astro.group import AstroGroup
 from sql_cli.cli import config as cli_config
 from sql_cli.cli.utils import resolve_project_dir
-from sql_cli.constants import DEFAULT_BASE_AIRFLOW_HOME, DEFAULT_DAGS_FOLDER, DEFAULT_DATA_DIR, STATE
+from sql_cli.constants import (
+    DEFAULT_BASE_AIRFLOW_HOME,
+    DEFAULT_DAGS_FOLDER,
+    DEFAULT_DATA_DIR,
+    EXT_LOGGER_NAMES,
+    LOGGER_NAME,
+)
 from sql_cli.exceptions import ConnectionFailed, DagCycle, EmptyDag, WorkflowFilesDirectoryNotFound
+from sql_cli.settings import STATE
 from sql_cli.utils.rich import RichHandler, rprint
 
 if TYPE_CHECKING:
@@ -28,7 +35,7 @@ app = typer.Typer(
     rich_markup_mode="rich",
 )
 app.add_typer(cli_config.app, name="config", hidden=True)
-log = logging.getLogger("sql_cli")
+log = logging.getLogger(LOGGER_NAME)
 
 
 def set_verbose_mode(verbose: bool) -> None:
@@ -53,7 +60,7 @@ def set_debug_mode(debug: bool) -> None:
     """
     STATE["debug"] = debug
 
-    for logger_name in ["airflow", "botocore", "snowflake"]:
+    for logger_name in EXT_LOGGER_NAMES:
         logger = logging.getLogger(logger_name)
         logger.propagate = debug
         if debug:
@@ -269,9 +276,11 @@ def init(
         help=f"(Optional) Set the data directory for local file databases such as sqlite. Default: {DEFAULT_DATA_DIR}",
         show_default=False,
     ),
+    verbose: bool = typer.Option(False, help="Whether to show verbose output", show_default=True),
 ) -> None:
     from sql_cli.project import Project
 
+    set_verbose_mode(verbose)
     project_dir_absolute = resolve_project_dir(project_dir)
     project = Project(project_dir_absolute, airflow_home, airflow_dags_folder, data_dir)
     project.initialise()
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -6,7 +6,6 @@ from typing import TYPE_CHECKING
 
 import typer
 from dotenv import load_dotenv
-from rich.logging import RichHandler
 
 import sql_cli
 from sql_cli.astro.command import AstroCommand
@@ -15,7 +14,7 @@ from sql_cli.cli import config as cli_config
 from sql_cli.cli.utils import resolve_project_dir
 from sql_cli.constants import DEFAULT_BASE_AIRFLOW_HOME, DEFAULT_DAGS_FOLDER, DEFAULT_DATA_DIR
 from sql_cli.exceptions import ConnectionFailed, DagCycle, EmptyDag, WorkflowFilesDirectoryNotFound
-from sql_cli.utils.rich import rprint
+from sql_cli.utils.rich import RichHandler, rprint
 
 if TYPE_CHECKING:
     from sql_cli.project import Project  # pragma: no cover
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING
 
 import typer
 from dotenv import load_dotenv
-from typer import Exit
+from rich.logging import RichHandler
 
 import sql_cli
 from sql_cli.astro.command import AstroCommand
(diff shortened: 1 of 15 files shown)


sql-cli/tests/test___main__.py
```diff
@@ -127,6 +128,37 @@ def test_generate_invalid(workflow_name, message, initialised_project_with_tests
...
+            "generate",
+            workflow_name,
+            "--project-dir",
+            initialised_project_with_tests_workflows.directory.as_posix(),
+            "--verbose",
+        ],
+    )
+    assert result.exit_code == exit_code
+    if result.exit_code != 0:
+        assert result.exception
```
tatiana, line 159: It is probably worth adding some checks on the output, which is only printed when the verbose mode is enabled

sql-cli/tests/test___main__.py
```diff
@@ -165,6 +197,32 @@ def test_validate_all(initialised_project_with_test_config):
...
+            "--connection",
+            connection,
+            "--verbose",
+        ],
+    )
+    assert result.exit_code == 0
+    output = get_stdout(result)
+    assert f"Validating connection(s) for environment '{env}'" in output
+    assert f"Validating connection {connection:{CONNECTION_ID_OUTPUT_STRING_WIDTH}} {status}" in output
+    assert f"  {message}" in output
```
tatiana, line 223: Are there any other statements we'd expect to see when verbose is enabled for the `validate` command?

sql-cli/tests/test___main__.py
```diff
@@ -267,6 +325,43 @@ def test_run_invalid(workflow_name, message, initialised_project_with_tests_work
...
+            "run",
+            workflow_name,
+            "--project-dir",
+            initialised_project_with_tests_workflows.directory.as_posix(),
+            "--verbose",
+        ],
+    )
+    assert result.exit_code == exit_code
+    if result.exit_code != 0:
+        assert result.exception
```
tatiana, line 362: Would this test fail if we did not have `--verbose`?

sql-cli/tests/test___main__.py
feluelle: I don't think so right now. If you want to see airflow logs, you would additionally need to run with `--debug`.

sql-cli/tests/test___main__.py
feluelle: Yes, thanks 👍 Will add more assertions.

sql-cli/sql_cli/connections.py
```diff
@@ -49,17 +52,19 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
...
             os.environ[f"AIRFLOW_CONN_{connection.conn_id.upper()}"] = connection.get_uri()
-            status = (
-                "[bold green]PASSED[/bold green]" if _is_valid(connection) else "[bold red]FAILED[/bold red]"
-            )
+            success_status, message = _is_valid(connection)
+            status = "[bold green]PASSED[/bold green]" if success_status else "[bold red]FAILED[/bold red]"
             rprint(f"Validating connection {connection.conn_id:{CONNECTION_ID_OUTPUT_STRING_WIDTH}}", status)
+            formatted_message = (
+                f"[bold green]{message}[/bold green]" if success_status else f"[bold red]{message}[/bold red]"
+            )
```
feluelle, line 60: @tatiana it probably makes sense to always print that message, no? 🤔 Otherwise we are not giving the users any hints on what's wrong with it.

sql-cli/sql_cli/__main__.py
```diff
@@ -29,16 +28,38 @@
...
+    """
+    Sets the logging level for all external dependencies.
+
+    :param debug: Whether debug logs should be enabled.
+    """
+    STATE["debug"] = debug
 
-snowflake_logger = logging.getLogger("snowflake")
-snowflake_logger.setLevel(logging.CRITICAL)
+    for logger_name in ["airflow", "botocore", "snowflake"]:
```
tatiana, line 56: It's probably worth having a constant with the values of the loggers we want to handle this way

sql-cli/sql_cli/connections.py
tatiana: Yes, it does, @feluelle ! We should keep the the SQL CLI standard output regardless of the verbose mode :)

sql-cli/sql_cli/connections.py
```diff
@@ -49,17 +52,19 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
...
+            log.debug("Connection Message: %s", formatted_message)
 
 
-def _is_valid(connection: Connection) -> bool:
+def _is_valid(connection: Connection) -> tuple[bool, str]:
     # Sqlite automatically creates the file if it does not exist,
     # but our users might not expect that. They are referencing a database they expect to exist.
     if connection.conn_type == "sqlite" and not Path(connection.host).is_file():
-        return False
+        return False, "Sqlite db does not exist!"
```
tatiana, line 68: Perhaps we could use the opportunity to cover this line with a test!

sql-cli/sql_cli/connections.py
```diff
@@ -10,6 +11,8 @@
 
 CONNECTION_ID_OUTPUT_STRING_WIDTH = 25
 
+log = logging.getLogger("sql_cli")
```
tatiana, line 14: It may make sense to have a constant with the `LOGGER_NAME`

sql-cli/sql_cli/__main__.py
```diff
@@ -150,19 +179,24 @@ def run(
...
-    verbose: bool = typer.Option(False, help="Whether to show airflow logs", show_default=True),
+    verbose: bool = typer.Option(False, help="Whether to show verbose output", show_default=True),
 ) -> None:
     from airflow.utils.state import State
 
     from sql_cli.dag_runner import run_dag
     from sql_cli.project import Project
     from sql_cli.utils.airflow import get_dag
 
+    set_verbose_mode(verbose)
```
tatiana, line 190: Don't we also need to add `set_verbose_mode(verbose)` to the init command?

sql-cli/sql_cli/constants.py
```diff
@@ -1,3 +1,5 @@
+STATE = {"debug": False}
```
tatiana, line 1: Since this value should change based on user configuration, it may make sense for us not to have it in `constants.py` but in a `settings.py` module, WDYT?

tatiana (commented):
Hi @feluelle ! I think we're very close to have this PR merged.

Some minor comments in-line, and two broader points:

1. Since we created the ticket, the scope has changed significantly. What do you think of updating the ticket description to explain why we have both debug and verbose, with examples of when we want to use each of them?

2. From now on, SQL CLI developers will need to keep in mind the following:
  - to use the correct logger ("sql_cli")
  - to use `set_verbose_mode` in the commands that we want to have this option (can you think of any scenarios we don't want it to?)
  - to understand how the log level used relates to the verbose option being set or not:
       - verbose=True: will display log  messages with level =< DEBUG
       - verbose=False: will display log messages with level >= CRITICAL

What do you think we adding a minor beginning of SQL CLI developer documentation so any new team members could benefit from this information?

sql-cli/sql_cli/__main__.py
tatiana: By the time this PR has been open, I'm also happy for these improvements to be done on a separate PR. It would be great to have a new ticket for this work before closing the existing one/merging this PR, though.

sql-cli/sql_cli/connections.py
feluelle: This is being removed in https://github.com/astronomer/astro-sdk/pull/1514.

sql-cli/sql_cli/__main__.py
feluelle: Great idea! I will add more logging to initialise.

feluelle:
> Since we created the ticket, the scope has changed significantly. What do you think of updating the ticket description to explain why we have both debug and verbose, with examples of when we want to use each of them?

Done - updated the linked ticket.

> to use set_verbose_mode in the commands that we want to have this option (can you think of any scenarios we don't want it to?)

As soon as we have something verbose to output, we can add that flag. Unfortunately this has to be added per command as global args such as debug have to come before the flow command i.e. `flow --debug init`.

> What do you think we adding a minor beginning of SQL CLI developer documentation so any new team members could benefit from this information?

Makes sense. Would it a follow-up ticket work for you?

tatiana (approved):
Thanks a lot for addressing all the feedback, @feluelle ! This will make our lives easier!

PR #1180, opened 2022-11-02
Implement Google Drive file location

### LINKED ISSUE #1044: Support for Google Drive as File source
Create support for google drive as a file location. A file(s) kept in google drive would be now loaded into the databases and can perform operations on it. 
For this, we use the Google API client to connect and retrieve the file.

One endpoint is used:

* [`files().list()`](https://developers.google.com/drive/api/v3/reference/files/list) to find files.

To use those file APIs you need to kind of visualise traversing the folders in a GUI. To get a file by path, you need to see what items are available in a folder and select the one with the name you want. If the item is a folder, you need to run the list against it again to see what items are in it. Not very efficient, but there are not many choices.

uranusjr:
Close #1044 (eventually?)

This mostly just uses the Google API client. Direct documentation on the client is sparse, but you can find the general HTTP API documentation (which backs the Python API).

Two endpoints are used:

* [`files().list()`](https://developers.google.com/drive/api/v3/reference/files/list) to find files.
* [`files().get()`](https://developers.google.com/drive/api/v3/reference/files/get) to retrieve file metadata.

To use those file APIs you need to kind of visualise traversing the folders in a GUI. To get a file by path, you need to see what items are available in a folder, and select the one with the name you want. If the item is a folder, you need to run list against it again to see what items are in it. Not very efficient, but there are not many choices.

I added some simple tests for `size`.

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/files/locations/google/gdrive.py
+++ b/python-sdk/src/astro/files/locations/google/gdrive.py
@@ -1,0 +1,131 @@
+from __future__ import annotations
+
+import pathlib
+import urllib.parse
+from typing import Any, Iterator, Sequence
+
+from airflow.providers.google.suite.hooks.drive import GoogleDriveHook
+
+from astro.constants import FileLocation
+from astro.files.locations.base import BaseFileLocation
+
+
+def _quote(s: str) -> str:
+    s = s.replace("'", "\\'")  # Escape single quote characters.
+    return f"'{s}'"  # Quote.
+
+
+def _find_item_id(path_parts: Sequence[str], *, conn: Any) -> str:
+    """Drill down the directory hierarchy to find the ID of an item.
+
+    Google Drive does not have a very good random access API by path, so we
+    must look at each directory, level by level, to find the item we want.
+    """
+    folder_id = "root"
+    for part in path_parts:
+        command = conn.files().list(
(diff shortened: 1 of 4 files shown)


python-sdk/src/astro/files/locations/google/gdrive.py
```diff
@@ -0,0 +1,131 @@
...
+from astro.constants import FileLocation
+from astro.files.locations.base import BaseFileLocation
+
+
+def _quote(s: str) -> str:
+    s = s.replace("'", "\\'")  # Escape single quote characters.
+    return f"'{s}'"  # Quote.
+
+
+def _find_item_id(path_parts: Sequence[str], *, conn: Any) -> str:
```
tatiana, line 18: Would it make sense to use the Drive hook's `get_file_id` method?
https://github.com/apache/airflow/blob/main/airflow/providers/google/suite/hooks/drive.py#L151

uranusjr: The function needs `folder_id`, so we still needs to call it recursively (or in a loop). It is more similar to one single `conn.files().list()` call in this function, but containing more fields and thus less efficient.

kaxil:
closed in favor of https://github.com/astronomer/astro-sdk/pull/1322

Issue #1152, opened 2022-10-28
Conn ID not found on Extraction with the @aql.transform() Decorator


ReadytoRocc:
**Describe the bug**
On a successful task completion with the `@aql.transform()` decorator, extraction is failing when utilizing the `astro.lineage.extractor.PythonSDKExtractor`.

**Version**
* Astro: 1.2.0
* astro-runtime:6.0.2

**To Reproduce**
1. Create a connection to a Snowflake Instance.
2. Write a DAG using the `@aql.transform()` operator, utilizing this connection.

**Expected behavior**
The extractor is able to find the `conn_id` of the task.

**Logs**
```
[2022-10-26, 02:40:59 UTC] {taskinstance.py:1401} INFO - Marking task as SUCCESS. dag_id=DAG_ID, task_id=TASK_ID, execution_date=20221021T202905, start_date=20221026T024051, end_date=20221026T024059
[2022-10-26, 02:40:59 UTC] {manager.py:63} ERROR - Failed to extract metadata The conn_id `` isn't defined task_type=TransformOperator airflow_dag_id=DAG_ID task_id=TASK_ID airflow_run_id=manual__2022-10-21T20:29:05.395735+00:00 
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/extractors/manager.py", line 46, in extract_metadata
    task_metadata = extractor.extract_on_complete(task_instance)
  File "/usr/local/lib/python3.9/site-packages/astro/lineage/extractor.py", line 59, in extract_on_complete
    open_lineage_facets: OpenLineageFacets = self.operator.get_openlineage_facets(task_instance)
  File "/usr/local/lib/python3.9/site-packages/astro/sql/operators/base_decorator.py", line 203, in get_openlineage_facets
    f"{self.output_table.openlineage_dataset_namespace()}"
  File "/usr/local/lib/python3.9/site-packages/astro/table.py", line 173, in openlineage_dataset_namespace
    database = create_database(self.conn_id)
  File "/usr/local/lib/python3.9/site-packages/astro/databases/__init__.py", line 30, in create_database
    conn_type = BaseHook.get_connection(conn_id).conn_type
  File "/usr/local/lib/python3.9/site-packages/airflow/hooks/base.py", line 70, in get_connection
    conn = Connection.get_connection_from_secrets(conn_id)
[...]

PR #1181, opened 2022-11-02
Fix the conn_id missing in output table

sunank200:
# Description

`conn_id` is missing in output table for transform operator. It should be checked for that

closes: #1152

## What is the new behavior?

- Check if `conn_id` is empty
- Push row count of output table to xcom
- Fixed some flaky test

No

### Checklist


python-sdk/src/astro/sql/operators/transform.py
```diff
@@ -60,6 +60,7 @@ def execute(self, context: Context):
             target_table=self.output_table,
             parameters=self.parameters,
         )
+        context["ti"].xcom_push(key="output_table_row_count", value=str(self.output_table.row_count))
```
pankajastro, line 63: does this only need for the transform operator only and not for other operators inherited from `BaseSQLDecoratedOperator`?

pankajastro: since if the output table is not passed then BaseSQLDecoratedOperator  assign it to Table()

rajaths010494: Yes, currently only for this operator open lineage is failing to fetch row_count

pankajastro: hmm, I see, look like we do not have lineage for `run_raw_sql`

pankajastro (approved):
LGTM

phanikumv (approved):

PR #1185, opened 2022-11-02
Upgrade snowflake-sqlalchemy package

utkarsharma2:
# Description
Due to [changes](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/350) in `sqlalchemy==1.4.42` our CI is failing, since snowflake-sqlalchemy's rfc imports are failing and we cannot change the version of snowflake-sqlalchemy to anything other than 1.2.4 because airflow 2.2.5 has a constraint.
## What is the new behavior?
Upgraded the package to 1.4.3 where the fix is added.

Nope

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -59,7 +59,7 @@ google = [
 ]
 snowflake = [
     "apache-airflow-providers-snowflake",
-    "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
+    "snowflake-sqlalchemy>=1.4.3,<1.5.0",
     "snowflake-connector-python[pandas]",
 ]
 postgres = [
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -79,7 +79,7 @@ all = [
     "apache-airflow-providers-snowflake",
     "smart-open[all]>=5.2.1",
     "snowflake-connector-python[pandas]",
-    "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
+    "snowflake-sqlalchemy>=1.4.3,<1.5.0",
     "sqlalchemy-bigquery>=1.3.0",
     "s3fs",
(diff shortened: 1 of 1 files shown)


python-sdk/pyproject.toml
```diff
@@ -59,7 +59,7 @@ google = [
 ]
 snowflake = [
     "apache-airflow-providers-snowflake",
-    "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
+    "snowflake-sqlalchemy>=1.4.3,<1.5.0",
```
pankajastro, line 62: does it cause any problems if we increase the version from 1.2.0 to 1.4.3 and a customer has some other things  which depend on 1.2.0 in the env?

python-sdk/pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     "pyarrow",
     "python-frontmatter",
     "smart-open",
-    "SQLAlchemy>=1.3.18"
+    "SQLAlchemy>=1.3.18,<1.4.42"
```
pankajastro, line 26: why we are restricting maybe a comment would be helpful

utkarsharma2: We have it as part of the PR description, but no harm in adding in the code as well, will do.

pankajastro (approved):

feluelle (commented):
Should we have a ticket to unpin this? Do we need the `rfc` imports?

utkarsharma2:
@feluelle 
1. We should add the ticket.
2. So we don't directly use `rfc` imports, `snowflake-sqlalchemy` uses it and we cannot change the version of `snowflake-sqlalchemy` to anything other than 1.2.4 because airflow 2.2.5 has a constraint.

feluelle:
Thanks for the explanation. Could you add the 2nd statement to the PR description, please?

utkarsharma2:
@feluelle Added it in the PR and comment

PR #1186, opened 2022-11-02
check in .airflowignore file

pankajastro:
# Description
We don't check in the .airflowignore file but this get used in the test

## What is the new behavior?
check-in the .airflowignore so that it tests respect the intended behaviour 

No

### Checklist


feluelle:
@pankajastro the `.airflowignore` is being created dynamically for the tests. We should not commit it to git.

PR #1192, opened 2022-11-03
Make flow usage environment aware

feluelle:
# Description

Currently, when running flow's `--help` commands they always show the following usage:
```
 Usage: flow [OPTIONS] COMMAND [ARGS]...
```

This is incorrect in case the command is called via astro CLI.

closes: #1190 

## What is the new behavior?

We check for the environment the command is running in and based on that format the usage output. The default will still be showing `flow` whereas when it is running in the astro cli it will show `astro flow`.

E.g.
```
Usage: astro flow [OPTIONS] COMMAND [ARGS]...
```

No.

### Checklist

### DIFF
--- a/.github/workflows/ci-sql-cli.yaml
+++ b/.github/workflows/ci-sql-cli.yaml
@@ -94,6 +94,7 @@ jobs:
     env:
       AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
       AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV: True
+      _TYPER_FORCE_DISABLE_TERMINAL: 1
 
 
   Code-Coverage:
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -16,6 +16,7 @@ app = typer.Typer(
     cls=AstroGroup,
     add_completion=False,
     context_settings={"help_option_names": ["-h", "--help"]},
+    rich_markup_mode="rich",
 )
 
 
@@ -157,30 +158,18 @@ def run(
     help="""
     Initialise a project structure to write workflows using SQL files.
 
-    \b\n
     Examples of usage:
-    \b\n
     $ flow init
-    \b\n
     $ flow init .
-    \b\n
     $ flow init project_name
 
-
-    \b\n
     By default, the project structure includes:
-
     ├── config: withholds configuration, e.g. database connections, within each environment directory
-    \b\n
     ├── data: directory which contains datasets, including SQLite databases used by the examples
-    \b\n
     └── workflows: directory where SQL workflows are declared, by default has two examples of workflow
 
-    \b\n
     Next steps:
-    \b\n
     * Update the file `config/default/configuration.yaml` to declare database connections.
-    \b\n
     * Create SQL workflows within the `workflows` folder.
     """,
 )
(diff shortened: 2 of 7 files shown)


.github/workflows/ci-sql-cli.yaml
```diff
@@ -94,6 +94,7 @@ jobs:
     env:
       AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
       AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV: True
+      _TYPER_FORCE_DISABLE_TERMINAL: 1
```
feluelle, line 97: This is also being used by typer for testing. See [here](https://github.com/tiangolo/typer/blob/master/scripts/test.sh#L8-L9).

feluelle: This seems to be required for tests which print the help output such as `test_usage` below.

sql-cli/sql_cli/__main__.py
```diff
@@ -6,28 +6,45 @@
...
 
 import sql_cli
+from sql_cli.astro.command import AstroCommand
+from sql_cli.astro.group import AstroGroup
 from sql_cli.constants import DEFAULT_AIRFLOW_HOME, DEFAULT_DAGS_FOLDER
 
 load_dotenv()
-app = typer.Typer(add_completion=False, context_settings={"help_option_names": ["-h", "--help"]})
+app = typer.Typer(
+    name="flow",
```
feluelle, line 15: This is useful for testing the help output as it contains the program name which if not set is `root`.

sql-cli/sql_cli/__main__.py
```diff
@@ -6,28 +6,45 @@
...
+    cls=AstroGroup,
+    add_completion=False,
+    context_settings={"help_option_names": ["-h", "--help"]},
+    rich_markup_mode="rich",
```
feluelle, line 19: This solves the issue below with the `\b\n` for new lines in the help output + we can use rich formatting in the help if needed.

feluelle: [Look](https://typer.tiangolo.com/tutorial/commands/help/#rich-markdown-and-markup) how beautiful it could look like 🤩

feluelle:
@tatiana @dimberman @pankajkoti please review :)

sql-cli/sql_cli/astro/command.py
```diff
@@ -0,0 +1,18 @@
...
+
+class AstroCommand(TyperCommand):
+    def format_usage(self, ctx: typer.Context, formatter: click.HelpFormatter) -> None:
+        """Writes the usage line into the formatter.
+
+        This is a low-level method called by :meth:`get_usage`.
+        """
+        pieces = self.collect_usage_pieces(ctx)
+        command_path = ctx.command_path
+        if os.getenv("ASTRO_CLI"):
```
pankajkoti, line 16: Perhaps we can add a comment here that this environment variable is set in the astro binary. Or we can reference https://github.com/astronomer/astro-cli/pull/855 somewhere? WDYT?

pankajkoti (approved):

sql-cli/sql_cli/astro/command.py
feluelle: Makes sense 👍

sql-cli/sql_cli/__main__.py
tatiana: This looks amazing, @feluelle ! :clap: :clap:

PR #1193, opened 2022-11-03
Use postgres instead of sqlite for dataset example

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1197

Dataset example test failing frequently with error 

```
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: imdb_movies
E       [SQL: CREATE TABLE IF NOT EXISTS top_animation AS 
E               SELECT title, rating
E               FROM imdb_movies
E               WHERE genre1='Animation'
E               ORDER BY rating desc
E               LIMIT 5;
E           ]
E       (Background on this error at: https://sqlalche.me/e/14/e3q8)

.nox/test-3-8-airflow-2-4/lib/python3.8/site-packages/sqlalchemy/engine/default.py:719: OperationalEr
```

## What is the new behavior?
Use Postgres instead of sqlite  

No

### Checklist


pankajastro:
> @pankajastro How will replacing DB help here?

I feel sqlite might be doing something not right, maybe a lock, when consumer dag started there might be still a lock on sqlite.db because of a transaction by producer DAG and as a result producer DAG tried to create the db file which does not have a table. This is what I suspect.

utkarsharma2 (approved):
@pankajastro Do we know the exact reason or it's a hunch?

phanikumv (approved):

pankajastro:
> @pankajastro Do we know the exact reason or it's a hunch?

hunch 😄

Issue #890, opened 2022-09-22
Issue with loading files using LoadFileOperator with native_support=True


phanikumv:
**Describe the bug**
Check why the `'use_native_support'=True` isn't loading all the files with the LoadFileOperator
Relevant thread https://astronomer.slack.com/archives/C02B8SPT93K/p1663783869151809
************

I'm seeing an issue in my job with the LoadFileOperator
The file was a GCS Parquet file - I physically checked and there is data in it
The table was a Snowflake TempTable - I checked and it's empty
It then gets merged with a MergeOperator, but given that the temp table was empty, nothing was added
No errors in the logs as far as I can tell.
I have use_native_support=True - should I turn it off and see if that helps?
This happened only occasionally - I loaded ~80 files each for 4 different tables (as fast as airflow could load them), and only some loads were affected
It does appear to happen consistently with the same records missing for at least one case if I rerun the tasks though
************

**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Write the DAG  '...'
2. Create connection '....'
3. Run using  '....'

**Expected behavior**

**Screenshots**

**Additional context**

utkarsharma2:
@fritz-astronomer I'm not able to reproduce this on my local, can you share the details like the GCS files and Snowflake Table details OR we can get on a call to try this out?

fritz-astronomer:
@utkarsharma2 following up in slack

utkarsharma2:
Need to try on the env @fritz-astronomer has shared with me.

fritz-astronomer:
Thank you!!!!

PR #1194, opened 2022-11-03
Bubble error when copy into query fails in snowflake

utkarsharma2:
# Description
Currently, we were not bubbling up the error raised by copying into command in snowflake native path GCS -> Snowflake.

closes: #890

## What is the new behavior?
Now we pass a handler that checks for the status of the query returned and if any is failing raise a `DatabaseCustomError` error.

Nope

### Checklist


python-sdk/src/astro/databases/snowflake.py
```diff
@@ -595,11 +597,20 @@ def load_file_to_table_natively(
...
             raise DatabaseCustomError from exe
         self.drop_stage(stage)
 
+    @staticmethod
+    def check_for_error(cur):
+        """
+        Handler to check the error state returned by snowflake when running `copy into` query.
+        """
+        results = cur.fetchall()
+        if any([True for result in results if result["status"] == COPY_INTO_COMMAND_FAIL_STATUS]):
```
feluelle, line 611:
```suggestion
        if any(result["status"] == COPY_INTO_COMMAND_FAIL_STATUS for result in results):
```

feluelle: Don't use `[...]` in `any` as this would first create a whole list of items before `any` actually runs.

The advantage of `any` is that it allows you to pass an iterable which also can be a generator which means that `any` will report `True` as soon as the first condition is `True`.

utkarsharma2: @feluelle Thanks, didn't know it.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -602,12 +604,29 @@ def load_file_to_table_natively(
...
         file_path = os.path.basename(source_file.path) or ""
         sql_statement = f"COPY INTO {table_name} FROM @{stage.qualified_name}/{file_path}"
+
+        # Below code is added due to breaking change in apache-airflow-providers-snowflake==3.2.0,
+        # we need to pass handler param to get the rows. But in version apache-airflow-providers-snowflake==3.1.0
+        # if we pass the handler provider raises an exception AttributeError
         try:
-            self.hook.run(sql_statement)
-        except (ValueError, AttributeError) as exe:
+            rows = self.hook.run(sql_statement, handler=lambda cur: cur.fetchall())
```
pankajastro, line 612: just for my understanding why do we need to fetch all always?

feluelle (approved):

python-sdk/src/astro/databases/snowflake.py
pankajastro: ok, you are checking individual row status

PR #1129, opened 2022-10-21
Improve error handling for undefined variables and cycles in Workflows

feluelle:
# Description

## What is the current behavior?

Currently, we are not catching the exception when a user has undefined variables or cycles in his workflows.

closes: #1127

## What is the new behavior?

Improve error handling for undefined variables and cycles in Workflows

- moved code for generate and run into separate functions due to too high complexity reported by flake8
- improve output for dag generation

## Does this introduce a breaking change?

No.

### Checklist
- [x] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1199, opened 2022-11-04
Revert UI changes from PR #1129

feluelle:
# Description

The flow run command also prints the file generated. This is a change in UI we don't want to have that soon. It needs more communication with stakeholders and users.

related: #1129

cc @tatiana 

## What is the new behavior?

Removes the log output from generate_dag in flow run command.

No.

### Checklist


tatiana (approved):
Looks great, @feluelle , by default SQL Analysts should not have to worry about the DAG generation itself during the `run` command.

One alternative could be to log this information as part of #1040

PR #1200, opened 2022-11-04
Add pre-defined order for dataset example dags

utkarsharma2:
Fix flaky CI test - example_datasets.py dag.


feluelle (commented):
@utkarsharma2 cant we just reorder the dags so that they are being executed in the right order?

Another alternative would be to use pytest-order or pytest-dependency for example and run these two examples separately. WDYT?

Excluding should be the very last solution.

kaxil:
> Excluding should be the very last solution.

100% Agreed

pankajastro:
> @utkarsharma2 cant we just reorder the dags so that they are being executed in the right order?
> 
> Another alternative would be to use pytest-order or pytest-dependency for example and run these two examples separately. WDYT?
> 
> Excluding should be the very last solution.

@feluelle @kaxil  we want to run this DAG as dataset dag so just was thinking if there is a way to run it programmatically, Ideally isn't should be the case that the consumer DAG should wait until the producer succeded automatically?   currently, we are getting it from DAG bag and triggering dag i.e we are running both dag separately which is not our intention here

feluelle:
I understand, but instead of excluding we can run them separately and make sure that they run in the correct order i.e. first the producer than the consumer dag. For example, with [pytest-order](https://pypi.org/project/pytest-order/) you do sth. like this:
```python
import pytest

def test_example_dag_producer():
    assert True

@pytest.mark.run(after='test_example_dag_producer')
def test_example_dag_consumer():
    assert True
```
We can also add the markers in `pytest.param(...)` and conditionally based on dag_id set the marker or not.

feluelle:
@utkarsharma2 @pankajastro WDYT of e70e7ce?

python-sdk/tests/test_example_dags.py
```diff
@@ -72,19 +71,28 @@ def get_dag_bag() -> DagBag:
...
+    "example_dataset_producer",
+    "example_dataset_consumer",
+]
+
+
+def order(dag_id: str) -> int:
+    try:
+        return PRE_DEFINED_ORDER.index(dag_id)
+    except ValueError:
+        return -1
```
feluelle, line 84:
```suggestion
    if dag_id in PRE_DEFINED_ORDER:
        return PRE_DEFINED_ORDER.index(dag_id)
    else:
        return -1
```

pankajkoti (approved):

Issue #1204, opened 2022-11-07
Fix `test_cleanup_non_temp_table`


feluelle:
See failed check's [summary](https://github.com/astronomer/astro-sdk/actions/runs/3395472641/jobs/5645401455#step:11:1364) for more information.

PR #1205, opened 2022-11-07
Make non-Temp tables unique across test runs

feluelle:
# Description

Currently, the `multiple_tables_fixture` and `database_table_fixture` are not working across runs as they are using the table names as passed by the user.

closes: https://github.com/astronomer/astro-sdk/issues/1204

## What is the new behavior?

With the proposed change, we always add a unique suffix to non-Temp tables.

No.

### Checklist

### DIFF
--- a/python-sdk/conftest.py
+++ b/python-sdk/conftest.py
@@ -133,7 +133,7 @@ def database_table_fixture(request):
     database = create_database(conn_id)
     table = params.get("table", Table(conn_id=database.conn_id, metadata=database.default_metadata))
     if not isinstance(table, TempTable):
-        # We add a unique suffix to the table name to make the name unique across runs
+        # We create a unique table name to make the name unique across runs
         table.name = create_unique_table_name(UNIQUE_HASH_SIZE)
     file = params.get("file")
 
@@ -196,7 +196,7 @@ def multiple_tables_fixture(request, database_table_fixture):
     for item in items:
         table = item.get("table", Table(conn_id=database.conn_id))
         if not isinstance(table, TempTable):
-            # We add a unique suffix to the table name to make the name unique across runs
+            # We create a unique table name to make the name unique across runs
             table.name = create_unique_table_name(UNIQUE_HASH_SIZE)
         file = item.get("file")
 
--- a/python-sdk/conftest.py
+++ b/python-sdk/conftest.py
@@ -132,6 +132,9 @@ def database_table_fixture(request):
     conn_id = DATABASE_NAME_TO_CONN_ID[database_name]
     database = create_database(conn_id)
     table = params.get("table", Table(conn_id=database.conn_id, metadata=database.default_metadata))
+    if not isinstance(table, TempTable):
+        # We add a unique suffix to the table name to make the name unique across runs
+        table.name += create_unique_table_name(UNIQUE_HASH_SIZE)
     file = params.get("file")
 
     database.populate_table_metadata(table)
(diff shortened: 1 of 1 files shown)


python-sdk/conftest.py
```diff
@@ -132,6 +132,9 @@ def database_table_fixture(request):
     conn_id = DATABASE_NAME_TO_CONN_ID[database_name]
     database = create_database(conn_id)
     table = params.get("table", Table(conn_id=database.conn_id, metadata=database.default_metadata))
+    if not isinstance(table, TempTable):
+        # We add a unique suffix to the table name to make the name unique across runs
+        table.name += create_unique_table_name(UNIQUE_HASH_SIZE)
```
utkarsharma2, line 137: @feluelle do we account for an upper limit on table name length?

feluelle: No, do we have one? 🤔 I am not aware of any validation on our side. We could probably at some point add one, but this is only for tests and the tests seem to be happy with that change.

pankajkoti: yes, we do. the `create_unique_table_name` function generates the table name setting the table name length limit to 62. But, since we're appending the output of `create_unique_table_name` to the existing name, it might exceed 62 characters. 
Do we need to append the suffix to existing table name or can we directly replace it with 
`table.name = create_unique_table_name(UNIQUE_HASH_SIZE)` . Or if if we need the existing table name then how about do it like this: `table.name = create_unique_table_name(prefix=table.name, UNIQUE_HASH_SIZE)` and redefine the function to accept prefix and prepend it like we do in https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/table.py#L75

feluelle: Thanks @pankajkoti! I think you're right, we actually don't need to append.

feluelle, line 137:
```suggestion
        table.name = create_unique_table_name(UNIQUE_HASH_SIZE)
```

python-sdk/conftest.py
```diff
@@ -192,6 +195,9 @@ def multiple_tables_fixture(request, database_table_fixture):
 
     for item in items:
         table = item.get("table", Table(conn_id=database.conn_id))
+        if not isinstance(table, TempTable):
+            # We add a unique suffix to the table name to make the name unique across runs
+            table.name += create_unique_table_name(UNIQUE_HASH_SIZE)
```
feluelle, line 200:
```suggestion
            table.name = create_unique_table_name(UNIQUE_HASH_SIZE)
```

python-sdk/conftest.py
```diff
@@ -132,6 +132,9 @@ def database_table_fixture(request):
     conn_id = DATABASE_NAME_TO_CONN_ID[database_name]
     database = create_database(conn_id)
     table = params.get("table", Table(conn_id=database.conn_id, metadata=database.default_metadata))
+    if not isinstance(table, TempTable):
+        # We add a unique suffix to the table name to make the name unique across runs
```
feluelle, line 136:
```suggestion
        # We create a unique table name to make the name unique across runs
```

python-sdk/conftest.py
```diff
@@ -192,6 +195,9 @@ def multiple_tables_fixture(request, database_table_fixture):
 
     for item in items:
         table = item.get("table", Table(conn_id=database.conn_id))
+        if not isinstance(table, TempTable):
+            # We add a unique suffix to the table name to make the name unique across runs
```
feluelle, line 199:
```suggestion
            # We create a unique table name to make the name unique across runs
```

pankajkoti (approved):

PR #1208, opened 2022-11-07
Fix or Skip `Deepsource` warning

pankajastro:
# Description
- BAN-B608: We are concatenating the user input table name to build the query string for now I have wrapped the table/schema name with a double quote 
- PYL-W0613: Skipping it because most of the unused param exists because the parent method has it and we are overloading 
- PYL-C0209: Fix some string-building statements but if a string has `{{` double curly bracket then the f-string was behaving weirdly so I'm skipping some 
- PYL-W0212: Mostly skipped because it third party lib private property and we do not have direct control over it.
-  PY-W2000: Fix it
- PTC-W0062: Fix it
- PTC-W0048: Fix it
- PYL-W1641: Fix it. Add `__hash__`
- PYL-W1508: Fix it
- PYL-W0223: Skip it. Anyways we will remove this class in future
- PYL-R0201: Skip it since we do not want to keep them static at least in this PR
- PYL-W0108: Skip it since we do not have a test for benchmarking so it would be difficult to catch any bug here

## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/python-sdk/dev/scripts/pre_commit_context_typing_compat.py
+++ b/python-sdk/dev/scripts/pre_commit_context_typing_compat.py
@@ -27,14 +27,16 @@ class ImportCrawler(NodeVisitor):
             return
 
         for alias in node.names:
-            if f"{node.module}.{alias.name}" == "airflow.utils.context.Context":
-                if not self.has_incompatible_context_imports:
-                    self.has_incompatible_context_imports = True
+            if (
+                f"{node.module}.{alias.name}" == "airflow.utils.context.Context"
+                and not self.has_incompatible_context_imports
+            ):
+                self.has_incompatible_context_imports = True
 
 
 def get_all_provider_files() -> list[Path]:
     """Retrieve all eligible provider module files."""
-    provider_files = []
+    _provider_files = []
     for (root, _, file_names) in os.walk(ASTRO_ROOT):
         for file_name in file_names:
             file_path = Path(root, file_name)
@@ -43,9 +45,9 @@ def get_all_provider_files() -> list[Path]:
                 and file_path.name.endswith(".py")
                 and TYPING_COMPAT_PATH not in str(file_path)
             ):
-                provider_files.append(file_path)
+                _provider_files.append(file_path)
 
-    return provider_files
+    return _provider_files
 
 
 def find_incompatible_context_imports(file_paths: list[Path]) -> list[str]:
--- a/python-sdk/docs/conf.py
+++ b/python-sdk/docs/conf.py
@@ -88,6 +88,7 @@ source_suffix = {
 }
 
 
+# TODO: We should consider removing param start _
 def skip_util_classes(_app, _what, _name, obj, skip, _options):
     """This allows us skipping certain objects (including functions & methods) from docs"""
     if ":sphinx-autoapi-skip:" in obj.docstring:
--- a/python-sdk/docs/conf.py
+++ b/python-sdk/docs/conf.py
@@ -88,7 +88,7 @@ source_suffix = {
 }
 
 
-def skip_util_classes(app, what, name, obj, skip, options):
+def skip_util_classes(_app, _what, _name, obj, skip, _options):
     """This allows us skipping certain objects (including functions & methods) from docs"""
     if ":sphinx-autoapi-skip:" in obj.docstring:
         skip = True
--- a/python-sdk/src/astro/custom_backend/serializer.py
+++ b/python-sdk/src/astro/custom_backend/serializer.py
@@ -47,19 +47,16 @@ def serialize(obj: Table | File | Any) -> dict | Any:  # noqa
     elif isinstance(obj, np.ndarray):
         return obj.tolist()
     elif isinstance(obj, SQLAlcRow):
-        keymap = obj._keymap  # skipcq PYL-W021
-        key_style = obj._key_style  # skipcq PYL-W021
+        serialized_obj = {
+            "class": "SQLAlcRow",
+            "key_map": obj._keymap,  # skipcq PYL-W021
+            "key_style": obj._key_style,  # skipcq PYL-W021
+        }
         if airflow.__version__ >= "2.3":
-            return {
-                "class": "SQLAlcRow",
-                "key_map": keymap,
-                "data": obj._data,  # skipcq PYL-W021
-                "key_style": key_style,
-            }
-        else:
-            return {"class": "SQLAlcRow", "key_map": keymap, "key_style": key_style}
+            serialized_obj["data"] = obj._data  # skipcq PYL-W021
+        return serialized_obj
 
-    elif isinstance(obj, str):  # pragma: no cover
+    elif isinstance(obj, str):
         return {"class": "string", "value": obj}
     else:
         return _attempt_to_serialize_unknown_object(obj)
--- a/python-sdk/src/astro/custom_backend/serializer.py
+++ b/python-sdk/src/astro/custom_backend/serializer.py
@@ -47,15 +47,17 @@ def serialize(obj: Table | File | Any) -> dict | Any:  # noqa
     elif isinstance(obj, np.ndarray):
         return obj.tolist()
     elif isinstance(obj, SQLAlcRow):
+        keymap = obj._keymap  # skipcq PYL-W021
+        key_style = obj._key_style  # skipcq PYL-W021
         if airflow.__version__ >= "2.3":
             return {
                 "class": "SQLAlcRow",
-                "key_map": obj._keymap,
-                "data": obj._data,
-                "key_style": obj._key_style,
+                "key_map": keymap,
+                "data": obj._data,  # skipcq PYL-W021
+                "key_style": key_style,
             }
         else:
-            return {"class": "SQLAlcRow", "key_map": obj._keymap, "key_style": obj._key_style}
+            return {"class": "SQLAlcRow", "key_map": keymap, "key_style": key_style}
 
     elif isinstance(obj, str):
         return {"class": "string", "value": obj}
--- a/python-sdk/src/astro/custom_backend/serializer.py
+++ b/python-sdk/src/astro/custom_backend/serializer.py
@@ -59,7 +59,7 @@ def serialize(obj: Table | File | Any) -> dict | Any:  # noqa
         else:
             return {"class": "SQLAlcRow", "key_map": keymap, "key_style": key_style}
 
-    elif isinstance(obj, str):
+    elif isinstance(obj, str):  # pragma: no cover
         return {"class": "string", "value": obj}
     else:
         return _attempt_to_serialize_unknown_object(obj)
--- a/python-sdk/src/astro/databases/aws/redshift.py
+++ b/python-sdk/src/astro/databases/aws/redshift.py
@@ -136,9 +136,6 @@ class RedshiftDatabase(BaseDatabase):
 
         :param table: The table we want to retrieve the qualified name for.
         """
-        # Initially this method belonged to the Table class.
-        # However, in order to have an agnostic table class implementation,
-        # we are keeping all methods which vary depending on the database within the Database class.
         if table.metadata and table.metadata.schema:
             qualified_name = f'"{table.metadata.schema}"."{table.name}"'
         else:
--- a/python-sdk/src/astro/databases/aws/redshift.py
+++ b/python-sdk/src/astro/databases/aws/redshift.py
@@ -251,16 +251,14 @@ class RedshiftDatabase(BaseDatabase):
 
         source_to_target_map_source_columns = list(source_to_target_columns_map.keys())
         source_to_target_map_target_columns = list(source_to_target_columns_map.values())
-        source_table_all_columns = self.hook.run(
-            "select col_name from pg_get_cols('%s') "
-            "cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int);",
-            parameters=(source_table_name,),
+        source_table_all_columns = self.hook.run(  # skipcq BAN-B608
+            f"select col_name from pg_get_cols('{source_table_name}') "
+            f"cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int);",
             handler=lambda x: [y[0] for y in x.fetchall()],
         )
-        target_table_all_columns = self.hook.run(
-            "select col_name from pg_get_cols('%s') "
-            "cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int);",
-            parameters=(source_table_name,),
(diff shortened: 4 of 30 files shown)


python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -238,12 +238,12 @@ def merge_table(
 
         source_to_target_map_source_columns = list(source_to_target_columns_map.keys())
         source_to_target_map_target_columns = list(source_to_target_columns_map.values())
-        source_table_all_columns = self.hook.run(
+        source_table_all_columns = self.hook.run(  # skipcq BAN-B608
```
pankajastro, line 241: This skip should we ok since we are wrapping table name with `''`

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -238,12 +238,12 @@ def merge_table(
...
             f"select col_name from pg_get_cols('{source_table_name}') "
             f"cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int);",
             handler=lambda x: [y[0] for y in x.fetchall()],
         )
-        target_table_all_columns = self.hook.run(
+        target_table_all_columns = self.hook.run(  # skipcq BAN-B608
```
pankajastro, line 246: same here as above

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -272,7 +272,7 @@ def merge_table(
             target_conflict_columns,
         )
         truncate_target_table = f"TRUNCATE {target_table_name}"
-        insert_into_target_table = f"INSERT INTO {target_table_name} SELECT * FROM {stage_table_name}"
+        insert_into_target_table = f"INSERT INTO '{target_table_name}' SELECT * FROM '{stage_table_name}'"
```
pankajastro, line 275: I have wrapped the table name with `''` lets see what CI think

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -253,7 +253,9 @@ def create_table_using_native_schema_autodetection(  # skipcq: PYL-R0201
         supported_config = self.NATIVE_AUTODETECT_SCHEMA_CONFIG.get(file.location.location_type)
         return supported_config["method"](table=table, file=file)  # type: ignore
 
-    def is_native_load_file_available(self, source_file: File, target_table: BaseTable) -> bool:
+    def is_native_load_file_available(
+        self, source_file: File, target_table: BaseTable
```
pankajastro, line 257: a override of base class method

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -292,7 +292,9 @@ def merge_table(
             for statement in statements:
                 cursor.execute(statement)
 
-    def is_native_load_file_available(self, source_file: File, target_table: BaseTable) -> bool:
+    def is_native_load_file_available(
+        self, source_file: File, target_table: BaseTable  # skipcq PYL-W0613
```
pankajastro, line 296: a override of base class method

python-sdk/src/astro/databases/aws/redshift.py
feluelle: But why are we not just using `params` instead? Or does `self.hook.run` not have this option?

python-sdk/dev/scripts/pre_commit_context_typing_compat.py
```diff
@@ -27,14 +27,16 @@ def visit_ImportFrom(self, node: ImportFrom) -> None:
...
+                f"{node.module}.{alias.name}" == "airflow.utils.context.Context"
+                and not self.has_incompatible_context_imports
+            ):
+                self.has_incompatible_context_imports = True
 
 
 def get_all_provider_files() -> list[Path]:
     """Retrieve all eligible provider module files."""
-    provider_files = []
+    _provider_files = []
```
utkarsharma2, line 39: @pankajastro What was the issue here?

python-sdk/docs/conf.py
```diff
@@ -88,7 +88,7 @@
 }
 
 
-def skip_util_classes(app, what, name, obj, skip, options):
+def skip_util_classes(_app, _what, _name, obj, skip, _options):
```
utkarsharma2, line 91: and here?

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -129,6 +129,19 @@ def table_exists(self, table: BaseTable) -> bool:
         inspector = sqlalchemy.inspect(self.sqlalchemy_engine)
         return bool(inspector.dialect.has_table(self.connection, table.name, schema=table.metadata.schema))
 
+    @staticmethod
+    def get_table_qualified_name(table: BaseTable) -> str:  # skipcq: PYL-R0201
```
utkarsharma2, line 133: @pankajastro Why did we implement this method here? Should come from the base class?

python-sdk/src/astro/databases/aws/redshift.py
utkarsharma2: why?

utkarsharma2: never mind got it. https://deepsource.io/gh/mindcurry/UniBorg/issue/PYL-W0613/occurrences

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -552,7 +552,9 @@ def create_table_using_schema_autodetection(
         # Since this method is used by both native and pandas path we cannot skip this step.
         self.truncate_table(table)
 
-    def is_native_load_file_available(self, source_file: File, target_table: BaseTable) -> bool:
+    def is_native_load_file_available(
```
utkarsharma2, line 555: @pankajastro maybe we should change this to static? in base class too?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -663,7 +665,7 @@ def load_pandas_dataframe_to_table(
 
     def get_sqlalchemy_template_table_identifier_and_parameter(
```
utkarsharma2, line 666: Change this to static?

python-sdk/src/astro/databases/sqlite.py
```diff
@@ -71,7 +71,7 @@ def create_schema_if_needed(self, schema: str | None) -> None:
         Since SQLite does not have schemas, we do not need to set a schema here.
         """
 
-    def schema_exists(self, schema: str) -> bool:
+    def schema_exists(self, schema: str) -> bool:  # skipcq PYL-W0613,PYL-R0201
```
utkarsharma2, line 74: @pankajastro Static method?

python-sdk/src/astro/files/types/csv.py
```diff
@@ -12,7 +12,9 @@
 class CSVFileType(FileType):
     """Concrete implementation to handle CSV file type"""
 
-    def export_to_dataframe(self, stream, columns_names_capitalization="original", **kwargs) -> pd.DataFrame:
+    def export_to_dataframe(
```
utkarsharma2, line 15: @pankajastro Static method?

python-sdk/src/astro/databases/aws/redshift.py
pankajastro: I want to have a different implementation for redshift.  Redshift accepts the table name like ```"db"."schema"."table"``` this is more safe. 
If I want a user to allow to run the query ```select * from db.schema.table``` in the current scenario if a user will pass the table name "table LIMIT 5" then the query will become ```select * from db.schema.table LIMIT 5``` which is incorrect but if we wrap the table name with a double quote then it would be more correct like ```select * from "db"."schema"."table LIMIT 5"```

python-sdk/src/astro/files/types/csv.py
pankajastro: override of a parent class method

python-sdk/dev/scripts/pre_commit_context_typing_compat.py
pankajastro: define in outer scope https://github.com/astronomer/astro-sdk/pull/1208/files#diff-93af335d96ed6e2ad0b9486f907098f5a5637f741d6d244833b226c4ed84fc29R67

python-sdk/docs/conf.py
pankajastro: ???

pankajastro: _app, _what, _name, _options unused param so mark them to start with `_`

utkarsharma2 (approved):

python-sdk/src/astro/databases/sqlite.py
pankajastro: two issue 
- override of base class method 
- schema is unused param

python-sdk/src/astro/databases/snowflake.py
pankajastro: isn't this is also an override of the base class method?

python-sdk/src/astro/databases/snowflake.py
pankajastro: isn't this is also an override of the base class method?

python-sdk/src/astro/databases/aws/redshift.py
pankajastro: good point, not sure how the redshift connector handles it but let me try to fix it

pankajastro: Look like the way the redshift connector accepts the param is depend on the value of `paramstyle` 
 https://github.com/aws/amazon-redshift-python-driver/blob/master/redshift_connector/cursor.py#L201 the default value is `format` so even if I'll do `"insert {}".format(table)` it would be string looks difficult to pass param in hook run method for redshift

feluelle: But [the tests passed](https://github.com/astronomer/astro-sdk/actions/runs/3417479740). So why the revert? It is recommended to use `parameters` to handle them properly. By doing that we let the underlying library decide how to deal with it and ensure security.

pankajastro: It didn't the test was failing here https://github.com/astronomer/astro-sdk/actions/runs/3417479740/jobs/5689302000   so reverted

python-sdk/src/astro/databases/aws/redshift.py
utkarsharma2: We should address this issue separately
ref: https://docs.aws.amazon.com/redshift/latest/dg/r_enable_case_sensitive_identifier.html

python-sdk/src/astro/databases/aws/redshift.py
```diff
@@ -238,13 +251,13 @@ def merge_table(
...
-        source_table_all_columns = self.hook.run(
+        source_table_all_columns = self.hook.run(  # skipcq BAN-B608
             f"select col_name from pg_get_cols('{source_table_name}') "
             f"cols(view_schema name, view_name name, col_name name, col_type varchar, col_num int);",
             handler=lambda x: [y[0] for y in x.fetchall()],
         )
-        target_table_all_columns = self.hook.run(
-            f"select col_name from pg_get_cols('{target_table_name}') "
+        target_table_all_columns = self.hook.run(  # skipcq BAN-B608
+            f"select col_name from pg_get_cols('{source_table_name}') "
```
feluelle, line 260:
```suggestion
            f"select col_name from pg_get_cols('{target_table_name}') "
```

python-sdk/src/astro/databases/aws/redshift.py
utkarsharma2: issue: https://github.com/astronomer/astro-sdk/issues/1212

python-sdk/src/astro/databases/aws/redshift.py
feluelle: My bad parameters is only for passing arguments - not table names to sql.

feluelle (approved):
LGTM in general. But what do you guys think if we always add a comment to `# skipcq`s so that we know why we have added it. Otherwise when refactoring or touching this part of code we don't know if it is still necessary. We always would have to look it up.

feluelle:
@pankajastro @utkarsharma2 WDYT of this^?

pankajastro:
> @pankajastro @utkarsharma2 WDYT of this^?

I 100% agree @feluelle but in some cases, we do have to do it like 
- In case of method overloading not using the same param is quite possible because if different DB has different nomenclature etc
-  At a few places, we are accessing third-party lib private attribute

Maybe we should look at bases classes once if it is possible to do some cleanup we will be able to fix a lot of it.

feluelle:
I am only saying we should always add comment. Otherwise we end up having useless `# skipcq` comments because we forgot what they were for.

feluelle:
Could you already do that for the comments you added here (as part of this PR) @pankajastro? This would be really great. 🙏

python-sdk/src/astro/custom_backend/serializer.py
```diff
@@ -47,17 +47,19 @@ def serialize(obj: Table | File | Any) -> dict | Any:  # noqa
...
+                "key_map": keymap,
+                "data": obj._data,  # skipcq PYL-W021
+                "key_style": key_style,
             }
         else:
-            return {"class": "SQLAlcRow", "key_map": obj._keymap, "key_style": obj._key_style}
+            return {"class": "SQLAlcRow", "key_map": keymap, "key_style": key_style}
 
-    elif isinstance(obj, str):
+    elif isinstance(obj, str):  # pragma: no cover
```
feluelle, line 62: This we should definitely cover.

python-sdk/src/astro/files/types/base.py
```diff
@@ -45,3 +45,6 @@ def __repr__(self) -> str:
 
     def __eq__(self, other):
         return self.name == other.name
+
+    def __hash__(self):  # pragma: no cover
+        return hash(self.name)
```
feluelle, line 50: Can we please have a simple test for this? Either don't add the comment or add a test. But we shouldn't ever put `no cover` anywhere except for `if TYPE_CHECKING`. Everything we can test, we should test.

feluelle:
Sorry I missed the `no cover` comments in the first review. Could you address them, please?

python-sdk/src/astro/custom_backend/serializer.py
pankajastro: covered

pankajastro:
> Could you already do that for the comments you added here (as part of this PR) @pankajastro? This would be really great. 🙏

Added the comments too in case making sense

pankajastro:
Merging it if any other feedback will take that separately

Issue #1171, opened 2022-11-01
`example_transform_file` failing for postgres db


pankajastro:
**Describe the bug**
Logs: https://astronomer.slack.com/files/U0300VBN3QU/F048H54G29Y/untitled.txt
Reported by: [Kelechi Egbosi](https://astronomer.slack.com/team/U02MR1CPQTC)

**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Write the DAG  '...'
2. Create connection '....'
3. Run using  '....'

**Expected behavior**

**Screenshots**

**Additional context**

pankajastro:
duplicate https://github.com/astronomer/astro-sdk/issues/1209

Issue #1209, opened 2022-11-07
[Example Dags] Fix condition in transform.sql file


kelechi2020:
**Describe the bug**
The `example_transform_file` dag throws an error

[SQL: CREATE TABLE IF NOT EXISTS airflow.test_is_20221027 AS SELECT title, rating FROM airflow._tmp_pr97v0fjsqkrsogzu2i5d38ck774whav6ivzxrogu2v2dnk9gwscro1kh WHERE genre1=='Animation' ORDER BY rating desc LIMIT 5; ]

**To Reproduce**
1. Run the example dag here: https://github.com/astronomer/astro-sdk/blob/main/python-sdk/example_dags/example_transform_file.py
2. Create connection 
3. Run using  '....'
4. See error:

`
c3bbb37f069e
*** Reading local file: /usr/local/airflow/logs/dag_id=example_transform_file/run_id=manual__2022-10-27T12:19:40.173973+00:00/task_id=transform/attempt=1.log
[2022-10-27T12:19:46.436+0000] {taskinstance.py:1165} INFO - Dependencies all met for <TaskInstance: example_transform_file.transform manual__2022-10-27T12:19:40.173973+00:00 [queued]>
[2022-10-27T12:19:46.445+0000] {taskinstance.py:1165} INFO - Dependencies all met for <TaskInstance: example_transform_file.transform manual__2022-10-27T12:19:40.173973+00:00 [queued]>
[2022-10-27T12:19:46.445+0000] {taskinstance.py:1362} INFO - 
--------------------------------------------------------------------------------
[2022-10-27T12:19:46.445+0000] {taskinstance.py:1363} INFO - Starting attempt 1 of 1
[2022-10-27T12:19:46.445+0000] {taskinstance.py:1364} INFO - 
--------------------------------------------------------------------------------
[2022-10-27T12:19:46.453+0000] {listener.py:27} INFO - TaskInstance Details: dag_id=example_transform_file, task_id=transform, dagrun_id=manual__2022-10-27T12:19:40.173973+00:00, map_index=-1, run_start_date=2022-10-27 12:19:46.437324+00:00, try_number=1, job_id=4937, op_classpath=astro.sql.operators.transform.TransformOperator
[2022-10-27T12:19:46.473+0000] {manager.py:67} WARNING - Unable to find an extractor. task_type=TransformOperator airflow_dag_id=example_transform_file task_id=transform airflow_run_id=manual__2022-10-27T12:19:40.173973+00:00 
[...]

PR #1210, opened 2022-11-07
[Docs] Update transform.sql example query

kelechi2020:
Update sql query

closes: #[ISSUE](https://github.com/astronomer/astro-sdk/issues/1209)

closes: https://github.com/astronomer/astro-sdk/issues/1171

- No

### Checklist

### DIFF
--- a/python-sdk/example_dags/demo_parse_directory/transform.sql
+++ b/python-sdk/example_dags/demo_parse_directory/transform.sql
@@ -1,5 +1,5 @@
 SELECT title, rating
 FROM {{ input_table }}
-WHERE genre1=='Animation'
+WHERE genre1='Animation'
 ORDER BY rating desc
 LIMIT 5;


python-sdk/example_dags/demo_parse_directory/transform.sql
```diff
@@ -1,5 +1,5 @@
 SELECT title, rating
 FROM {{ input_table }}
-WHERE genre1=='Animation'
```
pankajastro, line 3: This syntax is correct for Postgres but incorrect for SQLite we use this file to run test which uses SQLite if you want to update you will have to update conn here too https://github.com/astronomer/astro-sdk/blob/main/python-sdk/example_dags/example_transform_file.py#L22

pankajastro (commented):
This is not a documentation change but a test change though we use this for documentation too.

python-sdk/example_dags/demo_parse_directory/transform.sql
tatiana: @kelechi2020 @pankajastro would there be a way to represent this query that would work for both databases?
How should we move forward with this PR?

tatiana:
I'm closing this PR for now, and we can reopen it in future if we find it fit.

PR #1207, opened 2022-11-07
Add `.coveragerc` to the project

pankajastro:
# Description
## What is the current behavior?
Add frequently code coverage excluded line in `.coverragerc`

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1211, opened 2022-11-07
Add `no cover` to skip code-cov type_check and import_error

pankajastro:
# Description
Add `no cover` type_check and import_error to skip code coverage for an import statement. I'll fix it in more organized way in https://github.com/astronomer/astro-sdk/pull/1207

### Checklist

### DIFF
--- a/python-sdk/src/astro/utils/table.py
+++ b/python-sdk/src/astro/utils/table.py
@@ -101,7 +101,7 @@ def find_first_table(
 
     if op_args:
         first_table = _find_first_table_from_op_args(op_args=op_args, context=context)
-    if not first_table and op_kwargs and python_callable:
+    if not first_table and op_kwargs and python_callable is not None:
         first_table = _find_first_table_from_op_kwargs(
             op_kwargs=op_kwargs,
             python_callable=python_callable,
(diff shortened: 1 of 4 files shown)


python-sdk/src/astro/utils/table.py
```diff
@@ -101,7 +101,7 @@ def find_first_table(
 
     if op_args:
         first_table = _find_first_table_from_op_args(op_args=op_args, context=context)
-    if not first_table and op_kwargs and python_callable:
+    if not first_table and op_kwargs and python_callable is not None:
```
utkarsharma2, line 104: @pankajastro Why did we add this?

pankajastro: otherwise the static check is failing since `python_callable` which is not boolean

utkarsharma2 (approved):

python-sdk/src/astro/utils/table.py
feluelle: Interesting 🤔 🤷

pankajastro: also, I feel it does make sense if we pass empty callable 

```
val = [ ]
if val:   => won't go in this block
    print("fail") 
if val is not None:
    print("pass")
```

pankajastro:
Merging it if any further feedback will take it separately

python-sdk/src/astro/utils/table.py
feluelle:
```
if val:   => won't go in this block
    print("fail") 
```
this is expected

PR #1216, opened 2022-11-08
[sql-cli] Fix Airflow 2.2 issue #1162

tatiana:
Closes: #1162

While working on this ticket, I noticed we have a deeper issue with how we try to use and configure Airflow as a Python library.

We also now run tests for python 3.9 in CI and 4 different airflow versions to ensure compatibility.


sql-cli/noxfile.py
```diff
@@ -21,21 +21,35 @@ def dev(session: nox.Session) -> None:
...
     session.log("Installed Dependencies:")
-    session.run("pip3", "freeze")
-    airflow_home = f"~/airflow-latest-{session.python}"
+    session.run("poetry", "run", "pip3", "freeze")
+
+    # TODO: at the moment flow run depends on the Airflow global Airflow Home - we need to fix this
+    # Refactor so each test does this in their own sandboxed Airflow home
+    # From what I observed, the `run` command is the one using this, due to how the method
+    # utils.airflow.get_dag behaves under Airflow 2.2.0 (it defaults to $HOME/airflow)
+    airflow_home = f"~/airflow-latest-{session.python}-airflow-{airflow}"
```
feluelle, line 37:
```suggestion
    airflow_home = f"~/airflow-{airflow}-{session.python}"
```
This was meant to be the placeholder for the airflow version 😅

sql-cli/noxfile.py
```diff
@@ -21,21 +21,35 @@ def dev(session: nox.Session) -> None:
...
     session.run("airflow", "db", "init", env={"AIRFLOW_HOME": airflow_home})
+
     session.run(
+        "poetry",
+        "run",
         "pytest",
+        "-k",
+        "test_validate[test-sqlite_conn_invalid-FAILED]",
```
feluelle, line 45:
```suggestion
```

sql-cli/noxfile.py
feluelle: @tatiana WDYT of the following?
```suggestion
    airflow_home = f"~/python-{session.python}-airflow-{airflow}"
```

feluelle: Maybe we don't even need the python version, I am not 100% sure. Most errors we get if reuse the airflow home is due to the airflow database right? not due to the python version.

sql-cli/noxfile.py
```diff
@@ -21,15 +21,48 @@ def dev(session: nox.Session) -> None:
...
+    else:
+        session.run("poetry", "install", "--with", "dev")
+        session.run("poetry", "run", "pip", "install", f"apache-airflow=={airflow}")
+
     session.log("Installed Dependencies:")
     session.run("pip3", "freeze")
-    airflow_home = f"~/airflow-latest-{session.python}"
-    session.run("airflow", "db", "init", env={"AIRFLOW_HOME": airflow_home})
+
+    # At the moment flow run depends on the Airflow global Airflow Home - we need to fix this
```
dimberman, line 57: Would there be a way to fix this now or is this a bigger fix? At minimum I'd like to see a ticket tied to this.

sql-cli/sql_cli/run_dag.py
```diff
@@ -25,6 +26,11 @@
 log = logging.getLogger(__name__)
 log.setLevel(logging.INFO)
 
+# This was added to airflow.utils.session as part of Airflow 2.2.4, we're copying it here for backwwards
+# compatibility (currently we support Airflow 2.1 onwards)
+# It seems like a hack to avoid MyPy complaining we're assigning None to a variable that should be an Airflow session
+NEW_SESSION = airflow_settings.SASession = cast(airflow_settings.SASession, None)
```
dimberman, line 32: Can we not just use type: ignore instead? Seems iffy to make these kinds of hacks just for mypy

dimberman (dismissed):

sql-cli/noxfile.py
tatiana: @dimberman it's a bigger fix and I've logged a dedicated ticket for this last week:
https://github.com/astronomer/astro-sdk/issues/1219

sql-cli/sql_cli/run_dag.py
tatiana: @dimberman the `main` version of the SQL CLI is already using this same exact code, from the moment we relied on Airflow's `NEW_SESSION` during the implementation of the `run` command. The only difference is that for backwards compatibility, I brought this part of the code to our code base and realised it was a hack:
https://github.com/apache/airflow/blob/main/airflow/utils/session.py#L84

I agree this can be improved and I suggest it is done as part of a separate PR.

sql-cli/noxfile.py
tatiana: I think it is useful for us to have the ability to run concurrently the tests for multiple versions of Python - this protects us from possible issues until we solve the underlining problem of relying on this `$AIRFLOW_HOME`.

sql-cli/pyproject.toml
```diff
@@ -38,10 +38,12 @@ networkx = [
...
 Jinja2 = ">=2.10.1, <=3.1"
 python-frontmatter = "^1.0.0"
-PyYAML = ">5.4.1"
-apache-airflow = ">2.0"
+PyYAML = ">=5.4.1"
+# the Python SDK 1.2 enforces pandas >= 1.3.4,
+# but Airflow 2.1 has constraints pandas == 1.3.3
 python-dotenv = "^0.21.0"
 astro-sdk-python = {extras = ["all"], version = "^1.2.0"}
+apache-airflow = ">2.1"
```
tatiana, line 46: https://github.com/astronomer/astro-sdk/actions/runs/3450993943/jobs/5759928610

When we attempt to use the Airflow constraints for 2.1, these are the conflicts:
```
The conflict is caused by the following:
    apache-airflow 2.1.4 depends on pandas<2.0 and >=0.17.1; python_version >= "3.7"
    astro-sdk-python[all] 1.2.1 depends on pandas<2.0.0 and >=1.3.4
    apache-airflow 2.1.4 depends on pandas<2.0 and >=0.17.1; python_version >= "3.7"
    astro-sdk-python[all] 1.2.0 depends on pandas<2.0.0 and >=1.3.4
    The user requested (constraint) pandas==1.3.3
```

If we ignore the constraints for 2.1, there are a couple of errors:
```
tests/test___main__.py ........FF......F.F..FF..FF.........FFF...                      [ 60%]
tests/test_configuration.py ..                                                         [ 62%]
tests/test_connections.py FF                                                           [ 65%]
tests/test_dag_generator.py .......                                                    [ 75%]
tests/test_project.py ..                                                               [ 78%]
tests/test_run_dag.py ...ss...                                                         [ 90%]
tests/test_sql_directory_parser.py .......                                             [100%]
```

I consider it outside the scope of the current ticket to support 2.1.  I logged a ticket for when we decide to prioritise this:
https://github.com/astronomer/astro-sdk/issues/1240

.github/workflows/ci-sql-cli.yaml
```diff
@@ -63,6 +63,10 @@ jobs:
       (
         github.event_name == 'release'
       )
+    strategy:
+      fail-fast: false
+      matrix:
+        airflow: [ 2.2.5, 2.3.4, 2.4.2 ]
```
feluelle, line 69: I think it also makes sense to run on different python versions here.

sql-cli/pyproject.toml
```diff
@@ -38,10 +38,12 @@ networkx = [
 ]
 Jinja2 = ">=2.10.1, <=3.1"
 python-frontmatter = "^1.0.0"
-PyYAML = ">5.4.1"
-apache-airflow = ">2.0"
+PyYAML = ">=5.4.1"
+# the Python SDK 1.2 enforces pandas >= 1.3.4,
+# but Airflow 2.1 has constraints pandas == 1.3.3
```
feluelle, line 43: What is this comment referring to?

sql-cli/sql_cli/sql_directory_parser.py
```diff
@@ -135,15 +136,18 @@ def to_transform_operator(self) -> TransformOperator:
...
+        kwargs = {
+            "conn_id": self.metadata.get("conn_id"),
+            "parameters": None,
+            "handler": None,
+            "database": self.metadata.get("database"),
+            "schema": self.metadata.get("schema"),
+            "python_callable": lambda: (str(self.path), None),
+            "sql": self.content,
+        }
+        if airflow.__version__.startswith("2.2"):
```
feluelle, line 148: For dealing with versions we should use `packaging.version` module:

[python-sdk example](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/test_example_dags.py#L51-L55)

sql-cli/sql_cli/utils/airflow.py
```diff
@@ -25,7 +25,13 @@ def retrieve_airflow_database_conn_from_config(airflow_home: Path) -> str:
     parser = ConfigParser()
     parser.read(filename)
     confdict = {section: dict(parser.items(section)) for section in parser.sections()}
-    return confdict["database"]["sql_alchemy_conn"]
+    try:
+        # Airflow >= 2.3
+        sql_alchemy_conn = confdict["database"]["sql_alchemy_conn"]
+    except KeyError:
+        sql_alchemy_conn = confdict["core"]["sql_alchemy_conn"]
```
feluelle, line 32: If we know the version shouldn't we rather check on it via if-else ?

sql-cli/sql_cli/utils/airflow.py
```diff
@@ -50,8 +56,11 @@ def set_airflow_database_conn(airflow_meta_conn: str) -> None:
 
     :params airflow_db_conn: Similar to `sqlite:////tmp/project/airflow.db`
     """
-    # This is a hacky approcah we managed to find to make thigs work with Airflow 2.4
+    # This is a hacky approach we managed to find to make things work with Airflow 2.4
     os.environ["AIRFLOW__DATABASE__SQL_ALCHEMY_CONN"] = airflow_meta_conn
+    # This is necessary for Airflow <= 2.2
+    os.environ["AIRFLOW__CORE__SQL_ALCHEMY_CONN"] = airflow_meta_conn
```
feluelle, line 62: Maybe it makes sense here to also set depending on the current version. WDYT?

sql-cli/tests/test_run_dag.py
```diff
@@ -15,6 +17,16 @@
...
 
 
+def airflow_less_than(threshold: Iterable[int]) -> bool:
+    """
+    Check if Airflow is less than a given threshold version.
+
+    :param threshold: version tuple (e.g. (2, 3)).
+    """
+    current_version = tuple(int(item) for item in airflow.__version__.split("."))
+    return current_version < tuple(threshold)
```
feluelle, line 27: Let's use the `package.version` module (see comment above).

sql-cli/noxfile.py
```diff
@@ -21,15 +21,45 @@ def dev(session: nox.Session) -> None:
...
 @nox.session(python=["3.7", "3.8", "3.9"])
-def test(session: nox.Session) -> None:
+@nox.parametrize("airflow", ["2.2.5", "2.3.4", "2.4.2"])
+def test(session: nox.Session, airflow: str) -> None:
     """Run both unit and integration tests."""
+
     session.install("poetry")
-    session.run("poetry", "install", "--with", "dev")
-    # Log all the installed dependencies
+    if airflow.startswith("2.2"):
```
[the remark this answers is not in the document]
tatiana: I'd rather we do this for all 2.2 - other versions also need to be constraint, and in case we need to test any other specific 2.2, we would not need to change the comparison.

sql-cli/noxfile.py
```diff
@@ -21,15 +21,45 @@ def dev(session: nox.Session) -> None:
...
+            "https://raw.githubusercontent.com/apache/airflow/"
+            f"constraints-{airflow}/constraints-{session.python}.txt"
+        )
+
+        # Poetry does not support constraints:
+        # https://github.com/python-poetry/poetry/issues/3225
+        session.install("-e", ".", f"apache-airflow=={airflow}", "-c", constraints_url)
+
+        # We are duplicating the tests dependencies until we find a better solution.
+        # The solution might be to move out of poetry.
```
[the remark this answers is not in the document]
tatiana: I tried to automate this, and I didn't manage - since this ticket has been long-standing, I would really like us to just merge it and focus on higher priority ones.

sql-cli/sql_cli/sql_directory_parser.py
tatiana: @feluelle how do you think the packaging version could help in this specific line? I changed to use in the comparison, in the other part of the code, but it is not clear to be on the benefit in this particular case.

feluelle: I would say mainly for input validation.

feluelle: But the comment was more generally speaking. So it might make sense here to either use the specific string via `==` or the `Version()`. I know that `startswith` will probably work most of the time (except for `>2.20.x` 😅).

feluelle (approved):

Issue #482, opened 2022-06-22
`File` and `load_file` should take extra params to give to pandas


fritz-astronomer:
**Please describe the feature you'd like to see**
I may need to pass additional parameters for Pandas, describing my data
e.g. if I need to approximate 
```
pd.read_csv("s3://my_bucket/my_file.csv", sep='|')
```
I am currently unable to utilize `astro-sdk

**Describe the solution you'd like**
`File()` should take an `kwargs` or some other param which can be fed down to pandas or whatever is doing the underlying reading

**Are there any alternatives to this feature?**
🤷 

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [x] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [x] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [x] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [x] Improve the documentation (README, Sphinx, and any other relevant)
- [x] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

tatiana:
@fritz-astronomer, we are currently improving the load performance, and we won't be using Pandas for many of the load operations. Therefore, I believe it is unlikely that the Astro SDK would support any Pandas `kwargs`.

That said, I believe supporting different CSV delimiters is a good feature. We could add support for this as part of the `File` object, passed to the load operation. Would this solve the issue, or do you have additional needs to load this CSV?

fritz-astronomer:
@tatiana - I think that would suffice. 

A note though: I imagine that handling so much edge cases of file specifications could be a lot of code that may otherwise exist though. Maybe `File` can generally be some thin wrapper on top of some other file parsing python library? 🤷  Thinking of CSV, Avro, Parquet and how each of them can vary in really small annoying ways to parse.

fhoda:
[...]

PR #1225, opened 2022-11-09
Add pd param

pankajastro:
# Description
closes: #482 

## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -25,6 +25,8 @@ class File(LoggingMixin, Dataset):
     :param conn_id: Airflow connection ID
     :param filetype: constant to provide an explicit file type
     :param normalize_config: parameters in dict format of pandas json_normalize() function.
+    :param pd_kw: The Pandas related kwargs as a dict.
+        This will be supplied to Pandas ``read_csv`` function etc.
     """
 
     path: str
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -33,6 +33,7 @@ class File(LoggingMixin, Dataset):
     normalize_config: dict | None = None
     is_dataframe: bool = False
     is_bytes: bool = False
+    pd_kw: dict = {}
 
     uri: str = field(init=False)
     extra: dict | None = field(init=False, factory=dict)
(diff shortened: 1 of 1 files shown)


python-sdk/tests/files/type/test_base.py
```diff
@@ -11,7 +11,7 @@ def test_repr():
     file = File(path="astro", conn_id="local", filetype=FileType.CSV)
     assert file.__repr__() == (
         "File(path='astro', conn_id='local', filetype=<FileType.CSV: 'csv'>, "
-        "normalize_config=None, is_dataframe=False, is_bytes=False, "
+        "normalize_config=None, is_dataframe=False, is_bytes=False, pd_kw={}, "
```
pankajkoti, line 14: Can we provide an example test case of what these keyword arguments could look like?

python-sdk/src/astro/files/base.py
```diff
@@ -112,6 +115,8 @@ def export_to_dataframe(self, **kwargs) -> pd.DataFrame:
         """Read file from all supported location and convert them into dataframes."""
         mode = "rb" if self.is_binary() else "r"
         with smart_open.open(self.path, mode=mode, transport_params=self.location.transport_params) as stream:
+            # Update kwargs with the pandas kwargs supplied by user in File object
```
pankajkoti, line 118: In the case of Snowflake and Postgres implementations, we would need to accept this parameter and pass it to this function from where this function is getting called. 
https://github.com/astronomer/astro-sdk/blob/0e9d07b53bb2218dabecdbb991dd49cc2ff94cfe/python-sdk/src/astro/databases/snowflake.py#L537

https://github.com/astronomer/astro-sdk/blob/0e9d07b53bb2218dabecdbb991dd49cc2ff94cfe/python-sdk/src/astro/databases/postgres.py#L208

python-sdk/src/astro/files/base.py
```diff
@@ -25,6 +25,8 @@ class File(LoggingMixin, Dataset):
     :param conn_id: Airflow connection ID
     :param filetype: constant to provide an explicit file type
     :param normalize_config: parameters in dict format of pandas json_normalize() function.
+    :param pd_kw: The Pandas related kwargs as a dict.
+        This will be supplied to Pandas ``read_csv`` function etc.
```
tatiana, line 29: It feels that both the (previously existing) `normalize_config` and the (just introduced) `pd_kw` could have more explicit names. In both scenarios, we want to give arguments to underlying Pandas functions - it may make sense to have a single keyword argument. What if we had something like:
```
{
    "pandas_kwargs":
        "json_normalize": 
        "read_csv": 
}
```

tatiana (commented):
@pankajastro I am slightly worried about us exposing Pandas functions (something we have precedence for) - and not making a consistent interface across native and non-native transfers. What are your thoughts on this?

pankajastro:
> @pankajastro I am slightly worried about us exposing Pandas functions (something we have precedence for) - and not making a consistent interface across native and non-native transfers. What are your thoughts on this?

I 100% agree, I'm also not in favour of doing it!

Issue #1226, opened 2022-11-09
Install astro-sdk with Airflow constraint file


pankajastro:
**Describe the bug**
Unable to install astro-sdk with Airflow constraint file
```
The conflict is caused by:
  astro-sdk-python[amazon,google,postgres,snowflake] 1.2.1 depends on snowflake-sqlalchemy<=1.2.4 and >=1.2.0; extra == "snowflake"
  The user requested (constraint) snowflake-sqlalchemy==1.4.1
```
Refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1667942573693039

**To Reproduce**
```
pip install "astro-sdk-python[snowflake,postgres,amazon,google]==1.2.0" -c https://raw.githubusercontent.com/apache/airflow/constraints-2.4.0/constraints-3.8.txt
```

**Expected behavior**

**Screenshots**

**Additional context**

phanikumv:
Re-opening as discussed with @kaxil

kaxil:
You can use the following constraints file to install SDK using Python 3.8 and Airflow 2.4.2

```
aiobotocore==2.4.0
aiofiles==0.8.0
aiohttp==3.8.3
aioitertools==0.11.0
aiosignal==1.3.1
alembic==1.8.1
anyio==3.6.2
apache-airflow==2.4.2
apache-airflow-providers-amazon==6.1.0
apache-airflow-providers-common-sql==1.3.0
apache-airflow-providers-ftp==3.2.0
apache-airflow-providers-google==8.5.0
apache-airflow-providers-http==4.1.0
apache-airflow-providers-imap==3.1.0
apache-airflow-providers-postgres==5.3.0
apache-airflow-providers-snowflake==4.0.0
apache-airflow-providers-sqlite==3.3.0
apispec==3.3.2
argcomplete==2.0.0
asgiref==3.5.2
asn1crypto==1.5.1
astro-sdk-python==1.3.0.dev1
async-timeout==4.0.2
attrs==22.1.0
azure-common==1.1.[28](https://github.com/astronomer/astro-sdk/actions/runs/3535931961/jobs/5934488828#step:10:29)
azure-core==1.26.1
azure-storage-blob==12.14.1
Babel==2.11.0
backoff==2.2.1
beautifulsoup4==4.11.1
blinker==1.5
boto3==1.24.59
botocore==1.27.59
build==0.9.0
cachelib==0.9.0
cachetools==5.2.0
cattrs==22.2.0
[...]

Issue #1187, opened 2022-11-02
Remove snowflake pined version


pankajastro:
**Please describe the feature you'd like to see**

related to: https://github.com/astronomer/astro-sdk/pull/1185

PR #1228, opened 2022-11-09
Update snowflake-sqlalchemy version

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1226
closes: #1187 

## What is the new behavior?
Remove snowflake-sqlalchemy version restrion

### Checklist

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -62,7 +62,7 @@ google = [
 ]
 snowflake = [
     "apache-airflow-providers-snowflake",
-    "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
+    "snowflake-sqlalchemy>=1.2.0",
     "snowflake-connector-python[pandas]",
 ]
 postgres = [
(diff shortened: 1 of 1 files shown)


python-sdk/pyproject.toml
```diff
@@ -62,7 +62,7 @@ google = [
 ]
 snowflake = [
     "apache-airflow-providers-snowflake",
-    "snowflake-sqlalchemy>=1.2.0,<=1.2.4",
+    "snowflake-sqlalchemy>=1.2.0",
```
pankajastro, line 65: Not sure why this was restricted but because of this when we are installing `astro-sdk` with airflow constraints file getting below error
```
The conflict is caused by:
    astro-sdk-python[amazon,google,postgres,snowflake] 1.0.0 depends on snowflake-sqlalchemy<=1.2.4 and >=1.2.0; extra == "snowflake"
    The user requested (constraint) snowflake-sqlalchemy==1.4.2
```

feluelle: related to https://github.com/astronomer/astro-sdk/pull/1185 cc @utkarsharma2

pankajastro: In #1185  look like we have changed `SQLAlchemy` only so `snowflake-sqlalchemy` version was restricted even before

pankajastro: Also look like import issue has been fixed in latest release https://github.com/snowflakedb/snowflake-sqlalchemy/releases/tag/v1.4.3

tatiana (approved):

python-sdk/pyproject.toml
feluelle: Cool. Could you remove the version restriction there then as well? I think there is then no reason to limit that version.

feluelle (approved):

PR #1231, opened 2022-11-09
Remove duplicated print statements in sql-cli

dimberman:
# Description

Currently we have some duplicated print statements in generate & run.

related #1140 

## What is the new behavior?

This PR removes the duplications.

No

### Checklist


sql-cli/sql_cli/__main__.py
```diff
@@ -168,20 +171,34 @@ def run(
...
-    cli.run_dag(
-        project=project,
-        env=env,
-        dag=dag,
-        verbose=verbose,
-    )
+    try:
+        dr = cli.run_dag(
+            project=project,
+            dag=dag,
```
tatiana, line 180: don't we need the `env` parameter anymore?

feluelle: No we only need it for the project.

sql-cli/sql_cli/dag_generator.py
```diff
@@ -103,15 +103,16 @@ def generate_dag(directory: Path, dags_directory: Path, generate_tasks: bool) ->
         start_date=datetime(2020, 1, 1),
         sql_files=sql_files,
     )
-    template_dag_name = "gen_tasks_dag"
     if generate_tasks:
+        template_file = Path("templates/gen_tasks_dag.py.jinja2")
```
pankajkoti, line 107: How about we define a constant `TEMPLATES_DIRECTORY = Path("templates")` and set here `templates_file = TEMPLATES_DIRECTORY / "gen_tasks_dag.py.jinja2"`

sql-cli/sql_cli/dag_generator.py
```diff
@@ -103,15 +103,16 @@ def generate_dag(directory: Path, dags_directory: Path, generate_tasks: bool) ->
...
+    else:
         for sql_file in sql_files_dag.sorted_sql_files():
             sql_file.write_raw_content_to_target_path()
-        template_dag_name = "render_dag"
+        template_file = Path("templates/render_dag.py.jinja2")
```
pankajkoti, line 111: Based on above suggestion for defining a constant for the templates directory,
```suggestion
        template_file = TEMPLATES_DIRECTORY / "render_dag.py.jinja2"
```

sql-cli/sql_cli/__main__.py
```diff
@@ -229,5 +256,41 @@ def init(
     rprint("Initialized an Astro SQL project at", project.directory)
 
 
+def _generate_dag(project: Project, workflow_name: str, generate_tasks: bool) -> Path:
```
pankajkoti, line 259: Is it possible to move this method to some other module. I think @tatiana 's idea was to keep only commands in this __main__.py file and we keep it clean by keeping implementation methods in other corresponding modules.

sql-cli/sql_cli/__main__.py
```diff
@@ -168,20 +181,34 @@ def run(
...
+
     rprint(f"\nRunning the workflow [bold blue]{dag.dag_id}[/bold blue] for [bold]{env}[/bold] environment\n")
-    cli.run_dag(
-        project=project,
-        env=env,
-        dag=dag,
-        verbose=verbose,
-    )
+    try:
+        dr = dag_runner.run_dag(
```
pankajkoti, line 190: I am thinking that do we not need here the `env` flag value to decide which environment to run the DAG in? Do we not need to have environment specific runs?

feluelle: 🤔 @dimberman? Do we need the env in `run_dag`. Probably at some point, but I think we are not there yet.

feluelle: Currently we only need the env to identify which connections to use.

sql-cli/sql_cli/__main__.py
feluelle: Let's do that as part of the newly created ticket #1235

feluelle (approved):

Issue #1147, opened 2022-10-27
Specifying table of different DB in Output_table param fails silently


utkarsharma2:
**Describe the bug**
Refer to slack thread - https://astronomer.slack.com/archives/C02B8SPT93K/p1664979020078039

**Version**
* Astro: 1.2.0

**Expected behavior**
We should raise an exception in case the table is from a different DB.

phanikumv:
Seems to be risky for 1.2.1 release considering other priorities

pankajastro:
just wondering why didn't throw an error in this case.

pankajastro:
@utkarsharma2 what we should check here source and target table conn_type or db_name also?
```
if source_table.conn_type != target_table.conn_type:
   raise
```

PR #1233, opened 2022-11-10
Raise error db mismatch of operation 

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1147

## What is the new behavior?
Raise expectation if the input and output DB source is not the same

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/append.py
+++ b/python-sdk/src/astro/sql/operators/append.py
@@ -55,6 +55,9 @@ class AppendOperator(AstroSQLBaseOperator):
                 f"columns is not a valid type. Valid types: [tuple, list, dict], Passed: {type(columns)}"
             )
         self.columns = columns or {}
+        # currently, cross database operation is not supported
+        if self.source_table.sql_type != self.target_table.sql_type:
+            raise ValueError("source and target table must belongs from same datasource")
         task_id = task_id or get_unique_task_id("append_table")
         super().__init__(
             task_id=task_id,
--- a/python-sdk/src/astro/sql/operators/append.py
+++ b/python-sdk/src/astro/sql/operators/append.py
@@ -55,6 +55,11 @@ class AppendOperator(AstroSQLBaseOperator):
                 f"columns is not a valid type. Valid types: [tuple, list, dict], Passed: {type(columns)}"
             )
         self.columns = columns or {}
+        # currently, cross database operation is not supported
+        if (source_table and self.target_table) and (
+                source_table.sql_type != self.target_table.sql_type
+        ):
+            raise ValueError("source and target table must belongs from same datasource")
         task_id = task_id or get_unique_task_id("append_table")
         super().__init__(
             task_id=task_id,
@@ -65,9 +70,6 @@ class AppendOperator(AstroSQLBaseOperator):
         db = create_database(self.target_table.conn_id, table=self.source_table)
         self.source_table = db.populate_table_metadata(self.source_table)
         self.target_table = db.populate_table_metadata(self.target_table)
-        # currently, cross database operation is not supported
-        if self.source_table.sql_type != self.target_table.sql_type:
-            raise ValueError("source and target table must belongs from same datasource")
         db.append_table(
             source_table=self.source_table,
             target_table=self.target_table,
--- a/python-sdk/src/astro/sql/operators/append.py
+++ b/python-sdk/src/astro/sql/operators/append.py
@@ -55,11 +55,6 @@ class AppendOperator(AstroSQLBaseOperator):
                 f"columns is not a valid type. Valid types: [tuple, list, dict], Passed: {type(columns)}"
             )
         self.columns = columns or {}
-        # currently, cross database operation is not supported
-        if (source_table and self.target_table) and (
-                source_table.sql_type != self.target_table.sql_type
-        ):
-            raise ValueError("source and target table must belongs from same datasource")
         task_id = task_id or get_unique_task_id("append_table")
         super().__init__(
             task_id=task_id,
@@ -68,6 +63,11 @@ class AppendOperator(AstroSQLBaseOperator):
 
     def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
         db = create_database(self.target_table.conn_id, table=self.source_table)
+        # currently, cross database operation is not supported
+        if (self.source_table and self.target_table) and (
+            self.source_table.sql_type != self.target_table.sql_type
+        ):
+            raise ValueError("source and target table must belongs from same datasource")
         self.source_table = db.populate_table_metadata(self.source_table)
         self.target_table = db.populate_table_metadata(self.target_table)
         db.append_table(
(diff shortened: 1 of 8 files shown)


python-sdk/src/astro/sql/operators/append.py
```diff
@@ -63,6 +63,11 @@ def __init__(
 
     def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
         db = create_database(self.target_table.conn_id, table=self.source_table)
+        # currently, cross database operation is not supported
+        if (self.source_table and self.target_table) and (
```
pankajkoti, line 67: Would `self.source_table` or `self.target_table` be not set? If yes, do we need a separate error for them if they are unset?
e.g 
```
if not self.source_table or not self.source_table:
    raise some_error 
```

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -86,6 +86,14 @@ def execute(self, context: Context) -> None:
                 raise ValueError("You need to provide a table or a connection id")
         self.database_impl = create_database(self.conn_id, first_table)
 
+        # currently, cross database operation is not supported
+        if (
+            (first_table and self.output_table)
```
pankajkoti, line 91: Same question as above.

python-sdk/src/astro/sql/operators/append.py
pankajastro: In the case of a merge, an append operator source_table and target_table are required so the constructor should take care of it. Maybe if you want we can reduce it to.

```
if self.source_table.sql_type != self.target_table.sql_type:
    raise error
```

python-sdk/src/astro/sql/operators/base_decorator.py
pankajastro: These are not required param always in the case of BaseSQLDecoratedOperator

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -63,6 +63,9 @@ def __init__(
 
     def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
         db = create_database(self.target_table.conn_id, table=self.source_table)
+        # currently, cross database operation is not supported
+        if self.source_table.sql_type != self.target_table.sql_type:
+            raise ValueError("source and target table must belongs from same datasource")
```
pankajkoti, line 68:
```suggestion
            raise ValueError("source and target table must belong to the same datasource")
```

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -86,6 +86,14 @@ def execute(self, context: Context) -> None:
...
+            and (first_table.sql_type and self.output_table.sql_type)
+            and (first_table.sql_type != self.output_table.sql_type)
+        ):
+            raise ValueError("source and target table must belongs from same datasource")
```
pankajkoti, line 95:
```suggestion
            raise ValueError("source and target table must belong to the same datasource")
```

python-sdk/src/astro/sql/operators/merge.py
```diff
@@ -75,6 +75,9 @@ def __init__(
 
     def execute(self, context: Context) -> BaseTable:
         db = create_database(self.target_table.conn_id, table=self.source_table)
+        # currently, cross database operation is not supported
+        if self.source_table.sql_type != self.target_table.sql_type:
+            raise ValueError("source and target table must belongs from same datasource")
```
pankajkoti, line 80:
```suggestion
            raise ValueError("source and target table must belong to the same datasource")
```

python-sdk/tests/sql/operators/test_append.py
```diff
@@ -172,6 +171,7 @@ def test_append_on_tables_on_different_db(sample_dag, database_table_fixture):
                 source_table=load_append,
             )
         test_utils.run_dag(sample_dag)
+    assert exec_info.value.args[0] == "source and target table must belongs from same datasource"
```
pankajkoti, line 174:
```suggestion
    assert exec_info.value.args[0] == "source and target table must belong to the same datasource"
```

python-sdk/tests/sql/operators/test_merge.py
```diff
@@ -353,3 +353,17 @@ def test_inlets_outlets_non_supported_ds():
...
+    target_table = Table(conn_id="bigquery", name="test2", metadata=Metadata(schema="test"))
+    with pytest.raises(ValueError) as exec_info:
+        MergeOperator(
+            source_table=source_table,
+            target_table=target_table,
+            if_conflicts="ignore",
+            target_conflict_columns=["list"],
+            columns=["set_item_1", "set_item_2", "set_item_3"],
+        ).execute({})
+    assert exec_info.value.args[0] == "source and target table must belongs from same datasource"
```
pankajkoti, line 369:
```suggestion
    assert exec_info.value.args[0] == "source and target table must belong to the same datasource"
```

python-sdk/tests/sql/operators/transform/test_transform.py
```diff
@@ -261,3 +261,26 @@ def select(input_table: Table):
...
+        """
+
+    with sample_dag:
+        input_table = Table(conn_id="snowflake_conn", name="test1", metadata=Metadata(schema="test"))
+        output_table = Table(conn_id="bigquery", name="test2", metadata=Metadata(schema="test"))
+
+        top_five_animations(input_table=input_table, output_table=output_table)
+    with pytest.raises(ValueError) as exec_info:
+        test_utils.run_dag(sample_dag)
+    assert exec_info.value.args[0] == "source and target table must belongs from same datasource"
```
pankajkoti, line 286:
```suggestion
    assert exec_info.value.args[0] == "source and target table must belong to the same datasource"
```

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -63,6 +63,9 @@ def __init__(
 
     def execute(self, context: Context) -> BaseTable:  # skipcq: PYL-W0613
         db = create_database(self.target_table.conn_id, table=self.source_table)
```
pankajkoti, line 65: Can we move this line below the if condition, so that we avoid another DB call with that method in case tables do not belong to the same datasource?

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -86,6 +86,14 @@ def execute(self, context: Context) -> None:
                 raise ValueError("You need to provide a table or a connection id")
         self.database_impl = create_database(self.conn_id, first_table)
```
pankajkoti, line 87: Can we move this line below the if condition, so that we avoid another DB call with that method in case tables do not belong to the same datasource?

python-sdk/src/astro/sql/operators/merge.py
```diff
@@ -75,6 +75,9 @@ def __init__(
 
     def execute(self, context: Context) -> BaseTable:
         db = create_database(self.target_table.conn_id, table=self.source_table)
```
pankajkoti, line 77: Can we move this line below the if condition, so that we avoid another DB call with that method in case tables do not belong to the same datasource?

pankajkoti (approved):
Looks good to me. Please consider the minor suggestions.

PR #1237, opened 2022-11-10
Tidy up `Table.name` attr handling for better 2.5 XCom support

ashb:
We don't need to make it a property, we can just create the temp name in postinit.

(We "could" set the name via `@name.default`, but the default for this attribute is set in "field definition order" which means that we'd try to set the default before `self.metadata` has been set, which causes problems, and we don't really want to re-order the attributes as that affects how our users use this class, and name has to remain as the first arg there)

### DIFF
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -4,7 +4,7 @@ import random
 import string
 from typing import Any
 
-import attrs
+import attr.setters
 from attr import define, field, fields_dict
 from sqlalchemy import Column, MetaData
 
@@ -55,7 +55,7 @@ class BaseTable:
     # TODO: discuss alternative names to this class, since it contains metadata as opposed to be the
     # SQL table itself
     # Some ideas: TableRef, TableMetadata, TableData, TableDataset
-    name: str = field(default="", on_setattr=attrs.setters.validate)
+    name: str = field(default="", on_setattr=attr.setters.validate)
     conn_id: str = field(default="")
     # Setting converter allows passing a dictionary to metadata arg
     metadata: Metadata = field(
--- a/python-sdk/src/astro/table.py
+++ b/python-sdk/src/astro/table.py
@@ -4,6 +4,7 @@ import random
 import string
 from typing import Any
 
+import attrs
 from attr import define, field, fields_dict
 from sqlalchemy import Column, MetaData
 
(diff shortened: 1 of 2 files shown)


dimberman (approved):

kaxil (approved):
Once the failing test pass, lgtm

python-sdk/tests/sql/test_table.py
```diff
@@ -137,8 +137,8 @@ def test_table_to_datasets_extra():
     "table",
     [
         Table(),
-        Table("_tmp"),
-        Table(name="_tmp", conn_id="test_conn"),
+        Table("_tmp_a"),
+        Table(name="_tmp_a", conn_id="test_conn"),
```
tatiana, line 141: @ashb Why did we change these names?

python-sdk/src/astro/table.py
```diff
@@ -12,7 +13,7 @@
 from astro.settings import OPENLINEAGE_EMIT_TEMP_TABLE_EVENT
 
 MAX_TABLE_NAME_LENGTH = 62
-TEMP_PREFIX = "_tmp_"
+TEMP_PREFIX = "_tmp"
```
tatiana, line 16: Why did we changed the prefix?

python-sdk/tests/sql/test_table.py
kaxil: We had inconsistency in `_tmp_` (defined by `TEMP_PREFIX`) and what we checked for in Table's `__init__` as shown in the code-blocks below

https://github.com/astronomer/astro-sdk/blob/5d714ad468e3f8ddb4f14741b6385d0d3f89b85d/python-sdk/src/astro/table.py#L16

https://github.com/astronomer/astro-sdk/blob/0e9d07b53bb2218dabecdbb991dd49cc2ff94cfe/python-sdk/src/astro/table.py#L72

In general, users are “NOT” passing names at all for temp tables, so I had suggested to just make it consistent and use the value in `TEMP_PREFIX` because that is what we use for generating table names. (probably >99.5% cases)

python-sdk/src/astro/table.py
kaxil: It was inconsistent before so we had to just between the above and this https://github.com/astronomer/astro-sdk/pull/1237#discussion_r1020094281

python-sdk/tests/sql/test_table.py
ashb: And I've now reverted this change (or rather force pushed without this commit) in favour of the next commit.

Issue #810, opened 2022-09-09
aql.dataframe does not tolerate empty datafames


rossturk:
**Describe the bug**
When using `aql.dataframe()` I noticed that it does not gracefully tolerate functions that return an empty dataframe.

**Version**
* Astro: 1.0.2
* OS: Debian 11

**To Reproduce**

1. Run this DAG:

```
CONN_ID = "dwh"

@aql.dataframe
def get_empty_dataframe():
    zilch = []
    return pd.DataFrame(zilch)

with DAG(
    "empty-dataframe",
    schedule_interval="@daily",
    start_date=datetime(2022, 8, 1),
) as dag:
    dataframe = get_empty_dataframe(
        output_table=Table(
            conn_id=CONN_ID,
        )
    )

    aql.cleanup()
```

2. See the resulting error:

```
[2022-09-09T03:05:08.615+0000] {standard_task_runner.py:92} ERROR - Failed to execute job 8277 for task get_empty_dataframe ((snowflake.connector.errors.ProgrammingError) 001003 (42000): 01a6d839-0605-617c-0000-68210fdaed92: SQL compilation error:
syntax error line 2 at position 0 unexpected ')'.
[SQL: 
CREATE TABLE "ROSSTURK"._tmp_jntrpgqnu65eaf3e1c94noiz1pd5gt70wlniesbnklb67i7eezkk9pxy1 (
)

]
```

**Expected behavior**
I think it should be possible to work with empty dataframes - both with this decorator and `aql.append()`.

utkarsharma2:
@rossturk for creating this issue, We'll soon be back with the updates on this.

pankajastro:
Does snowflake allow the creation of an empty table? I think `CREATE TABLE t ();` will not work in the case of the snowflake. In the case of an empty dataframe what is the table structure we are expecting here?

pankajastro:
In the case of snowflake improve logging message @sunank200

PR #1238, opened 2022-11-11
Raise error if dataframe is empty

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/810
## What is the new behavior?

-
-
-

### Checklist


python-sdk/src/astro/databases/snowflake.py
```diff
@@ -645,7 +645,8 @@ def load_pandas_dataframe_to_table(
         :param if_exists: Strategy to be used in case the target table already exists.
         :param chunk_size: Specify the number of rows in each batch to be written at a time.
         """
-
+        if source_dataframe.empty:
```
pankajkoti, line 648: Should this check also be added in the base class method? https://github.com/astronomer/astro-sdk/blob/0e9d07b53bb2218dabecdbb991dd49cc2ff94cfe/python-sdk/src/astro/databases/base.py#L532

python-sdk/src/astro/databases/sqlite.py
```diff
@@ -84,6 +85,26 @@ def get_merge_initialization_query(parameters: tuple) -> str:
...
+        """
+        Create a table with the dataframe's contents.
+        If the table already exists, append or replace the content, depending on the value of `if_exists`.
+
+        :param source_dataframe: Local or remote filepath
+        :param target_table: Table in which the file will be loaded
+        :param if_exists: Strategy to be used in case the target table already exists.
+        :param chunk_size: Specify the number of rows in each batch to be written at a time.
+        """
+        if source_dataframe.empty:
```
pankajkoti, line 104: If this check is added to the base class method, we can remove this method override altogether, no?

pankajkoti (commented):
Do we need to add this check for `postgres`, `redshift` and `bigquery` too? Since we're overriding the `load_pandas_dataframe_to_table` method in those implementations too?

python-sdk/src/astro/databases/snowflake.py
pankajastro: I'll not mind adding it in the base class too but some do DB allow to table without a column.

tatiana (approved):

python-sdk/src/astro/databases/snowflake.py
pankajkoti: In my opinion, the method is called `load_pandas_dataframe_to_table`. So in case the dataframe is empty, there is nothing to load, shouldn't we alert in that case?

feluelle: > I'll not mind adding it in the base class too but some DB do allow to table without a column.

Interesting. I am curious which db does that and why 🤔

pankajastro: I don't know the exact reason but Postgres and bigquery do allow to creation of a table without a column.  You might like to read some more opinions here https://stackoverflow.com/questions/18024585/why-to-create-empty-no-rows-no-columns-table-in-postgresql

pankajkoti:
In my opinion, we should not be creating tables as part of `load_pandas_dataframe_to_table` if an empty dataframe is passed as to me it sounds like a side effect. 
If the user really wants to create a table by passing an empty dataframe, we could add a separate method for that. But `load_pandas_dataframe_to_table` would mean load a dataframe and if that is empty in my opinion, we should alert instead of silently just creating the table.

python-sdk/src/astro/databases/snowflake.py
sunank200: I feel that we should have a separate method which should check this in the base class. That method can be overridden for Postgres and Bigquery.

python-sdk/tests/sql/operators/test_dataframe.py
```diff
@@ -435,3 +435,32 @@ def validate(df):  # skipcq: PY-D0003
...
+    "conn_id",
+    [
+        "bigquery",
+        "postgres_conn",
+        "redshift_conn",
+        "snowflake_conn",
+        "sqlite_conn",
+    ],
+)
+def test_empty_dataframe_fail(sample_dag, conn_id):
```
sunank200, line 450: This seems integration test. It would be good to mark it as integration

sunank200 (commented):
I feel that we should have a separate method which should check if dataframe is empty in the base class. That method can be overridden for Postgres and Bigquery.

sunank200:
> In my opinion, we should not be creating tables as part of `load_pandas_dataframe_to_table` if an empty dataframe is passed, as to me, it sounds like a side effect. If the user really wants to create a table by passing an empty dataframe, we could add a separate method for that. But `load_pandas_dataframe_to_table` would mean load a dataframe and if that is empty, in my opinion, we should alert instead of silently just creating the table.

+1 on this. I feel we should not do multiple things as part of same method.

python-sdk/src/astro/databases/base.py
```diff
@@ -458,6 +458,15 @@ def get_dataframe_from_file(file: File):
 
         return file.export_to_dataframe()
 
+    @staticmethod
+    def _assert_not_empty_df(df):
```
sunank200, line 462: Shouldn't this method be overridden for PostGres and Bigquery like you had mentioned?

sunank200 (commented):
Shouldn't `_assert_not_empty_df(df)` be overridden for PostGres and Bigquery like you had mentioned?

python-sdk/src/astro/databases/snowflake.py
pankajastro: I have wrapped the error with the function `_assert_not_empty_df` in the base database and also add the check in base database too

Issue #1243, opened 2022-11-14
Issue with openlineage import


phanikumv:
**Describe the bug**
We shouldn’t import the openlineage package at top level if it is optional.
Refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1668186729410429

**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Write the DAG  '...'
2. Create connection '....'
3. Run using  '....'

**Expected behavior**

**Screenshots**

**Additional context**

PR #1252, opened 2022-11-15
Make `openlineage` an optional dependency

sunank200:
# Description

We shouldn’t import the openlineage package at top level if it is optional.
Refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1668186729410429

closes: #1243

## What is the new behavior?

- Add a try except for import error
- Remove the open lineage from test in pyproject

No

### Checklist

### DIFF
--- a/python-sdk/noxfile.py
+++ b/python-sdk/noxfile.py
@@ -67,7 +67,7 @@ def test_examples_by_dependency(session: nox.Session, extras):
     pytest_args = ["-k", pytest_options]
 
     session.install("-e", f".[{pypi_deps}]")
-    session.install("-e", ".[tests]")
+    session.install("-e", ".[openlineage, tests]")
     session.run("airflow", "db", "init")
 
     # Since pytest is not installed in the nox session directly, we need to set `external=true`.
--- a/python-sdk/noxfile.py
+++ b/python-sdk/noxfile.py
@@ -67,7 +67,7 @@ def test_examples_by_dependency(session: nox.Session, extras):
     pytest_args = ["-k", pytest_options]
 
     session.install("-e", f".[{pypi_deps}]")
-    session.install("-e", ".[openlineage, tests]")
+    session.install("-e", ".[tests]")
     session.run("airflow", "db", "init")
 
     # Since pytest is not installed in the nox session directly, we need to set `external=true`.
--- a/python-sdk/src/astro/lineage/__init__.py
+++ b/python-sdk/src/astro/lineage/__init__.py
@@ -1,0 +1,23 @@
+import logging
+
+log = logging.getLogger(__name__)
+
+try:
+    from openlineage.airflow.extractors import TaskMetadata
+    from openlineage.airflow.extractors.base import BaseExtractor
+    from openlineage.airflow.utils import get_job_name
+    from openlineage.client.facet import (
+        BaseFacet,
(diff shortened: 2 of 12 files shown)


python-sdk/noxfile.py
```diff
@@ -67,7 +67,7 @@ def test_examples_by_dependency(session: nox.Session, extras):
     pytest_args = ["-k", pytest_options]
 
     session.install("-e", f".[{pypi_deps}]")
-    session.install("-e", ".[tests]")
+    session.install("-e", ".[openlineage, tests]")
```
kaxil, line 70: We don't need to install `openlineage` here, right?

python-sdk/src/astro/sql/operators/append.py
```diff
@@ -1,19 +1,24 @@
...
+        DataQualityMetricsInputDatasetFacet,
+        DataSourceDatasetFacet,
+        OutputStatisticsOutputDatasetFacet,
+        SchemaDatasetFacet,
+        SchemaField,
+        SqlJobFacet,
+    )
+    from openlineage.client.run import Dataset as OpenlineageDataset
+except ImportError:
+    logging.warning("Install openlineage-airflow")
```
kaxil, line 21: Hmm, I don't like the `try..except` in all the modules, since it is used everywhere.. We can just try to import it once in some common module, and use that.

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -1,12 +1,17 @@
...
 import pandas as pd
 from airflow.decorators.base import get_unique_task_id
 from airflow.models.xcom_arg import XComArg
-from openlineage.client.facet import BaseFacet, DataSourceDatasetFacet, SchemaDatasetFacet, SchemaField
-from openlineage.client.run import Dataset as OpenlineageDataset
+
+try:
+    from openlineage.client.facet import BaseFacet, DataSourceDatasetFacet, SchemaDatasetFacet, SchemaField
+    from openlineage.client.run import Dataset as OpenlineageDataset
+except ImportError:
```
kaxil, line 13: Line 21 still will import openlineage library

kaxil:
I have pushed a different way of handling this in https://github.com/astronomer/astro-sdk/pull/1252/commits/a89689674a8800f9774164ecbc9fe7bf35645cf7

python-sdk/src/astro/lineage/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+        OutputStatisticsOutputDatasetFacet,
+        SchemaDatasetFacet,
+        SchemaField,
+        SqlJobFacet,
+    )
+    from openlineage.client.run import Dataset as OpenlineageDataset
+
+    has_openlineage = True
+except ImportError:
+    has_openlineage = False
```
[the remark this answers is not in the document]
feluelle: Can we not use `has_openlineage` in the other moduls to decide if we want to import or not?

kaxil (approved):

python-sdk/src/astro/lineage/__init__.py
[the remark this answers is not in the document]
kaxil: Removed.

>Can we not use has_openlineage in the other modules to decide if we want to import or not?

yeah, just another way -- no strong opinion

PR #1253, opened 2022-11-15
Refactor cli tests to rely on large terminal

feluelle:
Instead of changing the stdout via get_stdout, we just set the TERMINAL_WIDTH to a very large number to avoid new lines. The configuration has been copied from https://github.com/tiangolo/typer/blob/master/scripts/test.sh

**EDIT:**

I have also created https://github.com/tiangolo/typer/pull/509 to fix that in typer.

### DIFF
--- a/.github/workflows/ci-sql-cli.yaml
+++ b/.github/workflows/ci-sql-cli.yaml
@@ -99,6 +99,7 @@ jobs:
     env:
       AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
       AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV: True
+      TERMINAL_WIDTH: 3000
       _TYPER_FORCE_DISABLE_TERMINAL: 1
 
 
--- a/sql-cli/sql_cli/utils/rich.py
+++ b/sql-cli/sql_cli/utils/rich.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import IO, Any, Optional, Union
 
 import click
 from rich.panel import Panel
@@ -39,9 +39,6 @@ def rich_format_error(self: click.ClickException) -> None:
             title_align=ALIGN_ERRORS_PANEL,
         )
     )
-from typing import IO, Any, Optional
-
-from typer.rich_utils import _get_rich_console
 
 
 def rprint(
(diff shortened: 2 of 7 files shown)


.github/workflows/ci-sql-cli.yaml
```diff
@@ -99,6 +99,7 @@ jobs:
     env:
       AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
       AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV: True
+      TERMINAL_WIDTH: 3000
```
tatiana, line 102: @feluelle do you see any downsides to increasing the terminal width to this value?

feluelle: Not really. The advantage is that we can start having tests checking for explicit new lines whereas if we don't set the terminal width we need to either replace the new lines always, but then we cannot check for new lines in output, to make it consistent or check for the new lines in conjunction with the actual string/output. The question is what would be ideal terminal width be? I think there is none, hence we are setting it to a large number to avoid terminal doing terminal things.

tatiana: Sounds like a good plan! I'm happy for us to merge this once the checks are passing!

feluelle: It seems setting `TERMINAL_WIDTH` does not change reflect the terminal size in the tests. I can reproduce on a small local terminal :(

sql-cli/sql_cli/utils/rich.py
```diff
@@ -0,0 +1,29 @@
+from typing import IO, Any, Optional
+
+from typer.rich_utils import _get_rich_console
+
+
+def rprint(
```
pankajkoti, line 6: Should we also use this function as an import instead of
1. https://github.com/astronomer/astro-sdk/blob/f50be8d590ff29b44b81e3ba340152bc316a429b/sql-cli/sql_cli/run_dag.py#L19

2. https://github.com/astronomer/astro-sdk/blob/f50be8d590ff29b44b81e3ba340152bc316a429b/sql-cli/sql_cli/connections.py#L7

pankajkoti (approved):
Looks good to me, have a minor question inline above using the same utils function across other usages of rich.print across other modules.

sql-cli/sql_cli/utils/rich.py
feluelle: Yes, makes sense. Good catch! 👍

PR #1254, opened 2022-11-15
Add run command to python-sdk so users do not need SQL CLI on their a…

dimberman:
…iflow instances to run DAGs

# Description
Currently run_dag only exists in the SQL CLI library. This means that users do not have the sql-cli installed can not take advantage of the run_dag function. 

## What is the new behavior?
This function makes the run_dag function a part of the astro library so users can directly call it via the astro SDK in their DAGs. It also means that the astro CLI can use the run_dag function without installing the SQL CLI.

### Checklist


tatiana (commented):
@dimberman, I have mixed feelings about having this as part of the Python SDK.

On the one hand, it makes sense - since we're running Python DAGs, backporting a feature available in Airflow.

On the other hand, I'm worried about how we'll align this implementation with the SQL CLI run implementation - we may have to support two very similar things.

Is this PR mostly about transferring the existing functionality of the SQL CLI, or does it introduce other changes we should be aware of?

python-sdk/src/astro/run_dag.py
```diff
@@ -0,0 +1,235 @@
...
+    from airflow.utils.cli import get_dag
+
+    parser = ArgumentParser()
+    parser.add_argument("dag_id")
+    parser.add_argument("-d", "--dags_file", default=None)
+    parser.add_argument("-c", "--connections-file")
+    args = parser.parse_args()
+    connections = _convert_connections_yaml_to_airflow_objs(args.connections_file)
+    dag = get_dag(args.dags_file, dag_id=args.dag_id)
+    run_dag(dag=dag, connections=connections)
```
tatiana, line 235: What are our alternatives to exposing a CLI in the Python SDK?

feluelle:
Yes, I agree with @tatiana. 

It would be so much nicer if we could have that in core airflow or as separate python package. Then we would not have to maintain two implementations - `run_dag` is not specific to _astro-sdk-python_ DAGs nor _astro-sql-cli_ workflows. It works for all kind of DAGs, right? If we put it into the sdk, I am afraid, we are being slowed down by astro-sdk-python releases. If we have it separately, we can more quickly iterate and release changes.

tatiana:
We decided to not go ahead with this proposal, and we moved the feature to a separate repo:
https://github.com/astronomer/astro-run-dag

The discussion and reasoning can be found at:
https://docs.google.com/document/d/19rSL2nQljdg0SI5NREKpVls-uSXxYVFk4Nq_6VLQO8A/edit#

PR #1257, opened 2022-11-16
Fix temp tables snowflake

rajaths010494:
# Description
## What is the new behavior?

-
-
-

### Checklist


python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
```diff
@@ -86,19 +88,20 @@ def example_amazon_s3_snowflake_transform():
...
-            metadata=Metadata(
-                schema=os.environ["SNOWFLAKE_SCHEMA"],
-                database=os.environ["SNOWFLAKE_DATABASE"],
-            ),
-            conn_id="snowflake_conn",
-        ),
+        output_table=snowflake_output_table,
     )
     # [END dataframe_example_2]
-    aql.cleanup()
```
feluelle, line 101: @rajaths010494 don't we need that line?

rajaths010494: yes we need it.  I missed this when resolving conflicts.

rajaths010494: added it back

feluelle (approved):

python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
```diff
@@ -58,6 +58,7 @@ def example_amazon_s3_snowflake_transform():
             schema=os.environ["SNOWFLAKE_SCHEMA"],
         ),
         conn_id="snowflake_conn",
+        temp=True,
```
pankajastro, line 61: if you want to create tmp table then the below would be sufficient or do we need the table name for some reason? in our context the tmp usually contains prefix _tmp 
```
input_table_1 = Table(
        metadata=Metadata(
            database=os.environ["SNOWFLAKE_DATABASE"],
            schema=os.environ["SNOWFLAKE_SCHEMA"],
        ),
        conn_id="snowflake_conn",
)
```

rajaths010494: we would not need the table name. Even _tmp should suffice. Let me change that and run once the dag.

sunank200: I think we need named temp tables in this case. Also this `temp=True` is valid way to create temp tables as suggested in our docs too: https://astro-sdk-python.readthedocs.io/en/stable/guides/concepts.html

python-sdk/example_dags/example_snowflake_partial_table_with_append.py
```diff
@@ -102,10 +102,12 @@ def example_snowflake_partial_table_with_append():
     extracted_data = extract_data(
         homes1=homes_data1,
         homes2=homes_data2,
-        output_table=Table(name="combined_homes_data"),
+        output_table=Table(name="combined_homes_data", temp=True),
```
sunank200, line 105: We should name this table uniquely so that concurrent run on CI doesn't fail it.

python-sdk/example_dags/example_snowflake_partial_table_with_append.py
```diff
@@ -102,10 +102,12 @@ def example_snowflake_partial_table_with_append():
...
     )
 
-    transformed_data = transform_data(df=extracted_data, output_table=Table(name="homes_data_long"))
+    transformed_data = transform_data(
+        df=extracted_data, output_table=Table(name="homes_data_long", temp=True)
```
sunank200, line 109: We should name this table uniquely so that concurrent run on CI doesn't fail it.

sunank200 (changes requested):
Added few comments for changes

python-sdk/example_dags/example_snowflake_partial_table_with_append.py
rajaths010494: changed and also made table name unique

python-sdk/example_dags/example_snowflake_partial_table_with_append.py
rajaths010494: changed and also made table name unique

phanikumv (approved):

sunank200 (approved):
LGTM now

Issue #1227, opened 2022-11-09
Integrate load file operator together with generate and run in SQL CLI


pankajkoti:
**Please describe the feature you'd like to see**
The Load file task would be defined using an agreed upon YAML template

**Describe the solution you'd like**
To generate DAGs based on the YAML for the load file operator and establish appropriate dependencies. Generate and Run commands should be able to work appropriately with this operator.

**Acceptance Criteria**

- [ ] By default, debug is enabled, and we only see the clean output
- [ ] If debug is enabled, we should see INFO logs 
- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

pankajkoti:
I am working on the implementation right now. Discussed few refactoring with the team on the team channel and planning to create a draft PR by Wednesday afternoon IST.

PR #1258, opened 2022-11-16
[sql-cli] Implement load file operator

pankajkoti:
# Description

Currently, we are not having a way for users to load data into databases.

closes: #1227

## What is the new behavior?

Add `load_file` / `LoadFileOperator` to the sql-cli.

No.

### Checklist


python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -49,8 +49,11 @@ def __init__(
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        task_id: str | None = None,
         **kwargs,
     ) -> None:
+        self.task_id = task_id or get_unique_task_id("load_file")
+        kwargs["task_id"] = self.task_id
         super().__init__(
             **kwargs_with_datasets(
```
feluelle, line 58:
```suggestion
        task_id = task_id or get_unique_task_id("load_file")
        super().__init__(
            task_id=task_id,
            **kwargs_with_datasets(
```
I think we don't need the task_id in `self` or?

sql-cli/conftest.py
```diff
@@ -52,6 +52,7 @@ def sql_file(root_directory, dags_directory):
         root_directory=root_directory,
         path=root_directory / "a.sql",
         target_directory=dags_directory,
+        file_type="sql",
```
feluelle, line 55: This should be set if necessary inside the class `SqlFile`.

sql-cli/conftest.py
```diff
@@ -115,7 +119,7 @@ def sample_dag():
     yield DAG(dag_id, start_date=DEFAULT_DATE)
     with create_session() as session_:
         session_.query(DagRun).delete()
-        session_.query(TI).delete()
+        session_.query(Ti).delete()
```
feluelle, line 122: Why this change?

sql-cli/sql_cli/dag_generator.py
```diff
@@ -7,56 +7,60 @@
...
-from sql_cli.sql_directory_parser import SqlFile, get_sql_files
 from sql_cli.utils.jinja import render
+from sql_cli.workflow_directory_parser import WorkflowFile, get_workflow_files
 
 TEMPLATES_DIRECTORY = Path("templates")
 
 
 @dataclass(frozen=True)
-class SqlFilesDAG:
+class WorkflowFilesDAG:
```
feluelle, line 17: How about we call this just `Workflow`?

sql-cli/sql_cli/operators/load_file.py
```diff
@@ -0,0 +1,27 @@
...
+
+from astro.custom_backend.serializer import deserialize
+from astro.sql import LoadFileOperator
+
+
+def get_load_file_instance(operator_content: dict[str, Any], operator_file_name: str) -> LoadFileOperator:
+    input_file_content = operator_content.pop("input_file")
+    input_file_content["class"] = "File"
+    if "is_dataframe" not in input_file_content:
+        input_file_content["is_dataframe"] = False
```
feluelle, line 13:
```suggestion
    input_file_content.setdefault("is_dataframe", False)
```

feluelle: See https://docs.python.org/3/library/stdtypes.html#dict.setdefault

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,242 @@
...
+    return {
+        SqlFile(root_directory=directory, path=child, target_directory=target_directory)  # type: ignore
+        for child in directory.rglob(f"*.{SqlFile.file_type}")
+        if child.is_file() and not child.is_symlink()
+    }
+
+
+def get_yaml_files(directory: Path, target_directory: Path | None) -> set[YamlFile]:
+    """
+    Get all sql files within a directory.
```
feluelle, line 225: Please don't forget to update the doc strings.

python-sdk/src/astro/sql/operators/load_file.py
tatiana: Yes, I agree, the `BaseOperator` will take care of setting it to self (https://github.com/apache/airflow/blob/main/airflow/models/baseoperator.py#L775)

sql-cli/conftest.py
```diff
@@ -61,6 +62,7 @@ def sql_file_with_parameters(root_directory, dags_directory):
         root_directory=root_directory,
         path=root_directory / "c.sql",
         target_directory=dags_directory,
+        file_type="sql",
```
tatiana, line 65: This feels redundant with the class name. Do we need to expose it?

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,242 @@
...
+    :param directory: The directory look in for yaml files.
+    :param target_directory: The target directory path for the executable yaml workflow.
+
+    :returns: the workflow files found in the directory.
+    """
+    return {
+        YamlFile(root_directory=directory, path=child, target_directory=target_directory)  # type: ignore
+        for child in directory.rglob(f"*.{YamlFile.file_type}")
+        if child.is_file() and not child.is_symlink()
+    }
```
tatiana, line 236: These two methods are super similar. Would it make sense to unify them into a single `get_files_by_type`?

sql-cli/conftest.py
pankajkoti: Pycharm complains that camelCase imported as constant, so had made that change. But, reverted now.

Screenshot 2022-11-17 at 12 56 22 PM

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,231 @@
...
+        return TransformOperator(**kwargs)
+
+
+class YamlFile(WorkflowFile):
+    operator_instance_builder_callable_map = {LOAD_FILE_OPERATOR: get_load_file_instance}
+
+    def __init__(self, root_directory: Path, path: Path, target_directory: Path) -> None:
+        super().__init__(root_directory, path, target_directory, YAML_FILE_TYPE)
+        with open(path) as yaml_file:
+            self.yaml_content = yaml.load(yaml_file, Loader=yaml.Loader)
```
feluelle, line 188:
```suggestion
        with path.open() as yaml_file:
            data = yaml.safe_load(yaml_file)
```

feluelle: Path attributes can be opened using `.open`.

For loading yaml, always use `safe_load`. (see [this](https://stackoverflow.com/a/63916315) for example)

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,231 @@
...
+        self.operator = self._get_operator()
+        self.yaml_file_name = self.path.stem
+
+    def _get_operator(self) -> str:
+        operator = list(self.yaml_content.keys())[0]
```
feluelle, line 193: I think we should validate that only one key is allowed, right? Otherwise the operator returned here could be random as keys is a dict and the order is not guaranteed to be the same.

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,231 @@
...
+        if operator not in YamlFile.operator_instance_builder_callable_map:
+            raise NotImplementedError(f"Operator support for {operator} not available")
+        return operator
+
+    def to_operator(self) -> LoadFileOperator:
```
feluelle, line 198:
```suggestion
    def to_operator(self) -> BaseOperator:
```
maybe? Or we manually extend e.g. `LoadFileOperator | ExportFileOperator | ...`.

WDYT?

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,231 @@
...
+            / "/".join(self.get_sub_directories())
+        )
+        target_full_directory.mkdir(parents=True, exist_ok=True)
+
+        target_path = target_full_directory / self.path.name
+
+        target_path.write_text(self.raw_content)
+
+    @abstractmethod
+    def to_operator(self) -> Any:
```
feluelle, line 150:
```suggestion
    def to_operator(self) -> BaseOperator:
```
maybe?

sql-cli/sql_cli/constants.py
```diff
@@ -4,3 +4,11 @@
 DEFAULT_ENVIRONMENT = "default"
 DEFAULT_AIRFLOW_HOME = f".airflow/{DEFAULT_ENVIRONMENT}"
 DEFAULT_DAGS_FOLDER = ".airflow/dags"
+
+GENERATED_WORKFLOW_INCLUDE_DIRECTORY = "include"
+
+SQL_FILE_TYPE = "sql"
+YAML_FILE_TYPE = "yaml"
```
feluelle, line 11: We should also allow `yml` files - not only `yaml`.

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,231 @@
...
+            raise NotImplementedError(f"Operator support for {operator} not available")
+        return operator
+
+    def to_operator(self) -> LoadFileOperator:
+        return self.operator_instance_builder_callable_map[self.operator](
+            self.yaml_content[self.operator], self.yaml_file_name
+        )
+
+
+FILE_TYPE_CLASS_MAP = {SQL_FILE_TYPE: SqlFile, YAML_FILE_TYPE: YamlFile}
```
feluelle, line 204:
```suggestion
SUPPORTED_FILES = {"*.sql": SqlFile, "*.yml|*.yaml": YamlFile}
```
We should have file types together with the file classes - not in separate files. I think this makes it more obvious how we map file types to classes. We can also move `SUPPORTED_FILES` in `constants`.

You could then even have sth like the following to exclude specific yaml config files:
```python
SUPPORTED_FILES = {"*.sql": SqlFile, "[!config]*.yml|[!config]*.yaml": YamlFile}
```

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,231 @@
...
+        )
+        for child in directory.rglob(f"*.{file_type}")
+        if child.is_file() and not child.is_symlink()
+    }
+
+
+def get_workflow_files(directory: Path, target_directory: Path | None) -> set[WorkflowFile]:
+    sql_files = get_files_by_type(directory, SQL_FILE_TYPE, target_directory=target_directory)
+    yaml_files = get_files_by_type(directory, YAML_FILE_TYPE, target_directory=target_directory)
+    return sql_files.union(yaml_files)
```
feluelle, line 231: Here we should loop over all `SUPPORTED_FILES`.

sql-cli/sql_cli/workflow_directory_parser.py
pankajkoti: I tried using that pattern, but looks like rglob does not support the pattern you suggested. Added another entry for *.yml in the map.

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,245 @@
...
+        self.yaml_file_name = self.path.stem
+
+    def _get_operator(self) -> str:
+        top_level_keys = list(self.yaml_content.keys())
+        top_level_keys_count = len(top_level_keys)
+        if top_level_keys_count > 1:
+            raise ScannerError(
+                f"Only one top level operator expected. Got {top_level_keys_count} operators: {top_level_keys}"
+            )
+        operator = list(self.yaml_content.keys())[0]
```
feluelle, line 202:
```suggestion
        operator = top_level_keys[0]
```

sql-cli/sql_cli/templates/gen_tasks_dag.py.jinja2
```diff
@@ -14,6 +15,10 @@ with DAG(
     start_date=timezone.parse("{{ dag.start_date }}"),
     schedule_interval=None,
 ) as dag:
-    {%- for sql_file in dag.sorted_sql_files() %}
-    {{ define_task(sql_file, dag)|indent(width=4) }}
+    {%- for workflow_file in dag.sorted_workflow_files() %}
+    {% if workflow_file.__class__.__name__ == 'SqlFile'  %}
+    {{ define_transform_task(workflow_file, dag)|indent(width=4) }}
+    {% elif workflow_file.operator == 'load_file'  %}
+    {{ define_load_file_task(workflow_file, dag, workflow_file.get_yaml_content(), workflow_file.yaml_file_name)|indent(width=4) }}
```
feluelle, line 22:
```suggestion
    {% elif workflow_file.__class__.__name__ == 'YamlFile'  %}
    {{ define_yaml_task(workflow_file, dag)|indent(width=4) }}
```
Can we make this work^?

Because then we would not need to add this for every operator.

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,245 @@
...
+            "sql": self.content,
+        }
+        if airflow.__version__.startswith("2.2."):
+            kwargs["op_args"] = []
+        return TransformOperator(**kwargs)
+
+
+class YamlFile(WorkflowFile):
+    operator_instance_builder_callable_map = {LOAD_FILE_OPERATOR: get_load_file_instance}
+    operator_content_callable_map = {LOAD_FILE_OPERATOR: build_load_file_operator_content}
```
feluelle, line 186:
```suggestion
```

sql-cli/sql_cli/workflow_directory_parser.py
```diff
@@ -0,0 +1,245 @@
...
+    :param file_type: The type of the workflow file. e.g. sql or yaml
+    :param target_directory: The target directory path for the executable workflow file.
+    """
+
+    def __init__(self, root_directory: Path, path: Path, target_directory: Path, file_type: str) -> None:
+        self.root_directory = root_directory
+        self.path = path
+        self.target_directory = target_directory
+        self.raw_content = self.path.read_text()
+        self.file_type = file_type
```
feluelle, line 40: `WorkflowFile` should not have the file_type if it does not use it. We instantiate different WorkflowFile objects based on the file type already.

sql-cli/sql_cli/templates/gen_tasks_dag.py.jinja2
pankajkoti: I think would be good to keep operator level macros to keep it easier to maintain.

sql-cli/include/base/config/default/configuration.yml
```diff
@@ -7,19 +7,17 @@ connections:
...
-#    login: null
-#    password: null
-#    port: null
-#    schema: null
-
+  - conn_id: bigquery
+    conn_type: bigquery
+    description:
+    extra:
+      project: astronomer-dag-authoring
```
tatiana, line 15: WDYT if we retrieve this from an envvar?

sql-cli/sql_cli/macros/load_file_tasks.py.jinja2
```diff
@@ -0,0 +1,7 @@
+{% macro define_load_file_task(workflow_file, dag, yaml_content, yaml_file_name) -%}
+from sql_cli.operators.load_file import build_load_file_operator_content
+operator_content = build_load_file_operator_content({{yaml_content|safe}}, '{{yaml_file_name}}')
+{{ workflow_file.get_variable_name() }} = aql.load_file(
+    **operator_content
```
tatiana, line 5: This is much simpler! :raised_hands: did it work as expected? If yes, should we use a similar strategy for the `tranform_task` template?

pankajkoti: We would not like to use this approach as the generated DAG is inconsistent to read wrt to Python SQL aql tasks.
@feluelle and I are pairing right not to instead expand the Jinja template to make the generated task look similar to Python SDK example

.github/workflows/ci-sql-cli.yaml
```diff
@@ -100,6 +100,9 @@ jobs:
       AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
       AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV: True
       _TYPER_FORCE_DISABLE_TERMINAL: 1
+      GOOGLE_APPLICATION_CREDENTIALS_JSON: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS_JSON }}
+      GOOGLE_APPLICATION_CREDENTIALS: /tmp/google_credentials.json
```
pankajkoti, line 104:
```suggestion
```

.github/workflows/ci-sql-cli.yaml
```diff
@@ -100,6 +101,9 @@ jobs:
...
+      GOOGLE_CLOUD_PORJECT: "astronomer-dag-authoring"
```
feluelle, line 106:
```suggestion
```

feluelle (approved):
LGTM 👍 

Let's only fix the remaining failed SQL CLI test.

All other failing tests are related to python-sdk. :(

sql-cli/conftest.py
```diff
@@ -147,6 +154,38 @@ def initialised_project_with_test_config(initialised_project: Project):
...
+    render(
+        template_file=Path("transform_homes_main.sql.jinja2"),
+        output_file=initialised_project.directory
+        / "workflows"
+        / "example_load_file"
+        / "transform_homes_main.sql",
+        **render_kwargs,
+    )
+
+    return initialised_project
```
tatiana, line 186: It feels that it would be worth having an example workflow with the load. If we had that, would it still make sense to render it in this fixture?

sql-cli/include/base/config/default/configuration.yml
```diff
@@ -7,19 +7,15 @@ connections:
     schema:
     login:
     password:
-
 ## Example of Bigquery connection
-#  - conn_id: bigquery_conn
-#    conn_type: bigquery
-#    description: null
-#    extra:
-#      project: "my-gcp-project"
```
tatiana, line 16: Why did we remove the project from here? Should we adjust the `dev` configuration.yml as well?

sql-cli/include/base/config/default/configuration.yml
```diff
@@ -7,19 +7,15 @@ connections:
...
-#    schema: null
-
+  - conn_id: bigquery
+    conn_type: bigquery
+    description:
+    host:
+    login:
+    password:
+    port:
+    schema:
```
tatiana, line 18: Do we want to have this as a pre-defined connection for the end-users - or should we comment it out - as before? Would this work if they didn't have BQ/GCP providers?

sql-cli/pyproject.toml
```diff
@@ -39,7 +39,7 @@ Jinja2 = ">=2.10.1, <=3.1"
 python-frontmatter = "^1.0.0"
 PyYAML = ">=5.4.1"
 python-dotenv = "^0.21.0"
-astro-sdk-python = {extras = ["all"], version = "^1.2.0"}
+astro-sdk-python = {extras = ["all"], version = "^1.2.2"}
```
tatiana, line 42: It would be great to add a comment to why we only support the latest version of the Astro Python SDK

sql-cli/sql_cli/operators/load_file.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from typing import Any
+
+from astro.custom_backend.serializer import deserialize
+from astro.sql import LoadFileOperator
+
+
+def build_load_file_operator_content(
```
tatiana, line 9: Docstring, please!

sql-cli/sql_cli/operators/load_file.py
```diff
@@ -0,0 +1,30 @@
...
+    output_table_content = operator_content.pop("output_table")
+    output_table_content["class"] = "Table"
+    output_table_content.setdefault("name", operator_file_name)
+    output_table_content.setdefault("temp", False)
+    output_table = deserialize(output_table_content)
+    operator_content["output_table"] = output_table
+    return operator_content
+
+
+def get_load_file_instance(yaml_content: dict[str, Any], operator_file_name: str) -> LoadFileOperator:
```
tatiana, line 27: docstring, please

tatiana (commented):
Great progress on the load file, @pankajkoti and @feluelle! Left some questions and minor comments

feluelle:
Thanks @tatiana for your feedback!

sql-cli/conftest.py
feluelle: Yes, because what we do in that fixture is that we dynamically change the output table name as when running it concurrently the jobs can interfere on the same table.

sql-cli/include/base/config/default/configuration.yml
feluelle: Because it currently does not work. See https://github.com/astronomer/astro-sdk/issues/1277

sql-cli/include/base/config/default/configuration.yml
feluelle: > Would this work if they didn't have BQ/GCP providers?

We currently install all providers i.e. the user does not have a choice.

sql-cli/sql_cli/macros/load_file.py.jinja2
```diff
@@ -0,0 +1,58 @@
+{% macro load_file(yaml_file, dag) -%}
+{% set yaml_content = yaml_file.get_yaml_content() -%}
+{{ yaml_file.get_variable_name() }} = aql.load_file(
+    input_file=File(
```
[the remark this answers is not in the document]
feluelle: Thanks @tatiana - good point. I can add a comment there.

sql-cli/conftest.py
pankajkoti: We do have an example workflow with the load that gets added on project initialisation. We added this fixture as tests running concurrently across Airflow version were creating the same output table and were conflicting across different runs of Airflow version. Here, we manipulate the created workflow to have output table with unique name so that concurrent runs succeed well.

sql-cli/include/base/config/default/configuration.yml
pankajkoti: Added a comment in the same ticket to check whether we need to update the dev configuration.yml too.

sql-cli/include/base/config/default/configuration.yml
pankajkoti: Yes, since we install astro-sdk-python[all], the providers would be available.

feluelle (approved):

sql-cli/Makefile
```diff
@@ -31,7 +31,7 @@ pre-commit:  # Run pre-commit
 flow:  # Build and run the latest version of flow
 
 test:  # Run tests
-	@_TYPER_FORCE_DISABLE_TERMINAL=1 PYTHONWARNINGS="ignore" poetry run pytest -s --cov --cov-branch --cov-report=term-missing tests
+	@_TYPER_FORCE_DISABLE_TERMINAL=1 PYTHONWARNINGS="ignore" poetry run pytest
```
tatiana, line 34: @pankajkoti why did we remove the coverage stats?

pankajkoti: @feluelle can we add them back or you have a different opinion?

feluelle: It is now in the project.toml

pankajkoti: I guess Felix moved them to pyproject.toml here: https://github.com/astronomer/astro-sdk/pull/1258/files#diff-f92d64397e38a0a17a38a3b24f90b96020f886a2f8db484165658088325e0c9bR88-R93

feluelle: This way we have the options in all pytest runs - local and in CI.

sql-cli/sql_cli/utils/airflow.py
```diff
@@ -10,7 +10,7 @@
 from packaging.version import Version
 
 if TYPE_CHECKING:
-    from airflow.models.dag import DAG
+    from airflow.models.dag import DAG  # pragma: no cover
```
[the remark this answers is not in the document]
tatiana: It surprises me it would be a false positive. Did we confirm that tests are going through this if? Could it be the case TYPE_CHECKING is being false?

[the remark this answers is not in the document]
feluelle: Maybe [this](https://github.com/nedbat/coveragepy/issues/831) helps explaining it.

sql-cli/sql_cli/macros/load_file.py.jinja2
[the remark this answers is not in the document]
feluelle: Completely forgot this. Will add a comment now.

sql-cli/conftest.py
```diff
@@ -147,6 +182,38 @@ def initialised_project_with_test_config(initialised_project: Project):
...
+@pytest.fixture()
+def initialised_project_with_load_file_workflow(
+    root_directory_templates_load_file, initialised_project: Project
+):
+    output_table_name = create_unique_table_name(UNIQUE_HASH_SIZE)
+    render_kwargs = {
+        "context": {"output_table_name": output_table_name},
+        "searchpath": root_directory_templates_load_file,
+    }
+
```
pankajkoti, line 194: Write a comment here on why we need this manipulations.

sql-cli/sql_cli/__main__.py
```diff
@@ -13,10 +13,10 @@
 from sql_cli.astro.command import AstroCommand
 from sql_cli.astro.group import AstroGroup
 from sql_cli.constants import DEFAULT_AIRFLOW_HOME, DEFAULT_DAGS_FOLDER
-from sql_cli.exceptions import ConnectionFailed, DagCycle, EmptyDag, SqlFilesDirectoryNotFound
+from sql_cli.exceptions import ConnectionFailed, DagCycle, EmptyDag, WorkflowFilesDirectoryNotFound
 
 if TYPE_CHECKING:
-    from sql_cli.project import Project
+    from sql_cli.project import Project  # pragma: no cover
```
pankajkoti, line 19: Add a ticket to check on whether we can cover this line and remove the no cover.

sql-cli/sql_cli/macros/load_file.py.jinja2
```diff
@@ -0,0 +1,64 @@
+{% macro load_file(yaml_file, dag) -%}
+{% set yaml_content = yaml_file.get_yaml_content() -%}
+{{ yaml_file.get_variable_name() }} = aql.load_file(
+    {#- Note: Unfortunately, we cannot simply use **kwargs as some of the deserialized fields
```
pankajkoti, line 4: Create a ticket to evaluate if we can support **kwargs with a combination of python helper methods to tranform the yaml content as is needed by Jinja templates.

sql-cli/sql_cli/__main__.py
pankajkoti: Created ticket: https://github.com/astronomer/astro-sdk/issues/1283

sql-cli/include/base/config/default/configuration.yml
```diff
@@ -7,19 +7,15 @@ connections:
...
-#    description: null
-#    extra:
-#      project: "my-gcp-project"
-#    host: null
-#    login: null
-#    password: null
-#    port: null
-#    schema: null
-
+  - conn_id: bigquery
```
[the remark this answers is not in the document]
pankajkoti: Created ticket: #1282

sql-cli/sql_cli/macros/load_file.py.jinja2
pankajkoti: Created ticket: #1284

tatiana (approved):

PR #1259, opened 2022-11-16
Pass task_id to be used for parent class on LoadFileOperator initialisation

pankajkoti:
The SQL CLI render task for the load_file interface when calling LoadFileOperator needs task_id to be available
as part of the initialisation, hence generate a task_id if it is not provided already and pass it to the parent class for initialisation.
No

Issue screenshot:
Screenshot 2022-11-16 at 10 12 26 PM

### DIFF
--- a/python-sdk/src/astro/sql/operators/load_file.py
+++ b/python-sdk/src/astro/sql/operators/load_file.py
@@ -31,6 +31,7 @@ class LoadFileOperator(AstroSQLBaseOperator):
     :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
             in the resulting dataframe
     :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
+    :param task_id: ID of the task instance of the operator
 
     :return: If ``output_table`` is passed this operator returns a Table object. If not
         passed, returns a dataframe.
@@ -49,14 +50,17 @@ class LoadFileOperator(AstroSQLBaseOperator):
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        task_id: str | None = None,
         **kwargs,
     ) -> None:
+        task_id = task_id or get_unique_task_id("load_file")
         super().__init__(
+            task_id=task_id,
             **kwargs_with_datasets(
                 kwargs=kwargs,
                 input_datasets=input_file,
                 output_datasets=output_table,
-            )
+            ),
         )
         self.output_table = output_table
         self.input_file = input_file
--- a/python-sdk/src/astro/sql/operators/load_file.py
+++ b/python-sdk/src/astro/sql/operators/load_file.py
@@ -31,7 +31,6 @@ class LoadFileOperator(AstroSQLBaseOperator):
     :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
             in the resulting dataframe
     :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
-    :param task_id: ID of the task instance of the operator
 
     :return: If ``output_table`` is passed this operator returns a Table object. If not
         passed, returns a dataframe.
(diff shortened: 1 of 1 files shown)


python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -49,14 +50,17 @@ def __init__(
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        task_id: str | None = None,
         **kwargs,
     ) -> None:
+        task_id = task_id or get_unique_task_id("load_file")
         super().__init__(
+            task_id=task_id,
```
feluelle, line 58: @pankajkoti maybe we also need the task_id actually in `kwargs_with_datasets`? Did you test this?

feluelle: Or maybe only there? And it will be passed to the `super().__init__()` automatically.

pankajkoti: yes, tested @feluelle . We don't need it there :)

feluelle:
Also, isn't it possible to add task_id as part of `**kwargs` to the constructor?

I think the change you propose has the benefit of not having to set it i.e. by setting a default. But it should also work by explicitly setting it, no?

pankajkoti:
> Also, isn't it possible to add task_id as part of `**kwargs` to the constructor?
> 
> I think the change you propose has the benefit of not having to set it i.e. by setting a default. But it should also work by explicitly setting it, no?

yes, like we tested, we can pass it from SQL CLI too, and it is good to have this in the operator too that this PR implements.

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -49,14 +50,16 @@ def __init__(
...
         **kwargs,
     ) -> None:
+        kwargs["task_id"] = task_id or get_unique_task_id("load_file")
         super().__init__(
             **kwargs_with_datasets(
                 kwargs=kwargs,
                 input_datasets=input_file,
                 output_datasets=output_table,
-            )
+            ),
```
pankajkoti, line 62:
```suggestion
            )
```

feluelle (approved):

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -49,8 +50,10 @@ def __init__(
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        task_id: str | None = None,
         **kwargs,
     ) -> None:
+        kwargs["task_id"] = task_id or get_unique_task_id("load_file")
```
pankajkoti, line 56:
```suggestion
        kwargs.setdefault("task_id", get_unique_task_id("load_file"))
```

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -49,8 +50,10 @@ def __init__(
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        task_id: str | None = None,
```
pankajkoti, line 53:
```suggestion
```

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -31,6 +31,7 @@ class LoadFileOperator(AstroSQLBaseOperator):
     :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
             in the resulting dataframe
     :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
+    :param task_id: ID of the task instance of the operator
```
pankajkoti, line 34:
```suggestion
```

kaxil (approved):

PR #1265, opened 2022-11-17
Avoid append when `if_exists` is set to replace

### LINKED ISSUE #1260: dataframe operator is now appending to it's output table instead of replacing.
**Describe the bug**
 the dataframe operator is now appending to it's output table instead of replacing.
refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1668643987896139

**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Write the DAG  '...'
2. Create connection '....'
3. Run using  '....'

**Expected behavior**

**Screenshots**

**Additional context**

pankajastro:
# Description

closes #1260
The `load_pandas_dataframe_to_table` does not respect `if_exists` param

## What is the new behavior?
`if_exists` is set to `replace` replace the table rather than append

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -647,8 +647,10 @@ class SnowflakeDatabase(BaseDatabase):
         """
         self._assert_not_empty_df(source_dataframe)
 
-        auto_create_table = True
-        if if_exists == "replace":
+        auto_create_table = False
+        if not self.table_exists(target_table):
+            auto_create_table = True
+        elif if_exists == "replace":
             self.create_table(target_table, dataframe=source_dataframe)
 
         # We are changing the case of table name to ease out on the requirements to add quotes in raw queries.
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -647,9 +647,9 @@ class SnowflakeDatabase(BaseDatabase):
         """
         self._assert_not_empty_df(source_dataframe)
 
-        auto_create_table = False
-        if if_exists == "replace" or not self.table_exists(target_table):
-            auto_create_table = True
+        auto_create_table = True
+        if if_exists == "replace":
+            self.create_table(target_table, dataframe=source_dataframe)
 
         # We are changing the case of table name to ease out on the requirements to add quotes in raw queries.
         # ToDO - Currently, we cannot to append using load_file to a table name which is having name in lower case.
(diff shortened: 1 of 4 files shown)


python-sdk/src/astro/databases/snowflake.py
```diff
@@ -648,8 +648,10 @@ def load_pandas_dataframe_to_table(
         self._assert_not_empty_df(source_dataframe)
 
         auto_create_table = False
-        if if_exists == "replace" or not self.table_exists(target_table):
+        if not self.table_exists(target_table):
```
sunank200, line 651: If the table doesn't exist, don't we require to create a table?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -648,8 +648,10 @@ def load_pandas_dataframe_to_table(
...
             auto_create_table = True
+        elif if_exists == "replace":
```
sunank200, line 653: What would happen when the table doesn't exist and the user has passed in `replace`?

sunank200 (commented):
Added few comments

python-sdk/src/astro/databases/snowflake.py
pankajastro: `auto_create_table` would be true and pandas will take care of creating table

python-sdk/src/astro/databases/snowflake.py
pankajastro: auto_create_table would be true and pandas will take care of creating table

sunank200 (approved):
LGTM

python-sdk/tests/sql/operators/test_dataframe.py
```diff
@@ -465,3 +465,33 @@ def get_empty_dataframe():
...
+        name="ci_df_replace_table",
+        conn_id=conn_id,
+    )
+    with sample_dag:
+        get_empty_dataframe(output_table=output_tb)
+    test_utils.run_dag(sample_dag)
+    assert output_tb.row_count == 2
+    # re-run dag to and make sure it is replacing table
+    test_utils.run_dag(sample_dag)
+    assert output_tb.row_count == 2
```
kaxil, line 497: This is a brittle test, it checks that there were 2 rows when run multiple times, but it doesn't check that the rows are "replaced"

pankajastro: hmm, add more tests https://github.com/astronomer/astro-sdk/pull/1293

PR #1266, opened 2022-11-17
Fix error help for astro-cli

feluelle:
# Description

Currently, when using an unknown flag in astro-cli, it raises (under the condition that the flag is implemented in cobra but not in sql-cli):
> Try 'flow -h' for help.

In case the flag is not implemented in astro-cli cobra, we currently see cobra error output
> Error: unknown flag: --foo

^this needs to be fixed on astro-cli cobra side.

## What is the new behavior?

Monkey-patch rich_format_error to make it environment aware. So that when running in astro-cli it says:
> Try 'astro flow -h' for help.

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/utils/rich.py
+++ b/sql-cli/sql_cli/utils/rich.py
@@ -1,0 +1,41 @@
+from typing import Union
+
+import click
+from rich.panel import Panel
+from typer.rich_utils import (
+    ALIGN_ERRORS_PANEL,
+    ERRORS_PANEL_TITLE,
+    STYLE_ERRORS_PANEL_BORDER,
+    STYLE_ERRORS_SUGGESTION,
+    _get_rich_console,
+    highlighter,
+)
+
+from sql_cli.astro.utils import resolve_command_path
+
+
+def rich_format_error(self: click.ClickException) -> None:
+    """Print richly formatted click errors.
+
+    Called by custom exception handler to print richly formatted click errors.
+    Mimics original click.ClickException.echo() function but with rich formatting.
+    """
+    console = _get_rich_console(stderr=True)
+    ctx: Union[click.Context, None] = getattr(self, "ctx", None)
+    if ctx is not None:
+        console.print(ctx.get_usage())
+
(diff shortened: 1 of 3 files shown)


sql-cli/sql_cli/utils/rich.py
```diff
@@ -0,0 +1,41 @@
...
+
+def rich_format_error(self: click.ClickException) -> None:
+    """Print richly formatted click errors.
+
+    Called by custom exception handler to print richly formatted click errors.
+    Mimics original click.ClickException.echo() function but with rich formatting.
+    """
+    console = _get_rich_console(stderr=True)
+    ctx: Union[click.Context, None] = getattr(self, "ctx", None)
+    if ctx is not None:
```
tatiana, line 25: in which cases is the context empty? It may be worth adding tests if this is a valid scenario.

feluelle: This is actually a copy-paste of the original code in rich. 😅

sql-cli/sql_cli/utils/rich.py
```diff
@@ -0,0 +1,41 @@
...
+        console.print(ctx.get_usage())
+
+    if ctx is not None and ctx.command.get_help_option(ctx) is not None:
+        console.print(
+            f"Try [blue]'{resolve_command_path(ctx.command_path)} {ctx.help_option_names[0]}'[/] for help.",
```
feluelle, line 30: @tatiana the only line I changed is here.

```diff
+f"Try [blue]'{resolve_command_path(ctx.command_path)} {ctx.help_option_names[0]}'[/] for help."
-f"Try [blue]'{ctx.command_path} {ctx.help_option_names[0]}'[/] for help."
```
^this is what changed.

feluelle: I am open for alternatives to fix this. Unfortunately `ctx.command_path` cannot be set.

tatiana (approved):
@feluelle should we rebase and have this as part of the 0.2 release?

tatiana (approved):

pankajkoti (approved):

PR #1271, opened 2022-11-18
Enable benchmark GKE cluster teardown

pankajastro:
# Description
currently, the benchmark GKE cluster is up all time

## What is the new behavior?
Delete it since we are running the job just once a week

No

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/Makefile
+++ b/python-sdk/tests/benchmark/Makefile
@@ -1,4 +1,4 @@
-all: clean setup_gke container run_job wait_for_completion
+all: clean setup_gke container run_job wait_for_completion teardown_gke
 .PHONY: all
 
 GCP_PROJECT ?= astronomer-dag-authoring


tatiana (commented):
@pankajastro would it be worth changing `teardown_gke` to check if there are any other jobs running, before tearing down the cluster? My concern is that a Github action instance may affect others if we don't check this in advance.

pankajastro:
> @pankajastro would it be worth changing `teardown_gke` to check if there are any other jobs running, before tearing down the cluster? My concern is that a Github action instance may affect others if we don't check this in advance.

I'm more worried about the cost side. We are using a bit costly machine. Currently,  the job runs once a week. Our current infra will have issues running multiple jobs (resource limitation).  But I liked the idea! and would be happy to include it here maybe if some job is running then not run another job until it is complete. How we can check the currently running job in this case?

kaxil:
> > @pankajastro would it be worth changing `teardown_gke` to check if there are any other jobs running, before tearing down the cluster? My concern is that a Github action instance may affect others if we don't check this in advance.
> 
> I'm more worried about the cost side. We are using a bit costly machine. Currently, the job runs once a week. Our current infra will have issues running multiple jobs (resource limitation). But I liked the idea! and would be happy to include it here maybe if some job is running then not run another job until it is complete. How we can check the currently running job in this case?

`kubectl get jobs` ?

tatiana:
@pankajastro since the job is triggered and processed asynchronously, we need to at least confirm it was completed with failure/success before tearing down the cluster.

Something like this may work:
```
kubectl wait --for=condition=complete --timeout=30s benchmark/benchmark
```

pankajastro:
> @pankajastro since the job is triggered and processed asynchronously, we need to at least confirm it was completed with failure/success before tearing down the cluster.
> 
> Something like this may work:
> 
> ```
> kubectl wait --for=condition=complete --timeout=30s benchmark/benchmark
> ```

currently, we monitor the job until it is finished (https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/benchmark/Makefile#L65). The problem is when will run multiple benchmarking jobs together so I was thinking if I could have a way to check if there is already benchmarking job running then do not submit the new request?

tatiana (approved):
> currently, we monitor the job until it is finished (https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/benchmark/Makefile#L65). The problem is when will run multiple benchmarking jobs together so I was thinking if I could have a way to check if there is already benchmarking job running and then not submit the new request?

I hadn't realised that @pankajastro , sorry! I am happy for us to merge this PR as it stands and we can improve it over time, if we see we are having issues running concurrent instances of the benchmark

pankajastro:
I'm merging this, if any further feedback will take that separately

Issue #905, opened 2022-09-23
Add support for Azure Blob Storage


kaxil:

luanmorenomaciel:
Any timeline by when this support is going to be added into the astro sdk>

kaxil:
Hey, @luanmorenomaciel , in few months but we can prioritize if you have interest and planning to use SDK

tatiana:
I just agreed with @luanmorenomaciel that the focus of this ticket would be to implement the Azure Blob Storage FileLocation & the non-native transfer.

We'll have separate tickets for addressing the native transfers.

luanmorenomaciel:
starting to work on this item, narrowing the scope to send an small PR where we can list files from WASB and ABFSS protocol using the get_file_list operator

https://astro-sdk-python.readthedocs.io/en/stable/astro/files/operators/get_file_list.html

phanikumv:
Hey @luanmorenomaciel - feel free to let us know if you need any help with the implementation

luanmorenomaciel:
First PR submitted 
https://github.com/astronomer/astro-sdk/pull/1275

PR #1275, opened 2022-11-18
Initial Support for Azure Blob Storage Object

luanmorenomaciel:
# Description
Currently, the Astro Python SDK doesn't offer an interface to communicate with the Azure Blob Storage system through the WASB protocol.

related: https://github.com/astronomer/astro-sdk/issues/905

## What is the new behavior?
Add support for the WASB protocol 

- Interact with the file system

As our next goal we're going to implement the non-native load file for azure.

No

### Checklist
- [x] Establish connectivity with WASB Hook
- [x] Build list function with prefix intelligence
- [x] Collect size metadata information
- [x] Apply tests for coverage


rajaths010494:
Test case for Size and Openlineage is missing .

python-sdk/pyproject.toml
```diff
@@ -82,7 +87,9 @@ all = [
     "sqlalchemy-bigquery>=1.3.0",
     "s3fs",
     "protobuf<=3.20", # Google bigquery client require protobuf <= 3.20.0. We can remove the limitation when this limitation is removed
-    "openlineage-airflow>=0.17.0"
+    "openlineage-airflow>=0.17.0",
+    "apache-airflow-providers-microsoft-azure",
+    "azure-storage-blob"
```
phanikumv, line 92: I think we need to specify the minimum version for these two libraries

python-sdk/src/astro/constants.py
```diff
@@ -21,6 +21,8 @@ class FileLocation(Enum):
     HTTPS = "https"
     GS = "gs"  # Google Cloud Storage
     S3 = "s3"  # Amazon S3
+    WASB = "wasb"  # Azure Blob Storage
+    WASBS = "wasbs"  # Azure Blob Storage
```
phanikumv, line 25: dont we need abfss protocol to be added here ?

https://stackoverflow.com/questions/60277545/what-is-the-difference-between-abfss-and-wasbs-in-azure-storage

python-sdk/conftest.py
```diff
@@ -294,6 +296,16 @@ def _upload_or_delete_remote_file(file_create, object_prefix, provider, source_p
             hook.load_file(source_path, object_prefix, bucket_name)
         else:
             hook.delete_objects(bucket_name, object_prefix)
+    elif provider == "azure":
+        from airflow.providers.microsoft.azure.hooks.wasb import WasbHook
+
+        bucket_name = os.getenv("AZURE_BUCKET", "astro-sdk")
+        object_path = f"wasb://{bucket_name}/{object_prefix}"
```
phanikumv, line 303: we need to provide `abfss` url as well because wasb is going to be deprecated.

https://learn.microsoft.com/en-us/azure/databricks/external-data/wasb-blob

python-sdk/conftest.py
```diff
@@ -248,25 +249,26 @@ def remote_files_fixture(request):
...
-    elif provider == "amazon":
-        for object_prefix in object_prefix_list:
+        elif provider == "amazon":
             hook.delete_objects(bucket_name, object_prefix)
+        elif provider == "azure":
+            hook.delete_file(bucket_name, object_prefix)
 
 
-def _upload_or_delete_remote_file(file_create, object_prefix, provider, source_path):
+def _upload_or_delete_remote_file(file_create, object_prefix, provider, source_path, conn_id):  # noqa: C901
```
sunank200, line 271: Are you overloading this method? How does it handle the other file location using this function with `conn_id` added?

python-sdk/tests/files/locations/test_wasb.py
```diff
@@ -0,0 +1,43 @@
+from pathlib import Path
```
sunank200, line 1: We should add more tests and integration test for various scenarios for loading a file and folder from Azure blob storage

sunank200 (commented):
Added few comments

python-sdk/pyproject.toml
mateusoliveiraowshq: @phanikumv do you have any recommendations for the minimum version?

We noticed that other providers, such as:
- apache-airflow-providers-amazon
- apache-airflow-providers-postgres
- apache-airflow-providers-snowflake
- apache-airflow-providers-databricks
are not pinned at the moment.
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/pyproject.toml#L77-L89

And we are not sure what to do.

python-sdk/pyproject.toml
```diff
@@ -69,6 +69,11 @@ amazon = [
     "s3fs",
     "smart-open[s3]>=5.2.1",
 ]
+azure = [
+    "apache-airflow-providers-microsoft-azure",
+    "azure-storage-blob",
+    "smart-open[azure]>=5.2.1",
```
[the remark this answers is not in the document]
mateusoliveiraowshq: @sunank200, we noticed that in Amazon and Google, we are pinning to this version or higher, and we were just consistent with that version.

https://github.com/astronomer/astro-sdk/blob/main/python-sdk/pyproject.toml#L57
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/pyproject.toml#L70

python-sdk/conftest.py
mateusoliveiraowshq: @sunank200 at the moment, the fixture uses the default connections for AWS and GCP

- https://github.com/astronomer/astro-sdk/blob/main/python-sdk/conftest.py#L280
- https://github.com/astronomer/astro-sdk/blob/main/python-sdk/conftest.py#L293

Would you rather we use the azure default connection?

And it may be good to add support to this fixture pass the conn_id to any provider as well

python-sdk/tests/files/locations/test_wasb.py
mateusoliveiraowshq: @sunank200 the OpenLineage function since is repeated in other locations, would it make sense to have it in the location base class?
We will add the size method and test it.

python-sdk/conftest.py
mateusoliveiraowshq: we going to change to use the default connection for azure, and we will remove this logic.

`export AIRFLOW_CONN_WASB_DEFAULT='wasb://blob%20username:blob%20password@myblob.com?tenant_id=tenant+id'
`
https://airflow.apache.org/docs/apache-airflow-providers-microsoft-azure/stable/connections/wasb.html

python-sdk/src/astro/constants.py
luanmorenomaciel: @phanikumv the **WASB** protocol has been marked as **legacy** and may be retired in the upcoming years however, the current [Microsoft Azure Provider](https://airflow.apache.org/docs/apache-airflow-providers-microsoft-azure/stable/index.html) does not support the ADLS2 (ABFSS) protocol.

@mateusoliveiraowshq, @tatiana and I submitted the following [issue](https://github.com/apache/airflow/issues/28223) on the Apache Airflow codebase that implements a new Hook under the Microsoft Azure Provider.

Once we sort that out, the ABFS support is going to be added/implemented on the Astro Python SDK as new issue to be developed.

phanikumv: thats good, I think there is a PR already out there for ABFS support https://github.com/apache/airflow/pull/28262

python-sdk/tests/files/locations/test_wasb.py
sunank200: The name and namespace for openlineage can be in different formats. That's the reason that its overridden

python-sdk/conftest.py
sunank200: Yes whenever we have azure default connection, we can use that.

python-sdk/pyproject.toml
mateusoliveiraowshq: @phanikumv we make the test with `apache-airflow-providers-microsoft-azure>=4.0.0`, but we have the following issue during the test.

- https://github.com/astronomer/astro-sdk/actions/runs/3716469536/jobs/6302839191#logs

```
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
nox > Session test-3.8(airflow='2.2.5') failed.
Error: Process completed with exit code 1.
```
We remove the version tag to keep testing okay. Let me know if we need to make more tests in this case.

python-sdk/conftest.py
tatiana: https://github.com/astronomer/astro-sdk/issues/1446

tatiana:
Follow up tickets:
* https://github.com/astronomer/astro-sdk/issues/1446
* https://github.com/astronomer/astro-sdk/issues/1447
* https://github.com/astronomer/astro-sdk/issues/1448

tatiana:
The newly introduced Github secret `AZURE_WASB_CONN_STRING` is misbehaving in this PR (created from a fork). This behaviour could also be observed when creating a new PR from the original repo using the same commits:
https://github.com/astronomer/astro-sdk/pull/1455

However, when creating a new commit with the same changes, by a repo maintainer, from the original original, things seem to be working:
Screenshot from 2022-12-17 08-38-49
https://github.com/astronomer/astro-sdk/actions/runs/3719231698/jobs/6307958744
https://github.com/astronomer/astro-sdk/pull/1454

tatiana (approved):
Great work @luanmorenomaciel and @mateusoliveiraowshq ! Thank you very much for working so hard to contribute to the Astro SDK!

I am convinced there is an issue/misconfiguration on Github actions to capture recently introduced secrets - and we have a test failing.

I created a new PR with the same changes, and everything is passing:
https://github.com/astronomer/astro-sdk/pull/1454/files

We've logged follow up tickets for this work, including one to improve further the implementation - so we can have Azure fully operational, but I think this PR is in a good shape to be merged.

PR #1276, opened 2022-11-21
Fix database setting in PostgresDatabase

pankajastro:
# Description

Currently, the CI is failing for postgres related tests, because the postgres database is set to the table schema.

**More details:**

In https://github.com/astronomer/astro-sdk/commit/313f6da5fc572a8a35c6fa3db90373a7586be4de#diff-9fcbc9dc6fc68bf0a0d62a485508779f5b1f6d749a197e0f1ab1758983d1eb1e we added `kwargs.update({"database"` but PostgresHook did not support `database` kwarg at that time. This is why it basically did not do anything. Now after https://github.com/apache/airflow/pull/26744 got released in the latest provider, it is being picked up and it shows that we wrongly set `{"database": self.table.metadata.schema}`

## What is the new behavior?

We correctly set the database to the table's database.

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/settings.py
+++ b/python-sdk/src/astro/settings.py
@@ -5,8 +5,6 @@ from airflow.configuration import conf
 from astro.constants import DEFAULT_SCHEMA
 
 SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=DEFAULT_SCHEMA)
-# Airflow hook does have option to set the postgres search path or schema's.
-# Let's set postgres_default_schema fallback to public
 POSTGRES_SCHEMA = conf.get("astro_sdk", "postgres_default_schema", fallback=SCHEMA)
 BIGQUERY_SCHEMA = conf.get("astro_sdk", "bigquery_default_schema", fallback=SCHEMA)
 SNOWFLAKE_SCHEMA = conf.get("astro_sdk", "snowflake_default_schema", fallback=SCHEMA)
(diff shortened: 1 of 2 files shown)


python-sdk/src/astro/settings.py
```diff
@@ -5,7 +5,9 @@
 from astro.constants import DEFAULT_SCHEMA
 
 SCHEMA = conf.get("astro_sdk", "sql_schema", fallback=DEFAULT_SCHEMA)
-POSTGRES_SCHEMA = conf.get("astro_sdk", "postgres_default_schema", fallback=SCHEMA)
+# Airflow hook does have option to set the postgres search path or schema's.
```
feluelle, line 8: That is not true. We can run an explicit query to set the search path. See [here](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH).

pankajastro: in airflow, Postgres hook can we set schema?

pankajastro: right, my intention for this comment was to explain that the Postgres constructor has a way to set schema since most of our operators depend on hook initialization. But look like the root cause for CI failure may be something else that needs further investigation.

feluelle (approved):

phanikumv (approved):

sunank200:
Created an issue to check the backward compatibility for PostGresHook: https://github.com/astronomer/astro-sdk/issues/1291

Issue #1272, opened 2022-11-18
Change DAG generation so it does not depend on the sql_cli as a library


tatiana:
**Context**
We do not want to expose the SQL CLI as a library, only as a CLI tool. This means that the DAGs we build should rely on Airflow and the Astro Python SDK - but not have any Python imports.

As of 0.1, the DAG generated respected this. During the development of 0.2, some tasks ended up inadvertedly changing this behaviour.

**Acceptance criteria**
* [ ] The DAGs we generate should not depend on the `sql_cli`
* [ ] Once the renderer DAG generation no longer depends on the `sql_cli` as a library, we can explore for it to be the default way of generating DAGs

tatiana:
We agreed the scope of this will be rendering the DAG, since the expanded tasks already do not support on the `sql_cli`, as part of https://github.com/astronomer/astro-sdk/pull/1258

PR #1278, opened 2022-11-21
Move render function to python sdk

dimberman:
# Description
Currently running render requires that a user installs the SQL CLI on their airflow. This goes against the core SQL CLI philosophy that it be a lightweight DAG generator. This PR ensures that a user will only need to have the python-sdk installed on their airflow machine to run DAGs generated by the SQL CLI.

closes: https://github.com/astronomer/astro-sdk/issues/1272

## What is the new behavior?
Now render is a part of the astro python SDK, which can even be used by users who wish to render SQL files without creating individual transform operators (though we can wait until this is more fleshed out before explicitly exposing to astro python SDK users).

This creates a dependency on astro-python-sdk 1.3.0 for SQL CLI 0.3.

### Checklist


sql-cli/pyproject.toml
```diff
@@ -39,7 +39,7 @@ Jinja2 = ">=2.10.1, <=3.1"
 python-frontmatter = "^1.0.0"
 PyYAML = ">=5.4.1"
 python-dotenv = "^0.21.0"
-astro-sdk-python = {extras = ["all"], version = "^1.2.0"}
+astro-sdk-python = {extras = ["all"], version = "^1.3.0.dev1"}
```
pankajkoti, line 42: Does this work? 1.3.0.dev1 isn't a release right?

python-sdk/src/astro/utils/jinja.py
```diff
@@ -0,0 +1,26 @@
...
+def find_template_variables(file_path: Path) -> set[str]:
+    """
+    Find template variables in given file path which needed to be declared when rendering.
+
+    :param file_path: The file path to check for variables.
+
+    :returns: all undeclared variables.
+    """
+    env = Environment(
+        loader=FileSystemLoader(file_path.parent),
```
pankajkoti, line 20: Does this work well when the templates are not kept in this parent directory but are in the sql-cli directory? We were facing some challenges today where in it was giving TemplateNotFound Error.

cc: @feluelle

python-sdk/src/astro/sql/operators/render.py
```diff
@@ -3,7 +3,7 @@
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from sql_cli.sql_directory_parser import SqlFile, get_sql_files
+from astro.utils.sql_directory_parser import SqlFile, get_sql_files
```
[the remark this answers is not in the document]
dimberman: @pankajkoti this PR is not yet merged so we will have to change this when that PR is merged (or if this is merged first you can adjust in your PR)

sql-cli/pyproject.toml
dimberman: @pankajkoti I've modified how this all works so your local dev uses the local python-sdk. We won't be able to release 0.3 until python SDK 1.3.0 comes out.

python-sdk/src/astro/utils/jinja.py
dimberman: @feluelle can you explain this further?

sql-cli/sql_cli/dag_generator.py
```diff
@@ -6,8 +6,8 @@
 
 from networkx import DiGraph, depth_first_search, find_cycle, is_directed_acyclic_graph
 
+from astro.utils.sql_directory_parser import SqlFile, get_sql_files
```
tatiana, line 9: @dimberman why are we placing it inside `utils`? I have the feeling it would be better to have it alongside the other tasks operators, WDYT?

dimberman: @tatiana but it's not an operator? Maybe have a "render" directory?

python-sdk/src/astro/utils/jinja.py
feluelle: @pankajkoti this is `find_template_variables` - not the `render` function. It does not run `get_template`. This is completely fine like that.

sql-cli/pyproject.toml
feluelle: I am also wondering how this would work 🤔 

Did you test this in a new python venv? I understand that this works within the project, but don't quite understand how this would work outside of the project.

sql-cli/sql_cli/dag_generator.py
feluelle: Yes, `render` works for me. I think it is not fair to call this `utils`. It deserves a better name.

tatiana:
@dimberman it's important to update the Python SDK Sphinx docs with this new operator as well!

dimberman:
@tatiana I wasn't sure if we wanted to explicitly point this feature out in docs yet or not. Do we want python SDK users using this yet?

python-sdk/src/astro/utils/workflow_directory_parser.py
```diff
@@ -10,11 +10,53 @@
 from airflow.models import BaseOperator
 from yaml.scanner import ScannerError
 
+from astro.constants import GENERATED_WORKFLOW_INCLUDE_DIRECTORY
+from astro.custom_backend.serializer import deserialize
 from astro.sql import LoadFileOperator
 from astro.sql.operators.transform import TransformOperator
-from sql_cli.constants import GENERATED_WORKFLOW_INCLUDE_DIRECTORY
-from sql_cli.operators.load_file import get_load_file_instance
```
pankajkoti, line 16: It made sense to me to keep operator related methods separate (we may not want to call them operators, I agree), because with more operators coming in , we might populate this module with more operator specific builder methods.

dimberman: @pankajkoti ultimately we need to keep all of this stuff in the python SDK if we want render_file to work without the SQL CLI. If you would like me to put it somewhere else please suggest an alternative location.

pankajkoti: I think we can keep it like in this PR and move it to alternative location if needed in upcoming implementations.

python-sdk/src/astro/sql/operators/render.py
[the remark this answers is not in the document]
pankajkoti: yes, PR #1258 and I see you're changing it here.

kaxil:
Should we get this in Astro SDK 1.4?

sql-cli/noxfile.py
```diff
@@ -42,6 +42,12 @@ def test(session: nox.Session, airflow: str) -> None:
             "https://raw.githubusercontent.com/apache/airflow/"
             f"constraints-{airflow}/constraints-{session.python}.txt"
         )
+
+        # Poetry does not support constraints:
+        # https://github.com/python-poetry/poetry/issues/3225
+        session.install("-e", "../python-sdk")
+        session.install("-e", ".", f"apache-airflow=={airflow}", "-c", constraints_url)
```
feluelle, line 49:
```suggestion
        session.run("poetry", "run", "pip", "install", "-e", "../python-sdk")
```

sql-cli/tests/test___main__.py
```diff
@@ -1,3 +1,4 @@
+import os
```
feluelle, line 1:
```suggestion
```

sql-cli/tests/test___main__.py
```diff
@@ -15,6 +16,7 @@
 runner = CliRunner()
 
 CWD = pathlib.Path(__file__).parent
+os.environ["TERMINAL_WIDTH"] = "3000"
```
feluelle, line 19:
```suggestion
```

python-sdk/src/astro/constants.py
```diff
@@ -67,3 +67,5 @@ def __str__(self) -> str:
 MergeConflictStrategy = Literal["ignore", "update", "exception"]
 
 ColumnCapitalization = Literal["upper", "lower", "original"]
+
+GENERATED_WORKFLOW_INCLUDE_DIRECTORY = "include"
```
tatiana, line 71: Why do we need this constant?

sql-cli/conftest.py
```diff
@@ -8,10 +8,10 @@
 from airflow.utils import timezone
 from airflow.utils.session import create_session
 
+from astro.render.workflow_directory_parser import SqlFile, WorkflowFile
```
tatiana, line 11: Is the `SqlFile` also used by the implementation in which we create expanded DAGs?

feluelle (changes requested):
@dimberman I am not sure about moving the whole `workflow_directory_parser.py`.

Do we need all of this to be in the python-sdk? For example the `WorkflowFIle` abstraction.

I am a bit afraid that this would slow our development of the sql-cli as every change needs to be released as part of python-sdk.

python-sdk/src/astro/render/workflow_directory_parser.py
```diff
@@ -10,11 +10,51 @@
 from airflow.models import BaseOperator
 from yaml.scanner import ScannerError
 
+from astro.constants import GENERATED_WORKFLOW_INCLUDE_DIRECTORY
```
kaxil, line 13: uuffff yeah, I agree, I thought we would just have the ability to convert SQL files to task but this files involves a lot more than just that for example loading using YAML file -- that should just be in SQL CLI

PR #1286, opened 2022-11-22
Silence astronomer logger

feluelle:
# Description

Currently, in when running sql-cli in the astro-runtime environment, we see the following logs:
```
[2022-11-22 12:12:08,174] {plugin.py:79} INFO - Creating DB tables for astronomer.airflow.version_check.plugin
[2022-11-22 12:12:08,182] {plugin.py:88} INFO - Create
```

## What is the new behavior?

We silence these logs as we want, by default, to only expose our logs.

No.

### Checklist


feluelle:
Closing in favor of https://github.com/astronomer/astro-cli/pull/895

PR #1293, opened 2022-11-22
Add if_exists append and replace tests for `load_pandas_dataframe_to_table`

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1295
The test case of databases method `load_pandas_dataframe_to_table` with load strategy "append" and "replace"

## What is the new behavior?
Add the tests for databases `load_pandas_dataframe_to_table`  for  load strategy "append" and "replace". Tests are adding different dataframe to make sure it works as expected. Removed the old snowflake test 

No

### Checklist


python-sdk/tests/databases/test_bigquery.py
```diff
@@ -155,6 +155,95 @@ def test_load_pandas_dataframe_to_table(database_table_fixture):
...
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "database_table_fixture",
+    [
+        {"database": Database.BIGQUERY},
+    ],
+    indirect=True,
+    ids=["bigquery"],
+)
+def test_load_pandas_dataframe_to_table_with_replace(database_table_fixture):
```
sunank200, line 213: Can we not parameterise the test for all databases instead of writing the same test multiple times across databases?

python-sdk/tests/databases/test_bigquery.py
```diff
@@ -155,6 +155,95 @@ def test_load_pandas_dataframe_to_table(database_table_fixture):
...
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "database_table_fixture",
+    [
+        {"database": Database.BIGQUERY},
+    ],
+    indirect=True,
+    ids=["bigquery"],
+)
+def test_load_pandas_dataframe_to_table_with_append(database_table_fixture):
```
sunank200, line 167: Can we not parameterise the test for all databases instead of writing the same test multiple times across databases?

sunank200 (commented):
Can we not parameterise the test for all databases instead of writing the same test multiple times across databases? WDYT?

Tests flow looks good to me.

python-sdk/tests/databases/test_bigquery.py
pankajastro: yeah, we can do but a couple of other tests in a similar fashion (i.e not parameterise all DB together) I have moved tests to [test_all_databases.py](https://github.com/astronomer/astro-sdk/pull/1293/commits/470f41daa13532fcd705c077dcdb887ac8fe6ada) 470f41daa13532fcd705c077dcdb887ac8fe6ada

sunank200 (approved):
LGTM now

Issue #1296, opened 2022-11-23
Change the http URL in example DAG to s3 URI


sunank200:
**Describe the bug**
Change the HTTP URL for IRIS dataset to s3 URL in example_load_file

PR #1297, opened 2022-11-23
Change the file location to s3

sunank200:
# Description

HTTP URL in example DAG is failing. 

closes: #1296

## What is the new behavior?

- Change the HTTP URL for IRIS dataset to s3 URL in example_load_file

No

### Checklist

### DIFF
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -33,7 +33,7 @@ dag = DAG(
 with dag:
     # [START load_file_example_1]
     my_homes_table = aql.load_file(
-        input_file=File(path="s3://astro-sdk/sample.csv"),
+        input_file=File(path="s3://astro-sdk/python_sdk/example_dags/data/sample.csv"),
         output_table=Table(
             conn_id="postgres_conn",
         ),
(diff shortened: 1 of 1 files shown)


pankajastro (approved):

python-sdk/example_dags/example_load_file.py
```diff
@@ -215,7 +214,7 @@
     # [END load_file_example_17]
 
     # [START load_file_example_18]
-    dataframe = aql.load_file(input_file=File(path=data_url, filetype=FileType.CSV))
+    dataframe = aql.load_file(input_file=File(path="s3://astro-sdk/sample_csv.data", filetype=FileType.CSV))
```
phanikumv, line 217: what if a user wants to use this file ? It is not accessible unless thats a public s3 bucket.

phanikumv: can we store it in `python_sdk/example_dags/data` folder?

sunank200: But the case we are handling as part of this example is that the file is of a different extension than `csv` but the filetype is `csv`.

phanikumv (approved):

python-sdk/example_dags/example_load_file.py
sunank200: Changed the folder location as suggested here: https://github.com/astronomer/astro-sdk/pull/1297/commits/91eab7b161577a51bcc0445c17cdcf9a2d96f6ab

sunank200: I can create a separate issue for the following

> I am more concerned about the fact that we are making the data not available to users who want to try out this DAG for the first time

sunank200: Created a issue for this https://github.com/astronomer/astro-sdk/issues/1299
@phanikumv

python-sdk/example_dags/example_load_file.py
```diff
@@ -34,7 +33,7 @@
 with dag:
     # [START load_file_example_1]
     my_homes_table = aql.load_file(
-        input_file=File(path="s3://astro-sdk/sample.csv"),
+        input_file=File(path="s3://astro-sdk/python_sdk/example_dags/data/sample.csv"),
```
pankajkoti, line 36: Perhaps we can `s3://astro-sdk/python_sdk/example_dags/data` as a prefix constant

sunank200: Created a issue for this https://github.com/astronomer/astro-sdk/issues/1299

Issue #1135, opened 2022-10-21
Cache nox (and poetry) environment in CI for sql-cli project for improving speed


feluelle:
**Please describe the feature you'd like to see**

Our CI for sql-cli project is quite slow as it reinstalls all dependencies every time it runs. We can cache the nox (and poetry) environment. So it does not have to reinstall every time.

**Describe the solution you'd like**

We can have a look at the the following Actions from the community:
* [Python Poetry Action](https://github.com/marketplace/actions/python-poetry-action)
* [Setup nox](https://github.com/marketplace/actions/setup-nox)

**Are there any alternatives to this feature?**

Replace nox by using poetry with [Github matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs).

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1300, opened 2022-11-23
Correct directory path for caching nox installed dependencies in CI

pankajkoti:
# Description

The cache is not referring to the right directory path for caching nox installed environments.

closes: #1135 

## What is the new behavior?

Correct the path for nox dependencies to cache.

No


pankajkoti:
Although the cache is right referring to the nox directory path now, the tests themselves take a lot longer as we can see in the pytest summary. We will need to check if we can optimise the tests run duration.

sql-cli/sql_cli/constants.py
```diff
@@ -6,3 +6,5 @@
 DEFAULT_DAGS_FOLDER = ".airflow/dags"
 
 GENERATED_WORKFLOW_INCLUDE_DIRECTORY = "include"
+
+TEST_CONSTANT = 1
```
pankajkoti, line 10:
```suggestion

```

pankajkoti:
We can see an improvement in the Type-Check job where it is 3x faster now because of caching nox dependencies correctly.

Before: 2min41s
Screenshot 2022-11-23 at 9 30 02 PM

Now: 46s
Screenshot 2022-11-23 at 9 30 21 PM

pankajkoti:
Attaching screenshots for showing cache size before and after this change and that nox does not reinstall the packages upon using the cache.

Cache size before: ~25MB
Screenshot 2022-11-23 at 7 39 45 PM

Cache size after: ~228MB
Screenshot 2022-11-23 at 7 38 55 PM

tatiana (approved):
Great improvement, @pankajkoti !

.github/workflows/ci-sql-cli.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
         with:
            path: |
              ~/.cache/pip
-             .nox
+             sql-cli/.nox
```
pankajkoti, line 47:
```suggestion
             sql-cli/.nox
             .local
```

.github/workflows/ci-sql-cli.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
         with:
           path: |
             ~/.cache/pip
-            .nox
+            sql-cli/.nox
```
pankajkoti, line 90:
```suggestion
            sql-cli/.nox
            .local/
```

kaxil (approved):

Issue #1282, opened 2022-11-22
Use hosted file and sqlite database instead of bigquery for load_file operator example workflow


pankajkoti:
Currently, as part of PR https://github.com/astronomer/astro-sdk/pull/1258, we're introducing dependency
on bigquery connection. We can simplify this by using instead load from a hosted file into SQLite database so that we have minimum external dependencies and a bare minimum initialised directory. 

Reference example here: https://github.com/astronomer/astro-sdk/blob/d5aa768b2d4bca72ef98f8d533fe3f99624b172f/example_dags/calculate_popular_movies.py#L27

PR #1303, opened 2022-11-23
Use sqlite conn for load file workflow example

pankajkoti:

Previously, we were using bigquery conn to load data from GCS to Bigquery

closes: #1282 

## What is the new behavior?

With #1282 we aim to keep it simple by not having to create complex tests for using our example load workflow. The PR replaces the existing load file workflow to load data from a github file to load data into sqlite database.

We additionally are fixing minor issues where the File and Table deserializers are pased void data values for their optional keys so that the deserializer does not raise KeyError.

No

### Checklist


mikeshwe:
For the purposes of demoing SQL CLI, it would be best if the data file were included somewhere in directory structure, and any connection information to whatever blob storage we're using is well documented or configured by default.

IOW, it would be best if the demo works out of the box, or with as few configuration steps as possible.

tatiana (approved):
LGTM, thanks, @pankajkoti !

PR #1306, opened 2022-11-24
OpenLineage: Rename the name and namespace along with URI

### LINKED ISSUE #1281: Rename the name and namespace for sqlite as per new convention introduced
**Describe the bug**
There have been new conventions introduced: [Naming.md#local-file-system](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#local-file-system)
I believe sqlite counts as local file, therefore you could adjust naming convention in astro as well.

A word of explanation: both IP and port are mandatory. It means that if port is unknown (e. g. if file exists in Airflow's local file system only) you should default to some value. In one of OL extractors it is defaulted to `paramiko.config.SSH_PORT`, in most cases it's equal to 22.

Reference: https://github.com/astronomer/astro-sdk/pull/1142#issuecomment-1321563229

rajaths010494:
Describe the bug
- There have been new conventions introduced: [Naming.md#local-file-system](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#local-file-system)
I believe sqlite counts as local file, therefore you could adjust naming convention in astro as well.

A word of explanation: both IP and port are mandatory. It means that if port is unknown (e. g. if file exists in Airflow's local file system only) you should default to some value. In one of OL extractors it is defaulted to paramiko.config.SSH_PORT, in most cases it's equal to 22.

Reference: https://github.com/astronomer/astro-sdk/pull/1142#issuecomment-1321563229
 - Add openlineage_dataset_uri() method to create URI as per each database.

closes #1281


python-sdk/src/astro/databases/sqlite.py
```diff
@@ -144,6 +144,8 @@ def openlineage_dataset_namespace(self) -> str:
         """
         Returns the open lineage dataset namespace as per
         https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
-        Example: sqlite://127.0.0.1
+        Example: file://127.0.0.1:22
         """
-        return f"{self.sql_type}://{socket.gethostbyname(socket.gethostname())}"
+        conn = self.hook.get_connection(self.conn_id)
+        port = conn.port if conn.port else 22
```
JDarDagran, line 150: simpler is `conn.port or 22`

JDarDagran (approved):
LGTM

python-sdk/src/astro/databases/sqlite.py
rajaths010494: changed

rajaths010494:
Attaching a screenshot of the URI for reference on Marquez UI.
image

sunank200 (approved):
LGTM

sunank200:
@rajaths010494 @JDarDagran is the `dataSource` `uri` in the correct format as aspected?

JDarDagran:
Actually why do you need this format:
```
f"{self.input_data.openlineage_dataset_namespace()}"
f"://{self.input_data.openlineage_dataset_name()}"
```
with two slashes before dataset name? The path should be absolute so I'd rather see URI as `file://172.26.0.2:22/tmp/sqlite.db.top_animation`.

sunank200 (changes requested):
> Actually why do you need this format:
> 
> ```
> f"{self.input_data.openlineage_dataset_namespace()}"
> f"://{self.input_data.openlineage_dataset_name()}"
> ```
> 
> with two slashes before dataset name? The path should be absolute so I'd rather see URI as `file://172.26.0.2:22/tmp/sqlite.db.top_animation`.
> 
> [This](https://www.rfc-editor.org/rfc/rfc3986.txt) spec, 3.3. Path explains more.

@rajaths010494 we should do this change across.

rajaths010494:
> > Actually why do you need this format:
> > ```
> > f"{self.input_data.openlineage_dataset_namespace()}"
> > f"://{self.input_data.openlineage_dataset_name()}"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with two slashes before dataset name? The path should be absolute so I'd rather see URI as `file://172.26.0.2:22/tmp/sqlite.db.top_animation`.
> > [This](https://www.rfc-editor.org/rfc/rfc3986.txt) spec, 3.3. Path explains more.
> 
> @rajaths010494 we should do this change across.

Yeah saw i will make the necessary changes

rajaths010494:
> > > Actually why do you need this format:
> > > ```
> > > f"{self.input_data.openlineage_dataset_namespace()}"
> > > f"://{self.input_data.openlineage_dataset_name()}"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > with two slashes before dataset name? The path should be absolute so I'd rather see URI as `file://172.26.0.2:22/tmp/sqlite.db.top_animation`.
> > > [This](https://www.rfc-editor.org/rfc/rfc3986.txt) spec, 3.3. Path explains more.
> > 
> > 
> > @rajaths010494 we should do this change across.
> 
> Yeah saw i will make the necessary changes

image
Does this look fine.  @JDarDagran @sunank200 . If yes i will make the necessary changes.

JDarDagran:
> > > > Actually why do you need this format:
> > > > ```
> > > > f"{self.input_data.openlineage_dataset_namespace()}"
> > > > f"://{self.input_data.openlineage_dataset_name()}"
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > with two slashes before dataset name? The path should be absolute so I'd rather see URI as `file://172.26.0.2:22/tmp/sqlite.db.top_animation`.
> > > > [This](https://www.rfc-editor.org/rfc/rfc3986.txt) spec, 3.3. Path explains more.
> > > 
> > > 
> > > @rajaths010494 we should do this change across.
> > 
> > 
> > Yeah saw i will make the necessary changes
> 
> image
> 
> Does this look fine. @JDarDagran @sunank200 . If yes i will make the necessary changes.

👍

sunank200 (changes requested):
@rajaths010494 Change looks good to me.But lets add the test for this change.

sunank200:
We should create a method for URI for each database and use that. I will do that change and push it

sunank200:
Ran and tested the example DAGs for operators on Marquez.
image
image (1)

rajaths010494:
LGTM

sunank200 (approved):

phanikumv (approved):

PR #1311, opened 2022-11-28
Run jobs when PRs are synchronized from forks

kaxil:
Currently, PRs from forks are only run when label is created (or recreated) but not when more commits are added to the existing PR that already has the label as shown in https://github.com/astronomer/astro-sdk/pull/1275

### Checklist

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -14,7 +14,7 @@ on:
   # Run on PRs from forks
   pull_request_target:
     branches: [ 'main' ]
-    types: ['labeled']
+    types: ['labeled', 'synchronize']
     paths:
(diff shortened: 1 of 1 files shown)


tatiana (approved):

tatiana:
Thanks for the extremely fast fix, @kaxil !

.github/workflows/ci-python-sdk.yaml
```diff
@@ -14,7 +14,7 @@ on:
   # Run on PRs from forks
   pull_request_target:
     branches: [ 'main' ]
-    types: ['labeled']
+    types: ['labeled', 'synchronize']
```
pankajkoti, line 17: Should we also add this to the SQL CLI CI ?

Issue #1263, opened 2022-11-17
Improve the logs in case native transfers fallbacks to Pandas


magdagultekin:
**Please describe the feature you'd like to see**
When loading a CSV file from S3 to Snowflake a problem occurs, task fails but the table gets created and the data populated. However logs are not clear enough about it, please see below: 
```
[2022-11-16, 11:07:46 EST] {base.py:517} WARNING - Loading files failed with Native Support. Falling back to Pandas-based load
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/astro/databases/base.py", line 508, in load_file_to_table_natively_with_fallback
    self.load_file_to_table_natively(
  File "/usr/local/lib/python3.9/site-packages/astro/databases/snowflake.py", line 623, in load_file_to_table_natively
    self.evaluate_results(rows)
  File "/usr/local/lib/python3.9/site-packages/astro/databases/snowflake.py", line 630, in evaluate_results
    raise DatabaseCustomError(rows)
astro.exceptions.DatabaseCustomError: [{'file': 's3://s3-dev-etldata-001/inbound/test/Out_CM_CU.csv', 'status': 'LOAD_FAILED', 'rows_parsed': 168598, 'rows_loaded': 0, 'error_limit': 168598, 'errors_seen': 168598, 'first_error': 'Numeric value \'"RECORDID"\' is not recognized', 'first_error_line': 1, 'first_error_character': 1, 'first_error_column_name': '"OUT_CM_CU"["RECORDID":1]'}]
```
The logs show `0` rows loaded but it's not true. 

Refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1668615614891269
**Describe the solution you'd like**
1. I'd love to see the logs improved to reflect the action properly - meaning mention fallback to Pandas and show how many rows were populated in fact. 

2. The customer came back and said that he used `enable_native_fallback=False` and the data got loaded nevertheless. It's not an expected behaviour, right?
He also mentioned that the documentation confuses him, particularly this part (it doesn't mention fallbacking to Python - and by the way, what do you mean by that?):

**Additional context**
Task that was used:
```python
s3_to_snowflake = aql.load_file(
    task_id="s3_to_snowflake",
    input_file=File(path=f"s3://{S3_BUCKET_NAME}/{S3_FILE_NAME}", filetype=FileType.CSV),
    output_table=Table(
        conn_id=SNOWFLAKE_CONN_ID,
[...]

PR #1312, opened 2022-11-28
Improve logging for load_file

feluelle:
# Description

Currently, we don't have any logging for native nor pandas _load_file_. This makes it harder for users to follow the operations.

related: #1263

## What is the new behavior?

We now have log statements for native, pandas as well as fallback indication.

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -59,7 +59,6 @@ class BaseDatabase(ABC):
     illegal_column_name_chars: list[str] = []
     illegal_column_name_chars_replacement: list[str] = []
     NATIVE_PATHS: dict[Any, Any] = {}
-    NATIVE_LOAD_EXCEPTIONS: Any = DatabaseCustomError
     DEFAULT_SCHEMA = SCHEMA
     NATIVE_AUTODETECT_SCHEMA_CONFIG: Mapping[FileLocation, Mapping[str, list[FileType] | Callable]] = {}
     FILE_PATTERN_BASED_AUTODETECT_SCHEMA_SUPPORTED: set[FileLocation] = set()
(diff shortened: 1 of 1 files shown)


tatiana (approved):
@feluelle would it make sense to have a test covering this change? Approving it regardless, since the change already helps users.

kaxil (approved):

kaxil (commented):
One test failed for Airflow 2.4 and 2.2.5:

```
google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table astronomer-dag-authoring:first_table_schema._tmp_gcuqy9jbp3wxal8q0uqwb8jrk2a2zr6zage4t0g. Field id has changed mode from REQUIRED to NULLABLE
```

pankajkoti (approved):

pankajkoti:
> One test failed for Airflow 2.4 and 2.2.5:
> 
> ```
> google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table astronomer-dag-authoring:first_table_schema._tmp_gcuqy9jbp3wxal8q0uqwb8jrk2a2zr6zage4t0g. Field id has changed mode from REQUIRED to NULLABLE
> ```

I checked this and seems that it is because we explicitly specify ID is primary key in the target table and hence it becomes REQUIRED. Whereas in the source table, we directly load data from file without specifying the schema. 

Screenshot 2022-11-29 at 1 38 23 PM

As a result, the schema for the target table has ID as REQUIRED:
Screenshot 2022-11-29 at 1 40 22 PM


whereas, the source table has ID as NULLABLE:
Screenshot 2022-11-29 at 1 40 59 PM


So perhaps should we add the columns key in the source Table to specify that it is a primary key?

pankajkoti:
I tried doing that(specifying columns), but got the following error, so reverting my commit.

Screenshot 2022-11-29 at 2 12 42 PM

PR #1325, opened 2022-11-29
add load file using autoloader

dimberman:
# Description
## What is the new behavior?

Creates support for `aql.load_file` using autoloader. This first step will only work for s3 buckets, but we will eventually support GCS and others.

A note on the design of this solution: To allow for arbitrary user configurations and real-time s3/GCS configurations, we felt it was best to generate a python file and push it to DBFS. This method will serve as the basis for other functions in the future.

closes: https://github.com/astronomer/astro-sdk/issues/1248

No

### Checklist


kaxil:
Can you rebase the PR please

PR #1334, opened 2022-11-30
Fix bug with datetime not being serialized

kaxil:
Since Airflow still doesn't support serializing sets or datetime, it would be better to serialize dataframes to string instead of dict.

This PR also adds pre-commit hook so we don't miss `test_`  prefix in the test files

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,16 @@ repos:
       - id: trailing-whitespace
       - id: pretty-format-json
         args: ['--autofix']
-
+      - id: name-tests-test
+        args: ['--pytest-test-first']
+        exclude: |
+          (?x)
+          ^python-sdk/tests/benchmark/.*|
+          ^python-sdk/tests/utils/.*|
+          ^python-sdk/tests/sql/operators/utils.py|
+          ^python-sdk/tests/integration_test_dag.py|
+          ^sql-cli/tests/test_dag/.*$|
+          ^sql-cli/tests/utils.py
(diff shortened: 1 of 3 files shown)


pankajkoti (commented):
We will need to correct the assert in the tests here: https://github.com/astronomer/astro-sdk/blob/53d6bb7f4f531dba4c03f656890c63690d037e62/python-sdk/tests/dataframes/pandas.py#L70

to `assert s_df == {"data": {"id": {0: 1}, "name": {0: "xyz"}}}`

Also look like the tests are not running for that file as the file name is missing the `test_` prefix, we will need to rename it to test_pandas.py and hence codecoverage is reporting that the patched lines are not covered

pankajkoti:
Also seeing another issue where self.to_json() output is different thatn self.to_dict(). self.to_json() has the inner keys as strings whereas self.to_dict has the inner keys as integers.

Screenshot 2022-11-30 at 11 46 12 PM

kaxil:
>Also seeing another issue where self.to_json()s output is different than that of self.to_dict(). self.to_json() has the inner keys as strings whereas self.to_dict() has the inner keys as integers. Will this have an impact somewhere?

Airflow's XCom still can't serialize datetime, set and other types, so we can't use `self.to_dict()` whereas `self.to_json()` converts it to string, so we are good.

>Also look like the tests are not running for that file as the file name is missing the test_ prefix, we will need to rename it to test_pandas.py and hence codecoverage is reporting that the patched lines are not covered

Added pre-commit hook so that we detect such cases :)

Issue #1230, opened 2022-11-09
Generated DAGS are going into .airflow/dags instead of dags directory


mikeshwe:
**Describe the bug**

**Version**
* Astro CLI 1.7
* SQL CLI 0.1.1
* OS: [eg. Debian]

**To Reproduce**
`astro flow init --airflow-dags-folder=dags `
`astro flow run example_basic_transform
`astro flow run example_templating --env=dev`

the dag for `example_basic_transform` shows up in the correct spot, but not for `example_templating`

Guessing it's because of `--env=dev`

**Expected behavior**

**Screenshots**
![Screen Shot 2022-11-09 at 12 09 32 PM](https://user-images.githubusercontent.com/4237498/200930987-0d5bafaa-deaf-4ad4-8cf2-a85481293f77.png)

**Additional context**

vikramkoka:
Ah, that's really weird. 
cc: @utkarsharma2

tatiana:
One resolution for this problem would be to have a global configuration to store the configuration which is not environment specific. But there may be other solutions @feluelle @pankajastro @dimberman @kaxil

kaxil:
You can probably add a `AIRFLOW__CORE__DAGS_FOLDER` in https://github.com/astronomer/astro-sdk/blob/main/sql-cli/sql_cli/__init__.py so any command run from SQL CLI will have that set.

tatiana:
We'll be releasing the SQL CLI python package with this fix, but - from an Astro CLI perspective - we'll also need to have this merged: https://github.com/astronomer/astro-cli/pull/924. We'll be negotiating to have a 1.8 patch release with this fix, but in the worse case, it should be part of the Astro CLI 1.9.

tatiana:
Since we completed the Python part relevant to the 0.2.2 release, I'll consider this ticket closed. We should have a patch release of Astro CLI 1.8 containing the mentioned change (probably 1.8.4).

PR #1338, opened 2022-12-01
Use global config to be shared across environments

pankajkoti:
# Description

Currently, we have only one `configuration.yml `for each environment. With project initialisation, if
`airflow_home` and `airflow_dags_folder` are specified, we write those config keys to the default environment
directory's configuration.yml. As a result when workflows with different environment are run, that environment's
`configuration.yml `does not have `airflow_home` and `airflow_dags_folder` set with what was supplied during project
initialisation and hence the dags go in the default directory which is `.airflow/dags` instead of the one given by user.

related: #1230 

## What is the new behavior?

We now have a separate `configuration.yml` under a `global` subdirectory in the `config` directory which records global config keys 
like `airflow_home` and `airflow_dags_folder` which are meant to be same across environments. 

Additionally, we are also exposing a config typer command meant to be used by Astro CLI to get the `airflow_home` and
`airflow_dags_folder` values set in SQL CLI so that those can be mounted in the docker container while creating in Astro CLI.

No

### Checklist


sql-cli/sql_cli/__main__.py
```diff
@@ -48,6 +48,32 @@ def about() -> None:
...
+        help="(Optional) Environment used to fetch the configuration key from.",
+    ),
+    key: str = typer.Option(
+        default=None,
+        help="(Optional) Key from the configuration whose value needs to be fetched.",
+    ),
+) -> None:
+    from sql_cli.configuration import Config
+
+    project_dir_absolute = project_dir.resolve() if project_dir else Path.cwd()
```
dimberman, line 72: Can you put a comment here to explain why we default to the current working directory?

sql-cli/sql_cli/configuration.py
```diff
@@ -88,34 +95,26 @@ def write_value_to_yaml(self, section: str, key: str, value: str) -> None:
...
-            - item1
-            - item2
-        ```
-
-        :param section: Section within the YAML file where the key/value will be recorded
-        :param values: List of items to be written to the YAML file
-        """
-        yaml_config = self.from_yaml_to_dict()
-        yaml_config["connections"] = self.connections
+        """Write Config instance's key-values to respective environment specific and global configuration.yml."""
```
dimberman, line 106: Can you write more description here? Just reading this function doc I'm not sure what this function is meant to do. Would like to know why we do this.

dimberman (changes requested):

sql-cli/sql_cli/__main__.py
tatiana: Since this behaviour happens in most - if not all of our commands - perhaps we could separate it into a function.

sql-cli/tests/test_project.py
```diff
@@ -16,6 +16,8 @@
     Path("config/default/configuration.yml"),
     Path("config/dev"),
     Path("config/dev/configuration.yml"),
+    Path("config/global"),
```
tatiana, line 19: @pankajkoti, what do you think if we wrote a test that tried to reproduce @mikeshwe 's initial problem - and confirmed that the current solution works?

pankajkoti: Added test `test_generate_custom_airflow_config` in `test___main__.py`

sql-cli/sql_cli/configuration.py
pankajkoti: Added more description.

sql-cli/sql_cli/__main__.py
pankajkoti: @dimberman If project directory is not given, we assume that the user is calling the command from within the project directory and hence we default to CWD. Moved it to a separate functions, since it is common across most commands and added a comment there on why we use CWD.

sql-cli/sql_cli/configuration.py
```diff
@@ -39,42 +39,49 @@ class Config:
...
     def from_yaml_to_config(self) -> Config:
-        """
-        Return a Config instance with the content of the configuration.yaml.
-
-        :returns: Contents of the YAML configuration file.
-        """
-        yaml_config = self.from_yaml_to_dict()
+        """Returns a Config instance with the contents of the environment specific and global configuration.yaml"""
+        env_yaml_config = self.from_yaml_to_dict(self.get_env_config_filepath())
+        global_yaml_config = self.from_yaml_to_dict(self.get_global_config_filepath()) or {}
```
feluelle, line 75: According to the `from_yaml_to_dict` function definition it always returns a dict. So `or {}` will never be reached. Either the return type is wrong or the `or {}` can be removed. But my guess is that the return type is wrong in case the global config does not exist, correct?

sql-cli/sql_cli/configuration.py
```diff
@@ -88,31 +95,38 @@ def write_value_to_yaml(self, section: str, key: str, value: str) -> None:
...
         :param value: Value associated to the key in the YAML file.
         """
-        yaml_config = self.from_yaml_to_dict()
+        yaml_config = self.from_yaml_to_dict(filepath)
         yaml_config.setdefault(section, {})
         yaml_config[section][key] = value
 
-        filepath = self.get_filepath()
-        with open(filepath, "w") as fp:
+        with filepath.open(mode="w") as fp:
```
feluelle, line 102: I like that you are using pathlib .open :)

sql-cli/sql_cli/configuration.py
```diff
@@ -88,31 +95,38 @@ def write_value_to_yaml(self, section: str, key: str, value: str) -> None:
...
+        using relative paths. However, for the connection to be established successfully, it needs absolute path, so the
+        project initialisation flow reads the default yaml into config instance, transforms the config instance to
+        expand those relative paths to absolute paths and then calls this utility to persists the transformed instance.
         """
-        yaml_config = self.from_yaml_to_dict()
-        yaml_config["connections"] = self.connections
+        env_config_filepath = self.get_env_config_filepath()
+        env_yaml_config = self.from_yaml_to_dict(env_config_filepath)
+        env_yaml_config["connections"] = self.connections
+        with open(env_config_filepath, "w") as fp:
```
feluelle, line 122: You could do the same here.

sql-cli/sql_cli/configuration.py
```diff
@@ -88,31 +95,38 @@ def write_value_to_yaml(self, section: str, key: str, value: str) -> None:
...
         if self.airflow_home:
-            yaml_config["airflow"]["home"] = self.airflow_home
+            global_yaml_config["airflow"]["home"] = self.airflow_home
         if self.airflow_dags_folder:
-            yaml_config["airflow"]["dags_folder"] = self.airflow_dags_folder
-        filepath = self.get_filepath()
-        with open(filepath, "w") as fp:
-            yaml.dump(yaml_config, fp)
+            global_yaml_config["airflow"]["dags_folder"] = self.airflow_dags_folder
+        with open(global_config_filepath, "w") as fp:
```
feluelle, line 131: And here.

sql-cli/sql_cli/project.py
```diff
@@ -76,22 +82,34 @@ def airflow_config(self) -> dict[str, Any]:
...
         """
-        Sets custom Airflow configuration in case the user is not using the default values.
+        Initialises global config file that includes configuration to be shared across environments including the
+        airflow config.
+        """
+        config = Config(environment=GLOBAL_CONFIG, project_dir=self.directory)
+        global_env_filepath = config.get_global_config_filepath()
+        config.write_value_to_yaml("airflow", "home", str(self._airflow_home.resolve()), global_env_filepath)
+        if not Path.exists(self._airflow_dags_folder):
+            raise FileNotFoundError(f"Specified DAGs directory {self._airflow_dags_folder} does not exist.")
```
feluelle, line 94: The directory does not exist - not the file. Or is it common to raise this error also for directories? 🤔

sql-cli/sql_cli/project.py
```diff
@@ -76,22 +82,34 @@ def airflow_config(self) -> dict[str, Any]:
...
-    def update_config(self, environment: str = DEFAULT_ENVIRONMENT) -> None:
+    def _initialise_global_config(self) -> None:
         """
-        Sets custom Airflow configuration in case the user is not using the default values.
+        Initialises global config file that includes configuration to be shared across environments including the
+        airflow config.
+        """
+        config = Config(environment=GLOBAL_CONFIG, project_dir=self.directory)
+        global_env_filepath = config.get_global_config_filepath()
+        config.write_value_to_yaml("airflow", "home", str(self._airflow_home.resolve()), global_env_filepath)
```
feluelle, line 92: Should we also validate here that the airflow home exists?

sql-cli/sql_cli/project.py
```diff
@@ -76,22 +82,34 @@ def airflow_config(self) -> dict[str, Any]:
...
+        converted to absolute paths once the project directory is initialised so that the connections work successfully.
 
-        config.connections[0]["host"] = str(self.directory / config.connections[0]["host"])
+        :param environment: the environment for which the configuration has to be updated
+        """
+        config = Config(environment=environment, project_dir=self.directory)
+        config = config.from_yaml_to_config()
+        for connection in config.connections:
+            if connection["conn_type"] == SQLITE_CONN_TYPE and not os.path.isabs(connection["host"]):
+                connection["host"] = str(self.directory / connection["host"])
```
feluelle, line 112: This seems to be quite specific. Could we move that into its own context/function? What does this do? If it makes the path absolute we can use pathlib .resolve to make it absolute.

sql-cli/sql_cli/configuration.py
pankajkoti: Yes, if the file content is empty, the return is None, hence was converting it to empty dictionary. But made the change to instead return empty dictionary from `from_yaml_to_dict` in case the return is `None`.

sql-cli/sql_cli/project.py
pankajkoti: I did not find an exception for DirectoryNotFound . Upon reading docs and comments found that this is the exception that we could use and hence used it. https://docs.python.org/3/library/exceptions.html#FileNotFoundError

sql-cli/sql_cli/project.py
feluelle: Right, thank you :) But can we add that as a comment please :)

sql-cli/sql_cli/project.py
feluelle: Thanks for looking this up. Sounds good to me then. :)

sql-cli/sql_cli/project.py
pankajkoti: Resolve does not help as it somehow resolves it to the directory from where the command is called and does not necessary prefix the path with the initialised project's path. We can move it to a separate function once we have more use cases of what all transformations need to be done. Right now the function is quite small enough.

pankajkoti: As discussed over call, we are also checking that the resolved path exists. If it does not exist, we raise an exception. It is expected that the user specifies an absolute path. So, if a user has specified an absolute path, the code is a no-operation. But, if it is not an absolute path, we try to resolve it with the project directory as the base path and if the resolved path does not exist we raise an exception. I think this is the best we can do (Do nothing if it is an absolute path. If not try the best to resolve by prefixing with project directory).

sql-cli/sql_cli/project.py
pankajkoti: Added.

sql-cli/sql_cli/project.py
```diff
@@ -76,22 +82,45 @@ def airflow_config(self) -> dict[str, Any]:
...
+                # The example workflows have relative paths for the host URLs for SQLite connections. Additionally, the
+                # user might also sometimes set relative paths for the host from the initialised project directory. Such
+                # paths need to be converted to absolute paths so that the connections work successfully.
+                resolved_host_path = self.directory / host_path
+                if not resolved_host_path.exists():
+                    raise FileNotFoundError(
+                        f"The relative file path {host_path} was resolved into {resolved_host_path} but it's a failed "
+                        f"resolution as the path does not exist."
+                    )
+                connection["host"] = str(resolved_host_path)
```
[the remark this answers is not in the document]
pankajkoti: Created a ticket: https://github.com/astronomer/astro-sdk/issues/1375

feluelle (approved):

tatiana (approved):

Issue #1337, opened 2022-12-01
Fix CI failures found while testing SDK with airflow 2.5.0


phanikumv:
**Describe the bug**
Failing jobs on CI:

- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039998660
- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039999158
- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039999242
- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039999316
- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039999397
- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039999475
- [ ] https://github.com/astronomer/astro-sdk/actions/runs/3588515052/jobs/6039999567

Related to  https://github.com/astronomer/astro-sdk/pull/1334 and https://github.com/astronomer/astro-sdk/pull/1331 and https://github.com/astronomer/astro-sdk/pull/1319

feluelle:
Fixed by #1317

PR #1347, opened 2022-12-01
Fix test: `test_col_case_is_preserved`

kaxil:
This test was checking a wrong thing, see below:

```python
In [1]: import pandas

In [2]: df = pandas.DataFrame({"Numbers": [1, 2, 3], "Colors": ["red", "white", "blue"]})

In [3]: df.columns
Out[3]: Index(['Numbers', 'Colors'], dtype='object')

In [4]: cols = list(df.columns)

In [5]: cols.sort()

In [6]: df.columns == ["Colors", "Numbers"]
Out[6]: array([False, False])
```

It was returning a numpy array with `False, False` values, instead the test actually was meant to test whether the column case was preserved or not.

part of https://github.com/astronomer/astro-sdk/issues/1337

### Checklist


sunank200 (approved):
Nice catch. <3

Issue #1219, opened 2022-11-08
[sql-cli] Improve how the Airflow configuration is set


tatiana:
**Context**

Airflow versions until 2.4 were not designed to be used as a Python library. They do not expect users to change the project settings/configuration during runtime/execution after the Airflow package is imported.

During the development of the SQL CLI 0.1, a few different approaches were used as a workaround to overcome this. They are illustrated below.

1. Use environment variables, as we do during the project initialization (excerpt from `sql_cli/project.py`), by invoking Airflow as a command line tool:
https://github.com/astronomer/astro-sdk/blob/7d560c330b26efcae0e0fdb293dffc1d7c1d05cd/sql-cli/sql_cli/project.py#L102-L113

2. Overriding the Airflow configuration file so these configurations are used in future executions.

Example from `sql_cli/utils/airflow.py`:

https://github.com/astronomer/astro-sdk/blob/7d560c330b26efcae0e0fdb293dffc1d7c1d05cd/sql-cli/sql_cli/utils/airflow.py#L37-L43

3. Overriding environment variables from within the Python process after the Airflow module was imported and reload the relevant Airflow python modules from `sql_cli/utils/airflow.py`:

https://github.com/astronomer/astro-sdk/blob/7d560c330b26efcae0e0fdb293dffc1d7c1d05cd/sql-cli/sql_cli/utils/airflow.py#L54-L60

4. Using incorrectly set settings.DAGS_FOLDER, from `sql_cli/utils/airflow.py:

https://github.com/astronomer/astro-sdk/blob/7d560c330b26efcae0e0fdb293dffc1d7c1d05cd/sql-cli/sql_cli/utils/airflow.py#L109

**Consequences**

At the moment (0.1), the SQL CLI approach to setting the Airflow configuration is prone to errors.

Some commands fall back to AIRFLOW_HOME, whereas we'd like to override and have control of this.

Example of lines set in the tests necessary for the `flow run` command:
https://github.com/astronomer/astro-sdk/blob/7d560c330b26efcae0e0fdb293dffc1d7c1d05cd/sql-cli/noxfile.py#L31-L32

Since we want a compatibility matrix with multiple versions of Airflow, addressing this issue becomes particularly important.

**Some possible solutions**

a) Change Airflow to allow users to change the settings during runtime
b) Change the SQL CLI code in a way we only import Airflow after setting all the necessary configuration
c) A consistent way of either using environment variables or an Airflow config file - in one single point in the code (entry point?)

PR #1350, opened 2022-12-02
Improve airflow configuration

feluelle:
# Description

Currently, our airflow configuration consists of settings in the config file as well as in env vars. This PR addresses it by using a single approach.

closes: https://github.com/astronomer/astro-sdk/issues/1219

## What is the new behavior?

We only use env vars for setting airflow configuration. Unfortunately every time we set those, we need to reload the modules for them to read the updated env var.

No.

### Checklist

### DIFF
--- a/sql-cli/noxfile.py
+++ b/sql-cli/noxfile.py
@@ -43,32 +43,20 @@ def test(session: nox.Session, airflow: str) -> None:
         session.install("pytest-cov")
         session.install("mypy")
         session.install("types-pyyaml")
+
+        session.log("Installed Dependencies:")
+        session.run("pip3", "freeze")
+
+        session.run("pytest")
     else:
         session.install("poetry")
         session.run("poetry", "install", "--with", "dev")
         session.run("poetry", "run", "pip", "install", f"apache-airflow=={airflow}")
 
-    session.log("Installed Dependencies:")
-    session.run("pip3", "freeze")
+        session.log("Installed Dependencies:")
+        session.run("pip3", "freeze")
 
-    # At the moment flow run depends on the Airflow global Airflow Home - we need to fix this
-    # TODO: refactor so each test does this in their own sandboxed Airflow home
-    # From what I observed, the `run` command is the one using this, due to how the method
-    # utils.airflow.get_dag behaves under Airflow 2.2.0 (it defaults to $HOME/airflow)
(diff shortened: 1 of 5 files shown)


sql-cli/noxfile.py
```diff
@@ -43,32 +43,20 @@ def test(session: nox.Session, airflow: str) -> None:
...
-
-    # At the moment flow run depends on the Airflow global Airflow Home - we need to fix this
-    # TODO: refactor so each test does this in their own sandboxed Airflow home
-    # From what I observed, the `run` command is the one using this, due to how the method
-    # utils.airflow.get_dag behaves under Airflow 2.2.0 (it defaults to $HOME/airflow)
-    airflow_home = f"~/sql-cli-python-{session.python}-airflow-{airflow}"
-    session.run(
-        "airflow", "db", "init", env={"AIRFLOW_HOME": airflow_home, "AIRFLOW__CORE__LOAD_EXAMPLES": "False"}
-    )
-
```
tatiana, line 62: I'm so emotional about this code going away :sob:  Thank you so much, @feluelle !

sql-cli/sql_cli/utils/airflow.py
```diff
@@ -24,61 +24,56 @@ def airflow_version() -> Version:
...
-    """
-    filename = airflow_home / "airflow.cfg"
-    parser = ConfigParser()
-    parser.read(filename)
-    parser["core"]["load_examples"] = "False"
-    parser["core"]["logging_level"] = "WARN"
-    with open(filename, "w") as out_fp:
-        parser.write(out_fp)
+    # Initialise the airflow database & hide all logs
+    os.system("airflow db init > /dev/null 2>&1")  # skipcq: BAN-B605,BAN-B607
```
tatiana, line 48: Should we improve this in the verbose/debug ticket - so we can have more visibility if we use the flag?

tatiana (approved):
This looks incredible, @feluelle , amazing work! Thank you very much!!!

feluelle:
Thank you ❤️

sql-cli/sql_cli/utils/airflow.py
feluelle: Yes, great idea! I will add this to the ticket.

PR #1352, opened 2022-12-02
Add load local file to databricks Delta Table

dimberman:
# Description

As a first pass for loading files into delta, we will take out any complexity of external data stores and autoloader. This first pass allows loading local files into DBFS that can then become delta tables. That said, everything here is built with the understanding that we will expand into these features.Future PRs.

## What is the new behavior?

Users can now load local files into delta tables using the aql.load_file function.

### Checklist

### DIFF
--- a/python-sdk/conftest.py
+++ b/python-sdk/conftest.py
@@ -14,6 +14,7 @@ from airflow.utils.session import create_session, provide_session
 
 from astro.constants import Database, FileLocation, FileType
 from astro.databases import create_database
+from astro.databricks.load_options import default_delta_options
 from astro.table import MAX_TABLE_NAME_LENGTH, Table, TempTable
 from tests.sql.operators import utils as test_utils
 
@@ -140,8 +141,13 @@ def database_table_fixture(request):
 
     database.populate_table_metadata(table)
     database.create_schema_if_needed(table.metadata.schema)
+
     if file:
-        database.load_file_to_table(file, table)
+        database.load_file_to_table(
+            file,
+            table,
+            load_options=default_delta_options,
+        )
     yield database, table
     database.drop_table(table)
 
@@ -204,7 +210,11 @@ def multiple_tables_fixture(request, database_table_fixture):
         database.populate_table_metadata(table)
         database.create_schema_if_needed(table.metadata.schema)
         if file:
-            database.load_file_to_table(file, table)
+            database.load_file_to_table(
+                file,
+                table,
+                load_options=default_delta_options,
+            )
         tables.append(table)
 
     yield tables if len(tables) > 1 else tables[0]
--- a/python-sdk/docs/guides/databricks.rst
+++ b/python-sdk/docs/guides/databricks.rst
@@ -1,0 +1,121 @@
+.. _databricks:
+
+======================
+Databricks Integration
+======================
+Introduction
+
+The Astro Python SDK now allows users to easily access and work with data stored in Databricks through the Delta API.
+
+The Astro Python SDK provides a simple, yet powerful interface for working with Spark and Delta tables in Databricks, allowing users to take advantage of the scalability and performance of Databricks while maintaining the flexibility of Airflow DAG authoring and all the benefits of Airflow scheduling.
+
+Installation
+============
+To use the Astro Python SDK with Databricks, complete the following steps:
+
+The first step is to install the databricks submodule of the astro-sdk-python pip library. This can be done by running the following command:
+
+.. code-block:: bash
+
+    pip install 'astro-sdk-python[databricks]'
+
+The second step is to create a connection to your Databricks cluster.
+This requires creating a `personal access token <https://docs.databricks.com/dev-tools/api/latest/authentication.html>`_ in Databricks and creating a cluster with an http_endpoint. Once you have these, you can create a connection using the following syntax:
+
+
+.. code-block:: yaml
+
+    - conn_id: databricks_conn
+      conn_type: databricks
+      host: https://my-cluster.cloud.databricks.com/
+      password: my-token
+      extra:
+        http_path: /sql/1.0/warehouses/foobar
+
+.. note::
+
+    You need to update the ``password`` with your token from Databricks.
+
+You can also set a default databricks cluster to run all jobs against with this environment variable:
+
+.. code-block:: bash
+
+    AIRFLOW__ASTRO_SDK__DATABRICKS_CLUSTER_ID=<my databricks cluster>
+
+Loading Data
+============
+Once you have installed the astro-sdk package and created a connection to your Databricks cluster, you can begin loading data into your Databricks cluster.
+
+To load data into your Databricks cluster, you can use the same ``aql.load_file()`` function that works for all other databases.
+The only thing that is different with delta is that you now can use the ``delta_options`` parameter to specify delta specific parameters(such as ``copy_options`` and ``format_options`` for the ``COPY INTO`` command).
+Please note that when loading data into Delta using ``COPY INTO``, you must specify the filetype as Databricks does not automatically infer the data (this is not the case for autoloader).
+
+Currently, only local files can be loaded using the ``aql.load_file()`` function. Support for loading data from S3 and GCS will be added soon.
+
+To use the ``aql.load_file()`` function, you will need to specify the path to the file you want to load, the target Delta table you want to pass the result to.g
+
+.. code-block:: python
+
+    aql.load_file(
+        input_file=File("data.csv"),
+        output_table=Table(conn_id="my_databricks_conn"),
+    )
+
+If you have extra options you would like to add, you can user the ``load_options`` parameter to pass ``copy_into_parameters`` into the ``COPY INTO`` command.
+
+Please note that we by default set ``header`` and ``inferSchema`` to true, so if you pass in your own commands you will need to set those values explicitly.
+
+.. code-block:: python
+
+    from astro.databricks.load_options import DeltaLoadOptions
+
+    aql.load_file(
+        input_file=File("data.csv"),
+        output_table=Table(conn_id="my_databricks_conn"),
+        databricks_options=DeltaLoadOptions(copy_into_format_options={"header": "true"}),
+    )
+
+We also offer a ``astro.databricks.load_options.default_delta_options`` for those who do not want to manually set options.
+If you wish to use the defaults, you only need to set the ``AIRFLOW__ASTRO_SDK__DATABRICKS_CLUSTER_ID`` env variable
+so the Astro SDK knows where to send your load_file job.
+
+
+Querying Data
+=============
+Once you have loaded your data into Databricks, you can use the ``aql.transform()`` functions to create queries against the Delta tables. We currently do not support arbitrary Spark Python, but users can pass resulting Delta tables into local Pandas DataFrames (though please be careful of how large of a table you are passing).
+
+For example, you can use the ``aql.transform()`` function decorator to create a query that selects all users over the age of 30 and returns the results as a Pandas DataFrame:
+
+.. code-block:: python
+
+    @aql.transform()
+    def get_eligible_users(user_table):
+        return "SELECT * FROM {{user_table}} WHERE age > 30"
+
+
+    with dag:
+        user_table = aql.load_file(
+            input_file=File("data.csv"),
+            output_table=Table(conn_id="my_databricks_conn"),
+            databricks_options={
+                "copy_into_options": {"format_options": {"header": "true"}}
+            },
+        )
+        results = get_eligible_users(user_table)
+
+Parameterized Queries
+=====================
+
+The aql.transform() function in the Astro Python SDK allows users to create parameterized queries that can be executed with different values for the parameters. This is useful for reusing queries and for preventing SQL injection attacks.
+
+To create a parameterized query, you can use double brackets ({{ and }}) to enclose the parameter names in the query string. The aql.transform() function will replace the parameter names with the corresponding values when the query is executed.
+
+For example, you can create a parameterized query to select all users over a specified age like this:
+
+.. code-block:: python
+
+    @aql.transform()
+    def my_query(table: Table, age: int):
+        return "SELECT * FROM {{ table }} WHERE age > {{ age }}"
+
+The aql.transform() function will replace {{ table }} with users and {{ age }} with 30, and then run the resulting query against the Delta table.
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -15,6 +15,7 @@ from astro.dataframes.pandas import PandasDataframe
 
 if TYPE_CHECKING:  # pragma: no cover
     from sqlalchemy.engine.cursor import CursorResult
+
 from sqlalchemy.sql import ClauseElement
 from sqlalchemy.sql.elements import ColumnClause
 from sqlalchemy.sql.schema import Table as SqlaTable
@@ -32,6 +33,7 @@ from astro.exceptions import DatabaseCustomError, NonExistentTableException
 from astro.files import File, resolve_file_path_pattern
 from astro.files.types import create_file_type
 from astro.files.types.base import FileType as FileTypeConstants
+from astro.options import LoadOptions
 from astro.settings import LOAD_FILE_ENABLE_NATIVE_FALLBACK, LOAD_TABLE_AUTODETECT_ROWS_COUNT, SCHEMA
 from astro.table import BaseTable, Metadata
 
@@ -398,12 +400,14 @@ class BaseDatabase(ABC):
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        load_options: LoadOptions = LoadOptions(),
         **kwargs,
     ):
         """
         Load content of multiple files in output_table.
         Multiple files are sourced from the file path, which can also be path pattern.
 
+        :param load_options: Options for loading into specific data warehouses. Currently supports: databricks
         :param input_file: File path and conn_id for object stores
         :param output_table: Table to create
         :param if_exists: Overwrite file if exists
@@ -732,6 +736,28 @@ class BaseDatabase(ABC):
             self.get_table_qualified_name(table),
             self.get_table_qualified_name(table),
         )
+
+    def row_count(self, table: BaseTable):
+        """
+        Returns the number of rows in a table.
+
+        :param table: table to count
+        :return: The number of rows in the table
+        """
+        result = self.run_sql(
+            f"select count(*) from {self.get_table_qualified_name(table)}"  # skipcq: BAN-B608
+        ).scalar()
+        return result
+
+    def parameterize_variable(self, variable: str):
+        """
+        While most databases use sqlalchemy, we want to open up how we paramaterize variables for databases
+        that a) do not use sqlalchemy and b) have different parameterization schemes (namely delta).
+
+        :param variable: The variable to parameterize.
+        :return: Variable with proper parameters
+        """
+        return ":" + variable
 
     def is_native_load_file_available(  # skipcq: PYL-R0201
         self, source_file: File, target_table: BaseTable  # skipcq: PYL-W0613
--- a/python-sdk/src/astro/databricks/api_utils.py
+++ b/python-sdk/src/astro/databricks/api_utils.py
@@ -1,0 +1,138 @@
+import fileinput
+import logging
+import pathlib
+import time
+from pathlib import Path
+
+from airflow.configuration import conf
+from airflow.exceptions import AirflowException
+from databricks_cli.dbfs.api import DbfsApi, DbfsPath
+from databricks_cli.jobs.api import JobsApi
+from databricks_cli.runs.api import RunsApi
+from databricks_cli.sdk.api_client import ApiClient
+
+from astro.databricks.load_file.load_file_python_code_generator import render
+from astro.databricks.load_options import DeltaLoadOptions
+
+cwd = pathlib.Path(__file__).parent
+log = logging.getLogger(__name__)
+
+
+def generate_file(
+    data_source_path: str,
+    table_name: str,
+    source_type: str,
+    output_file_path: Path,
+    load_options: DeltaLoadOptions,
+    file_type: str = "",
+):
+    """
+    In order to run autoloader jobs in databricks, we need to generate a python file that creates a pyspark job.
+    This function uses jinja templating to generate a file with all necessary user inputs
+
+    :param data_source_path: The path where the data we're injesting lives (e.g. the s3 bucket URL)
+    :param table_name: The delta table we would like to push the data into
+    :param source_type: the source location (e.g. s3)
+    :param load_options: any additional load options the user would like to inject into their job
+    :param output_file_path: where the generated file should be placed.
+    :param file_type: when using the COPY INTO command, Spark requires explicitly stating the data type you are loading.
+        For individual files we can infer the file type, but if you are loading a directory please explicitly state
+        the file type in the File object.
+    :return: output file path
+    """
+    render(
+        Path("jinja_templates/load_file_to_delta.py.jinja2"),
+        {
+            "source_type": source_type,
+            "table_name": table_name,
+            "load_options": load_options,
+            "input_path": data_source_path,
+            "file_type": file_type.upper(),
+        },
+        output_file_path,
+    )
+
+    # Since we had to use jinja2 safe strings, we want to inline replace \' with '
+    # This is not a security concern now that jinja2 is no longer running.
+    for line in fileinput.input(output_file_path, inplace=True):
+        print(line.replace("\\'", "'"), end="")
+    log.debug("Created file %s", output_file_path)
+    return output_file_path
+
+
+def load_file_to_dbfs(local_file_path: Path, file_name: str, api_client: ApiClient) -> Path:
+    # TODO we should allow arbitrary dbfs paths set by env vars/users
+    """Load a file into DBFS. Used to move a python file into DBFS, so we can run the jobs as pyspark jobs.
+    Currently saves the file to dbfs:/mnt/pyscripts/, but will eventually support more locations.
+
+    :param local_file_path: path of the file to upload
+    :param api_client: Databricks API client
+    :return: path to the file in DBFS
+
+    """
+    log.debug("loading file %s", str(local_file_path))
+    dbfs = DbfsApi(api_client=api_client)
+    file_path = DbfsPath("dbfs:/mnt/pyscripts/")
+    dbfs.mkdirs(file_path)
+    dbfs.delete(file_path.join(file_name), recursive=False)
+    dbfs.put_file(src_path=str(local_file_path), dbfs_path=file_path.join(file_name), overwrite=True)
+    return Path("dbfs:/mnt/pyscripts") / file_name
+
+
+def create_and_run_job(
+    api_client,
+    file_to_run: str,
+    databricks_job_name: str,
+    existing_cluster_id=None,
+    new_cluster_specs=None,
+) -> None:
+    """Creates a databricks job and runs it to completion.
+
+    :param databricks_job_name: Name of the job to submit
+    :param file_to_run: path to file we will run with the job
+    :param api_client: databricks API client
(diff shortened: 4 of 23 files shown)


python-sdk/src/astro/files/base.py
```diff
@@ -25,11 +25,13 @@ class File(LoggingMixin, Dataset):
     :param conn_id: Airflow connection ID
     :param filetype: constant to provide an explicit file type
     :param normalize_config: parameters in dict format of pandas json_normalize() function.
+    :param delta_options:
     """
 
     path: str
     conn_id: str | None = None
     filetype: constants.FileType | None = None
+    delta_options: dict = {}
```
tatiana, line 34: @dimberman as we spoke, I believe the `delta_options` we're setting relate to the transfer as opposed to the file definition - so, similar to other load operations (e.g. `native_support_kwargs`), I believe they should be part of the `load_file`.

dimberman: @tatiana fixed

python-sdk/conftest.py
```diff
@@ -141,7 +141,14 @@ def database_table_fixture(request):
     database.populate_table_metadata(table)
     database.create_schema_if_needed(table.metadata.schema)
     if file:
-        database.load_file_to_table(file, table)
+        database.load_file_to_table(
+            file,
+            table,
+            databricks_options={
```
tatiana, line 147: So far, we've avoided naming arguments of generic methods with implementation-specific names. Would it make sense to call this: `native_support_kwargs` - similar to other methods?

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
...
+      host: https://my-cluster.cloud.databricks.com/
+      password: my-token
+      extra:
+        http_path: /sql/1.0/warehouses/foobar
+
+Loading Data
+============
+Once you have installed the astro-sdk package and created a connection to your Databricks cluster, you can begin loading data into your Databricks cluster.
+
+To load data into your databricks cluster, you can user the same ``aql.load_file()`` function that works for all other databases.
```
tatiana, line 37:
```suggestion
To load data into your Databricks cluster, you can use the same ``aql.load_file()`` function that works for all other databases.
```

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
...
+The only thing that is different with delta is that you now can use the ``delta_options`` parameter to specify delta specific parameters(such as ``copy_options`` and ``format_options`` for the ``COPY INTO`` command).
+Please note that when loading data into Delta using ``COPY INTO``, you must specify the filetype as Databricks does not automatically infer the data (this is not the case for autoloader).
```
tatiana, line 39: @dimberman would it make sense for us to infer by the extension if the user doesn't specify as we do with the other `load_file` implementations?

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
...
+Currently, only local files can be loaded using the aql.load_file() function. Support for loading data from S3 and GCS will be added soon.
+
+To use the aql.load_file() function, you will need to specify the path to the file you want to load, the target Delta table, and any additional COPY INTO arguments using the databricks_options parameter. For example:
+
+.. code-block:: python
+
+    aql.load_file(
+        input_file=File("data.csv"),
+        output_table=Table(conn_id="my_databricks_conn"),
+        databricks_options={"copy_into_options": {"format_options": {"header": "true"}}},
```
tatiana, line 50: Currently, for loading other CSV to other databases, I believe the default option is for the file to have a header. Perhaps we should make this the default in this case as well - so users don't need to specify it all the time.

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
+import fileinput
```
tatiana, line 1: Since this module is already in the `databricks` package, perhaps we could only name it `api` or give a more specific name as `delta`, if the classes/methods are specific to delta.

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    option_dict: dict = field(factory=dict)
+
+    def empty(self):
+        return len(self.option_dict or {}) == 0
+
+    def convert_to_option_string(self):
+        return ",".join([f"\\'{k}\\' = \\'{v}\\'" for k, v in self.option_dict.items()])
+
+
+def generate_file(
```
tatiana, line 33: WDYT think of a more specific name, eg. `generate_pyspark_job_file`?

python-sdk/conftest.py
kaxil: +1

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
+.. _databricks:
+
+======================
+Databricks Integration
+======================
+Introduction
+
+The Astro Python SDK now allows users to easily access and work with data stored in Databricks through the Delta API. The Astro Python SDK provides a simple, yet powerful interface for working with Spark and Delta tables in Databricks, allowing users to take advantage of the scalability and performance of Databricks while maintaining the flexibility of Airflow DAG authoring and all the benefits of Airflow scheduling.
```
kaxil, line 8: should we just link out to Delta API too - https://docs.delta.io/latest/api/python/index.html

kaxil: Can you also split this into multiple lines rather than a single line, please

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Installation
+============
+To use the Astro Python SDK with Databricks, complete the following steps:
+
+The first step is to install the databricks submodule of the astro-sdk-python pip library. This can be done by running the following command:
+
+.. code-block:: bash
+
+    pip install astro-sdk-python[databricks]
```
kaxil, line 18:
```suggestion
    pip install 'astro-sdk-python[databricks]'
```

So that it works on all shells

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
...
+
+    aql.load_file(
+        input_file=File("data.csv"),
+        output_table=Table(conn_id="my_databricks_conn"),
+        databricks_options={"copy_into_options": {"format_options": {"header": "true"}}},
+    )
+
+Querying Data
+=============
+Once you have loaded your data into Databricks, you can use the aql.transform() functions to create queries against the Delta tables. We currently do not support arbitrary Spark Python, but users can pass resulting Delta tables into local Pandas DataFrames (though please be careful of how large of a table you are passing).
```
kaxil, line 55:
```suggestion
Once you have loaded your data into Databricks, you can use the ``aql.transform()`` functions to create queries against the Delta tables. We currently do not support arbitrary Spark Python, but users can pass resulting Delta tables into local Pandas DataFrames (though please be careful of how large of a table you are passing).
```

python-sdk/src/astro/databases/base.py
```diff
@@ -733,6 +733,9 @@ def get_sqlalchemy_template_table_identifier_and_parameter(
             self.get_table_qualified_name(table),
         )
 
+    def parameterize_variable(self, variable: str):
+        return ":" + variable
```
kaxil, line 737: Can you add a description of why this function is needed and does this need to be in BaseDatabase class?

If so, when would someone need to override it?

python-sdk/src/astro/databricks/databricks_api_utils.py
kaxil: +1

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+@define
+class DeltaOption:
+    """
+    Helper class for rendering options into COPY_INTO Databricks SQL statements
+    """
+
+    option_dict: dict = field(factory=dict)
+
+    def empty(self):
+        return len(self.option_dict or {}) == 0
```
kaxil, line 27: `self.option_dict` is by default `{}` so we don't need the `or {}` part.


```python
In [3]: from attrs import define, field

In [4]: @define
   ...: class DeltaOption:
   ...:     """
   ...:     Helper class for rendering options into COPY_INTO Databricks SQL statements
   ...:     """
   ...:
   ...:     option_dict: dict = field(factory=dict)
   ...:

In [5]: DeltaOption().option_dict
Out[5]: {}
```

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def convert_to_option_string(self):
+        return ",".join([f"\\'{k}\\' = \\'{v}\\'" for k, v in self.option_dict.items()])
```
kaxil, line 30: Why do we need to escape?

```python
In [6]:     def convert_to_option_string(self):
   ...:         return ",".join([f"\\'{k}\\' = \\'{v}\\'" for k, v in self.option_dict.items()])
   ...:

In [7]:     def convert_to_option_string(self):
   ...:         return ",".join([f"'{k}' = '{v}'" for k, v in self.option_dict.items()])
   ...:
```

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    load_options: Dict[str, str],
+    output_file_path: Path,
+    copy_options: DeltaOption,
+    format_options: DeltaOption,
+    file_type: str = "",
+):
+    """
+    In order to run autoloader jobs in databricks, we need to generate a python file that creates a pyspark job.
+    This function uses jinja templating to generate a file with all necessary user inputs
+    :param format_options: FORMAT_OPTIONS options to pass to a COPY INTO command
```
kaxil, line 46:
```suggestion
    This function uses jinja templating to generate a file with all necessary user inputs

    :param format_options: FORMAT_OPTIONS options to pass to a COPY INTO command
```

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+        print("{}".format(line.replace("\\'", "'")), end="")
+    return output_file_path
+
+
+def load_file_to_dbfs(local_file_path: Path, file_name: str, api_client: ApiClient) -> Path:
+    # TODO we should allow arbitrary dbfs paths set by env vars/users
+    """
+    Load a file into DBFS. Used to move a python file into DBFS, so we can run the jobs as pyspark jobs.
+
+    Currently saves the file to dbfs:/mnt/pyscripts/, but will eventually support more locations.
```
kaxil, line 84:
```suggestion
    Currently saves the file to ``dbfs:/mnt/pyscripts/``, but will eventually support more locations.
```

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+
+    :param local_file_path: path of the file to upload
+    :param api_client: Databricks API client
+    :return: path to the file in DBFS
+    """
+    print("loading file " + str(local_file_path))
```
kaxil, line 90: Can be removed?

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    new_cluster_specs=None,
+) -> None:
+    """
+    Creates a databricks job and runs it to completion.
+    :param databricks_job_name: Name of the job to submit
+    :param file_to_run: path to file we will run with the job
+    :param api_client: databricks API client
+    :param existing_cluster_id: If you want to run this job on an existing cluster, you can give the cluster ID
+    :param new_cluster_specs: If you want to run this job on a new cluster, you can give the cluster specs
+    as a python dictionary.
```
kaxil, line 113:
```suggestion
       as a python dictionary.
```

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    json_info = {
+        "name": databricks_job_name,
+        "spark_python_task": {"python_file": file_to_run},
+    }
+    if existing_cluster_id:
+        json_info["existing_cluster_id"] = existing_cluster_id
+    elif new_cluster_specs:
+        json_info["new_cluster"] = new_cluster_specs
+
+    a = jobs_api.create_job(json=json_info)
```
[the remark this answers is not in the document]
kaxil: probably `response`?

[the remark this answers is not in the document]
kaxil: or `jobs` like Felix pointed out

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+        job_id=a["job_id"],
+        jar_params=None,
+        notebook_params=None,
+        python_params=None,
+        spark_submit_params=None,
+    )["run_id"]
+
+    runs_api = RunsApi(api_client)
+    while runs_api.get_run(run_id)["state"]["life_cycle_state"] == "PENDING":
+        print("job pending")
```
kaxil, line 137: Do you want to use a logger here instead of print? Or did you forget to remove after debugging?

python-sdk/src/astro/databricks/delta.py
```diff
@@ -92,12 +108,18 @@ def load_file_to_table(
...
+        databricks_job_name: str = "",
+        databricks_options: dict | None = None,
         **kwargs,
     ):
         """
         Load content of multiple files in output_table.
         Multiple files are sourced from the file path, which can also be path pattern.
 
+        :param databricks_options: options for passing into COPY INTO. currently we support `copy_into_options` with sub
+        dictionaries `format_options` and `copy_options`
```
kaxil, line 120:
```suggestion
        :param databricks_options: options for passing into COPY INTO. currently we support ``copy_into_options`` with sub
        dictionaries ``format_options`` and ``copy_options``
```

python-sdk/src/astro/databricks/delta.py
```diff
@@ -134,7 +161,10 @@ def create_table_from_select_statement(
         statement = self._create_table_statement.format(
             self.get_table_qualified_name(target_table), statement
         )
-        self.run_sql(sql=statement, parameters=None)
+        self.run_sql(sql=statement, parameters=parameters)
+
+    def parameterize_variable(self, variable: str):
+        return "%(" + variable + ")s"
```
kaxil, line 167:
```suggestion
    def parameterize_variable(self, variable: str) -> str:
        return "%(" + variable + ")s"
```

python-sdk/src/astro/databricks/load_file/load_file_job.py
```diff
@@ -0,0 +1,85 @@
...
+
+supported_file_locations = [FileLocation.LOCAL]
+
+
+def load_file_to_delta(
+    input_file: File, delta_table: BaseTable, databricks_job_name: str, databricks_options: dict | None = None
+):
+    """
+    Load a file object into a databricks delta table
+    :param input_file: File to load into delta
```
kaxil, line 27:
```suggestion
    Load a file object into a databricks delta table

    :param input_file: File to load into delta
```

python-sdk/src/astro/databricks/load_file/load_file_python_code_generator.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
```
kaxil, line 1: should we rename the module name from `load_file_python_code_generator.py` to `python_code_generator.py`

python-sdk/src/astro/sql/operators/base_decorator.py
```diff
@@ -188,7 +188,7 @@ def translate_jinja_to_sqlalchemy_template(self, context: dict) -> None:
                 context[k] = jinja_table_identifier
                 self.parameters[k] = jinja_table_parameter_value
             else:
-                context[k] = ":" + k
+                context[k] = self.database_impl.parameterize_variable(k)
```
kaxil, line 191: ok cool, I now know what that database function was about -- but can you still add description about that in the BaseDatabase class please

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         ndjson_normalize_sep: str = "_",
         use_native_support: bool = True,
         native_support_kwargs: dict | None = None,
+        databricks_options: dict | None = None,
```
kaxil, line 50: Let's avoid specific tech kwarg in LoadFile constructor

python-sdk/src/astro/table.py
```diff
@@ -117,6 +117,12 @@ def row_count(self) -> Any:
         Return the row count of table.
         """
         db = create_database(self.conn_id)
+        if db.sql_type == "delta":
+            result = db.run_sql(
+                f"SELECT COUNT(*) FROM {db.get_table_qualified_name(self)}",
+                handler=lambda x: x.fetchone(),
+            )
+            return result.asDict()["count(1)"]
```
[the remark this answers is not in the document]
kaxil: +1

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -114,6 +116,8 @@ def load_data_to_table(self, input_file: File) -> BaseTable:
             native_support_kwargs=self.native_support_kwargs,
             columns_names_capitalization=self.columns_names_capitalization,
             enable_native_fallback=self.enable_native_fallback,
+            databricks_job_name=f"Load data {context.get('dag_id')}_{context.get('task_id')}",
```
kaxil, line 119: Do we really need context for this?

You should just be able to do the following:

```suggestion
            databricks_job_name=f"Load data {self.dag_id}_{self.task_id}",
```

python-sdk/conftest.py
dimberman: Sounds good

python-sdk/docs/guides/databricks.rst
dimberman: That's a good call. It's annoying especially because having a header is default for autoloader but for some reason isn't for COPY INTO.

python-sdk/src/astro/databricks/databricks_api_utils.py
dimberman: Ah yes sounds good. This is a leftover of trying to get around some jinja weirdness when it comes to internal quotes. I should be able to fix this.

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+            "input_path": data_source_path,
+            "file_type": file_type.upper(),
+            "copy_options": copy_options,
+            "format_options": format_options,
+        },
+        output_file_path,
+    )
+
+    # Since we had to use jinja2 safe strings, we want to inline replace \' with '
+    # This is not a security concern now that jinja2 is no longer running.
```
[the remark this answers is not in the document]
dimberman: We don't, jinja does. Jinja doesn't like quotes because it offers an opportunity to do escapes (this is concerning since jinja can run arbitrary code). Because of this it automatically adds a \ to make the quotations "safe".

python-sdk/src/astro/databricks/databricks_api_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    for line in fileinput.input(output_file_path, inplace=True):
+        print("{}".format(line.replace("\\'", "'")), end="")
```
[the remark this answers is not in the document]
dimberman: Ah good point.

python-sdk/src/astro/databricks/databricks_api_utils.py
dimberman: I was running into issues where option_dict was None. I'll try it again and see if I get the same issue.

python-sdk/src/astro/databricks/databricks_api_utils.py
dimberman: I'll turn it into a debug log

python-sdk/conftest.py
dimberman: @tatiana @kaxil actually maybe it might be good to keep this as databricks options? The base class allows for arbitrary **kwargs so this could only matter if it's a databricks job. So if you're using delta it will matter and if you're not using delta it won't matter.

python-sdk/docs/guides/databricks.rst
dimberman: @tatiana I'm not sure what you mean here. Let's talk offline.

python-sdk/src/astro/sql/operators/load_file.py
dimberman: @kaxil let's discuss in our 1:1 I think it might honestly make sense to do it this way.

python-sdk/conftest.py
tatiana: My vote would still be to have a consistent interface across our implementations and try to have a container for platform-specific arguments - but I'm open to be persuaded otherwise :)

python-sdk/docs/guides/databricks.rst
```diff
@@ -0,0 +1,91 @@
...
+            databricks_options={
+                "copy_into_options": {"format_options": {"header": "true"}}
+            },
+        )
+        results = get_eligible_users(user_table)
+
+Parameterized Queries
+=====================
+
+The aql.transform() function in the Astro Python SDK allows users to create parameterized queries that can be executed with different values for the parameters. This is useful for reusing queries and for preventing SQL injection attacks.
```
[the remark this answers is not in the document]
kaxil: some on L109 and L111

python-sdk/src/astro/databases/base.py
```diff
@@ -733,6 +737,28 @@ def get_sqlalchemy_template_table_identifier_and_parameter(
             self.get_table_qualified_name(table),
         )
 
+    def row_count(self, table: BaseTable):
+        """
+        Returns the number of rows in a table.
+
+        :param table: table to count
+        :return:
```
kaxil, line 745: Add description to it or let's remove it

python-sdk/src/astro/databases/base.py
```diff
@@ -733,6 +737,28 @@ def get_sqlalchemy_template_table_identifier_and_parameter(
...
+        ).scalar()
+        return result
+
+    def parameterize_variable(self, variable: str):
+        """
+        While most databases use sqlalchemy, we want to open up how we paramaterize variables for databases
+        that a) do not use sqlalchemy and b) have different parameterization schemes (namely delta).
+
+        :param variable: The variable to parameterize.
+        :return:
```
kaxil, line 758: Add description to it or let's remove it

python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -0,0 +1,129 @@
...
+    )["run_id"]
+
+    runs_api = RunsApi(api_client)
+    while runs_api.get_run(run_id)["state"]["life_cycle_state"] == "PENDING":
+        log.info("job pending")
+        time.sleep(5)
+
+    while runs_api.get_run(run_id)["state"]["life_cycle_state"] == "RUNNING":
+        log.info("job running")
+        time.sleep(3)
```
kaxil, line 127: Should we put a parameterized timeout for these?

python-sdk/src/astro/databricks/load_options.py
```diff
@@ -0,0 +1,38 @@
...
+from astro.options import LoadOptions
+
+
+@define
+class DeltaLoadOptions(LoadOptions):
+    """Helper class for rendering options into COPY_INTO Databricks SQL statements"""
+
+    copy_into_format_options: dict = field(init=True, factory=dict)
+    copy_into_copy_options: dict = field(init=True, factory=dict)
+    existing_cluster_id: str = os.environ.get("AIRFLOW__ASTRO_SDK__DATABRICKS_CLUSTER_ID", "")
```
kaxil, line 16: Shouldn't this come from config and not environment variable?

kaxil: It should be in settings.py
```
conf.get("astro_sdk", "databricks_cluste_id")
```

python-sdk/src/astro/databricks/load_file/load_file_python_code_generator.py
kaxil: But it is under `load_file` folder already

python-sdk/src/astro/databricks/load_file/jinja_templates/load_file_to_delta.py.jinja2
```diff
@@ -0,0 +1,10 @@
+from pyspark.sql.functions import input_file_name, current_timestamp
+
+{% from 'jinja_templates/copy_into.py.jinja2' import copy_into_command -%}
+
+src_data_path = "{{ input_path }}"
+username = spark.sql("SELECT regexp_replace(current_user(), '[^a-zA-Z0-9]', '_')").first()[0]
+table_name = f"{{ table_name }}" # This can be generated based on task ID and dag ID or just entirely random
```
kaxil, line 7:
```suggestion
table_name = "{{ table_name }}" # This can be generated based on task ID and dag ID or just entirely random
```

python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    run_id = jobs_api.run_now(
+        job_id=job["job_id"],
+        jar_params=None,
+        notebook_params=None,
+        python_params=None,
+        spark_submit_params=None,
+    )["run_id"]
+
+    runs_api = RunsApi(api_client)
+    ping_interval = int(os.getenv("AIRFLOW__ASTRO_SDK__DATABRICKS_PING_INTERVAL", "5"))
```
kaxil, line 123: This should come from the config as well, not as an environment variable

kaxil: similar to https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/settings.py#L12

kaxil: Should we also have a timeout?

python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -0,0 +1,135 @@
...
+            "input_path": data_source_path,
+            "file_type": file_type.upper(),
+        },
+        output_file_path,
+    )
+
+    # Since we had to use jinja2 safe strings, we want to inline replace \' with '
+    # This is not a security concern now that jinja2 is no longer running.
+    for line in fileinput.input(output_file_path, inplace=True):
+        print(line.replace("\\'", "'"), end="")
```
kaxil, line 58: Do we want print here? or log?

python-sdk/src/astro/databases/base.py
```diff
@@ -398,12 +400,14 @@ def load_file_to_table(
...
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        load_options: LoadOptions = LoadOptions(),
         **kwargs,
     ):
         """
         Load content of multiple files in output_table.
         Multiple files are sourced from the file path, which can also be path pattern.
 
+        :param load_options:
```
kaxil, line 410: Missing description

kaxil (commented):
Good work @dimberman 👏 

Almost here, minor comments needs to be addressed

python-sdk/src/astro/databricks/api_utils.py
dimberman: @kaxil this is not printing anything. It's just how fileinput works. this is how we replace the lines in-place.

python-sdk/src/astro/databricks/api_utils.py
dimberman: @kaxil I don't think we should have a timeout. These things can take a while.

tatiana (approved):
I'm happy for the PR to be merged when the checks pass - and we can always improve things on separate PRs. Thanks for addressing the feedback, @dimberman !

feluelle:
@dimberman deepsource was failing. :/

PR #1353, opened 2022-12-03
Add load to spark with S3 support

dimberman:
# Description
Currently, the astro-python-sdk does not support s3 for the databricks provider. Additionally, there is no way for users to upload secrets to databricks to use for uploading files to s3 without explicitly linking their spark cluster to s3.

## What is the new behavior?
This PR adds s3 support to the databricks provider of the astro-python-sdk. It also supports uploading secrets to databricks so that users can upload files without needing to pre-load credentials to their spark cluster.

To do so, users only need to use an s3 connection ID like this:

```python
aql.load_file(
        input_file=File(
           "s3://tmp9/databricks-test/", conn_id="default_aws", filetype=FileType.CSV
       ),
        output_table=Table(conn_id="my_databricks_conn"),
  )

```

### Checklist


python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -18,6 +21,47 @@
...
+    Delete the scope we created to prevent littering the databricks secret store with one-time scopes.
+    :param scope_name: name of scope to delete
+    :param api_client: populated databricks client
+    """
+    secrets = SecretApi(api_client)
+    try:
+        secrets.delete_scope(scope_name)
+    except HTTPError as h:
+        # We don't care if the resource doesn't exist since we're trying to delete it
+        if not h.response.json()["error_code"] == "RESOURCE_DOES_NOT_EXIST":
```
kaxil, line 35:
```suggestion
        if not h.response.json().get("error_code", "") == "RESOURCE_DOES_NOT_EXIST":
```

python-sdk/src/astro/databricks/load_file/load_file_job.py
```diff
@@ -44,17 +50,26 @@ def load_file_to_delta(
     dbfs_file_path = None
     if input_file.location.location_type == FileLocation.LOCAL:
         dbfs_file_path = _load_load_file_to_dbfs(api_client, input_file)
+    elif input_file.location.location_type == FileLocation.S3:
```
kaxil, line 53: Any better thoughts on moving the if/else to more OOP paradigm?

python-sdk/src/astro/databricks/load_file/load_file_job.py
```diff
@@ -92,3 +101,34 @@ def _load_load_file_to_dbfs(api_client: ApiClient, input_file: File):
...
+    either set by the user, or inferable in the URL
+    :param input_file: file to parse
+    :return: file type of the uploaded file
+    """
+    if input_file.filetype:
+        return str(input_file.filetype)
+    if not input_file.filetype:
+        if "." in input_file.path:
+            return input_file.path.split(".")[1]
+    raise ValueError("For COPY INTO, you need to supply a file type.")
```
kaxil, line 134: Needs unit tests

python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -18,6 +21,47 @@
...
+        scope=scope_name,
+        initial_manage_principal=None,
+        backend_azure_keyvault=None,
+        scope_backend_type=None,
+    )
+
+    for k, v in filesystem_secrets.items():
+        # Add "astro_sdk_" to the front, so we know which secrets are hadoop settings
+        secret_key = "astro_sdk_" + k
+        secrets.put_secret(scope=scope_name, key=secret_key, string_value=v, bytes_value=None)
```
kaxil, line 62: Please add unit tests too so that we don't have to fully rely on integration tests

python-sdk/src/astro/databricks/load_file/jinja_templates/load_secrets.py.jinja2
```diff
@@ -0,0 +1,8 @@
+{% macro load_secrets_command(secret_scope) -%}
+secret_key_list = dbutils.secrets.list("{{ secret_scope }}")
+for secret_key in secret_key_list:
+    if 'astro_sdk_' in secret_key.key:
+        key_name = secret_key.key.replace("astro_sdk_","")
+        print(f"setting {key_name}")
+        sc._jsc.hadoopConfiguration().set(key_name, dbutils.secrets.get("{{ secret_scope }}", secret_key.key))
+{%- endmacro -%}
```
kaxil, line 8: Do we have a unit test covering the rendered template?

python-sdk/src/astro/databricks/load_file/jinja_templates/load_secrets.py.jinja2
```diff
@@ -0,0 +1,8 @@
+{% macro load_secrets_command(secret_scope) -%}
+secret_key_list = dbutils.secrets.list("{{ secret_scope }}")
+for secret_key in secret_key_list:
+    if 'astro_sdk_' in secret_key.key:
+        key_name = secret_key.key.replace("astro_sdk_","")
+        print(f"setting {key_name}")
```
kaxil, line 6: Do we need the print statement here?

python-sdk/src/astro/databricks/api_utils.py
dimberman: Oh weird I thought I had made one. Yes totally.

python-sdk/src/astro/databricks/load_file/jinja_templates/load_secrets.py.jinja2
kaxil: Yes a log statement would be better

python-sdk/src/astro/databricks/load_file/load_file_job.py
dimberman: Added a unit test for this

python-sdk/src/astro/databricks/api_utils.py
dimberman: Done

python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -18,6 +21,49 @@
...
+    :param scope_name: name of scope to delete
+    :param api_client: populated databricks client
+    """
+    secrets = SecretApi(api_client)
+    try:
+        secrets.delete_scope(scope_name)
+    except HTTPError as h:
+        # We don't care if the resource doesn't exist since we're trying to delete it
+        if not h.response.json().get("error_code", "") == "RESOURCE_DOES_NOT_EXIST":
+            raise h
```
tatiana, line 37: It would be great to have a test covering this scenario. It would be nice as well if we didn't use single-characters to represent variables, `err` is probably more meaningful.

python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -18,6 +21,49 @@
...
+
+
+def create_secrets(scope_name: str, filesystem_secrets: Dict[str, str], api_client: ApiClient) -> None:
```
tatiana, line 40: Is this the only way we support users loading from S3 into Datbricks?
Are we too opinionated here? Would it be a possibility for the user to manage their secrets and just give us the key?

python-sdk/src/astro/files/locations/base.py
```diff
@@ -132,6 +132,14 @@ def exists(self) -> bool:
         except OSError:
             return False
 
+    def databricks_settings(self) -> dict:
```
tatiana, line 135: It feels odd to have a method specific to `databricks` on the base class. Do you see any alternative ways of handling this?

python-sdk/src/astro/databricks/load_file/load_file_job.py
```diff
@@ -86,9 +95,40 @@ def _create_load_file_pyspark_file(
...
+    """
+    Since COPY INTO does not automatically detect schema, we need to ensure that the file type is
+    either set by the user, or inferable in the URL
+    :param input_file: file to parse
+    :return: file type of the uploaded file
+    """
+    if input_file.filetype:
+        return str(input_file.filetype)
+    if not input_file.filetype:
+        if "." in input_file.path:
```
tatiana, line 132: Perhaps we could use pathlib to handle this: `Path(input_file.path).suffix`

python-sdk/src/astro/databricks/api_utils.py
dimberman: @tatiana this is how we can ensure that users can use connections that exist in the Airflow DB for loading to databricks. We can make an "assume_secrets_exist" mode for users who want to pre-load these values into their databricks cluster, but I think that would be a good follow-up ticket.

python-sdk/src/astro/files/locations/base.py
dimberman: @tatiana I really don't. We need to have databricks settings specific to each file type. I don't see this as weird because databricks requires settings that others don't.

python-sdk/src/astro/databricks/load_file/load_file_job.py
dimberman: @tatiana if there is no suffix and the user does not specify the file type I don't think there's much that we can do besides raise an error. Unless we want to try to list files and see if any of those files have a suffix?

kaxil (approved):
There are some comments that needs addressing, happy for you to merge after that

python-sdk/src/astro/databricks/api_utils.py
dimberman: @tatiana I've added the ability to turn off secret loading and made it a global environment variable. I've also added documentation showing the two options so users can decide for themselves.

python-sdk/src/astro/files/locations/base.py
dimberman: I spoke with @tatiana and this conversation can be resolved as soon as I create a ticket to look into how best to support `databricks_settings` and link it here

dimberman: https://github.com/astronomer/astro-sdk/issues/1429 Added issue here

PR #1361, opened 2022-12-05
Fix validate_connection not respecting connection_id

feluelle:
# Description

Currently, we always show `Error: Config file does not contain given connection` if the user does not provide a specific connection id.

related: #1125

## What is the new behavior?

Only print the error message if there is not any connection id matching with the already registered connections.

No.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/connections.py
+++ b/sql-cli/sql_cli/connections.py
@@ -15,17 +15,13 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
     Validates that the given connections are valid and registers them to Airflow with replace policy for existing
     connections.
     """
-    config_file_contains_connection = False
-
     for connection in connections:
-        if connection.conn_id == connection_id:
-            config_file_contains_connection = True
         os.environ[f"AIRFLOW_CONN_{connection.conn_id.upper()}"] = connection.get_uri()
         status = "[bold green]PASSED[/bold green]" if _is_valid(connection) else "[bold red]FAILED[/bold red]"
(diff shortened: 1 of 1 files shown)


sql-cli/sql_cli/connections.py
```diff
@@ -15,17 +15,13 @@ def validate_connections(connections: list[Connection], connection_id: str | Non
...
     for connection in connections:
-        if connection.conn_id == connection_id:
-            config_file_contains_connection = True
         os.environ[f"AIRFLOW_CONN_{connection.conn_id.upper()}"] = connection.get_uri()
         status = "[bold green]PASSED[/bold green]" if _is_valid(connection) else "[bold red]FAILED[/bold red]"
         rprint(f"Validating connection {connection.conn_id:{CONNECTION_ID_OUTPUT_STRING_WIDTH}}", status)
 
-    if not config_file_contains_connection:
-        rprint("Error: Config file does not contain given connection", connection_id)
+    if connection_id and not any(connection.conn_id == connection_id for connection in connections):
```
kaxil, line 23: Should this be `all`?

```suggestion
    if connection_id and not all(connection.conn_id == connection_id for connection in connections):
```

pankajkoti: any looks correct to me as we want to check that any of the connection in the configuration file matches the one that we're trying to validate and if none match, then display the error.

pankajkoti (approved):

sql-cli/sql_cli/connections.py
kaxil: you are right

kaxil (approved):

PR #1362, opened 2022-12-05
Astro SQL CLI creates black compliant DAG files

dimberman:
# Description
Currently, the Astro SQL CLI generates tasks using jinja, which means that complex files can end up with a "generated" look since many lines are generated one line at a time. You will see things like:

```
    a = aql.transform_file(
        file_path=f"{DAGS_FOLDER}/include/basic/a.sql",
        parameters={
        },
```

Instead of
```
    a = aql.transform_file(
        file_path=f"{DAGS_FOLDER}/include/basic/a.sql",
        parameters={},
```
## What is the new behavior?
Now all generated files will be black compliant. This makes the DAGs both easier to read, and gives the SQL CLI a more "polished" feel to users.

### Checklist


kaxil (approved):

sql-cli/sql_cli/dag_generator.py
```diff
@@ -122,4 +123,11 @@ def generate_dag(directory: Path, dags_directory: Path, generate_tasks: bool) ->
         context={"dag": workflow_files_dag},
         output_file=output_file,
     )
+
+    # format DAG to be black compliant
+    with open(output_file, "r+") as file:
+        output_str = format_str(file.read(), mode=FileMode())
```
feluelle, line 129: Please use pathlib's [read_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text)

Could we create a small helper func for this, please? :)

sql-cli/tests/test_dag_generator.py
```diff
@@ -39,3 +39,66 @@ def test_generate_dag_invalid_directory(root_directory, dags_directory, generate
...
+        task_id="c",
+    )
+"""
+
+
+def test_generate_dag_black_compliant(root_directory, dags_directory):
+    dag_file = generate_dag(directory=root_directory, dags_directory=dags_directory, generate_tasks=True)
+    with open(dag_file) as file:
+        file_str = file.read()
+        assert file_str == compliant_dag
```
feluelle, line 104:
```suggestion
    assert dag_file.read_text() == compliant_dag
```

dimberman: done

PR #1372, opened 2022-12-06
Run SQL CLI tests with Airflow 2.5

tatiana:


.github/workflows/ci-sql-cli.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       fail-fast: false
       matrix:
         python: [ 3.8, 3.9 ]
-        airflow: [ 2.2.5, 2.3.4, 2.4.2 ]
+        airflow: [ 2.2.5, 2.3.4, 2.4.2, 2.5 ]
```
feluelle, line 71:
```suggestion
        airflow: [ 2.2.5, 2.3.4, 2.4.2, 2.5.0 ]
```

tatiana: @feluelle I believe we should be only adding the micro release if we're using constraints, WDYT?

tatiana: this would allow us to test with the latest micro releases - when possible - without us having any effort

feluelle: I am totally for it. Could you change this for 2.3 and 2.4 here as well?

kaxil: We are not using constraints file though for >2.2, or are you talking about different constraint?

tatiana: It's the same constraint, @kaxil !
I attempted removing constraints of 2.3.4 and 2.4.2 - it worked for 2.4, but in the case of 2.3, without constraints, it raised an issue due to an outdated click version:
```
 ImportError while loading conftest '/home/runner/work/astro-sdk/astro-sdk/sql-cli/conftest.py'.
conftest.py:12: in <module>
    from sql_cli.dag_generator import Workflow
sql_cli/dag_generator.py:7: in <module>
    from black import FileMode, format_str
src/black/__init__.py:31: in <module>
    ???
E   ImportError: cannot import name 'ParameterSource' from 'click.core' (/home/runner/work/astro-sdk/astro-sdk/sql-cli/.nox/test-3-9-airflow-2-3/lib/python3.9/site-packages/click/core.py)
```

E.g.:
https://github.com/astronomer/astro-sdk/actions/runs/3630597057

.github/workflows/ci-sql-cli.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
             ~/.cache/pip
             sql-cli/.nox
             .local/
-          key: ${{ runner.os }}-2.3-${{ hashFiles('sql-cli/pyproject.toml') }}
+          key: ${{ runner.os }}-${{ hashFiles('sql-cli/pyproject.toml') }}
```
pankajkoti, line 93:
```suggestion
          key: sql-cli-tests-${{ runner.os }}-${{ hashFiles('sql-cli/pyproject.toml') }}
```

can we add the job name prefix like suggested to the key.

pankajkoti (approved):
Have a minor suggestion.

kaxil (approved):

.github/workflows/ci-sql-cli.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
             ~/.cache/pip
             sql-cli/.nox
             .local/
-          key: ${{ runner.os }}-2.3-${{ hashFiles('sql-cli/pyproject.toml') }}
+          key: ${{ runner.os }}-${{ matrix.python }}-${{ matrix.airflow }}-${{ hashFiles('sql-cli/pyproject.toml') }}
```
pankajkoti, line 93: This looks good. 👍🏽

.github/workflows/ci-sql-cli.yaml
feluelle: It fails because of recently added `black`.

Instead of:
```
black = { version = "*", allow-prereleases = true }
```

we should have:
```
black = ">=22.10.0"
```

feluelle: Ah it was using the [latest version](https://github.com/astronomer/astro-sdk/actions/runs/3630597057/jobs/6124794996#step:7:496) already.

.github/workflows/ci-sql-cli.yaml
feluelle, line 93:
```suggestion
          key: os-${{ runner.os }}-python-${{ matrix.python }}-airflow-${{ matrix.airflow }}-project-${{ hashFiles('sql-cli/pyproject.toml') }}
```

feluelle: @tatiana wdyt? In case python and airflow version is the same 😅

PR #1373, opened 2022-12-06
Modify example dag test case to make sure all the dags are running

rajaths010494:
Make sure all the example dags are running.

### DIFF
--- a/python-sdk/tests/test_example_dags.py
+++ b/python-sdk/tests/test_example_dags.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import airflow
 import pytest
 from airflow.models.dagbag import DagBag
+from airflow.models.dagrun import DagRun
 from airflow.utils.db import create_default_connections
 from airflow.utils.session import provide_session
 from packaging.version import Version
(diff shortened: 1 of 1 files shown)


python-sdk/tests/test_example_dags.py
```diff
@@ -89,7 +90,11 @@ def order(dag_id: str) -> int:
 @pytest.mark.parametrize("dag_id", sorted(dag_bag.dag_ids, key=order))
 def test_example_dag(session, dag_id: str):
     dag = dag_bag.get_dag(dag_id)
-    wrapper_run_dag(dag)
+    dag_run = wrapper_run_dag(dag)
+    print(dag_run)
+    for ti in dag_run.get_task_instances():
+        if ti.state == "failed":
+            assert False
```
kaxil, line 97: This doesn't look right, 🤔 

Did we confirm that not all example DAGs were running?

rajaths010494: Yeah i am trying to do that only. This check is for if dag has run but has task failure check. 
i am yet to add a check for all dags ran or not.

PR #1374, opened 2022-12-06
Fix #1370 flow validate only desired connection

### LINKED ISSUE #1370: All connections tested even though one's been specified
**Describe the bug**
I ran `astro flow validate --connection=<connection_id>` and all connections were tested even though I passed one conn id specifically. 

**Version**
* Astro Runtime: 7.0.0
* Astro CLI: 1.8.3

**To Reproduce**
1. Update file `config/default/configuration.yml` as shown below:
```
connections:
  - conn_id: sqlite_conn
    conn_type: sqlite
    host: /Users/magdagultekin/magda-dev/data/imdb.db
    login: null
    password: null
    schema: null
  - conn_id: sqlite_default
    conn_type: sqlite
    host: /tmp/sqlite.db
    login: null
    password: null
    schema: null
```
4. Run `astro flow validate --connection=sqlite_default`
5. See message:
```
Validating connection(s) for environment 'default'
Validating connection sqlite_conn               PASSED
Validating connection sqlite_default            FAILED
```

**Expected behavior**
Only `sqlite_default` should be tested.

**Screenshots**
image

tatiana:
Fix #1370


sql-cli/tests/test_connections.py
```diff
@@ -41,3 +41,33 @@ def test_validate_connections_config_file_does_not_contain_connection(
     captured = capsys.readouterr()
     assert f"Config file does not contain given connection {unknown_connection_id}" in captured.out
     assert f"AIRFLOW_CONN_{unknown_connection_id.upper()}" not in os.environ
+
+
+@patch("sql_cli.connections.Path.is_file", return_value=True)
+@patch("sql_cli.connections.Connection.test_connection", return_value=(True, ""))
+@patch.dict(os.environ, {}, clear=True)
```
feluelle, line 48: Please also add an `assert`ion on `os.environ`.

sql-cli/tests/test_connections.py
```diff
@@ -41,3 +41,33 @@ def test_validate_connections_config_file_does_not_contain_connection(
...
+    capsys,
+):
+    validate_connections(connections=multiple_connections)
+    captured = capsys.readouterr()
+    assert captured.out.count("PASSED") == 2
+
+
+@patch("sql_cli.connections.Path.is_file", return_value=True)
+@patch("sql_cli.connections.Connection.test_connection", return_value=(True, ""))
+@patch.dict(os.environ, {}, clear=True)
```
feluelle, line 63: Same here.

sql-cli/tests/test_connections.py
```diff
@@ -41,3 +41,37 @@ def test_validate_connections_config_file_does_not_contain_connection(
...
+    assert "AIRFLOW_CONN_SQLITE_CONN2" in os.environ
+
+
+@patch("sql_cli.connections.Path.is_file", return_value=True)
+@patch("sql_cli.connections.Connection.test_connection", return_value=(True, ""))
+@patch.dict(os.environ, {}, clear=True)
+def test_validate_connections_multiple_connections_single_specified(
+    mock_test_connection,
+    mock_is_file,
+    multiple_connections,
```
kaxil, line 69: hmm why have this as a fixture and not part of this test directly?

I feel like duplication of code is better in this case than to jump to look for definition of that fixture unless we plan to use that at several places

kaxil: A good read about the same is: https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

tatiana: @kaxil, I agree. I was naively being consistent with what we already had in this test file (connections fixture) - but I agree with your view. I already changed it, please, have a look

PR #1381, opened 2022-12-07
Move `check_if_connection_exists` to `load_file`

kaxil:
This moves `check_if_connection_exists` in `python-sdk/src/astro/sql/operators/load_file.py` which is the only place it is ever used

### DIFF
--- a/python-sdk/src/astro/sql/operators/load_file.py
+++ b/python-sdk/src/astro/sql/operators/load_file.py
@@ -4,6 +4,7 @@ from typing import Any
 
 import pandas as pd
 from airflow.decorators.base import get_unique_task_id
+from airflow.hooks.base import BaseHook
 from airflow.models.xcom_arg import XComArg
 
 from astro.airflow.datasets import kwargs_with_datasets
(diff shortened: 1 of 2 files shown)


python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -327,3 +328,15 @@ def load_file(
...
+
+def check_if_connection_exists(conn_id: str) -> bool:
+    """
+    Given an Airflow connection ID, identify if it exists.
+    Return True if it does or raise an AirflowNotFoundException exception if it does not.
+
+    :param conn_id: Airflow connection ID
+    :return bool: If the connection exists, return True
+    """
+    BaseHook.get_connection(conn_id)
```
feluelle, line 341: @kaxil don't we need to catch the exception?

kaxil: No, the intention is if the connection doesn't exist, `BaseHook.get_connection(conn_id)` will raise `AirflowNotFoundException` :) 

This PR just moves the method without changing the logic

feluelle: Okay, but we have a similar function somewhere in the code base which wraps it in a try-except.

feluelle (approved):

python-sdk/src/astro/sql/operators/load_file.py
kaxil: hmm interesting, I couldn't find it in Python SDK

PR #1382, opened 2022-12-07
Automatically add `from __future__ import annotations`

kaxil:
This PR automatically adds `from __future__ import annotations` to all the Python files so that we can take advantage of the lazy imports

It uses https://pycqa.github.io/isort/docs/configuration/options.html#add-imports from isort

### DIFF
--- a/python-sdk/example_dags/example_dataframe_api.py
+++ b/python-sdk/example_dags/example_dataframe_api.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import os
 from datetime import datetime
 from io import StringIO
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -6,3 +6,4 @@ profile=black
 line_length=110
 combine_as_imports=true
 known_first_party=astro,tests,sql_cli
+add_imports=from __future__ import annotations
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -122,7 +122,7 @@ repos:
       rev: v3.3.1
       hooks:
       -   id: pyupgrade
-          args: [--py37-plus]
+          args: [--py37-plus,--keep-runtime-typing]
           # Exclude auto-generated example files from being changed
           exclude: ^sql-cli/include/base/.airflow/dags
 
--- a/python-sdk/conftest.py
+++ b/python-sdk/conftest.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import random
 import string
 from copy import deepcopy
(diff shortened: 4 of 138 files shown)


python-sdk/example_dags/example_dataframe_api.py
```diff
@@ -41,7 +42,7 @@ def load_and_group_covid_data():
 
 
 @aql.dataframe(columns_names_capitalization="original")
-def find_worst_covid_month(dfs: List[pd.DataFrame]):
+def find_worst_covid_month(dfs: list[pd.DataFrame]):
```
pankajkoti, line 45: Did you have to make such changes explicitly or we were able to do this automatically?

pankajkoti (approved):
@feluelle @tatiana SQL CLI tests are failing with the below error: 
```FAILED tests/test_run_dag.py::test_run_dag_with_conn_id - AttributeError: 'Table' object has no attribute 'iloc'```
 Any idea?

python-sdk/example_dags/example_dataframe_api.py
pankajkoti: Looks like the test is failing only on this PR. It seems stable on other PRs and on main.

kaxil: I have rebased the PR, let's see :)

feluelle:
I don't get it - why is that failing?!

I can debug this tomorrow, because I think it is really useful to have.

feluelle:
Simply removing the `from __future__ import annotations` only from the sqlite_test.py DAG file fixes the issue. Continuing to debug the issue..

feluelle:
@kaxil Okay, I found out what the issues is. We basically are checking for the class type in [here](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/dataframe.py#L60). So we need the imported module at this time. If we use future annotations i.e. do a lazy import the types will be translated into `'pandas.Dataframe'`, but we need it as class type. So we either import it there(if not already imported) or we cannot use lazy importing.

tatiana:
@kaxil I have mixed feelings about this PR!
What are we gaining by adding this to all files? How often do we need lazy import? Wouldn't it be better to be more explicit about it - if and when we need it?

kaxil:
>What are we gaining by adding this to all files? How often do we need lazy import? Wouldn't it be better to be more explicit about it - if and when we need it?

There is absolutely no reason to not have lazy imports. We should not need to do if TYPE_CHECKING, it was the limitation of python earlier, but the devs are addressing the same issue.

I will see where I can get to, currently there are conflicts on the PR and tests failing

feluelle:
Yes, if it works it is a nice feature, but as you already noticed currently for some features we rely on direct imports.

tatiana:
Since the SQL CLI was discontinued, I'm closing this PR.

PR #1384, opened 2022-12-07
Redirect default airflow_home to tmp directory in tests

feluelle:
# Description

Currently, when testing with different versions of airflow locally, sometimes airflow still writes to the default airflow home directory. However, we are not using the default.

**Note:**
We are setting the `AIRFLOW_HOME` on project init, but not before that. Therefore any airflow import before that will write to the default airflow home directory.

## What is the new behavior?

Set the default airflow home to a new tmp directory. So that any call before calling project init writes to the tmp directory.

We also migrate all installs to explicitly use `poetry run` to ensure we are always installing to the correct venv. And we move typer env vars from workflow file to noxfile to be able to run nox locally with the same required env vars.

No.

### Checklist

### DIFF
--- a/sql-cli/noxfile.py
+++ b/sql-cli/noxfile.py
@@ -42,8 +42,8 @@ def test(session: nox.Session, airflow: str) -> None:
         )
         # We are duplicating the tests dependencies until we find a better solution.
         # The solution might be to move out of poetry.
-        dev = ("pytest", "pytest-cov", "mypy", "types-pyyaml")
-        session.run("poetry", "run", "pip", "install", "-e", ".", *dev, "-c", constraints_url)
+        dev_deps = ("pytest", "pytest-cov", "mypy", "types-pyyaml")
+        session.run("poetry", "run", "pip", "install", "-e", ".", *dev_deps, "-c", constraints_url)
         session.run("poetry", "run", "pip", "install", f"apache-airflow=={airflow}", "-c", constraints_url)
     else:
         session.run("poetry", "install", "--with", "dev")
--- a/sql-cli/noxfile.py
+++ b/sql-cli/noxfile.py
@@ -54,7 +54,11 @@ def test(session: nox.Session, airflow: str) -> None:
     session.log("Installed Dependencies:")
     session.run("poetry", "run", "pip", "freeze")
 
-    session.run("poetry", "run", "pytest", *session.posargs)
+    pytest_env = {
+        "TERMINAL_WIDTH": "3000",
+        "_TYPER_FORCE_DISABLE_TERMINAL": "1",
+    }
+    session.run("poetry", "run", "pytest", *session.posargs, env=pytest_env)
 
 
 @nox.session(python=["3.8"])
--- a/sql-cli/noxfile.py
+++ b/sql-cli/noxfile.py
@@ -25,11 +25,13 @@ def dev(session: nox.Session) -> None:
 def test(session: nox.Session, airflow: str) -> None:
     """Run both unit and integration tests."""
 
-    session.virtualenv.env = {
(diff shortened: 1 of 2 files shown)


sql-cli/noxfile.py
```diff
@@ -25,6 +25,12 @@ def dev(session: nox.Session) -> None:
 def test(session: nox.Session, airflow: str) -> None:
     """Run both unit and integration tests."""
 
+    # Airflow automatically creates the ~/airflow directory on importing airflow
+    # but we are not using this directory hence we change the home to a tmp directory to avoid side effects.
+    env = {
+        "AIRFLOW_HOME": session.create_tmp(),
+    }
```
feluelle, line 32:
```suggestion
    env = {
        # Airflow automatically creates the ~/airflow directory on importing airflow
        # but we are not using this directory hence we change the home to a tmp directory to avoid side effects.
        "AIRFLOW_HOME": session.create_tmp(),
    }
```

sql-cli/noxfile.py
```diff
@@ -25,37 +25,42 @@ def dev(session: nox.Session) -> None:
 def test(session: nox.Session, airflow: str) -> None:
     """Run both unit and integration tests."""
 
+    session.virtualenv.env.update(
+        {
+            # Airflow automatically creates the ~/airflow directory on importing airflow
+            # but we are not using this directory hence we change the home to a tmp directory to avoid side effects.
+            "AIRFLOW_HOME": session.create_tmp(),
```
tatiana, line 32: It's a shame we need to add this again :(

It would be great if we had some way of guaranteeing that when users are using the SQL CLI commands we're not depending on this home directory. Do you have any thoughts on this, @feluelle?

tatiana (approved):

sql-cli/noxfile.py
feluelle: I think it is possible to address this when using airflow as a library has been improved. The env var introduced in 2.4 is a great step in the right direction.

feluelle: Let me test using this in the `__init__.py` instead of in here.

feluelle: Init does not seem to be enough. It seems it needs to be set before installation 😅 or some other file/modules besides `__init__.py`.

sql-cli/noxfile.py
```diff
@@ -25,37 +25,42 @@ def dev(session: nox.Session) -> None:
...
-        session.run("poetry", "run", "pip3", "freeze")
-        session.run("poetry", "run", "pytest", *session.posargs)
+    session.log("Installed Dependencies:")
+    session.run("poetry", "run", "pip", "freeze")
+
+    pytest_env = {
+        "TERMINAL_WIDTH": "3000",
+        "_TYPER_FORCE_DISABLE_TERMINAL": "1",
+    }
+    session.run("poetry", "run", "pytest", *session.posargs, env=pytest_env)
```
pankajkoti, line 63: I like this section ❤️ 

I was wondering if we should modify the make target call nox instead of poetry run here: https://github.com/astronomer/astro-sdk/blob/main/sql-cli/Makefile#L34

so that we're consistent with the CI run and local run make target.

pankajkoti (approved):

sql-cli/noxfile.py
feluelle: Yes, I also thought about that, but which python version and airflow version should we choose?

pankajkoti: Python 3.9 and Airflow 2.5 ?  :) 
I think runtime image uses Python 3.9 so and observed that my poetry somehow has Airflow 2.5

feluelle: poetry will always use the latest version possible specified in `pyproject.toml`.

PR #1386, opened 2022-12-07
[sql-cli] Fix run_dag tests

feluelle:
# Description

Currently, when running sql-cli tests, we are not always setting up airflow - only in the case of e2e tests in `__main__.py`. But we have `test_run_dag.py` which relies on an airflow database to be set up.

## What is the new behavior?

Set up an airflow database for `test_run_dag.py` tests.

No.

### Checklist


feluelle:
Thank you @pankajkoti for your big help on identifying and fixing this issue ❤️

pankajkoti (approved):

Issue #1277, opened 2022-11-21
[sql-cli] Connection configuration having extra field causing tests to fail on Airflow 2.2.5


pankajkoti:
Upon adding the `bigquery` example connection in the configuration files, the `extra -> project` field is causing tests
to fail on Airflow 2.2.5. Upon removing this field the tests run fine.
We need to check if this happens for all the connections having extra field or it happens only for `bigquery` type connection.

pankajkoti:
Check whether the commented out config in dev configuration.yaml also needs to be updated .

PR #1388, opened 2022-12-08
Add working examples and tests for all supported connection types

feluelle:
# Description

Currently, we don't have any working example(s) of bigquery connections, because we thought the `extra` is broken. But apparently we did not configure it right.

**Prior to Airflow 2.3**, the `extra` has to be of type string and also in case of bigquery all of the extras needed to be prefixed with `extra__google_cloud_platform__`. Another example is snowflake where the `extra` needed to be a json as well to work for airflow 2.2.5 which we support.

closes: #1277

## What is the new behavior?

We add examples and tests for all currently supported connection types. Moreover configure the ci pipeline to be able to test them.

No.

### Checklist

### DIFF
--- a/.github/workflows/ci-sql-cli.yaml
+++ b/.github/workflows/ci-sql-cli.yaml
@@ -91,6 +91,7 @@ jobs:
             sql-cli/.nox
             .local/
           key: sql-cli-tests-os-${{ runner.os }}-python-${{ matrix.python }}-airflow-${{ matrix.airflow }}-deps-${{ hashFiles('sql-cli/poetry.lock') }}
+      - run: python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
       - run: pip3 install nox
       - run: nox -s "test-${{ matrix.python }}(airflow='${{ matrix.airflow }}')"
       - name: Upload coverage
@@ -99,6 +100,8 @@ jobs:
           name: coverage-${{ matrix.python }}-${{ matrix.airflow }}
           path: ./sql-cli/.coverage
     env:
+      GOOGLE_APPLICATION_CREDENTIALS_JSON: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS_JSON }}
+      GOOGLE_APPLICATION_CREDENTIALS: /tmp/google_credentials.json
       AIRFLOW__CORE__XCOM_BACKEND: astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend
       AIRFLOW__ASTRO_SDK__STORE_DATA_LOCAL_DEV: True
       TERMINAL_WIDTH: 3000
--- a/.github/workflows/ci-sql-cli.yaml
+++ b/.github/workflows/ci-sql-cli.yaml
@@ -46,6 +46,19 @@ jobs:
   Type-Check:
     if: github.event.action != 'labeled'
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: dimberman/pagila-test
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-sql-cli.yaml
```diff
@@ -26,6 +26,22 @@ concurrency:
...
   cancel-in-progress: true
 
+# This list should only have non-sensitive env vars
+# Env vars with secrets should be in the specific jobs
+env:
+  GOOGLE_CLOUD_PROJECT: astronomer-dag-authoring
+  POSTGRES_HOST: postgres
+  POSTGRES_PORT: 5432
+  REDSHIFT_DATABASE: dev
+  REDSHIFT_HOST: utkarsh-cluster.cdru7mxqmtyx.us-east-2.redshift.amazonaws.com
```
tatiana, line 36: Doesn't it feel a bit dodgy we have a team members' name in the cluster name :joy:

.github/workflows/ci-sql-cli.yaml
```diff
@@ -26,10 +26,39 @@ concurrency:
...
+  SNOWFLAKE_REGION: us-east-1
+  SNOWFLAKE_ROLE: AIRFLOW_TEST_USER
+
 jobs:
   Type-Check:
     if: github.event.action != 'labeled'
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: dimberman/pagila-test
```
tatiana, line 51: Just like here - I wonder if we could use the standard pagila image

.github/workflows/ci-sql-cli.yaml
feluelle: It does. But at least it is consistent 😅 We should change this for the python-sdk as well.

tatiana: Should we log a ticket for this?

tatiana (approved):
@feluelle would there be a way for us avoiding having issues with connection extras changes in future? Does Airflow have some strategy to verify if the users' extras are adequate for the hooks we're using ATM?

.github/workflows/ci-sql-cli.yaml
pankajkoti: but what if our team member wants to have a cluster on their name :D ?

sql-cli/tests/config/test/configuration.yml
```diff
@@ -1,9 +1,33 @@
...
+    password: $REDSHIFT_PASSWORD
+
+  - conn_id: snowflake_conn
+    conn_type: snowflake
+    host: $SNOWFLAKE_HOST
+    port: 443
+    login: $SNOWFLAKE_ACCOUNT_NAME
+    password: $SNOWFLAKE_PASSWORD
+    schema: $SNOWFLAKE_SCHEMA
+    extra: '{"account": "$SNOWFLAKE_ACCOUNT", "region": "$SNOWFLAKE_REGION", "role": "$SNOWFLAKE_ROLE", "warehouse": "$SNOWFLAKE_WAREHOUSE", "database": "$SNOWFLAKE_DATABASE"}'
```
pankajkoti, line 33: Can we also add an aws_conn ?

feluelle: For S3?

feluelle: Done caa0061

.github/workflows/ci-sql-cli.yaml
feluelle: For consistency, I kept the same image.

feluelle:
> @feluelle would there be a way for us avoiding having issues with connection extras changes in future? Does Airflow have some strategy to verify if the users' extras are adequate for the hooks we're using ATM?

@tatiana what do you mean exactly? We are using [test_connection](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#testing-connections) to verify that the connections works, in case it doesn't the plan of [this issue](https://github.com/astronomer/astro-sdk/issues/1040) is to also print the error message to the user. Sometimes it is useful, sometimes not :/

Issue #1329, opened 2022-11-30
Expose Astro CLI .env environment variables to the SQL CLI 


tatiana:
**Please describe the feature you'd like to see**
When running the SQL CLI from the Astro CLI, we do not have access to .env variables because they are not mounted in the docker container.

**Describe the solution you'd like**
To expose the content of .env in the SQL CLI docker container when it is run from Astro CLI.

**Are there any alternatives to this feature?**
To stop using Docker to integrate between Astro CLI and SQL CLI.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1391, opened 2022-12-08
Create .env file on project initialisation + important refactorings (read desc)

pankajkoti:
# Description

Currently, there is no way for Astro CLI users to specify environment
variables that can be used for configuring connections as we launch
docker containers from Astro CLI commands which do not have access
to host environment variables

related: #1329 

## What is the new behavior?

As part of project initialisation, we create a .env file in the initialised project and
ask the user to configure environment variables in this file. Astro CLI will mount
this .env file in the docker containers so that those environment variables are 
available in the docker container since we use `dotenv` library to load 
environment variables from .env file. Such environment variables can be used
to expand yaml files to create connections including secrets as environment variables.
e.g. https://legacy.docs.greatexpectations.io/en/latest/guides/how_to_guides/configuring_data_contexts/how_to_use_a_yaml_file_or_environment_variables_to_populate_credentials.html

Remove transform env specific config implementation and move the
needed relevant connection transformation into the convert_to_connection
method.

Remove the method to write config to yaml as it overwrites environment
vars placeholder with the expanded vars when read into config instance.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .git
 coverage.xml
-.env
 **/.coverage*
 **/.python-version
 python-sdk/integration-tests/test_data*
@@ -58,6 +57,7 @@ sql-cli/tests/.airflow/dags/*
 sql-cli/include/base/.airflow/dags/*
 !sql-cli/include/base/.airflow/dags/include/.gitkeep
 !sql-cli/include/base/.airflow/dags/.gitkeep
+!/sql-cli/include/base/.env
 
 # Airflow
 **/.airflowignore
--- a/sql-cli/sql_cli/configuration.py
+++ b/sql-cli/sql_cli/configuration.py
@@ -50,7 +50,7 @@ class Config:
         :returns: Content of the YAML configuration file as a python dictionary.
         """
         load_dotenv(self.project_dir / ".env")
-        with open(filepath) as fp:
+        with filepath.open() as fp:
             yaml_with_env = os.path.expandvars(fp.read())
             yaml_config = yaml.safe_load(yaml_with_env)
         return yaml_config or {}
--- a/sql-cli/sql_cli/configuration.py
+++ b/sql-cli/sql_cli/configuration.py
@@ -7,23 +7,9 @@ from typing import Any
 
 import yaml
 from airflow.models.connection import Connection
+from dotenv import load_dotenv
 
 from sql_cli.constants import CONFIG_DIR, CONFIG_FILENAME, GLOBAL_CONFIG
-
-
-def convert_to_connection(conn: dict[str, Any]) -> Connection:
-    """
-    Convert the SQL CLI connection dictionary into an Airflow Connection instance.
-
-    :param conn: SQL CLI connection dictionary
-    :returns: Connection object
-    """
-    from airflow.api_connexion.schemas.connection_schema import connection_schema
-
-    c = conn.copy()
-    c["connection_id"] = c["conn_id"]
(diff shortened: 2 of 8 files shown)


feluelle (dismissed):

sql-cli/sql_cli/configuration.py
```diff
@@ -7,25 +7,11 @@
 
 import yaml
 from airflow.models.connection import Connection
+from dotenv import load_dotenv
 
 from sql_cli.constants import CONFIG_DIR, CONFIG_FILENAME, GLOBAL_CONFIG
 
 
-def convert_to_connection(conn: dict[str, Any]) -> Connection:
```
pankajkoti, line 14: Moved this method to `connections.py` module

sql-cli/sql_cli/configuration.py
```diff
@@ -55,16 +41,16 @@ def get_global_config_filepath(self) -> Path:
...
+    def from_yaml_to_dict(self, filepath: Path) -> dict[str, Any]:
         """
         Return a dict with the contents of the given configuration.yaml
 
         :param filepath: Path of the desired configuration.yaml to read contents from.
 
         :returns: Content of the YAML configuration file as a python dictionary.
         """
-        with open(filepath) as fp:
+        load_dotenv(self.project_dir / ".env")
```
pankajkoti, line 52: load environment variables from the initialised project's .env file to expand the YAML with env vars's values in the subsequent lines.

sql-cli/sql_cli/configuration.py
```diff
@@ -102,32 +88,3 @@ def write_value_to_yaml(self, section: str, key: str, value: str, filepath: Path
 
         with filepath.open(mode="w") as fp:
             yaml.dump(yaml_config, fp)
-
-    def write_config_to_yaml(self) -> None:
```
pankajkoti, line 106: We're removing this as I believe we should not be doing this as it has side effects. E.g. When the config object is loaded from YAML, with loading the YAML with expand the env vars to replace the placeholders with their values. Now, in our existing implementation, we call this method to write such config instance back to the YAML which has a side effect of replacing the env var placeholders with the actual values that the Config instance holds.

sql-cli/sql_cli/project.py
```diff
@@ -99,30 +94,6 @@ def _initialise_global_config(self) -> None:
             "airflow", "dags_folder", str(self._airflow_dags_folder.resolve()), global_env_filepath
         )
 
-    def transform_env_config(self, environment: str = DEFAULT_ENVIRONMENT) -> None:
```
pankajkoti, line 102: At the moment this method was only expanding host paths for SQLite connection and persisting back the expanded host path to the YAML. Persisting back is risky. Instead we should resolve this every time in the config instance alone without persisting back to the YAML and the logic to transform the connection to resovle host path has been moved to the `connections.py` module's method `convert_to_connection`

.gitignore
```diff
@@ -58,6 +57,7 @@ sql-cli/tests/.airflow/dags/*
 sql-cli/include/base/.airflow/dags/*
 !sql-cli/include/base/.airflow/dags/include/.gitkeep
 !sql-cli/include/base/.airflow/dags/.gitkeep
+!/sql-cli/include/base/.env
```
tatiana, line 60: I really liked the double negation!

tatiana (approved):
Thanks for addressing the concerns from yesterday's discussion, @pankajkoti !

PR #1393, opened 2022-12-08
Integration test should count rows for in Operators

### LINKED ISSUE #1273: Integration test should count rows for any database operation on Operators
**Describe the bug**
- Check the row count, metadata, column names as part of test
- Check and fix the integration test accordingly for all the operators

Operators affected:
1. Append
2. Merge
3. Transform
4. Dataframe
5. LoadFile

Tests affected:
Start with the data frame operator.

rajaths010494:
**Describe the bug**
- Check the row count, metadata, column names as part of test
- Check and fix the integration test accordingly for the operators

closes #1273

### DIFF
--- a/python-sdk/tests/sql/operators/test_merge.py
+++ b/python-sdk/tests/sql/operators/test_merge.py
@@ -198,7 +198,7 @@ def test_merge_with_the_same_schema(database_table_fixture, multiple_tables_fixt
             {"id": 4, "name": "Czwarte imię"},
         ]
     )
-
+    assert first_table.row_count == 4
     test_utils.assert_dataframes_are_equal(computed, expected)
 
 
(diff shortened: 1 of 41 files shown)


python-sdk/tests/sql/operators/test_merge.py
```diff
@@ -198,7 +198,7 @@ def test_merge_with_the_same_schema(database_table_fixture, multiple_tables_fixt
             {"id": 4, "name": "Czwarte imię"},
         ]
     )
-
+    assert first_table.row_count == 4
```
tatiana, line 201: @rajaths010494 in this particular test, we are comparing if the values match the expected values (so the content of all rows and columns) - `test_utils.assert_dataframes_are_equal`. Which benefit do you think we gain by adding this additional assert?

tatiana (approved):

python-sdk/tests/sql/operators/test_merge.py
rajaths010494: this is to check in first table in the database for the merge operation has the data. As you mentioned we export the table and validate it with dataframe.

rajaths010494:
Having issues with merge since integrations have been separated. 
https://github.com/astronomer/astro-sdk/pull/1405

PR #1397, opened 2022-12-08
Add the ability to interact between local dataframes and Delta tables…

dimberman:
… in databricks

# Description
Currently there is no established path for moving a dataframe into a databricks delta table except to manually save that dataframe as a local file and then push that file into delta using aql.load_file. 

## What is the new behavior?
This PR allows users to pass local dataframes into delta files using all of the existing methods withing the astro SDK (e.g. passing it into a transform function). Users should not need to learn any new functionality as all of this will work out of the box with existing aql based DAGs.

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -389,6 +389,21 @@ class BaseDatabase(ABC):
                 columns_names_capitalization=columns_names_capitalization,
             )
 
+    def fetch_all_rows(self, table: BaseTable, row_limit: int = -1) -> list:
+        """
+        Fetches all rows for a table and returns as a list. This is needed because some
+        databases have different cursors that require different methods to fetch rows
+
+        :param row_limit: Limit the number of rows returned, by default return all rows.
+        :param table: The table metadata needed to fetch the rows
+        :return: a list of rows
+        """
+        statement = f"SELECT * FROM {self.get_table_qualified_name(table)}"
+        if row_limit > 0:
+            statement = statement + f" LIMIT {row_limit}"
+        response = self.run_sql(statement)
+        return response.fetchall()  # type: ignore
+
     def load_file_to_table(
         self,
         input_file: File,
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -399,7 +399,7 @@ class BaseDatabase(ABC):
         :return: a list of rows
         """
         statement = f"SELECT * FROM {self.get_table_qualified_name(table)}"
-        if row_limit > 0:
+        if row_limit > -1:
             statement = statement + f" LIMIT {row_limit}"
         response = self.run_sql(statement)
         return response.fetchall()  # type: ignore
--- a/python-sdk/src/astro/databricks/delta.py
+++ b/python-sdk/src/astro/databricks/delta.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import tempfile
+import uuid
 from textwrap import dedent
 
-import pandas
+import pandas as pd
 from airflow.providers.databricks.hooks.databricks import DatabricksHook
 from airflow.providers.databricks.hooks.databricks_sql import DatabricksSqlHook
 from databricks.sql.client import Cursor
@@ -14,6 +16,7 @@ from astro.constants import DEFAULT_CHUNK_SIZE, ColumnCapitalization, LoadExistS
 from astro.databases.base import BaseDatabase
 from astro.databricks.load_file.load_file_job import load_file_to_delta
 from astro.databricks.load_options import DeltaLoadOptions
+from astro.dataframes.pandas import PandasDataframe
 from astro.files import File
 from astro.options import LoadOptions
 from astro.table import BaseTable, Metadata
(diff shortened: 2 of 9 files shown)


python-sdk/tests/databases/test_all_databases.py
```diff
@@ -14,6 +15,15 @@
 CWD = pathlib.Path(__file__).parent
 
 
+def fetch_result(database: BaseDatabase, statement):
+    if database.sql_type == "delta":
+        rows = database.run_sql(statement, handler=lambda x: x.fetchall())
```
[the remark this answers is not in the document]
dimberman: @kaxil fixed

python-sdk/tests_integration/databases/test_all_databases.py
```diff
@@ -153,50 +190,9 @@ def test_load_pandas_dataframe_to_table_with_replace(database_table_fixture):
...
-            "table": Table(metadata=Metadata(schema=SCHEMA)),
-        },
-        {"database": Database.SQLITE, "table": Table()},
-    ],
-    indirect=True,
-    ids=["bigquery", "postgres", "snowflake", "sqlite"],
-)
-@mock.patch("astro.files.base.File.export_to_dataframe")
-@mock.patch("astro.files.base.File.export_to_dataframe_via_byte_stream")
-def test_export_to_dataframe_via_byte_stream_is_called_for_postgres(
```
kaxil, line 189: Why did we delete this?

kaxil (approved):

python-sdk/src/astro/databases/base.py
```diff
@@ -389,6 +389,18 @@ def create_schema_and_table_if_needed(
                 columns_names_capitalization=columns_names_capitalization,
             )
 
+    def fetch_all_rows(self, table: BaseTable) -> list:
```
tatiana, line 392: Couldn't it be risky to return all the rows of a specific table?
What if, by default, we had a maximum number of rows - which users could configure?

python-sdk/src/astro/databricks/delta.py
```diff
@@ -100,6 +102,17 @@ def create_schema_if_needed(self, schema: str | None) -> None:
         # Schemas do not need to be created for delta, so we don't need to do anything here
         return None
 
+    def fetch_all_rows(self, table: BaseTable) -> list:
```
tatiana, line 105: Same comment as in the base table - should we have an upper bound here?

python-sdk/tests_integration/databases/test_all_databases.py
```diff
@@ -66,6 +71,45 @@ def test_export_table_to_pandas_dataframe(
...
+        },
+        {"database": Database.SQLITE, "table": Table()},
+        {"database": Database.DELTA, "table": Table()},
+    ],
+    indirect=True,
+    ids=["bigquery", "postgres", "snowflake", "sqlite", "delta"],
+)
+@mock.patch("astro.files.base.File.export_to_dataframe")
+@mock.patch("astro.files.base.File.export_to_dataframe_via_byte_stream")
+def test_export_to_dataframe_via_byte_stream_is_called_for_postgres(
```
tatiana, line 97: it's probably better to rename this test since it is not Postgres-specific.

python-sdk/tests_integration/databases/test_all_databases.py
tatiana: It was just moved, @kaxil !

kaxil: aah although why :D ?

python-sdk/src/astro/databricks/delta.py
dimberman: I don't think having an upper bound is a good idea as it will create false information for users, but I can see having upper_bound as an option. As to the risk of users loading too much data, I think the function is pretty self explanatory and they should consider the risks (especially since the Database object isn't really exposed to users unless they look for it).

python-sdk/src/astro/databases/base.py
dimberman: @tataiana I added a row limit option

python-sdk/tests_integration/databases/test_all_databases.py
dimberman: @kaxil this probably happened during a merge from main. I'll move it back tomorrow morning

tatiana (approved):
Thanks @dimberman, I'm happy for this to be merged once:
* row limit is -1 by default (and that is when it is unlimited)
* we remove Postgres from the test name

python-sdk/src/astro/databases/base.py
```diff
@@ -389,15 +389,18 @@ def create_schema_and_table_if_needed(
                 columns_names_capitalization=columns_names_capitalization,
             )
 
-    def fetch_all_rows(self, table: BaseTable) -> list:
+    def fetch_all_rows(self, table: BaseTable, row_limit: int = 0) -> list:
```
kaxil, line 392: `row_limit` on the function called `fetch_all_rows` doesn't make sense to me so I agree with @dimberman , the function is self explanatory and needed for internal usage only

kaxil: but that's fine, feel free to merge @dimberman

Issue #1379, opened 2022-12-07
Enhance dataframe operator to insert into existing table


phanikumv:
**Please describe the feature you'd like to see**
Find a way to insert into an existing table (output table) from a Dataframe? It seems that `if_exists` is always set to replace: https://github.com/astronomer/astro-sdk/blob/1.3.0/python-sdk/src/astro/sql/operators/dataframe.py#L188

Refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1670370753724899

**Describe the solution you'd like**
A clear and concise description of what you want to happen, if possible with code examples.

**Are there any alternatives to this feature?**
Is there another way we could solve this problem or enable this use-case?

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1399, opened 2022-12-09
Add append strategy for Dataframe operator

sunank200:
# Description

Insert into an existing table (output table) from a Dataframe? It seems that if_exists is always set to replace: https://github.com/astronomer/astro-sdk/blob/1.3.0/python-sdk/src/astro/sql/operators/dataframe.py#L188

Refer to https://astronomer.slack.com/archives/C02B8SPT93K/p1670370753724899

closes: #1379

## What is the new behaviour?

- Introduce a parameter `if_exists" 
- Add tests for append

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/dataframe.py
+++ b/python-sdk/src/astro/sql/operators/dataframe.py
@@ -120,7 +120,7 @@ class DataframeOperator(AstroSQLBaseOperator, DecoratedOperator):
         database: str | None = None,
         schema: str | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
-        if_exists: LoadExistStrategy | None = "replace",
+        if_exists: LoadExistStrategy = "replace",
         **kwargs,
     ):
         self.conn_id: str = conn_id or ""
@@ -135,7 +135,7 @@ class DataframeOperator(AstroSQLBaseOperator, DecoratedOperator):
             self.output_table = None
         self.op_args = self.kwargs.get("op_args", ())  # type: ignore
         self.columns_names_capitalization = columns_names_capitalization
-        self.if_exists = if_exists if if_exists else "replace"
+        self.if_exists = if_exists
 
         # We purposely do NOT render upstream_tasks otherwise we could have a case where a user
         # has 10 dataframes as upstream tasks and it crashes the worker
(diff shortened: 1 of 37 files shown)


python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -119,6 +120,7 @@ def __init__(
         database: str | None = None,
         schema: str | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
+        if_exists: LoadExistStrategy | None = "replace",
```
kaxil, line 123:
```suggestion
        if_exists: LoadExistStrategy = "replace",
```

python-sdk/src/astro/sql/operators/dataframe.py
```diff
@@ -133,6 +135,7 @@ def __init__(
             self.output_table = None
         self.op_args = self.kwargs.get("op_args", ())  # type: ignore
         self.columns_names_capitalization = columns_names_capitalization
+        self.if_exists = if_exists if if_exists else "replace"
```
kaxil, line 138:
```suggestion
        self.if_exists = if_exists
```

kaxil: Will it ever return `None`?

python-sdk/src/astro/sql/operators/dataframe.py
sunank200: Wouldn't it break if the user doesn't specify `if_exists` for users in the older version of Astro-SDK?

python-sdk/src/astro/sql/operators/dataframe.py
sunank200: Wouldn't it break if the user doesn't specify `if_exists` for users in the older version of Astro-SDK? that's the reason I kept it optional

python-sdk/src/astro/sql/operators/dataframe.py
kaxil: `"replace"` is already there as the default

python-sdk/src/astro/sql/operators/dataframe.py
kaxil: Can you give an example of that please?

python-sdk/src/astro/sql/operators/dataframe.py
sunank200: Yes, I tired it locally now. While using annotations, even when the value is not passed at all it picks up the default value.

kaxil (approved):

PR #1403, opened 2022-12-09
Separate Unit tests and Integration tests

kaxil:
This has bugged me from a long time where I was not able to run unit tests without creds and there was no clear separation between unit tests and integration tests.

This PR does that.

Now you can simply run unit tests without creds locally using the following:

```python
❯ pytest tests/
....
(29 durations < 1s hidden.  Use -vv to show these durations.)
=================================================================================== 268 passed, 6 skipped, 176 warnings in 10.12s ====================================================================================
```
As you can see all the unit tests were run in under 11 seconds

I have planned few more follow up PRs to run them separately on the CI too. There are lot of integration tests that do "duplicate testing" which can be overtime moved to unit test by mocking and testing properly


utkarsharma2 (approved):

tatiana:
@kaxil did you consider using the marker integration? We had started with this approach - some of the integration tests were marked using it, and it may have involved less changes.

kaxil:
> @kaxil did you consider using the marker integration? We had started with this approach - some of the integration tests were marked using it, and it may have involved less changes.

I wanted a clear separation of how the tests are setup, run and make it easily discoverable. For example, unit tests should not need the `test_connectsion.yaml` file nor should it need other secrets that we have configured as GitHub secrets.

We easily forget to add markers to all the integration tests too, now it should be easy and if we forget the CI will fail because it won't have creds.

PR #1404, opened 2022-12-09
Add load to spark gcs

dimberman:
# Description

## What is the new behavior?
This PR adds the ability for GCP users to load GCS files to delta for processing in databricks automatically. It is worth noting that this is optional, as users can turn off secret loading and manually add GCS settings to their databricks cluster using [this guide](https://docs.gcp.databricks.com/external-data/gcs.html) if they prefer. The main benefit here is that users do not need to set these settings ahead of time.

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/databricks/load_file/jinja_templates/load_secrets.py.jinja2
+++ b/python-sdk/src/astro/databricks/load_file/jinja_templates/load_secrets.py.jinja2
@@ -3,6 +3,7 @@ secret_key_list = dbutils.secrets.list("{{ secret_scope }}")
 for secret_key in secret_key_list:
     if 'astro_sdk_' in secret_key.key:
         key_name = secret_key.key.replace("astro_sdk_","")
-        logger.info(f"setting {key_name}")
+        # We are using print here as we have yet to find a way to make logs surface in the databricks UI
+        print(f"setting {key_name}")
         sc._jsc.hadoopConfiguration().set(key_name, dbutils.secrets.get("{{ secret_scope }}", secret_key.key))
 {%- endmacro -%}
--- a/python-sdk/src/astro/files/locations/google/gcs.py
+++ b/python-sdk/src/astro/files/locations/google/gcs.py
@@ -1,11 +1,32 @@
 from __future__ import annotations
 
+import json
+import os
 from urllib.parse import urlparse, urlunparse
 
 from airflow.providers.google.cloud.hooks.gcs import GCSHook
 
 from astro.constants import FileLocation
 from astro.files.locations.base import BaseFileLocation
+
+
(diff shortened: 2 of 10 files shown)


python-sdk/src/astro/files/locations/google/gcs.py
```diff
@@ -43,6 +43,21 @@ def size(self) -> int:
             object_name = object_name[1:]
         return int(self.hook.get_size(bucket_name=bucket_name, object_name=object_name))
 
+    def databricks_settings(self) -> dict:
```
kaxil, line 46: Should this be changed to `databricks_auth_settings` on all file locations?

dimberman: @kaxil I like this idea.

dimberman:
@kaxil @tatiana I'm not sure what's going on with codecov. I've explicitly made tests for the lines it is claiming are untested.

python-sdk/src/astro/databricks/load_file/jinja_templates/load_secrets.py.jinja2
```diff
@@ -3,6 +3,6 @@ secret_key_list = dbutils.secrets.list("{{ secret_scope }}")
 for secret_key in secret_key_list:
     if 'astro_sdk_' in secret_key.key:
         key_name = secret_key.key.replace("astro_sdk_","")
-        logger.info(f"setting {key_name}")
+        print(f"setting {key_name}")
```
tatiana, line 6: Please, could you add a comment on why we're using print here :)

python-sdk/src/astro/files/locations/google/gcs.py
```diff
@@ -43,6 +63,23 @@ def size(self) -> int:
             object_name = object_name[1:]
         return int(self.hook.get_size(bucket_name=bucket_name, object_name=object_name))
 
+    def databricks_auth_settings(self) -> dict:
+        """
+        Required settings to upload this file into databricks. Only needed for cloud storage systems
+        like S3
+        :return: A dictionary of settings keys to settings values
+        """
+        creds_dict = self.hook.get_connection(self.conn_id).extra_dejson
```
tatiana, line 72: @dimberman, out of curiosity, did you try this with the google default connection when the user is not using a service account JSON?

tatiana (approved):
I'm happy for this PR to be merged once the tests and checks are passing!

python-sdk/src/astro/files/locations/google/gcs.py
dimberman: Yes @kaxil and I looked into it. I have added more detailed documentation of what requirements people will need to use this feature.

Issue #1315, opened 2022-11-28
Be able to run SQL CLI workflows on Airflow / Astro CLI locally


tatiana:
**Please describe the feature you'd like to see**
As of SQL CLI 0.2, all the workflows rely on SQLite. We mount those during `astro flow` commands but are not mounted during `astro dev start`. This means users cannot have an end-to-end journey using the SQL CLI / Astro CLI, where they'd visualise the workflow they built with SQL.

**Describe the solution you'd like**
It would be great to find a strategy so users can visualise/run example DAGs in Astro CLI - until we define the documentation journey.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

feluelle:
What astro-cli already has:
* `$AIRFLOW_HOME/include` directory for mounting additional directories/files

What flow depends on:
* data directory for local file (sqlite) dbs

One option could be to copy the flow projects into the include directory:

This could be done using something like:
```
astro dev start --include path/to/flow_project_1,another_path/to/flow_project_2
```
which would copy to astro's `$AIRFLOW_HOME/include` (natively). We create a new feature for copying on demand directories/files.

OR by using something like:
```
cp path/to/flow_project_1/data ./include/flow_project_1_data
cp path/to/flow_project_2/data ./include/flow_project_2_data
```
executed in the astro-cli project directory (non-natively). Note we only need `data` directory here, because we can be explicit as we control where the files/directories lands.

Then we create connections in airflow (as a starting point manually) and set the host path accordingly i.e. to where we have moved the files to.

pankajkoti:
Another option suggested by Tatiana:
We accept another parameter during projection initialisation that the user can specify as their example data directory.

[...]

PR #1407, opened 2022-12-12
Add configurable data dir to project init

feluelle:
# Description

Currently, users are unable to use a different data directory than the default.

related: #1315

## What is the new behavior?

Users, from now on, can choose a different data directory.

No.

### Checklist


sql-cli/sql_cli/connections.py
```diff
@@ -6,39 +6,36 @@
...
 
 
-def convert_to_connection(conn: dict[str, Any], project_dir: Path | None) -> Connection:
+def convert_to_connection(conn: dict[str, Any], data_dir: Path) -> Connection:
     """
     Convert the SQL CLI connection dictionary into an Airflow Connection instance.
 
     :param conn: SQL CLI connection dictionary
-    :param project_dir: Path to the initialised project
+    :param data_dir: Path to the initialised project data directory
```
pankajkoti, line 19:
```suggestion
    :param data_dir: Path to the initialised project's data directory
```

pankajkoti (commented):
Looks good to me, waiting for PR to be ready for review, and have minor suggestions.

sql-cli/sql_cli/project.py
```diff
@@ -108,12 +128,21 @@ def initialise(self) -> None:
...
-            ignore=shutil.ignore_patterns(".gitkeep"),
+            ignore=shutil.ignore_patterns(*excludes),
             dirs_exist_ok=True,
         )
+        if self.data_dir != Path(self.directory, DEFAULT_DATA_DIR):
+            shutil.copytree(
+                src=BASE_SOURCE_DIR / DEFAULT_DATA_DIR,
+                dst=self.data_dir,
+                dirs_exist_ok=True,
+            )
```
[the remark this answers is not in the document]
feluelle: I checked again and it seems we can now. :) I tried it at first using a different copytree config which didn't work.

sql-cli/sql_cli/configuration.py
```diff
@@ -68,6 +69,7 @@ def from_yaml_to_config(self) -> Config:
             environment=self.environment,
             airflow_home=global_yaml_config.get("airflow", {}).get("home"),
             airflow_dags_folder=global_yaml_config.get("airflow", {}).get("dags_folder"),
+            data_dir=global_yaml_config.get("general", {}).get("data_dir"),
```
tatiana, line 72: Do we want the `data_dir` to be environment specific or global? It may simplify to have it global, as a starting point

pankajkoti (commented):
It just occurred to me now that we will need to expose the data dir via `config` typer command, call that config command in Astro CLI to mount the data dir on the docker containers to run, right?

sql-cli/include/base/config/global.yml
```diff
@@ -1,3 +1,5 @@
 airflow:
   dags_folder:
   home:
+general:
+  data_dir:
```
dimberman, line 5: ohhh I like this!

sql-cli/sql_cli/__main__.py
```diff
@@ -251,11 +251,17 @@ def init(
         help=f"(Optional) Set the DAGs Folder. Default: {DEFAULT_DAGS_FOLDER}",
         show_default=False,
     ),
+    data_dir: Path = typer.Option(
+        None,
+        dir_okay=True,
+        help=f"(Optional) Set the data directory. Default: {DEFAULT_DATA_DIR}",
```
dimberman, line 257: Can you please add more information here about what a "data directory" is? I might not immediately know if I'm just playing around with the CLI (and we can never fully depend on people to read the docs ;) )

sql-cli/sql_cli/configuration.py
dimberman: I think global is fine to start. We can make it environmental later if that matches user requests/usage patterns.

dimberman (dismissed):

sql-cli/sql_cli/__main__.py
feluelle: Thanks. Yes, will do! 👍

feluelle:
Yes @pankajkoti thanks for the reminder. Done in https://github.com/astronomer/astro-cli/pull/972.

pankajkoti (approved):

PR #1409, opened 2022-12-12
Move file config/global/configuration.yml -> config/global.yml

pankajkoti:
Based on previous discussions and observations from users, it seems that having a global directory creates confusion because global itself is not any environment and the directories within config directory are meant to be environments. So, we are no longer creating the global directory as part of the initialised project and moving the global configuration to top level global.yml file in the config directory.

related: #1230


sql-cli/tests/config/test/configuration.yml
```diff
@@ -22,7 +22,7 @@ connections:
     schema: pagila
     login: postgres
     password: $POSTGRES_PASSWORD
-    port: $POSTGRES_PORT
+    port: 5432
```
tatiana, line 25: Is this change intended, @pankajkoti ?

tatiana (approved):

sql-cli/tests/config/test/configuration.yml
pankajkoti: yes, the change is intentional. A validate test was failing on my local saying not a valid integer.

feluelle: @pankajkoti this seems to be an issue to your local environment/setup. I am not having the issue and using `POSTGRES_PORT=5433` in my `.env` file. Now I am unable to use the same db. I would be in favor of reverting this change, if that's okay for you both? @tatiana

sql-cli/sql_cli/project.py
```diff
@@ -82,7 +82,7 @@ def _initialise_global_config(self) -> None:
         Initialises global config file that includes configuration to be shared across environments including the
         airflow config.
         """
-        config = Config(environment=GLOBAL_CONFIG, project_dir=self.directory)
+        config = Config(environment=DEFAULT_ENVIRONMENT, project_dir=self.directory)
```
feluelle, line 85: @pankajkoti why is `environment=DEFAULT_ENVIRONMENT` for the "global" configuration?

sql-cli/tests/config/test/configuration.yml
pankajkoti: It is an issue if the environment variable in not set

sql-cli/sql_cli/project.py
pankajkoti: The Config instance has environment as a required key. We can set it to anything as it does not have any effect for the global config keys.

feluelle: > We can set it to anything as it does not have any effect

^this is huge code smell in my opinion.

sql-cli/tests/config/test/configuration.yml
pankajkoti: okay, will add back the env var.

PR #1412, opened 2022-12-12
Deprecate `export_file` before renaming it `export_table_to_file`

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1411

we have the `export_file` operator but we want to rename it to `export_table_to_file` operator since this is the last patch release before 1.4 so lets deprecate it in 1.3.1

## What is the new behavior?
Deprecate export_file before renaming it export_table_to_file

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/export_file.py
+++ b/python-sdk/src/astro/sql/operators/export_file.py
@@ -1,21 +1,18 @@
 from __future__ import annotations
 
+import warnings
 from typing import Any
 
 import pandas as pd
-from airflow.decorators.base import get_unique_task_id
 from airflow.models.xcom_arg import XComArg
 
-from astro.airflow.datasets import kwargs_with_datasets
 from astro.constants import ExportExistsStrategy
-from astro.databases import create_database
 from astro.files import File
-from astro.sql.operators.base_operator import AstroSQLBaseOperator
-from astro.table import BaseTable, Table
-from astro.utils.typing_compat import Context
+from astro.sql.operators.export_table_to_file import ExportTableToFileOperator, export_table_to_file
+from astro.table import BaseTable
 
 
-class ExportFileOperator(AstroSQLBaseOperator):
+class ExportFileOperator(ExportTableToFileOperator):
     """Write SQL table to csv/parquet on local/S3/GCS.
 
     :param input_data: Table to convert to file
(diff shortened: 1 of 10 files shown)


sunank200 (commented):
What about `export_file` decorator? Should we not just keep the warning in the decorator?

Also we should change the test correspondingly

pankajastro:
> What about `export_file` decorator? Should we not just keep the warning in the decorator?

@sunank200 I have added a warning here https://github.com/astronomer/astro-sdk/pull/1412/files#diff-1ac9e72fc085a7cadda08023ad72a39a6ed1353f12613688f22d0427be823305R75 isn't it ok or I'm missing something?

tatiana (approved):

python-sdk/tests/sql/operators/test_export_file.py
```diff
@@ -36,6 +36,22 @@ def make_df():
     assert df.equals(pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]}))
 
 
+def test_raise_exception_for_invalid_input_type():
```
utkarsharma2, line 39: @pankajastro Can we add a test case to test the warning message?

pankajastro: Added 62bd5b71dc75cb6fc10225e1ff9392deeceab277

pankajastro:
Merging it, and looking into code coverage part separately

python-sdk/src/astro/sql/operators/export_file.py
```diff
@@ -170,12 +72,14 @@ def export_file(
...
-        task_id=task_id,
-        output_file=output_file,
-        input_data=input_data,
-        if_exists=if_exists,
-        **kwargs,
-    ).output
+    warnings.warn(
+        """This decorator is deprecated.
+        Please use `astro.sql.operators.export_table_to_file.export_table_to_file`.
+        And, will be removed in astro-sdk-python>=1.4.0.""",
```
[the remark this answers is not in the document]
pankajastro: make sense fix the deprecation message https://github.com/astronomer/astro-sdk/pull/1424

pankajastro:
> Merging it, and looking into code coverage part separately

In this branch, the codecov reported value was incorrect. maybe we have merged some PR which is not fully green. the main branch code coverage is around 90% but the reported here was  97. https://github.com/astronomer/astro-sdk/commit/bf608683fef5588e6b7417d7a29a98f7c46c4198

Issue #1369, opened 2022-12-06
Tmp Schemas need to account for Location/Region when orchestrating BigQuery


ReadytoRocc:
**Describe the bug**
I am unable to run queries using the `astro.sql.transform` decorator when they do not exist in the region where `tmp_astro` is created. In the below example error, I am running a query on a table in `us-central1`, when `tmp_astro` exists in `europe-west2`:

```
sqlalchemy.exc.DatabaseError: (google.cloud.bigquery.dbapi.exceptions.DatabaseError) 404 Not found: Dataset PROJECT_ID:tmp_astro was not found in location us-central1
```

**Version**
* Astro: 1.3.0
* OS: Debian

**To Reproduce**

- Run the below DAG on using a TABLE_PATH in one region.
- Then rerun it again using a different table path in another region.

```
"""
gcp_bq
DAG auto-generated by Astro Build.
"""

from airflow.decorators import dag
from astro import sql as aql
from astro.sql.table import Table
import pendulum


@aql.transform(conn_id="bq_fe_sa", task_id="cell_2")
def cell_2_func():
    return """SELECT * FROM TABLE_PATH LIMIT 10;"""

@dag(
    schedule_interval="0 0 * * *",
    start_date=pendulum.from_format("2022-12-02", "YYYY-MM-DD").in_tz("America/Phoenix"),
)
def gcp_bq():
    cell_2 = cell_2_func()

dag_obj = gcp_bq()
```

**Expected behavior**
I would expect the Astro SDK to make a `tmp_astro` dataset per region, and ensure that it is using/refencing the appropriate tmp schema based on the source table in the query.

**Additional context**
The BigQueryHook supports a [location parameter](https://github.com/apache/airflow/blob/providers-google/8.6.0/airflow/providers/google/cloud/hooks/bigquery.py#L95), allowing you to specify the location/region for your query. I don't believe this can be set via a Connection parameter.

PR #1416, opened 2022-12-13
Create region specific datasets for bigquery

utkarsharma2:
# Description
Currently, we are assuming all the bigquery datasets to be in the default region - `US`
If the user tries to access a preexisting table in a different region say, `europe-west6` all the Temp tables will try to be created in `US` region, which will result in the error. 

closes: #1369

## What is the new behavior?
1. The solution is to create `tmp_astro` region specific. But we cannot do that since `<project_id>.<dataset_id>` is unique. So we are creating a unique dataset id like -  `<dataset_id>__<region_id>`, this region_id.
2.  Added a default setting to select the Bigquery default region.
3.  Added `location` attribute to Metadata class. 

Nope

### Checklist


python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -183,6 +212,20 @@ def load_pandas_dataframe_to_table(
...
         )
 
+    def create_schema_if_needed(self, schema: str | None, location: str | None = None) -> None:
+        """
+        This function checks if the expected schema exists in the database. If the schema does not exist,
+        it will attempt to create it.
+
+        :param schema: DB Schema - a namespace that contains named objects like (tables, functions, etc)
+        """
+        # We check if the schema exists first because snowflake will fail on a create schema query even if it
```
phanikumv, line 222:
```suggestion
        # We check if the schema exists first because BigQuery will fail on a create schema query even if it
```

phanikumv (approved):
Just a minor comment, other than that it LGTM

sunank200 (commented):
It would be good to add test for table property `schema` as well

sunank200 (approved):
Otherwise LGTM

python-sdk/src/astro/table.py
```diff
@@ -26,13 +26,24 @@ class Metadata:
...
+    location: str | None = None
 
     def is_empty(self) -> bool:
         """Check if all the fields are None."""
         return all(getattr(self, field_name) is None for field_name in fields_dict(self.__class__))
 
+    @property
+    def schema(self):
+        if self.location:
+            return f"{self._schema}__{self.location.replace('-', '_')}"
```
feluelle, line 40: How would this look when `self._schema` is empty?

feluelle: Why are we modifying `self.location`? Would using `-` break something?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -129,7 +132,11 @@ def default_metadata(self) -> Metadata:
 
         :return:
         """
-        return Metadata(schema=self.DEFAULT_SCHEMA, database=self.hook.project_id)  # type: ignore
+        return Metadata(
+            schema=self.DEFAULT_SCHEMA,  # type: ignore
+            database=self.hook.project_id,
+            location=BIGQUERY_SCHEMA_LOCATION,
+        )  # type: ignore
```
feluelle, line 139: Do we need both ignores?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -143,6 +150,28 @@ def schema_exists(self, schema: str) -> bool:
...
+        except NotFound:
+            return ""
+
+    def check_same_region(self, first_table: BaseTable, output_table: BaseTable):
+        """
+        Check if two tables are from the same database region
+        """
+        first_table_location = self.get_schema_region(schema=first_table.metadata.schema)
+        output_table_location = self.get_schema_region(schema=output_table.metadata.schema)
+        return first_table_location == output_table_location
```
feluelle, line 173:
```suggestion
    def check_same_region(self, table: BaseTable, other_table: BaseTable):
        """
        Check if two tables are from the same database region
        """
        table_location = self.get_schema_region(schema=table.metadata.schema)
        other_table_location = self.get_schema_region(schema=other_table.metadata.schema)
        return table_location == other_table_location
```

python-sdk/src/astro/databases/base.py
```diff
@@ -683,7 +683,7 @@ def export_table_to_file(
     # Schema Management
     # ---------------------------------------------------------
 
-    def create_schema_if_needed(self, schema: str | None) -> None:
+    def create_schema_if_needed(self, schema: str | None, location: str | None = None) -> None:
```
tatiana, line 686: In some parts of the code, we're using region, and here we're using location. Would it make sense to be consistent and only use region? Are there any circumstances when we want to use zone?

python-sdk/src/astro/databases/base.py
```diff
@@ -704,6 +704,20 @@ def schema_exists(self, schema: str) -> bool:
         """
         raise NotImplementedError
 
+    def get_schema_region(self, schema: str) -> str:
```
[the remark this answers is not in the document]
utkarsharma2: Yup, it should.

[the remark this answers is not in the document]
utkarsharma2: Yup, it should.

python-sdk/src/astro/table.py
utkarsharma2: Added a comment explaining the need for replacement.

python-sdk/src/astro/databases/google/bigquery.py
utkarsharma2: Nope. removed it

python-sdk/src/astro/databases/base.py
utkarsharma2: @tatiana Good point, I think we should replace it with the region will avoid confusion with the file's location attribute.

PR #1417, opened 2022-12-13
Make environment optional for Config instance

pankajkoti:
The Config instance has environment as a required param. However, for generating a config instance just to retrieve global config keys, we had a need to pass some environment key and if we need to set a random value like `global` it creates a problem if someone calls the method `get_env_config_filepath` because the directory `global` no longer exists. Hence, allow the creation of Config instance by making environment key optional.

Also, re-add env var for postgres password in test configuration.yml

### DIFF
--- a/sql-cli/sql_cli/configuration.py
+++ b/sql-cli/sql_cli/configuration.py
@@ -19,18 +19,20 @@ class Config:
     """
 
     project_dir: Path
-    environment: str
+    environment: str | None = None
 
     connections: list[dict[str, Connection]] = field(default_factory=list)
     airflow_home: str | None = None
     airflow_dags_folder: str | None = None
 
-    def get_env_config_filepath(self) -> Path:
+    def get_env_config_filepath(self) -> Path | None:
         """
         Return environment specific configuration.yaml filepath.
 
         :returns: The path to the desired YAML configuration file
         """
+        if not self.environment:
+            return None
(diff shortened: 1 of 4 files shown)


sql-cli/sql_cli/configuration.py
```diff
@@ -19,18 +19,20 @@ class Config:
...
 
-    def get_env_config_filepath(self) -> Path:
+    def get_env_config_filepath(self) -> Path | None:
         """
         Return environment specific configuration.yaml filepath.
 
         :returns: The path to the desired YAML configuration file
         """
+        if not self.environment:
+            return None
```
feluelle, line 35: This doesn't looks right to me. Why do we call this function if we don't need the env config filepath?

pankajkoti: Since we're allowing Config to be created without setting environment, we will need to take care of it in this method right? I mean if someone calls this method after they have created a Config with None environment.

feluelle (approved):
After talking with @pankajkoti about this change. I decide to be okay with it as long as we plan to improve/refactor the configuration going forward.

feluelle:
@pankajkoti maybe could you add some tests covering the changes?

pankajkoti:
> After talking with @pankajkoti about this change I decided to be okay with it as long as we plan to improve/refactor the configuration going forward.

Based on our discussion, I have created a ticket https://github.com/astronomer/astro-sdk/issues/1418 for the refactoring.

PR #1441, opened 2022-12-16
Add test case to verify the serialization working correctly

utkarsharma2:
# Description
RN, we don't have a test case to verify the serialization and deserialization are working correctly.

## What is the new behavior?
Added a test case to verify the serialization and deserialization

Nope

### DIFF
--- a/python-sdk/tests_integration/sql/operators/test_load_file.py
+++ b/python-sdk/tests_integration/sql/operators/test_load_file.py
@@ -4,6 +4,7 @@ from unittest import mock
 
 import pandas as pd
 import pytest
+from airflow.decorators import task
 from airflow.exceptions import AirflowNotFoundException
 from airflow.providers.amazon.aws.hooks.s3 import S3Hook
 from airflow.providers.google.cloud.hooks.gcs import GCSHook
@@ -16,7 +17,7 @@ from astro.exceptions import DatabaseCustomError
 from astro.files import File
 from astro.settings import SCHEMA
 from astro.sql import load_file
-from astro.table import Table
+from astro.table import Metadata, Table
 from tests.utils.airflow import create_context
 
 from ..operators import utils as test_utils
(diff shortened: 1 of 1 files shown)


python-sdk/tests_integration/sql/operators/test_load_file.py
```diff
@@ -1138,3 +1139,25 @@ def test_aql_nested_ndjson_file_to_database_explicit_illegal_sep_params(sample_d
...
+        def serialize():
+            return init_table
+
+        @task(task_id="verify")
+        def verify(t):
+            assert t == init_table
+
+        t = serialize()
+        verify(t)
+    test_utils.run_dag(sample_dag)
```
kaxil, line 1163: Does not have to been an integration test for sure!

python-sdk/tests_integration/sql/operators/test_load_file.py
```diff
@@ -1138,3 +1139,25 @@ def test_aql_nested_ndjson_file_to_database_explicit_illegal_sep_params(sample_d
     df = db.export_table_to_pandas_dataframe(test_table)
     assert df.shape == (1, 36)
     assert "payload_size" in df.columns
+
+
+def test_serialization_and_deserialization_of_table(sample_dag):
+    """
+    Test serialization and deserialization of table object
+    """
```
kaxil, line 1147: Doesn't add any value, says the same thing as the test name

python-sdk/tests_integration/sql/operators/test_load_file.py
dimberman, line 1163: @utkarsharma2 aas Kaxil mentioned. Ultimately we are just serializing data into json and then deserializing it from json. There is no need to create a DAG run and if anything that adds unnecessary overhead. Why not just serialize it into a json variable and ensure that the resulting json looks correct/can deserialize back to the same object?

dimberman (changes requested):

utkarsharma2:
No longer needed. Closing the PR.

PR #1442, opened 2022-12-16
Override GoogleBaseHook with BigqueryHook

utkarsharma2:
# Description
Because of below issue:
```
airflow.exceptions.AirflowException: You are trying to use `common-sql` with GoogleBaseHook, but its provider does not support it. Please upgrade the provider to a version that supports `common-sql`. The hook class should be a subclass of `airflow.providers.common.sql.hooks.sql.DbApiHook`. Got GoogleBaseHook Hook with class hierarchy: [<class 'airflow.providers.google.common.hooks.base_google.GoogleBaseHook'>, <class 'airflow.hooks.base.BaseHook'>, <class 'airflow.utils.log.logging_mixin.LoggingMixin'>, <class 'object'>]
```
We are using a work around and using Bigquey Hook

### DIFF
--- a/python-sdk/src/astro/sql/operators/data_validations/ColumnCheckOperator.py
+++ b/python-sdk/src/astro/sql/operators/data_validations/ColumnCheckOperator.py
@@ -6,6 +6,7 @@ from airflow.decorators.base import get_unique_task_id
 from airflow.providers.common.sql.operators.sql import SQLColumnCheckOperator
 
 from astro.databases import create_database
+from astro.settings import BIGQUERY_SCHEMA_LOCATION
 from astro.table import BaseTable
 from astro.utils.typing_compat import Context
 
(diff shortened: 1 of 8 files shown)


python-sdk/src/astro/sql/operators/data_validations/ColumnCheckOperator.py
```diff
@@ -77,8 +79,27 @@ def __init__(
...
+        db = create_database(
+            conn_id=self.conn_id, region=self.dataset.metadata.region or BIGQUERY_SCHEMA_LOCATION
+        )
+        if db.sql_type == "bigquery":
+            return db.hook
+        return super().get_db_hook()
+
     def execute(self, context: "Context"):
         if isinstance(self.dataset, BaseTable):
+            # Work around for GoogleBaseHook not inheriting from DBApi
```
feluelle, line 97: @utkarsharma2 you forgot to remove the comment here.

feluelle (commented):
Why did you merge without an approval?

PR #1449, opened 2022-12-16
Alternate approach to get location for Bigquery tables

kaxil:
An alternate approach to https://github.com/astronomer/astro-sdk/pull/1416/

Also closes https://github.com/astronomer/astro-sdk/pull/1460

**ToDo**:

- [x] Add more tests

The `region`  parameter passed to Table wasn't used anywhere except internally -- so going to remove it without deprecation


python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -134,9 +134,33 @@ def default_metadata(self) -> Metadata:
...
+            )
+            source_schema = source_table.metadata.schema
+            schema = (
+                f"{source_schema}__{source_location.replace('-', '_')}"
+                if not source_schema
+                else source_schema
+            )
+            source_db = source_table.metadata.database or self.hook.project_id
+            table.metadata = Metadata(schema=schema, database=source_db)
+            return table
```
kaxil, line 156: This can be moved to a separate helper method

python-sdk/src/astro/databases/base.py
```diff
@@ -195,14 +195,16 @@ def default_metadata(self) -> Metadata:
         """
         raise NotImplementedError
 
-    def populate_table_metadata(self, table: BaseTable) -> BaseTable:
+    def populate_table_metadata(self, table: BaseTable, source_table: None | BaseTable = None) -> BaseTable:
```
kaxil, line 198: This makes it pluggable, so that we can change different table attribtues for different databases  with different logic

python-sdk/src/astro/databases/base.py
```diff
@@ -198,11 +198,13 @@ def default_metadata(self) -> Metadata:
...
         """
         Given a table, check if the table has metadata.
-        If the metadata is missing, and the database has metadata, assign it to the table.
+        If the metadata is missing and source_table is defined, copy the attrs from that.
+        If the source_table is missing, and the database has metadata, assign it to the table.
         If the table schema was not defined by the end, retrieve the user-defined schema.
         This method performs the changes in-place and also returns the table.
 
         :param table: Table to potentially have their metadata changed
+        :param source_table: Table to copy metadata from
```
kaxil, line 207:
```suggestion
```

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -134,9 +134,37 @@ def default_metadata(self) -> Metadata:
...
+        source_schema = source_table.metadata.schema
+        schema = (
+            f"{self.DEFAULT_SCHEMA}__{source_location.replace('-', '_')}"
+            if not source_schema
+            else source_schema
+        )
+        source_db = source_table.metadata.database or self.hook.project_id
+        table.metadata = Metadata(schema=schema, database=source_db)
+        return table
+
```
kaxil, line 167: This entire code-block needs more tests :)

utkarsharma2 (approved):
A better approach will cater to other operators in the future.

python-sdk/src/astro/databases/base.py
```diff
@@ -362,7 +364,7 @@ def create_schema_and_table_if_needed(
         is_file_pattern_based_schema_autodetection_supported = (
             self.check_file_pattern_based_schema_autodetection_is_supported(source_file=file)
         )
-        if self.schema_exists(table.metadata.schema) and if_exists == "replace":
+        if if_exists == "replace":
```
dimberman, line 367: @kaxil is this safe? Do we have any documentation as to why we check if the schema exists first?

python-sdk/src/astro/databases/base.py
```diff
@@ -698,9 +700,7 @@ def export_table_to_file(
     # Schema Management
     # ---------------------------------------------------------
 
-    def create_schema_if_needed(
-        self, schema: str | None, location: str | None = None  # skipcq: PYL-W0613
-    ) -> None:
+    def create_schema_if_needed(self, schema: str | None) -> None:
```
dimberman, line 703: Is this a breaking change?

python-sdk/src/astro/databases/google/bigquery.py
```diff
@@ -134,9 +134,36 @@ def default_metadata(self) -> Metadata:
...
             database=self.hook.project_id,
-            region=BIGQUERY_SCHEMA_LOCATION,
         )  # type: ignore
 
+    def populate_table_metadata(self, table: BaseTable) -> BaseTable:
+        if (
+            table.temp
+            and (self.table and not self.table.metadata.is_empty())
+            and (table.metadata and table.metadata.is_empty())
+        ):
```
dimberman, line 144: Could this all be a function?

dimberman (changes requested):

python-sdk/src/astro/databases/base.py
kaxil: This was only added in https://github.com/astronomer/astro-sdk/commit/730d273b32e4f6a453f90837942a8490bc766cf9 last week and I don't think we need it -- drop query is idempotent. And if the schema doesn't exist it should "error" out rather than ignoring

python-sdk/src/astro/databases/base.py
kaxil: This should have only been used internally (was added last week) and was not used by anything apart from Bigquery that too internally

python-sdk/src/astro/databases/google/bigquery.py
kaxil: It is already a small function that calls another function, don't see the need to have another function here

PR #1452, opened 2022-12-17
Add the ability load files using autoloader

dimberman:
# Description
Currently all loading is done via the COPY INTO command. This command can work fine on small to mid-sized datasets but falls short when dealing with very large datasets. It also does not have a lot of popular autoloader features such as incremental loading.

CLOSES: https://github.com/astronomer/astro-sdk/issues/1445

## What is the new behavior?
Now the user will by default use autoloader unless a) they set the `load_options.load_mode` to COPY INTO or b) they load a single file (since autoloader can not handle single files).

### Checklist


python-sdk/src/astro/files/base.py
```diff
@@ -115,6 +116,11 @@ def openlineage_dataset_name(self) -> str:
         """
         return self.location.openlineage_dataset_name
 
+    def is_directory(self) -> bool:
+        """ """
```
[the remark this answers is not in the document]
dimberman: Oh shoot I must have stashed that part. fixing now.

python-sdk/docs/guides/databricks.rst
```diff
@@ -61,6 +61,35 @@ To use the ``aql.load_file()`` function, you will need to specify the path to th
...
+    from astro.databricks.load_options import DeltaLoadOptions
+    from astro.constants import DatabricksLoadMode
+
+    delta_options = DeltaLoadOptions.get_default_delta_options()
+    delta_options.load_mode = DatabricksLoadMode.COPY_INTO
+    aql.load_file(
+        input_file=File("data.csv"),
+        output_table=Table(conn_id="my_databricks_conn"),
+        load_options=delta_options,
+    )
```
feluelle, line 88: I would prefer having defaults set in the `__init__` so that we can do just this:
```suggestion
    aql.load_file(
        input_file=File("data.csv"),
        output_table=Table(conn_id="my_databricks_conn"),
        load_options=DeltaLoadOptions(load_mode=DatabricksLoadMode.COPY_INTO),
    )
```

python-sdk/src/astro/files/base.py
```diff
@@ -115,6 +116,11 @@ def openlineage_dataset_name(self) -> str:
...
+
+        return os.path.isdir(self.path)
```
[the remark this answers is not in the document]
feluelle: +1

python-sdk/tests/databricks/load_file/test_load_python_code_generator.py
```diff
@@ -24,11 +22,57 @@
...
+            data_source_path="foobar",
+            table_name="baz",
+            source_type="s3",
+            output_file_path=Path(tfile.name),
+            load_options=delta_options,
+            file_type="CSV",
+        )
+        with open(output_file) as file:
+            file_string = file.read()
+            assert file_string == expected_basic_file_autoloader
```
feluelle, line 68:
```suggestion
def test_generate_file_autoloader(tmp_path):
    delta_options = DeltaLoadOptions.get_default_delta_options()
    delta_options.load_secrets = False
    output_file = generate_file(
        data_source_path="foobar",
        table_name="baz",
        source_type="s3",
        output_file_path=tmp_path,
        load_options=delta_options,
        file_type="CSV",
    )
    assert output_file.read_text() == expected_basic_file_autoloader
```

python-sdk/tests/databricks/load_file/test_load_python_code_generator.py
```diff
@@ -24,11 +22,57 @@
...
+
 
-def test_generate_file_basic():
+def test_generate_file_copy_into():
```
feluelle, line 71: Same here. 
* You don't need `tempfile.NamedTemporaryFile()`. Using `tmp_path` should be sufficient.
* Also for a single file operation you can use pathlib.read_text instead of manually opening the file and reading from it

dimberman: @feluelle I was able to do that for all tests except one (because one of the functions makes assumptions about it being a NamedTemporaryFile). Thank you for the tip!

feluelle: I actually missed that `tmp_path` does not create the file - it creates a path (directory). So if the file needs to be created beforehand, you can call `tm_path_factory.mkstemp()`.

feluelle: For more information see [tmp_path_factory](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-factory-fixture)

feluelle (approved):

PR #1453, opened 2022-12-17
Disable Custom serialization for Back-compat

kaxil:
We only need AstroPandasDataframe and other custom serialization and deserialization if Airflow >= 2.5 and Pickling is not enabled and neither Custom XCom backend is used.

For others, we should leave it untouched.

(This PR is not strictly needed, I need it right now for internal purposes for Cloud IDE)


kaxil:
I should have left the comment, sorry - This was for Cloud IDE team, I cut an 1.3.3b1 version for them after merging this.

PR: https://github.com/astronomer/astro/pull/7339

PR #1461, opened 2022-12-19
Use correct file in example_load_file DAG

kaxil:
https://github.com/astronomer/astro-sdk/pull/1297 changed all the files to use S3, even though from PR description it should have only changed HTTP path.

Due to this we were trying to use S3 files with Bigquery connection ID weirdly

### DIFF
--- a/python-sdk/docs/configurations.rst
+++ b/python-sdk/docs/configurations.rst
@@ -124,8 +124,6 @@ or by updating Airflow's configuration
    [astro_sdk]
    auto_add_inlets_outlets = True
 
-.. _configure_native_fallback:
-
 Configuring to emit temp table event in openlineage
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Astro SDK has ability to create temporary tables see: :ref:`table`.
@@ -145,6 +143,7 @@ or by updating Airflow's configuration
    [astro_sdk]
    openlineage_emit_temp_table_event = True
 
+.. _configure_native_fallback:
 
 Configuring the native fallback mechanism
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


kaxil:
cc @sunank200  @pankajastro  @phanikumv

python-sdk/example_dags/example_load_file.py
```diff
@@ -159,7 +159,7 @@
     # [START load_file_example_12]
```
kaxil, line 159: This example is used in our docs to show how to use GCS's prefix but this is now completely wrong 

image

python-sdk/docs/configurations.rst
```diff
@@ -124,8 +124,6 @@ or by updating Airflow's configuration
    [astro_sdk]
    auto_add_inlets_outlets = True
 
-.. _configure_native_fallback:
```
kaxil, line 127: image

The docs was broken ! it had reference at other wrong place

python-sdk/example_dags/example_load_file.py
```diff
@@ -159,7 +159,7 @@
...
     aql.load_file(
         input_file=File(
-            "s3://astro-sdk/python_sdk/example_dags/data/sample_pattern.csv",
+            "gs://astro-sdk/workspace/sample_pattern.csv",
```
kaxil, line 162: This should be changed to a pattern but I don't see a folder with pattern.

cc @utkarsharma2 Can you take care of this in a separate PR, please

pankajastro (approved):
Look good to me but look we need a review for our docs

utkarsharma2:
@kaxil Why did we modify the example - load_file_example_9 to load_file_example_17?

kaxil:
> @kaxil Why did we modify the example - load_file_example_9 to load_file_example_17?

Coz that example in docs refer to `enable_native_fallback` which is example_17

utkarsharma2:
My bad I somehow got confused it with `use_native_support` param.

python-sdk/example_dags/example_load_file.py
utkarsharma2: @kaxil https://github.com/astronomer/astro-sdk/pull/1497

PR #1466, opened 2022-12-20
Add add pandas_options in load_file

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1519

currently, load_file do not have an option to pass the pandas-related param while reading file

## What is the new behavior?
use `load_options` and pass the given values while reading files using the pandas path 

No

### Checklist


python-sdk/src/astro/databases/base.py
```diff
@@ -724,8 +746,8 @@ def schema_exists(self, schema: str) -> bool:
     # ---------------------------------------------------------
 
     def get_sqlalchemy_template_table_identifier_and_parameter(
-        self, table: BaseTable, jinja_table_identifier: str
-    ) -> tuple[str, str]:  # skipcq PYL-W0613
+        self, table: BaseTable, jinja_table_identifier: str  # skipcq PYL-W0613
```
dimberman, line 749: Why are we bypassing code quality here?

python-sdk/src/astro/files/types/csv.py
```diff
@@ -15,14 +16,21 @@ class CSVFileType(FileType):
...
     ) -> pd.DataFrame:  # skipcq PYL-R0201
         """read csv file from one of the supported locations and return dataframe
 
         :param stream: file stream object
+        :param pandas_options: Pandas option to pass to the Pandas lib while reading csv
         :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
             in the resulting dataframe
         """
+        if pandas_options is not None:
+            kwargs.update(pandas_options.to_dict)
```
dimberman, line 33: Can you explain why are adding these to the kwargs? Is it safe to pass arbitrary kwargs to it?

dimberman (dismissed):

python-sdk/src/astro/databases/base.py
pankajastro: `jinja_table_identifier` param is unused here so the static check is complaining about it. Also, this skipcq was there but looks like it was not on the correct line number.

python-sdk/src/astro/files/types/csv.py
pankajastro: The kwargs here are not exposed to the public interface here (i.e in operator, decorator) so exposing a parameter `pandas_options` in load_file which will have attributes corresponding to pandas `read_csv` etc method and updating the kwargs with these values. https://github.com/astronomer/astro-sdk/pull/1466/files#diff-8927674ffbb5764b3494eb1634934aa428126e76fe1f43479de1b4cfae0b5334

utkarsharma2:
@pankajastro What sort of pandas do we want to pass using pandas_options?

pankajastro:
> @pankajastro What sort of options do we want to pass using pandas_options? is delimiter?

yes, currently delimiter, but in future maybe other params as well which pandas read_{csv, json} etc method accept.

utkarsharma2:
@pankajastro I was a little concerned because if we keep adding the internal libs-specific parameters we will soon need to expose every database-specific or file type-specific parameter. We are exposing internal details to the end user, kinda violating the interface we have provided. Instead, since load files only have to deal with the file(s) and tables can we keep the parameters specific to tables and files? Like in this scenario, the delimiter is specific to CSV files, maybe we can move this parameter to File Object? WDYT?

pankajastro:
> @pankajastro I was a little concerned because if we keep adding the internal libs-specific parameters we will soon need to expose every database-specific or file type-specific parameter. We are exposing internal details to the end user, kinda violating the interface we have provided. Instead, since load files only have to deal with the file(s) and tables can we keep the parameters specific to tables and files? Like in this scenario, the delimiter is specific to CSV files, maybe we can move this parameter to File Object? WDYT?

I understand your concern and probably we will expose only limited param. We don't want to make File object messy either  so I closed https://github.com/astronomer/astro-sdk/pull/1225 there is some discussion on https://github.com/astronomer/astro-sdk/issues/482 too

python-sdk/src/astro/databases/base.py
```diff
@@ -416,6 +429,8 @@ def load_file_to_table(
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
         load_options: LoadOptions | None = None,
+        databricks_job_name: str = "",  # skipcq PYL-W0613
+        pandas_options: PandasLoadOptions | None = None,
```
tatiana, line 433: @pankajastro What do you think of us using the `load_options` argument instead of introducing a pandas-specific one?

Are there any scenarios where we want both of them?

 @dimberman , since you recently worked on this, I'd love your thoughts as well

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -52,6 +54,7 @@ def __init__(
         load_options: LoadOptions | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
+        pandas_options: PandasLoadOptions | None = None,
```
tatiana, line 57: Would it make sense for us to deprecate the `ndjson_normalize_sep` and have it as part of  `PandasLoadOptions`?

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -74,6 +77,7 @@ def __init__(
         self.columns_names_capitalization = columns_names_capitalization
         self.enable_native_fallback = enable_native_fallback
         self.load_options = load_options
+        self.pandas_options = pandas_options
```
tatiana, line 80: It would be great if we avoided implementation-specific arguments, and had this as part of `load_options`

tatiana (commented):
@pankajastro, as we discussed, my only concern - which needs a bit of further look into the code - is if we are calling multiple pandas functions as part of load_file (e.g. to create a table if it doesn't exist), and the user would like to specify custom arguments to multiple of these functions.

I feel it is fine for us to focus on the `read_csv`, `read_json`, etc., but it would be great to look at the code and confirm this is not an issue.

python-sdk/src/astro/databases/base.py
kaxil: Yes, I was hoping that when I suggested @dimberman to add `load_options`, it would become more generic so we can add `PandasLoadOption` and pass it to `load_options` itself and not a separate param

python-sdk/src/astro/sql/operators/load_file.py
kaxil: +1

python-sdk/src/astro/sql/operators/load_file.py
kaxil: +1

python-sdk/src/astro/databases/base.py
feluelle: +1, would love if we could do that in this PR.

pankajastro: Let's say we are doing native load and want to create schema and table with auto-detection (method `create_schema_and_table_if_needed`) then we might need pandas options for this and the load_option for other native supported param?? @utkarsharma2 @tatiana WDYT please let me know after that I'm ok to remove this and use load_options

utkarsharma2: @pankajastro Can we address this in a separate PR? Since we haven't reached a conclusion on the earlier discussion.

pankajastro: sure, @utkarsharma2 thanks.

python-sdk/src/astro/sql/operators/load_file.py
pankajastro: @tatiana Deprecation make sense but would be better to keep it in separate PR tracking here https://github.com/astronomer/astro-sdk/issues/1518

python-sdk/src/astro/sql/operators/load_file.py
pankajastro: Not quite sure but I feel we may need a separate pandas param https://github.com/astronomer/astro-sdk/pull/1466#discussion_r1060364256

python-sdk/src/astro/databases/base.py
```diff
@@ -246,6 +247,7 @@ def create_table_using_schema_autodetection(
         file: File | None = None,
         dataframe: pd.DataFrame | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",  # skipcq
+        pandas_options: PandasLoadOptions | None = None,
```
tatiana, line 250: @pankajastro could we name this `load_options`, using the same argument as the Databricks implementation?

python-sdk/src/astro/databases/base.py
tatiana: Please, let's do it as part of this PR. We are moving from `native_kwargs` to `load_options`. To introduce a third way of doing this (`pandas_otions, even temporarily to our `main` branch, could be a step backwards). I don't believe we should merge this PR before renaming this argument.

tatiana: @utkarsharma2, if you could add a link to the follow-up ticket to change `load_options` to a list, it would be great!

utkarsharma2: @tatiana, Here - https://github.com/astronomer/astro-sdk/issues/1532

python-sdk/src/astro/dataframes/load_options.py
```diff
@@ -0,0 +1,34 @@
...
+from pandas._typing import DtypeArg
+
+from astro.options import LoadOptions
+
+
+@define
+class PandasLoadOptions(LoadOptions):
+    @property
+    def to_dict(self):
+        return dict(filter(lambda x: x[1], asdict(self).items()))
```
utkarsharma2, line 13: @pankajastro Can we use asdict(self) instead of this?

pankajastro: fixed it. remove to_dict and using asdict

python-sdk/src/astro/sql/operators/load_file.py
pankajastro: changed

utkarsharma2 (approved):
LGTM! :)

pankajastro:
Hi @dimberman Can you please re-review this PR

python-sdk/src/astro/dataframes/load_options.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@define
+class NdjsonLoadOptions(PandasLoadOptions):
+    ndjson_normalize_sep: str = "_"
+
+
+@define
+class ParquetLoadOptions(PandasLoadOptions):
+    columns: list[str] | None = None
```
tatiana, line 32: For now, we should prefix all of these with Pandas - once we have a generic representation of these, which works across DBs & pandas, then we can give a generic name and place them together with the file file types.

tatiana (approved):
Thanks a lot, @pankajastro , for addressing all the comments. I'm happy for this PR to be merged after this last comment we agreed on is addressed.

python-sdk/src/astro/dataframes/load_options.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@define
+class JsonLoadOptions(PandasLoadOptions):
+    encoding: str | None = None
+
+
+@define
+class NdjsonLoadOptions(PandasLoadOptions):
+    ndjson_normalize_sep: str = "_"
```
tatiana, line 27:
```suggestion
    normalize_sep: str = "_"
```

PR #1472, opened 2022-12-20
Generate constraints on releases and pushes (not PRs)

kaxil:
This will limit generating constraints on releases and pushes (not PRs). I don't think we need to generate constraints on every PR

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -364,6 +364,7 @@ jobs:
       AZURE_WASB_CONN_STRING: ${{ secrets.AZURE_WASB_CONN_STRING }}
 
   Generate-Constraints:
+    if: (github.event_name == 'release' || github.event_name == 'push')
     strategy:
       fail-fast: false
       matrix:


.github/workflows/ci-python-sdk.yaml
```diff
@@ -364,6 +364,7 @@ jobs:
       AZURE_WASB_CONN_STRING: ${{ secrets.AZURE_WASB_CONN_STRING }}
 
   Generate-Constraints:
+    if: (github.event_name == 'release' || github.event_name == 'push')
```
pankajkoti, line 367: Do we need to check that the push is on main branch? 
I see the the action is skipped here but is it that the commit was pushed first and then the pull request was created and hence it got skipped. What if we push a commit to an existing pull request? Will it skip then also?

kaxil: We restrict pushes on `main` and release branches only in the following block

https://github.com/astronomer/astro-sdk/blob/50257f9cf5c09409a76805af9008e8af0800d500/.github/workflows/ci-python-sdk.yaml#L2-L4

Hence it won't send push event for other branches to this job

pankajkoti (approved):

kaxil:
Failures are unrelated, merging

PR #1476, opened 2022-12-20
Add merge function to databricks delta

### LINKED ISSUE #1451: Add merge function to databricks

I'd like to see a feature that allows users to run the aql.merge function to merge two Databricks Delta tables together using the aql library. This would enable us to more easily and efficiently merge data from multiple tables into a single table for further processing and analysis.

Describe the solution you'd like

A function or method that takes the names of the two Delta tables to be merged, along with a connection string for a Databricks cluster, and runs the aql.merge function on the tables using the aql library. The function should accept all of the same arguments as the traditional aql.merge function, except for an additional conn_id argument that specifies the connection to use.

Here's an example of what the function signature might look like:

```
target_table = Table(name="foo", conn_id="my_databricks_conn")
source_table = Table(name="bar", conn_id="my_databricks_conn")

aql.merge(
    target_table=target_table,
    source_table=source_table,
    target_conflict_columns=["id"],
    columns=["id", "name"],
    if_conflicts="update",
)
```

**Are there any alternatives to this feature?**
Another alternative would be to use a tool like Apache Spark to merge the data from the two tables into a third table. This would require users to set up and configure a Spark cluster in their databricks instance, whereas the proposed function would be more lightweight and easier to use.

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

dimberman:
# Description

Currently, the merge function returns a `NotImplementedError` if a user attempts to merge two delta tables. This leaves the delta astro SDK experience incomplete and can create confusion for users.

Closes #1451 

## What is the new behavior?
The new behavior is to add a merge function in a similar fashion to how we generate a merge function in snowflake. We have found these two interaces to be fairly similar so it might be worth looking into whether the functions can be merged in a future ticket.

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/databricks/delta.py
+++ b/python-sdk/src/astro/databricks/delta.py
@@ -85,16 +85,6 @@ class DeltaDatabase(BaseDatabase):
         # TODO Do we need to implement this function? It seems like databricks will handle schemas for us
         raise NotImplementedError("Not implemented yet.")
 
-    def merge_table(
-        self,
-        source_table: BaseTable,
-        target_table: BaseTable,
-        source_to_target_columns_map: dict[str, str],
-        target_conflict_columns: list[str],
-        if_conflicts: MergeConflictStrategy = "exception",
-    ) -> None:
-        raise NotImplementedError("We do not yet support merge for databricks")
-
     def schema_exists(self, schema: str) -> bool:
         # Schemas do not need to be created for delta, so we can assume this is true
         return True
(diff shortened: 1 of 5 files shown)


python-sdk/tests/airflow_test/test_datasets.py
```diff
@@ -1,7 +1,7 @@
 import os
 from unittest import mock
 
-import airflow
+import airflow_test
```
dimberman, line 4: @kaxil i've been running into issues in pycharm where it can't run tests in the IDE due to conflicts between the tests and the imported packages. I wonder if OSS users might run into the same issue.

utkarsharma2: @pankajastro Also faced this issue recently.

python-sdk/src/astro/databricks/delta.py
```diff
@@ -312,3 +302,106 @@ def load_pandas_dataframe_to_table(
             file = File(path=t + f"/dataframe_{uuid.uuid4()}.parquet", filetype=FileType.PARQUET)
             file.create_from_dataframe(source_dataframe)
             self.load_file_to_table(input_file=file, output_table=target_table, if_exists=if_exists)
+
+    @staticmethod
+    def get_merge_initialization_query(parameters: tuple) -> str:
+        """
+        Handles database-specific logic to handle constraints
+        for Delta. setting constraints is not required for merging tables in Delta so we only "RETURN 1"
```
feluelle, line 310:
```suggestion
        for Delta. setting constraints is not required for merging tables in Delta so we only return "1"
```

python-sdk/src/astro/databricks/delta.py
```diff
@@ -312,3 +302,106 @@ def load_pandas_dataframe_to_table(
...
+
+        values_to_check = [target_table_name, source_table_name]
+        values_to_check.extend(source_cols)
+        values_to_check.extend(target_cols)
+        if if_conflicts == "update":
+            statement += " when matched then UPDATE SET {merge_vals}"
+            merge_statement = ",".join(
+                [f"target_table.{t} = source_table.{s}" for s, t in source_to_target_columns_map.items()]
+            )
+            statement = statement.replace("{merge_vals}", merge_statement)
```
feluelle, line 391:
```suggestion
            merge_statement = ",".join(
                [f"target_table.{t} = source_table.{s}" for s, t in source_to_target_columns_map.items()]
            )
            statement += f" when matched then UPDATE SET {merge_statement}"
```
This can be simplified using f-string.

We should not use `replace` for formatting strings. This is really error-prone! Using `format` should do the same job better.

python-sdk/src/astro/databricks/delta.py
```diff
@@ -312,3 +302,106 @@ def load_pandas_dataframe_to_table(
...
+        statement += " when not matched then insert({target_columns}) values ({append_columns})"
+        statement = statement.replace(
+            "{target_columns}",
+            ",".join(f"{t}" for t in target_cols),
+        )
+        statement = statement.replace(
+            "{append_columns}",
+            ",".join(f"source_table.{s}" for s in source_cols),
+        )
```
feluelle, line 400: Please also change this to use either f-strings or `format`.

python-sdk/src/astro/databricks/delta.py
```diff
@@ -312,3 +302,106 @@ def load_pandas_dataframe_to_table(
...
+        merge_target_dict = {
+            f"`target_table`.`{x}`": f"`target_table`.`{x}`" for i, x in enumerate(target_conflict_columns)
+        }
+        merge_source_dict = {
+            f"`source_table`.`{x}`": f"`source_table`.`{x}`" for i, x in enumerate(target_conflict_columns)
+        }
+        statement = statement.replace(
+            "{merge_clauses}",
+            " AND ".join(f"{k}={v}" for k, v in zip(merge_target_dict.keys(), merge_source_dict.keys())),
+        )
```
feluelle, line 381: Same here. Please refactor this using either f-strings or `format`.

feluelle (changes requested):

python-sdk/src/astro/databricks/delta.py
dimberman: This is a great call. I'm going to make a separate PR to do do the same with the SNOWFLAKE queries.

python-sdk/src/astro/databricks/delta.py
```diff
@@ -312,3 +302,98 @@ def load_pandas_dataframe_to_table(
...
+
+        values_to_check = [target_table_name, source_table_name]
+        values_to_check.extend(source_cols)
+        values_to_check.extend(target_cols)
+        if if_conflicts == "update":
+            merge_statement = ",".join(
+                [f"target_table.{t} = source_table.{s}" for s, t in source_to_target_columns_map.items()]
+            )
+            statement += f" when matched then UPDATE SET {merge_statement}"
+        target_columns = ",".join(f"{t}" for t in target_cols)
```
feluelle, line 389:
```suggestion
        target_columns = ",".join(f"target_table.{t}" for t in target_cols)
```

feluelle (approved):

feluelle:
@dimberman if you face issues with module names locally e.g. `airflow`, this should be handled locally. I am okay with changing it, but ideally your local setup should not affect how production is set up.

dimberman:
@feluelle I'm not the only person running into these naming conflicts. This also might lead to a barrier to any community members who might want to develop on the project.

Issue #1166, opened 2022-10-31
Implement SQL CLI deploy


tatiana:
I'm closing this ticket since we broke this work into smaller tickets.

These are the detailed tickets on the SQL CLI deployment and split the remaining work with the team for us to accomplish the milestone 0.3
   ◦ https://github.com/astronomer/astro-sdk/issues/1473 - Review
   ◦ https://github.com/astronomer/astro-sdk/issues/1517 - Review
   ◦ https://github.com/astronomer/astro-sdk/issues/1520 - Review
   ◦ https://github.com/astronomer/astro-sdk/issues/1528 - In Progress (Draft)
   ◦ https://github.com/astronomer/astro-sdk/issues/1529 - Done
   ◦ https://github.com/astronomer/astro-sdk/issues/1582 - Pankaj to pair up with Daniel on this.
   ◦ https://github.com/astronomer/astro-sdk/issues/1601 - Review
   ◦ https://github.com/astronomer/astro-sdk/issues/1602 - Review

I also created future tickets related to deployment, currently assigned to milestone 0.4 (to be reviewed)
   ◦ https://github.com/astronomer/astro-sdk/issues/1522
   ◦ https://github.com/astronomer/astro-sdk/issues/1523
   ◦ https://github.com/astronomer/astro-sdk/issues/1524
   ◦ https://github.com/astronomer/astro-sdk/issues/1525
   ◦ https://github.com/astronomer/astro-sdk/issues/1526
   ◦ https://github.com/astronomer/astro-sdk/issues/1527

tatiana:
This is a duplicate of #1520

PR #1479, opened 2022-12-21
Expose get/set deployment config in SQL CLI

tatiana:
Refactor the retrieval of configuration to allow retrieving most of the settings by exposing them as JSON.
```
$ flow config get airflow_home
$ flow config get --as-json
```

Add command to set deployment configuration per environment, using:
```
$ flow config set deploy --env=dev --astro-workspace=cl123 --astro-deplyment-cl345
```

As part of this PR, I also refactored the Airflow home to be environment-specific.

Disclaimers:
1. The command config is hidden from end-users and is meant to be used to integrate with the Astro CLI.
2. We'll need a follow up ticket on the the Astro CLI side, changing the current call from ` flow config airflow_home` to  flow config get airflow_home`

Related: #1166

### DIFF
--- a/sql-cli/include/base/config/dev/configuration.yml
+++ b/sql-cli/include/base/config/dev/configuration.yml
@@ -9,3 +9,7 @@ connections:
   - conn_id: sqlite_conn
     conn_type: sqlite
     host: retail.db
+
+# Apache Airflow optional settings
+airflow:
+  home:
--- a/sql-cli/sql_cli/cli/config.py
+++ b/sql-cli/sql_cli/cli/config.py
@@ -1,26 +1,28 @@
 from __future__ import annotations
 
+import json
 from pathlib import Path
 
 import typer
 
 from sql_cli.astro.command import AstroCommand
 from sql_cli.cli.utils import resolve_project_dir
+from sql_cli.constants import DEFAULT_ENVIRONMENT
 
 app = typer.Typer()
 
 
-class InvalidGetConfigArg(Exception):
+class InvalidConfigException(Exception):
     pass
 
 
-def _get(key: str, project_dir: Path, env: str, as_json: bool) -> str | dict:
+def _get(key: str, project_dir: Path, env: str, as_json: bool) -> str:
     """
     Typer-agnostic implementation of the `config get` command.
 
-    :param key: (optional) Key to be fetched from the configuration file. If none is specified, return all.
-    :param project_dir: (optional) Path to the project directory
-    :param env: Environment configuration (default, dev)
+    :param key: Key to be fetched from the configuration file.
+    :param project_dir: Path to the project directory
+    :param env: SQL CLI environment (default, dev)
     :param as_json: If the output should be displayed as JSON.
     :returns: Either the string containing the key value or a JSON containing desired the key-value pair(s)
     """
@@ -31,17 +33,44 @@ def _get(key: str, project_dir: Path, env: str, as_json: bool) -> str | dict:
     if key:
         return getattr(project_config, key)
     elif as_json:
-        return project_config.as_json()
+        return json.dumps(project_config.to_dict())
     else:
-        raise InvalidGetConfigArg("Please, either define a key or use the --as-json flag")
+        raise InvalidConfigException("Please, either define a key or use the --as-json flag")
+
+
+def _set(key: str, project_dir: Path, env: str, astro_deployment: str, astro_workspace: str) -> None:
+    """
+    Set deployment configuration associated to a SQL CLI environment.
+
+    :param key: Configuration property (key) to be set. At the mmoment only "deploy" is accepted.
+    :param project_dir: Path to the project directory.
+    :param env: SQL CLI environment (default, dev).
+    :param astro_deployment: Astronomer Cloud deployment ID.
+    :param astro_workspace: Astronomer Cloud deployment workspace.
+    """
+    from sql_cli.configuration import Config
+
+    if key != "deploy":
+        raise InvalidConfigException(
+            f"The key {key} is not supported yet. Only deploy is currently supported."
+        )
+
+    project_dir_absolute = resolve_project_dir(project_dir)
+    project_config = Config(environment=env, project_dir=project_dir_absolute).from_yaml_to_config()
+    config_filepath = project_config.get_env_config_filepath()
+
+    project_config.write_value_to_yaml("deployment", "astro_deployment", astro_deployment, config_filepath)
+    project_config.write_value_to_yaml("deployment", "astro_workspace", astro_workspace, config_filepath)
 
 
 @app.command(
     cls=AstroCommand,
     help="""
-    Gets the project configuration. There are two was of using this:
-         1) astro config get airflow_home
-         2) astro config get --as-json
+    Get the project configuration.
+
+    Example of usages:
+    $ astro config get airflow_home
+    $ astro config get --as-json
 
     The first returns a key from the config whereas the second returns all the configuration as JSON.
     """,
@@ -50,16 +79,50 @@ def get(
     key: str = typer.Argument(
         default="",
         show_default=False,
-        help="Key from the configuration whose value needs to be fetched.",
+        help="Configuration key which value needs to be fetched.",
     ),
     project_dir: Path = typer.Option(
         None, dir_okay=True, metavar="PATH", help="(Optional) Default: current directory.", show_default=False
     ),
     env: str = typer.Option(
-        default="default",
+        default=DEFAULT_ENVIRONMENT,
         help="(Optional) Environment used to fetch the configuration key from.",
     ),
     as_json: bool = typer.Option(False, help="If the response should be in JSON format", show_default=True),
 ) -> None:
     value = _get(key, project_dir, env, as_json)
     print(value)
+
+
+@app.command(
+    cls=AstroCommand,
+    help="""
+   Set the project configuration.
+
+   Example:
+   $ astro config set deploy --env=dev --astro-workspace=cl123 --astro-deployment=cl345
+   """,
+)
+def set(  # noqa: A001
+    key: str = typer.Argument(
+        default="",
+        show_default=False,
+        help="Key from the configuration whose value needs to be fetched.",
+    ),
+    project_dir: Path = typer.Option(
+        None, dir_okay=True, metavar="PATH", help="(Optional) Default: current directory.", show_default=False
+    ),
+    astro_deployment: str = typer.Option(
+        ...,
+        help="Astro deployment deployment ID (e.g. cl8bqua474573873jwenjhb6bbo)",
+    ),
+    astro_workspace: str = typer.Option(
+        ...,
+        help="Astro deployment workspace ID (e.g. cl6geh889308371i01vscssm4q)",
+    ),
+    env: str = typer.Option(
+        default=DEFAULT_ENVIRONMENT,
+        help="(Optional) Environment used to fetch the configuration key from.",
+    ),
+) -> None:
+    _set(key, project_dir, env, astro_workspace=astro_workspace, astro_deployment=astro_deployment)
--- a/sql-cli/sql_cli/cli/config.py
+++ b/sql-cli/sql_cli/cli/config.py
@@ -64,18 +64,19 @@ def _set(key: str, project_dir: Path, env: str, astro_deployment: str, astro_wor
 
 
 @app.command(
+    "get",
     cls=AstroCommand,
     help="""
     Get the project configuration.
 
     Example of usages:
     $ flow config get airflow_home
-    $ flow config get --as-json
+    $ flow config get --json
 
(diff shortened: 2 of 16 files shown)


sql-cli/tests/test_configuration.py
```diff
@@ -19,5 +19,4 @@ def test_from_yaml_to_config_without_env():
     config_from_file = config_reference.from_yaml_to_config()
     assert isinstance(config_from_file, Config)
     assert config_from_file.project_dir == config_reference.project_dir
-    assert config_from_file.environment is None
-    assert config_from_file.connections == []
+    assert config_from_file.environment == "default"
```
tatiana, line 22: @feluelle, I'd love your thoughts on this change.

My feeling is that it is valid in the use case where we use the Configuration object and we don't care about the environment - but I thought it might be more graceful to have a default value than to have to propagate str/None possibilities through the code, handling scenarios where config would not have an environment.

sql-cli/include/base/config/dev/configuration.yml
```diff
@@ -9,3 +9,7 @@ connections:
   - conn_id: sqlite_conn
     conn_type: sqlite
     host: retail.db
+
+# Apache Airflow optional settings
+airflow:
+  home:
```
tatiana, line 15: @pankajkoti I'd love your thoughts on this change, since you worked a lot on this part of the code as well

sql-cli/tests/test_configuration.py
dimberman: I personally think this makes sense. Default is... well... a default so it might as well be reflected as such in the code.

feluelle: Yes, +1 on explicit `default` as a default environment.

sql-cli/sql_cli/constants.py
```diff
@@ -3,7 +3,7 @@
 CONFIG_FILENAME = "configuration.yml"
 
 DEFAULT_ENVIRONMENT = "default"
-DEFAULT_AIRFLOW_HOME = f".airflow/{DEFAULT_ENVIRONMENT}"
+DEFAULT_BASE_AIRFLOW_HOME = ".airflow/"
```
feluelle, line 6:
```suggestion
DEFAULT_BASE_AIRFLOW_HOME = ".airflow"
```
should work too, and makes it consistent with the other paths.

sql-cli/sql_cli/project.py
```diff
@@ -80,51 +83,85 @@ def data_dir(self) -> Path:
...
+
+        :param airflow_home: Path to Airflow home we want to clean up.
         """
-        logs_folder = self.airflow_home / "logs"
+        logs_folder = airflow_home / "logs"
         shutil.rmtree(logs_folder)
 
-        webserver_config = self.airflow_home / "webserver_config.py"
+        webserver_config = airflow_home / "webserver_config.py"
         webserver_config.unlink()
```
feluelle, line 153: This function should actually live in our airflow utils and maybe get called in `initialise_airflow` at the end. We get the airflow_home there as well as an arg. WDYT?

sql-cli/sql_cli/project.py
```diff
@@ -145,8 +182,9 @@ def initialise(self) -> None:
             dirs_exist_ok=True,
         )
         self._initialise_global_config()
-        initialise_airflow(self.airflow_home, self.airflow_dags_folder)
-        self._remove_unnecessary_airflow_files()
+        for env in self.environments_list:
+            self._initialise_env_config(env)
+            self._initialise_airflow_for_environment(env)
```
feluelle, line 187: Nice!

sql-cli/sql_cli/project.py
```diff
@@ -168,11 +206,18 @@ def load_config(self, environment: str = DEFAULT_ENVIRONMENT) -> None:
         config = Config(environment=environment, project_dir=self.directory).from_yaml_to_config()
         if config.airflow_home:
             self._airflow_home = Path(config.airflow_home).resolve()
+        else:
+            self._initialise_env_config(environment)
+            config = Config(environment=environment, project_dir=self.directory).from_yaml_to_config()
+
+        if not self.get_env_airflow_home(environment).exists():
+            self._initialise_airflow_for_environment(environment)
+
```
feluelle, line 215: Why did we add this? This should be done by `initialise`.

sql-cli/sql_cli/project.py
```diff
@@ -168,11 +206,18 @@ def load_config(self, environment: str = DEFAULT_ENVIRONMENT) -> None:
...
         if config.airflow_dags_folder:
             self._airflow_dags_folder = Path(config.airflow_dags_folder).resolve()
         if config.data_dir:
             self._data_dir = Path(config.data_dir).resolve()
-        reload_airflow(self.airflow_home)
+        reload_airflow(self.get_env_airflow_home(environment))
```
feluelle, line 220:
```suggestion
        reload_airflow(self._airflow_home)
```
?

sql-cli/tests/config/test/configuration.yml
```diff
@@ -22,7 +22,7 @@ connections:
     schema: pagila
     login: postgres
     password: $POSTGRES_PASSWORD
-    port: $POSTGRES_PORT
+    port: 5432
```
feluelle, line 25: Please revert this change. You can set the env vars in your `.env` file.

sql-cli/sql_cli/cli/config.py
```diff
@@ -0,0 +1,129 @@
...
+    cls=AstroCommand,
+    help="""
+   Set the project configuration.
+
+   Example:
+   $ flow config set deploy --env=dev --astro-workspace=cl123 --astro-deployment=cl345
+   """,
+)
+# skipcq: PYL-W0622
+def set(  # noqa: A001
```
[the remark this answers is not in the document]
feluelle: Note that it exposes it as `set` see line 98 in the suggestion.

sql-cli/sql_cli/project.py
tatiana: The following use case fails if we don't handle this afterwards:
```
$ flow init .
# user manually adds a new environment `stage` to the `config` folder
$ flow validate --env stage
```

I realised this use case with one of the existing tests!

feluelle: Right! Thanks 👍 Can we simplify this?

sql-cli/sql_cli/cli/config.py
```diff
@@ -0,0 +1,129 @@
...
+    Get the project configuration.
+
+    Example of usages:
+    $ flow config get airflow_home
+    $ flow config get --as-json
+
+    The first returns a key from the config whereas the second returns all the configuration as JSON.
+    """,
+)
+def get(
```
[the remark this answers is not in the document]
tatiana: Much better, changing it!

sql-cli/sql_cli/cli/config.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def _set(key: str, project_dir: Path, env: str, astro_deployment: str, astro_workspace: str) -> None:
+    """
+    Set deployment configuration associated to a SQL CLI environment.
+
+    :param key: Configuration property (key) to be set. At the mmoment only "deploy" is accepted.
+    :param project_dir: Path to the project directory.
+    :param env: SQL CLI environment (default, dev).
+    :param astro_deployment: Astronomer Cloud deployment ID.
```
pankajkoti, line 48: Should we rename this to `astro_deployment_id` ?

sql-cli/sql_cli/cli/config.py
```diff
@@ -0,0 +1,132 @@
...
+    :param astro_workspace: Astronomer Cloud deployment workspace.
```
pankajkoti, line 49: Will this be workspace ID?

sql-cli/sql_cli/cli/config.py
```diff
@@ -0,0 +1,132 @@
...
+    ),
+    env: str = typer.Option(
+        default=DEFAULT_ENVIRONMENT,
+        help="(Optional) Environment used to fetch the configuration key from.",
+    ),
+    as_json: bool = typer.Option(
+        False, "--json", help="If the response should be in JSON format", show_default=True
+    ),
+) -> None:
+    value = _get(key, project_dir, env, as_json)
```
pankajkoti, line 96: Should we check that both `key` and `as_json` are not set together?

sql-cli/sql_cli/configuration.py
```diff
@@ -94,3 +92,16 @@ def write_value_to_yaml(self, section: str, key: str, value: str, filepath: Path
...
+        """
+        Return configuration represented as JSON string.
+
+        :returns: Dictionary containing non-sensitive global and per-environment configuration.
+        """
+        config_as_dict = {"global": self.from_yaml_to_dict(self.get_global_config_filepath())}
+        if self.environment:
+            env_config = self.from_yaml_to_dict(self.get_env_config_filepath())
+            env_config.pop("connections", None)  # remove sensitive information
+            config_as_dict[self.environment] = env_config
```
pankajkoti, line 106:
```suggestion
            config_as_dict["local"] = env_config
```
Would it make sense to rename the key to "local", or the requestor will know how exactly to parse the response for environment values?

sql-cli/sql_cli/project.py
```diff
@@ -80,50 +87,71 @@ def data_dir(self) -> Path:
...
 
     def _initialise_global_config(self) -> None:
         """
-        Initialises global config file that includes configuration to be shared across environments including the
+        Initialises global config YAML file that includes configuration to be shared across environments including the
         airflow config.
         """
-        config = Config(project_dir=self.directory)
-        global_env_filepath = config.get_global_config_filepath()
+        config = Config(environment=DEFAULT_ENVIRONMENT, project_dir=self.directory)
```
pankajkoti, line 116: @feluelle had reservations on passing Default environment for initialising global config. His question was why do we need to pass environment for global config, which I also felt is a valid question.

sql-cli/sql_cli/project.py
```diff
@@ -145,8 +173,9 @@ def initialise(self) -> None:
             dirs_exist_ok=True,
         )
         self._initialise_global_config()
-        initialise_airflow(self.airflow_home, self.airflow_dags_folder)
-        self._remove_unnecessary_airflow_files()
+        for env in self.environments_list:
+            self._initialise_env_config(env)
+            self._initialise_airflow_for_environment(env)
```
pankajkoti, line 178: How will we take care of initialising Airflow when a user adds a new environment?

pankajkoti: I am thinking if we should provide a command to create a new environment which would take care of it.

sql-cli/sql_cli/project.py
pankajkoti, line 116:
```suggestion
        config = Config(project_dir=self.directory)
```
Why not just this since Config now already has environment = DEFAULT_ENVIRONMENT as the default value?

sql-cli/sql_cli/project.py
tatiana: @feluelle I added comments - how would you suggest simplifying it?

sql-cli/sql_cli/project.py
tatiana: They are not the same thing, `self._airflow_home` is used during setup if the user explicitly specifies it

sql-cli/sql_cli/configuration.py
tatiana: @pankajkoti, at the moment, it returns for all environments - and I'd expect the one using the endpoint to know which environment they are interested in. Do you see challenges with this approach?

sql-cli/sql_cli/project.py
tatiana: >  How will we take care of initialising Airflow when a user adds a new environment?
@pankajkoti that is being handled in the `load_config`

In future we can look into exposing creating a new environment in a separate command, but it feels we don't need this ATM

sql-cli/sql_cli/project.py
tatiana: I cleaned it up - by the time I made this change, that was not the default :)

tatiana: Since I set the default value in `Config`, this is now possible :)

sql-cli/sql_cli/cli/config.py
tatiana: :+1: I'm fine with it, @pankajkoti , it would make it more consistent with the Astro CLI
https://github.com/astronomer/docs/blob/main/astro/cli/astro-deploy.md

sql-cli/sql_cli/cli/config.py
tatiana: Same as before - changing it

pankajkoti (approved):

feluelle (approved):

Issue #1139, opened 2022-10-26
Implement Universal transfer operator [POC]


sunank200:
**Please describe the feature you'd like to see**
Define an interface for the universal transfer operator and implement the POC for the universal transfer operator based on Airflow 2.4 Dataset concept 

Depends on:
- [x] Outline and document the changes in classes based on [Notion doc](https://www.notion.so/astronomerio/Draft-Universal-Transfer-Operator-Proposal-3e3640c79e404ee3b28b39686968e9f7) and [doc](https://docs.google.com/document/d/1TCg7RAm9Vl0DPGmJqKClKxf7Jzscy77AXNsCObgY2Ic/edit) [Spike] - Goal is to define fixed scope for the release.
- [x] Dummy transfer using [FiveTran](https://www.fivetran.com/) [Spike]
- [x] Declare the interface for universal transfer operator
- [x] Use Airflow 2.4 Dataset concept to build more types of Datasets:
   - Table
   - File
   - Dataframe
   - API
- [x] Implement various Data providers based on [Notion doc](https://www.notion.so/astronomerio/Draft-Universal-Transfer-Operator-Proposal-3e3640c79e404ee3b28b39686968e9f7) and [doc]
     - [x] File Data Provider
     - [ ] Database Provider
- Ingestion Operator Handling
- [ ] Add the data transfer implementation with each one of the following use-case (pick first two for this POC)
   - FileSystem to Database transfers (GCS to bigquery)
   - FileSystem to FileSystem transfers (GCS to S3)   
   - API to FileSystem transfers (API to S3)
   - Database to FileSystem transfers (Bigquery to GCS)
   - Database to Database transfers
   - Other types (TBD)
        - Database to API transfers
        - API to Database transfers
        - API to API transfers

- [ ] Leverage any direct, high-speed point-to-point loading mechanisms if already existing such as Bigquery providing a built-in mechanism to load GCS files efficiently into Biqquery.

**Additional context**
- Re-use Python SDK’s code wherever possible
- https://lucid.app/lucidchart/b546293b-4565-4bcb-ae64-46f9cad5cd94/edit?view_items=6g~1jpzPNvvP&invitationId=inv_27271c07-9b7f-420f-baa8-ba087234346d

**Acceptance Criteria**
- [ ] Demo the transfer with an example DAG for each use-cases
- [ ] All checks and tests in the CI should pass
[...]

PR #1267, opened 2022-11-18
Add the interface for universal transfer operator 

sunank200:
# Description
## What is the current behavior?

Add interface for universal transfer operator with working example DAG.

closes: #1139 

## What is the new behavior?

- Define an interface for the universal transfer operator 
- Implement a POC for Fivetran integration
- Implement GCS to S3 transfer

## Does this introduce a breaking change?
No

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1492, opened 2022-12-23
Add universal transfer operator interfaces with working example

sunank200:
# Description

Closing the PR https://github.com/astronomer/astro-sdk/pull/1267 and continuing the change in this fresh PR.

closes: #1139 
closes: #1544 
closes: #1545
closes: #1546
closes: #1551

## What is the new behavior?

- Define interfaces
    -  Use Airflow 2.4 Dataset concept to build more types of Datasets:
        - Table
        - File
        - Dataframe
        - API
    
    - Define an interface for the universal transfer operator 
        - Add the `TransferParameters` class to pass transfer configurations.
        - Use context manager from DataProvider for clean up.
        - Introduce three transfer modes - `native`, `non-native` and `third-party`.
    
    - `DataProviders`
        - Add interface for `DataProvider`.
        - Add interface for `BaseFilesystemProviders`.
        - Add `read` and `write` methods in `DataProviders` with the context manager.
    
    - `TransferIntegrations` and third-party transfers
        - Add interface for `TransferIntegrations` and introduce the third-party transfer approach
 
- Non-native transfers
    - Add `Dataprovider` for S3 and GCS.
    - Add a transfer workflow for S3 to GCS using a non-native approach.
    - Add a transfer workflow for GCS to S3 using a non-native approach.
    - Add example DAG for S3 to GCS implementation.
    - Add example DAG for GCS to S3 implementation.

- Third-party transfers
    - Add `FivetranTransferIntegration` class for all transfers using Fivetran.
    - Implement `FivetranOptions` which inherits from `TransferParameters` class to pass transfer configurations.
    - Implement a POC for Fivetran integration
    - Add example DAG for Fivetran implementation
    - Fivetran POC with working DAG for transfer example (S3 to Snowflake) when `connector_id` is passed.

- Document the APIs for Fivetran transfers on the notion here: https://www.notion.so/astronomerio/Fivetran-3bd9ecfbdcae411faa49cb38595a4571

- MakeFile and Dockerfile along with docker-compose.yaml to build it locally and on the container

No

### Checklist
- [x] Example DAG
- [x] Extended the README/documentation, if necessary


universal_transfer_operator/src/datasets/base.py
```diff
@@ -0,0 +1,60 @@
...
+class UniversalDataset(LoggingMixin, Dataset):
+    """
+    Repersents all file dataset, and abstract away the details related to location and file types.
+    Intended to be used within library.
+
+    :param path: Path to a file in the filesystem/Object stores
+    :param conn_id: Airflow connection ID
+    """
+
+    path: str
```
feluelle, line 29: @sunank200 I thought again about this `path`. I think it is not a good idea to define the absolute path here. Can't we use the connection details to construct the path for operators? So we only require a table name for tables for example as we do in the astro-sdk when converting path to dataset.

feluelle: I am a bit worried about users defining database host in the path such as snowflake. In case of s3 it is not so much of a deal, because you dont authorize against the bucket. But for databases, in my opinion, I see the host as sensitive information.

universal_transfer_operator/src/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+
+        destination_dataprovider = create_dataprovider(
+            dataset=self.destination_dataset,
+            transfer_params=self.transfer_params,
+            transfer_mode=self.transfer_mode,
+            if_exists=self.if_exists,
+        )
+
+        destination_data = destination_dataprovider.write(source_data)
+        cleanup(source_data)
```
feluelle, line 77: It would be great if we could do the actual transfer in a context manager (with `try`, `finally`). So that we `cleanup` also when an error occurs. Something like:
```python
with source_dataprovider.read() as source_data:
    destination_dataprovider.write(source_data)
```

feluelle (changes requested):
It seems the UTO implements a lot of the concepts the astro-sdk already has. Why are we doing that if it still depends on the astro-sdk as a dependency?

Also, I thought the goal is to have UTO open sourced at some point not requiring the astro-sdk? @kaxil? This is related to [composable operator AIP](https://cwiki.apache.org/confluence/display/AIRFLOW/AIP-14+Create+composable+operators+using+common+interfaces). WDYT?

universal_transfer_operator/src/constants.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class TransferMode(Enum):
+    # [START TransferMode]
+    NATIVE = "native"
+    NONNATIVE = "nonnative"
+    THIRDPARTY = "thirdparty"
+    # [END TransferMode]
+
+    def __str__(self) -> str:
```
utkarsharma2, line 43: @sunank200 Can we implement __repr__ instead?

universal_transfer_operator/src/data_providers/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    "s3": "data_providers.filesystem.aws.s3",
+    "aws": "data_providers.filesystem.aws.s3",
+    "gs": "data_providers.filesystem.google.cloud.gcs",
+    "google_cloud_platform": "data_providers.filesystem.google.cloud.gcs",
+}
+
+
+def create_dataprovider(
+    dataset: Dataset,
+    transfer_params: dict = {},
```
utkarsharma2, line 21: We should never use mutable values as defaults. Use None instead and assign a default value within the function call if not provided by the user.

ref: https://medium.com/python-features/how-to-avoid-classic-pitfall-while-passing-default-values-in-python-7002c0dc4c7c

universal_transfer_operator/src/data_providers/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    transfer_mode: TransferMode = TransferMode.NONNATIVE,
+    if_exists: LoadExistStrategy = "replace",
+) -> DataProviders:
+    from airflow.hooks.base import BaseHook
```
utkarsharma2, line 25: Avoiding cyclic dependency?

universal_transfer_operator/src/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+        self,
+        *,
+        source_dataset: Dataset,
+        destination_dataset: Dataset,
+        transfer_params: dict | None = {},
+        transfer_mode: TransferMode | None = TransferMode.NONNATIVE,
+        if_exists: LoadExistStrategy = "replace",
+        **kwargs,
+    ) -> None:
+        self.source_dataset = source_dataset
```
utkarsharma2, line 40: Would it be helpful if we provide the default task_id here? like we have in other operators?

universal_transfer_operator/src/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+        self.destination_dataset = destination_dataset
+        self.transfer_mode = transfer_mode or TransferMode.NONNATIVE
```
utkarsharma2, line 42:
```suggestion
```

universal_transfer_operator/src/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+
+    :return: returns the destination dataset
+    """
+
+    def __init__(
+        self,
+        *,
+        source_dataset: Dataset,
+        destination_dataset: Dataset,
+        transfer_params: dict | None = {},
```
utkarsharma2, line 35: We should use None as default here.

utkarsharma2, line 35:
```suggestion
        transfer_params: dict | None = None,
```

universal_transfer_operator/src/integrations/fivetran.py
```diff
@@ -0,0 +1,288 @@
...
+    daily_sync_time: str = "00:00"
+    schedule_type: str = ""
+    trust_certificates: bool = False
+    trust_fingerprints: bool = False
+    run_setup_tests: bool = True
+
+
+@attr.define
+class FiveTranOptions:
+    retry_limit: int = 3
```
utkarsharma2, line 80: Minor comment, can we specify the units here? like 3 time / min,hour,day?

universal_transfer_operator/src/universal_transfer_operator.py
utkarsharma2, line 35:
```suggestion
        transfer_params: dict | None = None,
```

universal_transfer_operator/src/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+            transfer_integration = get_transfer_integration(self.transfer_params)
+            return transfer_integration.transfer_job(self.source_dataset, self.destination_dataset)
+
+        source_dataprovider = create_dataprovider(
+            dataset=self.source_dataset,
+            transfer_params=self.transfer_params,
+            transfer_mode=self.transfer_mode,
+            if_exists=self.if_exists,
+        )
+        source_data = source_dataprovider.read()
```
utkarsharma2, line 67: is this data stored in memory r.n?

universal_transfer_operator/src/universal_transfer_operator.py
utkarsharma2: +1

universal_transfer_operator/transfers/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+
+from airflow.models import BaseOperator
+from airflow.utils.context import Context
+from transfers.constants import TransferMode
+from transfers.data_providers import cleanup, create_dataprovider
+from transfers.datasets.base import UniversalDataset as Dataset
+from transfers.integrations import get_transfer_integration
+from transfers.utils import check_if_connection_exists
+
+from astro.constants import LoadExistStrategy
```
kaxil, line 13: Let's not depend/import from Astro SDK

kaxil:
> It seems the UTO implements a lot of the concepts the astro-sdk already has. Why are we doing that if it still depends on the astro-sdk as a dependency?
> 
> Also, I thought the goal is to have UTO open sourced at some point not requiring the astro-sdk? @kaxil? This is related to [composable operator AIP](https://cwiki.apache.org/confluence/display/AIRFLOW/AIP-14+Create+composable+operators+using+common+interfaces). WDYT?

Yup you are absolutely right, it is related to composable operator AIP and it won't be depending on Astro SDK; instead it will be the opposite (in an ideal world or a distant future) where Astro SDK will depend on UTO

sunank200:
> > It seems the UTO implements a lot of the concepts the astro-sdk already has. Why are we doing that if it still depends on the astro-sdk as a dependency?
> > Also, I thought the goal is to have UTO open sourced at some point not requiring the astro-sdk? @kaxil? This is related to [composable operator AIP](https://cwiki.apache.org/confluence/display/AIRFLOW/AIP-14+Create+composable+operators+using+common+interfaces). WDYT?
> 
> Yup you are absolutely right, it is related to composable operator AIP and it won't be depending on Astro SDK; instead it will be the opposite (in an ideal world or a distant future) where Astro SDK will depend on UTO

@kaxil @feluelle I will remove the dependency from astro-sdk completely. Its just constants and methods which I was using from astro-sdk. I will remove this on the commit i push today

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -0,0 +1,85 @@
...
+from universal_transfer_operator import UniversalTransferOperator
+
+from constants import TransferMode
+
+START_DATE = datetime(2000, 1, 1)
+with DAG(
+    "example_universal_transfer_operator",
+    schedule_interval=None,
+    start_date=START_DATE,
+    catchup=False,
```
kaxil, line 15:
```suggestion
with DAG(
    "example_universal_transfer_operator",
    schedule_interval=None,
    start_date=datetime(2022, 1, 1),
    catchup=False,
```

kaxil: nit

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -0,0 +1,85 @@
...
+) as dag:
+    transfer_non_native = UniversalTransferOperator(
+        task_id="transfer_non_native",
+        source_dataset=File("gs://uto-test/uto/", conn_id="google_cloud_default", extra={}),
+        destination_dataset=File("s3://astro-sdk-test/uto/", conn_id="aws_default", extra={}),
```
kaxil, line 20: How about just `source` and `destination` instead of `source_dataset` and `destination_dataset`?

What do you and others think? @feluelle @utkarsharma2

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    transfer_fivetran_with_connector_id = UniversalTransferOperator(
+        task_id="transfer_fivetran_with_connector_id",
+        source_dataset=File("s3://astro-sdk-test/uto/", conn_id="aws_default", extra={}),
+        destination_dataset=Table(
+            "snowflake://gp21411.us-east-1.snowflakecomputing.com/providers_fivetran_dev.s3.fivetran_ankit_test",
```
kaxil, line 27: There is a duplication of this info from connection -- schema and other things, no?

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -0,0 +1,85 @@
...
+) as dag:
+    transfer_non_native = UniversalTransferOperator(
+        task_id="transfer_non_native",
+        source_dataset=File("gs://uto-test/uto/", conn_id="google_cloud_default", extra={}),
+        destination_dataset=File("s3://astro-sdk-test/uto/", conn_id="aws_default", extra={}),
+    )
+
+    transfer_fivetran_with_connector_id = UniversalTransferOperator(
+        task_id="transfer_fivetran_with_connector_id",
+        source_dataset=File("s3://astro-sdk-test/uto/", conn_id="aws_default", extra={}),
```
kaxil, line 25: Since `extra` param in this entire file is empty, can we just not pass it, please

kaxil:
```suggestion
        source_dataset=File("s3://astro-sdk-test/uto/", conn_id="aws_default"),
```

universal_transfer_operator/transfers/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        source_dataset: Dataset,
+        destination_dataset: Dataset,
+        transfer_params: dict | None = {},
+        transfer_mode: TransferMode | None = TransferMode.NONNATIVE,
+        if_exists: LoadExistStrategy = "replace",
```
kaxil, line 37: This should be part of `transfer_params`

universal_transfer_operator/transfers/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+        **kwargs,
+    ) -> None:
+        self.source_dataset = source_dataset
+        self.destination_dataset = destination_dataset
+        self.transfer_mode = transfer_mode or TransferMode.NONNATIVE
```
kaxil, line 42: since we have already set a default in constructor, we can remove this:

```suggestion
        self.transfer_mode = transfer_mode
```

universal_transfer_operator/transfers/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+    :return: returns the destination dataset
+    """
+
+    def __init__(
+        self,
+        *,
+        source_dataset: Dataset,
+        destination_dataset: Dataset,
+        transfer_params: dict | None = {},
+        transfer_mode: TransferMode | None = TransferMode.NONNATIVE,
```
kaxil, line 36:
```suggestion
        transfer_mode: TransferMode = TransferMode.NONNATIVE,
```

universal_transfer_operator/transfers/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+        super().__init__(**kwargs)
+
+    def execute(self, context: Context) -> Any:
+        if self.source_dataset.conn_id:
+            if not check_if_connection_exists(self.source_dataset.conn_id):
+                raise ValueError("source_dataset connection does not exist.")
+
+        if self.destination_dataset.conn_id:
+            if not check_if_connection_exists(self.source_dataset.conn_id):
+                raise ValueError("destination_dataset connection does not exist.")
```
kaxil, line 55: I feel this should be part of a `validate` method on the `Table` / `File` dataset itself

universal_transfer_operator/transfers/universal_transfer_operator.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if self.transfer_mode is TransferMode.THIRDPARTY:
```
kaxil, line 57:
```suggestion
        if self.transfer_mode == TransferMode.THIRDPARTY:
```

check https://realpython.com/courses/python-is-identity-vs-equality/

universal_transfer_operator/src/datasets/base.py
kaxil: +1

universal_transfer_operator/src/universal_transfer_operator.py
kaxil: +1

universal_transfer_operator/transfers/data_providers/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    "s3": "data_providers.filesystem.aws.s3",
+    "aws": "data_providers.filesystem.aws.s3",
+    "gs": "data_providers.filesystem.google.cloud.gcs",
+    "google_cloud_platform": "data_providers.filesystem.google.cloud.gcs",
+}
+
+
+def create_dataprovider(
+    dataset: Dataset,
+    transfer_params: dict = {},
```
kaxil, line 21: same as @utkarsharma2 's comment on https://github.com/astronomer/astro-sdk/pull/1492#discussion_r1057553957

universal_transfer_operator/transfers/data_providers/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    transfer_mode: TransferMode = TransferMode.NONNATIVE,
+    if_exists: LoadExistStrategy = "replace",
```
kaxil, line 23: should be part of `transfer_params`

universal_transfer_operator/transfers/data_providers/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+        transfer_mode=transfer_mode,
+        if_exists=if_exists,
+    )
+    return data_provider
+
+
+def cleanup(file_list: list[TempFile]):
+    for file in file_list:
+        if os.path.exists(file.tmp_file.name):
+            os.remove(file.tmp_file.name)
```
kaxil, line 43: Shouldn't the cleanup be part of data-provider too? as how to do cleanup will vary on the type of provider

universal_transfer_operator/transfers/data_providers/base.py
```diff
@@ -0,0 +1,95 @@
...
+    a new module to the `uto/data_providers` directory, without the need of
+    changing other modules and classes.
+    """
+
+    def __init__(
+        self,
+        dataset: Dataset,
+        transfer_mode,
+        transfer_params: dict = {},
+        if_exists: LoadExistStrategy = "replace",
```
kaxil, line 27: `if_exists` should be part of `transfer_params`

universal_transfer_operator/transfers/data_providers/base.py
```diff
@@ -0,0 +1,95 @@
...
+    ):
+        self.dataset = dataset
+        self.transfer_params = transfer_params
+        self.transfer_mode = transfer_mode
+        self.if_exists = if_exists
+        self.transfer_mapping: Any = {}
```
kaxil, line 33: Can you add a comment on what this is for? I can't recall

universal_transfer_operator/transfers/data_providers/base.py
```diff
@@ -0,0 +1,95 @@
...
+    def check_if_exists(self) -> bool:
+        """Return true if the dataset exists"""
+        raise NotImplementedError
+
+    def check_if_transfer_supported(self, source_dataset: Dataset) -> bool:
+        """
+        Checks if the transfer is supported from source to destination based on source_dataset.
+        """
+        source_connection_type = get_dataset_connection_type(source_dataset)
+        return FileLocation(source_connection_type) in self.transfer_mapping
```
kaxil, line 53: Should `FileLocation` be here? in the base class?

universal_transfer_operator/transfers/data_providers/filesystem/aws/s3.py
```diff
@@ -0,0 +1,120 @@
...
+        super().__init__(
+            dataset=dataset,
+            transfer_params=transfer_params,
+            transfer_mode=transfer_mode,
+            if_exists=if_exists,
+        )
+        self.transfer_mapping: set = {
+            FileLocation.AWS,
+            FileLocation.google_cloud_platform,
+        }
```
kaxil, line 33: make this a class variable

universal_transfer_operator/transfers/data_providers/filesystem/aws/s3.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @property
```
kaxil, line 35: use `cached_property`

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,120 @@
...
+        super().__init__(
+            dataset=dataset,
+            transfer_params=transfer_params,
+            transfer_mode=transfer_mode,
+            if_exists=if_exists,
+        )
+        self.transfer_mapping: set = {
+            FileLocation.AWS,
+            FileLocation.google_cloud_platform,
+        }
```
kaxil, line 34: make this class var

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @property
```
kaxil, line 36: cached_property

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,120 @@
...
+    def hook(self) -> GCSHook:
+        """Return an instance of the database-specific Airflow hook."""
+        return GCSHook(
+            gcp_conn_id=self.dataset.conn_id,
+            delegate_to=self.dataset.extra.get("delegate_to", None),
+            impersonation_chain=self.dataset.extra.get("google_impersonation_chain", None),
```
kaxil, line 42: I am trying to think if we want to expose `hook_kwargs` -- might be a low-level details.

Not sure yet, just thiking

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+    def check_if_exists(self) -> bool:
+        """Return true if the dataset exists"""
+        return self.hook.exists(bucket_name=self.get_bucket_name(), object_name=self.get_blob_name())
```
kaxil, line 47: `self.get_bucket_name()` should be converted to `@property` instead I think

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,120 @@
...
+        logging.info(
+            "Getting list of the files. Bucket: %s; Delimiter: %s; Prefix: %s",
+            self.get_bucket_name(),  # type: ignore
+            self.dataset.extra.get("delimiter", None),
+            self.dataset.extra.get("prefix", None),
+        )
+        files = self.hook.list(
+            bucket_name=self.get_bucket_name(),  # type: ignore
+            prefix=self.dataset.extra.get("prefix", None),
+            delimiter=self.dataset.extra.get("delimiter", None),
```
kaxil, line 62: even this 2 should be property I think so we don't have to do `self.dataset.extra.get("prefix", None),` everytime

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+        local_filename = []
+        if files:
```
kaxil, line 65: should we error if no files are found, or expose something if no files found.

(No action needed right now, but for future)

universal_transfer_operator/transfers/datasets/base.py
```diff
@@ -0,0 +1,60 @@
...
+
+    DATASET_SUPPORT = True
+except ImportError:
+    # Airflow < 2.4
+    Dataset = object
+    DATASET_SUPPORT = False
+
+
+@define
+class UniversalDataset(LoggingMixin, Dataset):
```
kaxil, line 20: Where and why do we need this?

universal_transfer_operator/transfers/datasets/base.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Repersents all file dataset, and abstract away the details related to location and file types.
+    Intended to be used within library.
+
+    :param path: Path to a file in the filesystem/Object stores
+    :param conn_id: Airflow connection ID
+    """
+
+    path: str
+    conn_id: str
```
kaxil, line 30: shouldn't be required

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
sunank200: Yes, that's correct but we need a pointer to the table and Open lineage has a really good way of defining that: https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

kaxil: We can generate that internally but let's not duplicate those details

universal_transfer_operator/src/universal_transfer_operator.py
sunank200: Added the context manager here: https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0 Will add the logic for actual transfer in further commits.

universal_transfer_operator/src/data_providers/__init__.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/src/universal_transfer_operator.py
sunank200: `transfer_mode` is required as we support three modes: `native`, `non-native` and `third-party`

universal_transfer_operator/src/integrations/fivetran.py
sunank200: It's for the API call actually.

universal_transfer_operator/src/universal_transfer_operator.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/universal_transfer_operator.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/data_providers/__init__.py
```diff
@@ -0,0 +1,43 @@
+import importlib
+import os.path
+
+from transfers.constants import TransferMode
+from transfers.data_providers.base import DataProviders, TempFile
+from transfers.datasets.base import UniversalDataset as Dataset
+
+from astro.constants import LoadExistStrategy
+from astro.utils.path import get_class_name
```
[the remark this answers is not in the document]
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/universal_transfer_operator.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/universal_transfer_operator.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/universal_transfer_operator.py
sunank200: Added a validator for `conn_id`. [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/universal_transfer_operator.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/data_providers/__init__.py
sunank200: Added a context manager so that clean-up happens in `_exit__()`.  https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0

universal_transfer_operator/transfers/data_providers/base.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/data_providers/base.py
sunank200: No, i have added a class for all Locations supported. [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/831d627d1a86acab300968fc56d11bc7f016b4b0)

universal_transfer_operator/transfers/datasets/base.py
sunank200: File, Table, Dataframe and API classes are common methods that would be part of UniversalDataset. UniversalTransferOperator uses this UniversalDataset only like https://github.com/astronomer/astro-sdk/blob/bc459fa8756fb2612b6a064520a31995dac49c1d/universal_transfer_operator/transfers/universal_transfer_operator.py#L8

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
sunank200: That's a good idea but I wanted to keep the name explicitly clarify user that they need to pass Airflow Dataset URI

universal_transfer_operator/src/universal_transfer_operator.py
sunank200: Done

universal_transfer_operator/src/data_providers/__init__.py
sunank200: Yes but not needed now.

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead)

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead)

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead)

universal_transfer_operator/transfers/data_providers/filesystem/aws/s3.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead)

universal_transfer_operator/transfers/data_providers/__init__.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead)

universal_transfer_operator/src/constants.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead)

universal_transfer_operator/src/universal_transfer_operator.py
sunank200: Using this instead: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

```
from contextlib import contextmanager

@contextmanager
def managed_resource(*args, **kwds):
    # Code to acquire resource, e.g.:
    resource = acquire_resource(*args, **kwds)
    try:
        yield resource
    finally:
        # Code to release resource, e.g.:
        release_resource(resource)
```

sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/ac3af739b548c86636c4519e0c1c7bcadf321ead

universal_transfer_operator/transfers/data_providers/filesystem/google/cloud/gcs.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/22b0aeeb515b6e9eb72dd68f62a64cb807a5732a

universal_transfer_operator/transfers/data_providers/base.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/22b0aeeb515b6e9eb72dd68f62a64cb807a5732a

universal_transfer_operator/transfers/data_providers/__init__.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/22b0aeeb515b6e9eb72dd68f62a64cb807a5732a

universal_transfer_operator/transfers/universal_transfer_operator.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/22b0aeeb515b6e9eb72dd68f62a64cb807a5732a

universal_transfer_operator/src/universal_transfer_operator.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/22b0aeeb515b6e9eb72dd68f62a64cb807a5732a

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
```diff
@@ -0,0 +1,151 @@
...
+            for file in source_ref:
+                if os.path.exists(file.tmp_file.name):
+                    dest_key = os.path.join(dest_s3_key, os.path.basename(file.actual_filename))
+                    self.hook.load_file(
+                        filename=file.tmp_file.name,
+                        key=dest_key,
+                        replace="replace",
+                        acl_policy=self.s3_acl_policy,
+                    )
+                    destination_keys.append(dest_key)
```
dimberman, line 95: This level of nesting is hard to read. Can you separate this into a helper function?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,160 @@
...
+            if files:
+                for file in files:
+                    _, _, file_name = file.rpartition("/")
+                    with NamedTemporaryFile(suffix=file_name, delete=False) as tmp_file:
+                        self.hook.download(
+                            bucket_name=self.bucket_name,
+                            object_name=self.blob_name,
+                            filename=tmp_file.name,
+                        )
+                        local_file_paths.append(TempFile(tmp_file=tmp_file, actual_filename=file_name))
```
dimberman, line 86: This should be its own function for easier testing/readability

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/google/cloud/gcs.py
```diff
@@ -0,0 +1,160 @@
...
+                # There will always be a '/' before file because it is
+                # enforced at instantiation time
+                dest_gcs_object = object_prefix + file
+                self.hook.upload(
+                    bucket_name=bucket_name,
+                    object_name=dest_gcs_object,
+                    filename=file.name,
+                    gzip=self.gzip,
+                )
+                destination_objects.append(dest_gcs_object)
```
dimberman, line 110: same here

universal_transfer_operator/src/universal_transfer_operator/datasets/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # @conn_id.validator
+    # def _validate_conn(self, attr, conn_id: str):
+    #     """Validate the conn_id provided."""
+    #     try:
+    #         from airflow.hooks.base import BaseHook
+    #
+    #         BaseHook.get_connection(conn_id)
+    #     except ValueError:
+    #         raise ValueError("source_dataset connection does not exist.")
```
dimberman, line 44: Why is this commented out?

dimberman (changes requested):
@sunank200 this is looking great, but is there a way to break this into smaller PRs?

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
feluelle: I am fine either way. Normally, I would prefer `source` and `destination` too, but since the operator itself does not mention that it transfers datasets, I think I like it to be explicit in the args.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
feluelle: @sunank200 it would be nice if we somehow could use _pathlib_ here. This would simplify the logic a lot imo. 

For example, something like (pseudo-code):
```python
            for file in source_ref:
                if file.tmp_file.exists():  # checks if the pathlib.Path obj exists
                    self.hook.load_file(
                        filename=file.tmp_file.as_posix(),  # returns pathlib.Path str representation
                        key=dest_s3_key / file.actual_filename.name,  # join path using pathlib "/" operator; `.name` on Path obj returns the last element of the path
                        replace="replace",
                        acl_policy=self.s3_acl_policy,
                    )
                    destination_keys.append(dest_key)
```

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -0,0 +1,88 @@
...
+        source_connection_type = get_dataset_connection_type(source_dataset)
+        method_name = self.LOAD_DATA_FROM_SOURCE.get(source_connection_type)
+        if method_name:
+            transfer_method = self.__getattribute__(method_name)
+            return transfer_method(
+                source_dataset=source_dataset,
+                destination_dataset=destination_dataset,
+            )
+        else:
+            raise ValueError(f"No transfer performed from {source_connection_type} to S3.")
```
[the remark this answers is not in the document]
sunank200: Changed it: https://github.com/astronomer/astro-sdk/pull/1492/commits/29e5ffaf685e27c9b5da3db1659b3442f78f68c9

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
```diff
@@ -0,0 +1,151 @@
...
+    def write(self, source_ref: list[TempFile]):
+        """Write the file from local file location to the dataset"""
+
+        dest_s3_key = self.dataset.path
+
+        if not self.keep_directory_structure and self.prefix:
+            dest_s3_key = os.path.join(dest_s3_key, self.prefix)
+
+        destination_keys = []
+        if source_ref:
```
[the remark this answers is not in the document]
sunank200: Removed it: https://github.com/astronomer/astro-sdk/pull/1492/commits/29e5ffaf685e27c9b5da3db1659b3442f78f68c9

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/google/cloud/gcs.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/29e5ffaf685e27c9b5da3db1659b3442f78f68c9 and https://github.com/astronomer/astro-sdk/pull/1492/commits/535465e155915816db932fb678a30687a400e304

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/google/cloud/gcs.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/29e5ffaf685e27c9b5da3db1659b3442f78f68c9 and https://github.com/astronomer/astro-sdk/pull/1492/commits/535465e155915816db932fb678a30687a400e304

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/29e5ffaf685e27c9b5da3db1659b3442f78f68c9 and https://github.com/astronomer/astro-sdk/pull/1492/commits/535465e155915816db932fb678a30687a400e304

universal_transfer_operator/src/universal_transfer_operator/datasets/base.py
sunank200: Removed it

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
sunank200: Made the change: https://github.com/astronomer/astro-sdk/pull/1492/commits/b59f087700a88b7f24989d0619c89bd06771c8e5

Removed `path` and `conn_id` from dataset. Added `name` to `Table`

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
sunank200: I think the operator itself does not mention that it transfers datasets, like @feluelle said its better to keep `source_dataset` and `destination_dataset`.

universal_transfer_operator/src/datasets/base.py
sunank200: Removed it: https://github.com/astronomer/astro-sdk/pull/1492/commits/b59f087700a88b7f24989d0619c89bd06771c8e5

universal_transfer_operator/transfers/datasets/base.py
sunank200: Removed it: https://github.com/astronomer/astro-sdk/pull/1492/commits/b59f087700a88b7f24989d0619c89bd06771c8e5

universal_transfer_operator/src/universal_transfer_operator/integrations/fivetran.py
```diff
@@ -0,0 +1,292 @@
...
+        if destination_dict is None:
+            raise ValueError("destination is none. Pass a valid destination")
+        destination = Destination(**destination_dict)
+        payload = {
+            "group_id": group_id,
+            "service": destination.service,
+            "region": destination.region if destination.region else "GCP_US_EAST4",
+            "time_zone_offset": destination.time_zone_offset if destination.time_zone_offset else "-5",
+            "config": destination.config,
+            "run_setup_tests": destination.run_setup_tests if destination.run_setup_tests else True,
```
feluelle, line 231: Can we set the defaults in the class definition?

Also why is the default `GCP_US_EAST4` if we don't know the destination will be gcp?

universal_transfer_operator/src/universal_transfer_operator/integrations/fivetran.py
```diff
@@ -0,0 +1,292 @@
...
+    retry_delay: int = 1
+    poll_interval: int = 15
+    schedule_type: str = "manual"
+    connector: Connector = attr.field(factory=Connector)
+
+
+class FivetranIntegration(TransferIntegration):
+    """
+    Class for FiveTran.
+    """
```
feluelle, line 91: Can we add a bit more documentation about what a Fivetran Integration is?

universal_transfer_operator/src/universal_transfer_operator/integrations/base.py
```diff
@@ -0,0 +1,34 @@
...
+from airflow.hooks.base import BaseHook
+
+from universal_transfer_operator.datasets.base import UniversalDataset as Dataset
+
+
+class TransferIntegration(ABC):
+    """
+    Class for third party transfer.
+
+    """
```
feluelle, line 14:
```suggestion
class TransferIntegration(ABC):
    """
    Basic implementation of a third party transfer.
    """
```

feluelle (commented):
Looking really good, but left some comments for you to look at.

I noticed that you also have a lot of ToDo's. It might make sense to create issues for those. So that others could also work on these. WDYT?

universal_transfer_operator/src/universal_transfer_operator/integrations/fivetran.py
sunank200: I have added the default locations to class definitions.

In this case the `region` here is the data processing location. This is where Fivetran will operate and run computation on data. More details can be found here: https://fivetran.com/docs/rest-api/destinations

universal_transfer_operator/src/universal_transfer_operator/integrations/base.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1492/commits/8943d29df9d6d565ce668ceb9c7497aa3775c959

universal_transfer_operator/src/universal_transfer_operator/integrations/fivetran.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/8943d29df9d6d565ce668ceb9c7497aa3775c959)

universal_transfer_operator/src/universal_transfer_operator/integrations/fivetran.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/8943d29df9d6d565ce668ceb9c7497aa3775c959)

feluelle: Thanks for the explanation :)

dimberman (approved):
LGTM. Great job Ankit!

python-sdk/tests/benchmark/analyse.py
```diff
@@ -9,6 +9,7 @@
 from urllib.parse import urlparse
 
 import pandas as pd
+import settings as benchmark_settings
```
feluelle, line 12:
```suggestion
```

python-sdk/tests/benchmark/dags/evaluate_load_file.py
```diff
@@ -6,6 +6,7 @@
 from pathlib import Path
 from urllib.parse import urlparse
 
+import settings as benchmark_settings
```
feluelle, line 9:
```suggestion
```

python-sdk/tests/benchmark/run.py
```diff
@@ -9,6 +9,7 @@
 import airflow
 import pandas as pd
 import psutil
+import settings as benchmark_settings
```
feluelle, line 12:
```suggestion
```

feluelle (approved):

universal_transfer_operator/transfers/datasets/base.py
sunank200: Removed this now as uri is not being passed by user: https://github.com/astronomer/astro-sdk/pull/1492/commits/0f409c31eed44985e27326c89164b5bdbf7f0284

universal_transfer_operator/dev/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM quay.io/astronomer/astro-runtime-dev:7.0.0-alpha5-base
```
kaxil, line 1:
```suggestion
FROM quay.io/astronomer/astro-runtime-dev:7.1.0-base
```

universal_transfer_operator/pyproject.toml
```diff
@@ -0,0 +1,149 @@
...
+tests = [
+    "pytest>=6.0",
+    "pytest-split",
+    "pytest-dotenv",
+    "requests-mock",
+    "pytest-cov",
+    "pytest-describe",
+    "types-requests",
+    "mypy",
+    "sqlalchemy-stubs", # Change when sqlalchemy is upgraded https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
```
kaxil, line 52: Do we need all of this?

universal_transfer_operator/pyproject.toml
```diff
@@ -0,0 +1,149 @@
...
+    "sqlalchemy-bigquery>=1.3.0",
+    "databricks-cli",
+    "apache-airflow-providers-databricks",
+    "s3fs",
+    "protobuf<=3.20", # Google bigquery client require protobuf <= 3.20.0. We can remove the limitation when this limitation is removed
+    "openlineage-airflow>=0.17.0",
+    "airflow-provider-fivetran>=1.1.3",
+    "apache-airflow-providers-microsoft-azure",
+    "azure-storage-blob"
+]
```
kaxil, line 102: can you take a pass at all the deps please -- and add only ones that are used by UTO in this PR

sunank200: Sure. Doing it now

universal_transfer_operator/pyproject.toml
```diff
@@ -0,0 +1,149 @@
...
+doc = [
+    "myst-parser>=0.17",
+    "sphinx>=4.4.0",
+    "sphinx-autoapi>=2.0.0",
+    "sphinx-rtd-theme"
+]
+
+[project.urls]
+Home = "https://astronomer.io/"
+Source = "https://github.com/astronomer/astro-sdk/universal_transfer_operator"
```
kaxil, line 112: This link is broken! even in SDK

universal_transfer_operator/src/universal_transfer_operator/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""A decorator that allows users to run SQL queries natively in Airflow."""
```
kaxil, line 1: Not true anymore for UTO

universal_transfer_operator/src/universal_transfer_operator/data_providers/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+from airflow.hooks.base import BaseHook
+
+from universal_transfer_operator.constants import TransferMode
+from universal_transfer_operator.data_providers.base import DataProviders
+from universal_transfer_operator.datasets.base import Dataset
+from universal_transfer_operator.utils import TransferParameters, get_class_name
+
+DATASET_CONN_ID_TO_DATAPROVIDER_MAPPING = {
+    "s3": "universal_transfer_operator.data_providers.filesystem.aws.s3",
+    "aws": "universal_transfer_operator.data_providers.filesystem.aws.s3",
```
kaxil, line 12: To remove duplication you could do the following -- this is not strictly needed but just an fyi

```suggestion
    **{dict.fromkeys(["aws", "s3"], "universal_transfer_operator.data_providers.filesystem.aws.s3"},
```

kaxil (approved):
Approving to unblock but I wasn't able to review it fully

universal_transfer_operator/src/universal_transfer_operator/data_providers/__init__.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/c2e08eaec89c463d8e83e7a8fa523231e56ede4f)

universal_transfer_operator/src/universal_transfer_operator/__init__.py
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/c2e08eaec89c463d8e83e7a8fa523231e56ede4f)

universal_transfer_operator/pyproject.toml
sunank200: Its not in main branch yet

universal_transfer_operator/pyproject.toml
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/c2e08eaec89c463d8e83e7a8fa523231e56ede4f)

universal_transfer_operator/pyproject.toml
sunank200: [Removed](https://github.com/astronomer/astro-sdk/pull/1492/commits/c2e08eaec89c463d8e83e7a8fa523231e56ede4f)

universal_transfer_operator/pyproject.toml
```diff
@@ -0,0 +1,149 @@
+[build-system]
+requires = ["flit_core ~=3.2"]
+build-backend = "flit_core.buildapi"
+
+[project]
+name = "universal-transfer-operator"
+dynamic = ["version"]
+description = """
+"""
```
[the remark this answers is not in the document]
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/c2e08eaec89c463d8e83e7a8fa523231e56ede4f)

universal_transfer_operator/dev/Dockerfile
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1492/commits/25d47803297b1fd7e4282addb7cf690becd58b11)

PR #1494, opened 2022-12-23
Fix the broken cleantables test caused by databricks config error

dimberman:
# Description

We've recently had failing tests on CI/CD with the following error

```
Py4JJavaError: An error occurred while calling o341.sql.
: java.util.concurrent.ExecutionException: com.databricks.sql.transaction.tahoe.DeltaFileNotFoundException: dbfs:/user/hive/warehouse/non_temp_table/_delta_log/00000000000000000000.json: Unable to reconstruct state at version 1 as the transaction log has been truncated due to manual deletion or the log retention policy (delta.logRetentionDuration=30 days) and checkpoint retention policy (delta.checkpointRetentionDuration=2 days)
```

To the best of my understanding, we are running into an issue where databricks does not like dropping/recreating tables if the underlying data has been changed unless the user sets versioning in their S3 bucket. This might have been caused during autoloader testing. I have tested this new changed table multiple times and it seems to be working fine. So if the tests pass here this should hold us over until we can set up versioning post-holidays.


feluelle (approved):

pankajastro (approved):

pankajastro:
I'm merging this PR as a couple of my PR's block because CI is failing.

Issue #913, opened 2022-09-26
Execute astro-sdk integration tests on astro cloud


phanikumv:
**Please describe the feature you'd like to see**
Enable astro sdk  integration tests to run on astro cloud, similar to providers team integration tests.

**Describe the solution you'd like**
A clear and concise description of what you want to happen, if possible with code examples.

**Are there any alternatives to this feature?**
Is there another way we could solve this problem or enable this use-case?

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

sunank200:
@kelechi2020 Can you please confirm that we are running the astro sdk integration tests on Astro-runtime?

kelechi2020:
We are able to run the astro-sdk example dags on Astro-runtime.

pankajastro:
> We are able to run the astro-sdk example dags on Astro-runtime.

How did you deploy it on astro, I mean do you have some automation script for deployment?

pankajastro:
@kaxil Any thoughts on how we can run integration tests (not example DAG)?

kaxil:
This ticket was more about running the Example DAGs on Astro Cloud similar to how we do those for astronomer-providers

pankajastro:
We would need to run a Postgres instance to test examples related to Postgres where should we run that?

phanikumv:
we can run it on azure or any cloud of your choice - consider adding a task containing a start up / shutdown script in order to not take up too much costs

phanikumv:
Implementation is 50% done. @pankajastro to connect with @ash / @uranusjr and discuss regarding the postgres connection issue.

sunank200:
[...]

PR #1499, opened 2022-12-28
Deploy python-sdk on astro

pankajastro:
# Description
closes: #913 

We want to run our example DAG on astro cloud and collect the result 

## What is the new behavior?
- Add a CI job which will trigger daily and deploy our example DAG and install astro-sdk from main branch in astro-cloud image
- Add a master DAG which will trigger all example DAG

No

### Checklist


dimberman:
@kaxil @tatiana @pankajastro is there any security concerns with having astro deployment as a part of CI/CD? Perhaps we should only do this with release branches instead of with  the main branch? WDYT?

pankajastro:
> @kaxil @tatiana @pankajastro is there any security concerns with having astro deployment as a part of CI/CD? Perhaps we should only do this with release branches instead of with the main branch? WDYT?

Hi @dimberman We will run this only on the main branch as a nightly job and not on each merge. Also, CI/CD will take care of the deployment of our example DAG on astro cloud only and we have a master DAG which will take care of running our example DAG at the scheduled time. We do similar things in the [provider](https://github.com/astronomer/astronomer-providers). Please do let me know if you think there can be any specific potential security issues.

python-sdk/tests_integration/astro_deploy/master_dag.py
```diff
@@ -0,0 +1,201 @@
...
+
+def get_report(dag_run_ids: List[str], **context: Any) -> None:
+    """Fetch dags run details and generate report"""
+    with create_session() as session:
+        last_dags_runs: List[DagRun] = session.query(DagRun).filter(DagRun.run_id.in_(dag_run_ids)).all()
+        message_list: List[str] = []
+
+        airflow_version = context["ti"].xcom_pull(task_ids="get_airflow_version")
+        airflow_version_message = f"Airflow version for the below run is `{airflow_version}` \n\n"
+        message_list.append(airflow_version_message)
```
pankajkoti, line 30: Maybe we can add another line at the beginning saying this alert is for Python SDK integration tests since we're sending report to the same provider-alert channel now.

.github/workflows/ci-benchmark.yaml
```diff
@@ -17,7 +18,8 @@ jobs:
       github.event_name == 'schedule' ||
       (
         github.event.label.name == 'run_benchmark' &&
-        github.event.name != 'pull_request_target'
+        github.event.name != 'pull_request_target' &&
+        github.event.schedule != '0 0 * * *'
```
pankajkoti, line 22: Are we sure with this that the Benchmark job will keep continue running on Monday at 00:00. Because, I think with this it may never run.

.github/workflows/ci-benchmark.yaml
```diff
@@ -56,3 +58,17 @@ jobs:
       - run: ( echo "cat <<EOF >python-sdk/test-connections.yaml"; cat .github/ci-test-connections.yaml; ) >python-sdk/test-connections.yaml && . python-sdk/test-connections.yaml
       - run: python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
       - run: cd python-sdk/tests/benchmark && make
+
+  Astro-deploy:
+    if: >-
+      github.event_name == 'schedule' && github.event.schedule != '0 0 * * 1'
```
pankajkoti, line 64: Similarly I think with this check, our deploy job may not run at all on Mondays.

pankajkoti (commented):
Maybe we should change the schedule for the benchmark job from `'0 0 * * 1'` to '30 3 * * 1' so that it is easier to separate out the jobs based on schedule events in the if conditionals and also we allow sufficient time for the deploy job to complete before running the benchmark script on Mondays.

python-sdk/tests_integration/astro_deploy/master_dag.py
pankajastro: rephrase it let me know it looks now

.github/workflows/ci-benchmark.yaml
pankajastro: earlier check condition was not right, please let me know I have updated the check

.github/workflows/ci-benchmark.yaml
pankajastro: earlier check condition was not right, please let me know I have updated the check

python-sdk/tests_integration/astro_deploy/deploy.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
```
feluelle, line 1: Can't we deploy using astro-cli? I think that is safer and we don't need to go into details that much e.g. doing http requests.

python-sdk/tests_integration/astro_deploy/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM quay.io/astronomer/astro-runtime:7.1.0-base
```
tatiana, line 1: @pankajastro would it make sense to use an earlier version of the Runtime (>= 7.2.0), which has the Python SDK included?

pankajastro: yup, we should use the latest runtime image

python-sdk/tests_integration/astro_deploy/deploy.sh
tatiana: @pankajastro @rajaths010494, did you consider the possibility of using the Astro CLI? I'm curious if this approach was considered or ruled out for any reason.

Since this PR has been long-standing, could we also have this as a separate ticket (to change the deployment to use Astro CLI) and try to close this PR ASAP - WDYT?

feluelle (commented):
Let's please try to use https://github.com/astronomer/deploy-action

cc @tatiana

python-sdk/tests_integration/astro_deploy/deploy.sh
rajaths010494: We had used the same approach in providers repo so we used the same here. If already ticket exists for enhancement. If not we can create one

python-sdk/tests_integration/astro_deploy/Dockerfile
rajaths010494: Updated to 7.2.0

python-sdk/tests_integration/astro_deploy/deploy.sh
feluelle: This script adds a lot of maintenance for us. Let us please implement the github action or cli, but not a script from scratch. I know this is being used in the providers repo. We should also replace this there.

rajaths010494: Yeah sure will create a ticket in provider repo as well to replace the script.

pankajastro: I tried to use it around 10 months back in the provider repo but at that time we had some roadblocks in making it work with full automation for custom images. But, I think it makes sense to look at docs and retry since lots have changed since then.

rajaths010494:
https://github.com/astronomer/astro-sdk/issues/1737

python-sdk/dev/docker-compose.yaml
```diff
@@ -18,7 +18,7 @@ x-airflow-common:
     AIRFLOW__WEBSERVER__EXPOSE_CONFIG: "True"
     AIRFLOW__SCHEDULER__DAG_DIR_LIST_INTERVAL: "5"
     ASTRONOMER_ENVIRONMENT: local
-    AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES: airflow\.* astro\.*
+    # AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES: airflow\.* astro\.*
```
sunank200, line 21: Shouldn't this be uncommented @rajaths010494 ?

rajaths010494: reverted back

utkarsharma2 (approved):

python-sdk/tests_integration/astro_deploy/deploy.sh
sunank200: Created a ticket to use astro-deploy in a separate ticket: https://github.com/astronomer/astro-sdk/issues/1742
cc: @rajaths010494 @phanikumv

sunank200:
> Let's please try to use https://github.com/astronomer/deploy-action
> 
> cc @tatiana

Created a separate ticket for this: https://github.com/astronomer/astro-sdk/issues/1742

PR #1500, opened 2022-12-28
Use `autospec=True` for mocks

kaxil:
https://docs.python.org/3.9/library/unittest.mock.html#autospeccing and https://nedbatchelder.com/blog/202202/why_your_mock_still_doesnt_work.html explains why

### DIFF
--- a/python-sdk/tests/databases/test_bigquery.py
+++ b/python-sdk/tests/databases/test_bigquery.py
@@ -125,7 +125,7 @@ def test_get_run_id():
         ),
     ],
 )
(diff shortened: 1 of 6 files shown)


kaxil:
cc @bharanidharan14 @rajaths010494 -- we should do the same for Providers too

bharanidharan14:
> cc @bharanidharan14 @rajaths010494 -- we should do the same for Providers too

@kaxil created a ticket for this in providers 
https://github.com/astronomer/astronomer-providers/issues/828

pankajastro (approved):

PR #1503, opened 2022-12-30
Deprecate `export_table_to_file` to `export_to_file`

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1530
Related to: https://astronomer.slack.com/archives/C03868KGF2Q/p1671621286602989

Recently, we deprecated `export_file` to `export_table_to_file` but looks like `export_to_file` is a better name fit for this

## What is the new behavior?
Deprecate  to `export_table_to_file` to  `export_to_file` 

### Checklist

### DIFF
--- a/python-sdk/docs/astro/sql/operators/export.rst
+++ b/python-sdk/docs/astro/sql/operators/export.rst
@@ -1,16 +1,16 @@
-.. _export_table_to_file:
+.. _export_to_file:
 
 ==================================================================================
-:py:mod:`export_table_to_file operator <astro.sql.operators.export_table_to_file>`
+:py:mod:`export_to_file operator <astro.sql.operators.export_to_file>`
 ==================================================================================
 
-.. _export_table_to_file_operator:
+.. _export_to_file_operator:
 
-When to use the ``export_table_to_file`` operator
+When to use the ``export_to_file`` operator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The ``export_table_to_file`` operator allows you to write SQL tables to CSV or parquet files and store them locally, on S3, or on GCS. The ``export_table_to_file`` function can export data from :ref:`supported_databases` or a Pandas dataframe.
+The ``export_to_file`` operator allows you to write SQL tables to CSV or parquet files and store them locally, on S3, or on GCS. The ``export_to_file`` function can export data from :ref:`supported_databases` or a Pandas dataframe.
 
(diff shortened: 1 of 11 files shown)


python-sdk/docs/astro/sql/operators/export.rst
```diff
@@ -1,16 +1,16 @@
...
+:py:mod:`export_to_file operator <astro.sql.operators.export_to_file>`
 ==================================================================================
 
-.. _export_table_to_file_operator:
+.. _export_to_file_operator:
 
-When to use the ``export_table_to_file`` operator
+When to use the ``export_to_file`` operator
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The ``export_table_to_file`` operator allows you to write SQL tables to CSV or parquet files and store them locally, on S3, or on GCS. The ``export_table_to_file`` function can export data from :ref:`supported_databases` or a Pandas dataframe.
```
tatiana, line 11: @pankajastro, what do you think if we deprecated `export_table_to_file` before removing it? 

Very few users would be using it since it has been available between 1.3.1-1.3.3 (from 14 December to 4 January), but that may be a less abrupt way of making this change.

We're planning to remove it by 1.5, so we have a deadline to remove this - and add a Github ticket associated with the 1.5 milestones, so we don't forget it.

tatiana (commented):
@pankajastro since this task allows users to export both from SQL tables to file and from Pandas Dataframes to file, the change is relevant and essential. Ideally, this should be part of 1.4.

The one open question is whether we want a smooth transition to deprecate it (my +1) or to make this change more abrupt.

python-sdk/docs/astro/sql/operators/export.rst
pankajastro: @utkarsharma2 @sunank200 @phanikumv @rajaths010494 @kaxil WDYT?

feluelle: I like that suggestion from tatiana. 👍

pankajastro: deprecated and created a ticket to remove it in 1.5.0 https://github.com/astronomer/astro-sdk/issues/1521

tatiana (approved):
Thanks, @pankajastro, for adding the deprecation warning - and I'm very sorry for not being more critical of the previous PR. No more renaming of this task for the next 2 releases at least :joy:

utkarsharma2 (approved):

PR #1504, opened 2022-12-30
Fix benchmark job

pankajastro:
# Description
## What is the new behavior?

-
-
-

### Checklist


.github/workflows/ci-benchmark.yaml
```diff
@@ -49,10 +49,15 @@ jobs:
       AIRFLOW__ASTRO_SDK__DATAFRAME_ALLOW_UNSAFE_STORAGE: True
       FORCE_COLOR: "true"
       ASTRO_PUBLISH_BENCHMARK_DATA: True
+      USE_GKE_GCLOUD_AUTH_PLUGIN: True
```
feluelle, line 52: Don't we need to set this env var also in the Dockerfile?

.github/workflows/ci-benchmark.yaml
```diff
@@ -49,10 +49,15 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
       - run: rm -f python-sdk/test-connections.yaml
       - run: ls -la
       - run: ( echo "cat <<EOF >python-sdk/test-connections.yaml"; cat .github/ci-test-connections.yaml; ) >python-sdk/test-connections.yaml && . python-sdk/test-connections.yaml
       - run: python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
+      - run: echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+      - run: curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+      - run: sudo apt update
```
feluelle, line 61: It would be nice if we could `name` the `- run:` jobs and split them into meaningful sections. WDYT?

python-sdk/tests/benchmark/Makefile
```diff
@@ -6,7 +6,6 @@ GIT_HASH = $(shell git log -1 --format=%h)
 APP ?= benchmark
 CONTAINER_REGISTRY=gcr.io/$(GCP_PROJECT)/$(APP)
 ASTRO_TARGET_PLATFORM ?= "linux/amd64"
-
```
feluelle, line 9: I think the empty line made sense, no? Could you revert this change, please?

.github/workflows/ci-benchmark.yaml
pankajastro: look like currently there is some issue while installing it in GitHub action https://github.com/actions/runner-images/issues/6778#issuecomment-1360360603

python-sdk/tests/benchmark/Makefile
```diff
@@ -6,7 +6,6 @@ GIT_HASH = $(shell git log -1 --format=%h)
 APP ?= benchmark
 CONTAINER_REGISTRY=gcr.io/$(GCP_PROJECT)/$(APP)
 ASTRO_TARGET_PLATFORM ?= "linux/amd64"
```
pankajastro, line 8:
```suggestion
ASTRO_TARGET_PLATFORM ?= "linux/amd64"

```

.github/workflows/ci-benchmark.yaml
pankajastro: I split it into `Install google-cloud-sdk-auth-plugin` `Export environment variable` and `Run benchmark `.

.github/workflows/ci-benchmark.yaml
feluelle: Not sure I understand. I meant we need to add the env var in the Dockerfile, too.

pankajastro: this is currently a workaround to install it in GitHub action and it does not require in docker

feluelle: I don't think it is. It is how it is supposed to work.

The following is from an email, I got this morning:

> To run kubectl with the new plugin prior to the release of v1.26
> 
> To have kubectl use the new binary plugin for authentication instead of using the default provider-specific code, use the following steps.
> 
> Set export USE_GKE_GCLOUD_AUTH_PLUGIN=True in ~/.bashrc (or in Environment variables for Windows).
> Run the following command:
> source ~/.bashrc
> For Windows environments, start a new terminal.
> Update gcloud to the latest version.
> gcloud components update
> Run the following command:
> gcloud container clusters get-credentials CLUSTER_NAME
> Replace the CLUSTER_NAME with the name of your cluster. This will force the config for this cluster to be updated to the Client-go Credential Plugin configuration.
> Full and latest instructions are available at the [Kubectl auth changes in GKE v1.26](https://notifications.google.com/g/p/AM0OveryzzeUSXvh-iaTJQA97-SCYO0kQZz_BS1_D1LqZwStEq7BffHgYZ_XBeZxBY0uptzkbje-62xt2AjtYwwKOU9fkjCZw2WAxufF0kbLQC4b2-T-aXC_uDtymP2NwQT4KHMYY_ZJj5NuteqIcwetMa1U04R5jPueF0aMYlGL6uMw79YXSzw4_7i1nAuferA0FC3g_9UO8wp0j327osRdTp4z4A1rTnmV_XbtqfIWugkM9-4c2eogmNLFsBf1ZhQVdqCMHjVINw) blogpost. Installation instructions above, are copied for convenience.

feluelle: You can also see it in the link of the quote above^

pankajastro: can't install it with a component update since the runner image comes with google-cloud-sdk  maybe because it install with apt and throws an error
https://github.com/actions/runner-images/blob/ubuntu20/20230109.1/images/linux/Ubuntu2004-Readme.md
```
Beginning update. This process may take several minutes.
ERROR: (gcloud.components.update) 
You cannot perform this action because the Google Cloud CLI component manager 
is disabled for this installation. You can run the following command 
to achieve the same result for this installation: 
```

I tried to install it with apt only but it was not able to find package 
```
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package google-cloud-cli-gke-gcloud-auth-plugin
```

.github/workflows/ci-benchmark.yaml
pankajastro: ok, we need this to run the gcloud/kubectl cmd which is a part of Run benchmark make file

pankajastro: moved this in `Run benchmark`

python-sdk/tests/benchmark/Dockerfile
```diff
@@ -17,16 +17,18 @@ ENV GIT_HASH ${GIT_HASH}
...
-
-RUN apt install -y libpq-dev python3-dev
+RUN apt-get update && \
+    apt-get install -y python3-pip && \
+    apt-get install -y wget && \
+    apt install -y jq && \
+    apt install -y vim && \
+    apt install -y curl && \
+    echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
+    curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - && \
```
tatiana, line 27: Do we need to add all these lines here? In particular, the `kubectl` related ones?
@pankajastro and I were looking superficially, and it seems the one point where we need `kubectl` is in the `Makefile`. It may be worth checking if these lines need to be changed - if not, we could remove them. If we need to keep them, then we probably should be setting the environment variable.

pankajastro: I have removed `google-cloud-sdk-auth-plugin` from dockerfile

.github/workflows/ci-benchmark.yaml
pankajastro: removed `google-cloud-cli-gke-gcloud-auth-plugin` from docker image since currently, we do not use any kubectl command in docker. We use it only in Makefile that we run on github action host

tatiana (approved):

.github/workflows/ci-benchmark.yaml
feluelle: Alright. Do we then need changes to the Dockerfile at all?

pankajastro: No, but currently there are lots of `RUN` commands I have just grouped them.

pankajastro:
Merging this but @feluelle please let me know if still disagree on https://github.com/astronomer/astro-sdk/pull/1504/files#r1071855675 I'll revisit and resolve it in a separate PR.

PR #1506, opened 2023-01-03
Scope pytest fixtures

feluelle:
# Description

Our tests are becoming significantly slow because we are not setting a proper scope for our pytest fixtures. 

## What is the new behavior?

With the new change, the time for nox session `test-3.8(airflow='2.5')` has  been reduced from ~6m down to ~2m i.e. a 3x speed improvement.

No.

### Checklist

### DIFF
--- a/sql-cli/conftest.py
+++ b/sql-cli/conftest.py
@@ -1,7 +1,5 @@
 import logging
-import random
 import shutil
-import string
 from pathlib import Path
 
 import pytest
@@ -9,11 +7,11 @@ from airflow.models import DAG, Connection, DagRun, TaskInstance as TI
 from airflow.utils import timezone
 from airflow.utils.session import create_session
 
-from astro.table import MAX_TABLE_NAME_LENGTH
 from sql_cli.constants import EXT_LOGGER_NAMES, LOGGER_NAME
 from sql_cli.dag_generator import Workflow
 from sql_cli.project import Project
 from sql_cli.workflow_directory_parser import SqlFile, WorkflowFile
+from tests.utils import create_unique_table_name
 
 CWD = Path(__file__).parent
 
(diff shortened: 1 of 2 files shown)


sql-cli/conftest.py
```diff
@@ -146,22 +144,14 @@ def sample_dag():
...
-    unique_id = random.choice(string.ascii_lowercase) + "".join(
-        random.choice(string.ascii_lowercase + string.digits) for _ in range(length - 1)
-    )
-    return unique_id
+@pytest.fixture(scope="module")
+def tmp_path_cached(tmp_path_factory):
+    return tmp_path_factory.mktemp("sql-cli")
 
 
-@pytest.fixture()
```
tatiana, line 162: It is a very impressive reduction in the time processing, @feluelle !

I'd like to understand a bit more about the side effects of this change. Does setting the scope to the `module` mean that all the test functions defined in a test module will be reusing the same fixture (same initialised project)? Could that cause any unexpected behaviours?

feluelle: > I'd like to understand a bit more about the side effects of this change. Does setting the scope to the module mean that all the test functions defined in a test module will be reusing the same fixture (same initialised project)? 

This is correct.

> Could that cause any unexpected behaviours?

Yes. It might not be obvious at first, but from the fixture name e.g. `initialised_project`, I think it makes sense to say this is an already initialised project. It is not called `initialise_project`. 😬 WDYT? I think currently it does not make sense to "always" rerun the fixture. The intention of it was that you get an initialised project when you use it. 🤔 

> Why did we set the scope as "session" in most of the other fixtures and "module" here?

Great question. I tested it out and it seems that we can only use `module` (init once per `test_` file) as different modules have quite different expectations/assertions as they do quite different things. E.g. The "generate" logic generates files in the dags folder. Now, if another test is `assert`ing on the files in the project, it will fail because there are already generated files. So we need to be careful when setting the `scope`.

FYI `session` means once per test run/session.

tatiana: @feluelle, I see. It is primarily a trade-off between speed and test setup/tear-down isolation. It is a fair compromise, for the time being, to have the speed gain - and review in future the fixture scope if this causes us issues.

Regarding the name, I don't think it is clear that `initialised_project` is a fixture that will be reused across all tests in the module :joy: But the scope decorator does clarify this. As always, it's always hard to name things :see_no_evil:  Would `module_initialised_project` make things more straightforward? I'm also happy to keep the name as it is!

tatiana (approved):
Happy for the PR to be merged after rebasing :)

pankajkoti (approved):
This is great. Overall across all Airflow versions, it is 5 min+ faster as I can see that previously, 2.3.4 used to take almost 18 mins I believe, it is down to 13 mins now.

dimberman (approved):
Great catch, @feluelle!

sql-cli/conftest.py
feluelle: Naming is hard - indeed. As hard as it is I can't think of a proper name for it.

Issue #1510, opened 2023-01-03
Fix databricks flaky tests


pankajastro:
**Describe the bug**
I'm hitting the below error frequently in CI
https://github.com/astronomer/astro-sdk/actions/runs/3829415846/jobs/6516092626
```
FAILED tests_integration/databricks_tests/test_delta.py::test_create_table_from_select_statement[delta] - AssertionError: assert 'Stanley' == 'Dancer'
  - Dancer
  + Stanley
ERROR tests_integration/databricks_tests/test_delta.py::test_delta_create_table_with_columns[delta] - databricks.sql.exc.ServerOperationError: org.apache.hadoop.hive.ql.metadata.HiveException: at least one column must be specified for the table
= 1 failed, 46 ***ed, 509 deselected, 39 warnings, 1 error in 614.05s (0:10:14) =
nox > Command pytest tests_integration/ --splits 12 --group 4 --cov=src --cov-report=xml --cov-branch failed with exit code 1
nox > Session test-3.8(airflow='2.5.0') failed.
Error: Process completed with exit code 1.
```

PR #1511, opened 2023-01-03
Fix databricks flaky tests

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1510

Look like select stmt does not return value in the same order always or load does not happen in the same order so sometimes value assertion is failing 

## What is the new behavior?
check value exists in any row of a column

No

### Checklist

### DIFF
--- a/python-sdk/tests_integration/databricks_tests/test_delta.py
+++ b/python-sdk/tests_integration/databricks_tests/test_delta.py
@@ -131,7 +131,6 @@ def test_create_table_from_select_statement(database_table_fixture):
 
     df = database.hook.get_pandas_df(f"SELECT * FROM {target_table.name}")
     assert len(df) == 26
-    assert df.NAME[0] == "Dancer"
     database.drop_table(target_table)
 
 
--- a/python-sdk/tests_integration/databricks_tests/test_delta.py
+++ b/python-sdk/tests_integration/databricks_tests/test_delta.py
@@ -131,7 +131,7 @@ def test_create_table_from_select_statement(database_table_fixture):
 
     df = database.hook.get_pandas_df(f"SELECT * FROM {target_table.name}")
     assert len(df) == 26
-    assert "Dancer" in df.NAME
+    assert "Dancer" in df["NAME"].tolist()
     database.drop_table(target_table)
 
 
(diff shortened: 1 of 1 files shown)


python-sdk/tests_integration/databricks_tests/test_delta.py
```diff
@@ -131,7 +131,6 @@ def test_create_table_from_select_statement(database_table_fixture):
 
     df = database.hook.get_pandas_df(f"SELECT * FROM {target_table.name}")
     assert len(df) == 26
-    assert df.NAME[0] == "Dancer"
```
feluelle, line 134: Does it make sense to check if the value is in any field - instead of always the first one?

pankajkoti (commented):
Can we please update the PR description regarding the new behaviour as per the latest commits?

python-sdk/tests_integration/databricks_tests/test_delta.py
pankajastro: make sense, checking if value is in any

dimberman: I'm a bit confused as to why this would be flaky? Seems like the table should be loaded the same way it was in the file

pankajastro: From the test looks like we are loading a folder. I checked the folder in s3 and it has two csv files, one file first row has the value "Dancer" for the field "Name" and the other file has the value "Stanley" for the field "Name".

error we are getting in CI
```
FAILED tests_integration/databricks_tests/test_delta.py::test_create_table_from_select_statement[delta] - AssertionError: assert 'Stanley' == 'Dancer'
  - Dancer
  + Stanley
```
https://github.com/astronomer/astro-sdk/actions/runs/3829415846/jobs/6516092626
https://github.com/astronomer/astro-sdk/actions/runs/3831362066/jobs/6520398696

pankajastro: @dimberman ^^

dimberman: Ohhh ok so I guess loading to databricks isn't deterministic. (I guess that it makes sense since it's using a distributed spark cluster to load)

dimberman (approved):

PR #1515, opened 2023-01-03
Create a Databricks Notebook Operator

dimberman:
# Description

There is no way to run an existing notebook in a Databricks cluster using the Astro SDK. While users can run notebooks using the DatabricksSubmitRunOperator, this operator is a part of the OSS airflow library and will not have the future benefits of Workflow support and lineage.

## What is the new behavior?

This PR introduces a `DatabricksNotebookOperator` that allows users to point to an existing notebook in their Databricks cluster and run it using the Astro SDK. We will later decide how this operator is interfaced to users using our functional API, but this PR should create the basis for future notebook support.

No, this PR does not introduce any breaking changes.

### Checklist

### DIFF
--- a/python-sdk/src/astro/databricks/api_utils.py
+++ b/python-sdk/src/astro/databricks/api_utils.py
@@ -168,7 +168,16 @@ def create_and_run_job(
         python_params=None,
         spark_submit_params=None,
     )["run_id"]
+    monitor_job(api_client=api_client, run_id=run_id)
 
+
+def monitor_job(api_client: ApiClient, run_id: str):
+    """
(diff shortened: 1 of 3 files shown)


tatiana (commented):
@dimberman I know this is in the draft, but since you asked for us to have a quick look... would it be possible to add a description to the PR to give us more context (could be how to use it, or links to Databrics notebooks)

dimberman:
@tatiana I fixed the description so hopefully that makes it clearer :)

dimberman:
Closing this as we are putting this work into cosmos

PR #1516, opened 2023-01-03
Add a snowflake load options

dimberman:
# Description
The current behavior for loading data to Snowflake is to depend on `native_kwargs`, which are loosely defined and lack type safety. There is also no way to specify `FILE_FORMAT` options in Snowflake, which can block users from being able to load their data.

## What is the new behavior?
This PR introduces a `SnowflakeLoadOptions` class that contains `file_format_options` and `copy_options` fields. These fields allow users to specify the necessary arguments to correctly load their data.

No, this does not introduce a breaking change.

### Checklist


python-sdk/src/astro/databases/base.py
```diff
@@ -415,7 +415,7 @@ def load_file_to_table(
         native_support_kwargs: dict | None = None,
         columns_names_capitalization: ColumnCapitalization = "original",
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
-        load_options: LoadOptions | None = None,
+        load_options: LoadOptions = LoadOptions(),
```
tatiana, line 418: I'm thrilled with how we're heading into unifying the `LoadOptions` and having subclasses specific to the database. It is a significant improvement since we introduced `native_kwargs`.

I was curious about the design decision here to create, by default, an instance of the class - instead of having `None`. Could you share the reason, @dimberman?

dimberman: @tatiana I was worried we'd have to create Loadoptions the whole way down as certain functions require one but I found a workaround so was able to get rid of this.

python-sdk/src/astro/databases/base.py
```diff
@@ -521,11 +522,13 @@ def load_file_to_table_natively_with_fallback(
         native_support_kwargs: dict | None = None,
         enable_native_fallback: bool | None = LOAD_FILE_ENABLE_NATIVE_FALLBACK,
         chunk_size: int = DEFAULT_CHUNK_SIZE,
+        load_options: LoadOptions | None = None,
         **kwargs,
     ):
         """
         Load content of a file in output_table.
 
+        :param load_options:
```
feluelle, line 531:
```suggestion
        :param load_options: Options for database specific loading
         parameters (e.g. SnowflakeLoadOptions or DeltaLoadOptions)
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -601,7 +609,13 @@ def load_file_to_table_natively(
         """
         native_support_kwargs = native_support_kwargs or {}
         storage_integration = native_support_kwargs.get("storage_integration")
-        stage = self.create_stage(file=source_file, storage_integration=storage_integration)
+        if not load_options:
+            load_options = SnowflakeLoadOptions()
+        if not isinstance(load_options, SnowflakeLoadOptions):
+            raise ValueError("Error: Requires a SnowflakeLoadOptions")
```
feluelle, line 615: Please cover this with a test! Thanks.

python-sdk/src/astro/options.py
```diff
@@ -5,3 +5,12 @@
...
         return NotImplementedError()
+
+
+@attr.define
+class SnowflakeLoadOptions(LoadOptions):
+    file_options: dict = attr.field(init=True, factory=dict)
+    copy_options: dict = attr.field(init=True, factory=dict)
+
+    def empty(self):
+        return not self.file_options and not self.copy_options
```
feluelle, line 16: Same here. I think it is important to cover this.

pankajastro (commented):
does it make sense to add one example in our docs? for example how can users set file_format (RECORD_DELIMITER in case of CSV) using the SnowflakeLoadOptions file_options param?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -601,7 +609,13 @@ def load_file_to_table_natively(
...
+        stage = self.create_stage(
+            file=source_file, storage_integration=storage_integration, load_options=load_options
+        )
 
         table_name = self.get_table_qualified_name(target_table)
         file_path = os.path.basename(source_file.path) or ""
```
pankajkoti, line 621: Can we move this block and the below line to construct the `sql_statement` into `_copy_into_table_from_stage` method. Because otherwise the `_copy_into_table_from_stage` is just executing a `sql_statement` and checking for exception; it does not give a hint of what the method name means, the `sql_statement` is abstract to the method

feluelle (approved):

dimberman:
@pankajastro I've created a ticket for wider documentation around LoadOptions. I don't want to hold up the 1.4 release over this so I think it's fine to do a more thorough job here. https://github.com/astronomer/astro-sdk/issues/1556

python-sdk/src/astro/options.py
dimberman: added test

python-sdk/src/astro/databases/snowflake.py
dimberman: done

PR #1531, opened 2023-01-05
Docs: add "Operators" to "Reference" 

### LINKED ISSUE #694: Docs: add "Operators" to "Reference" and to have "Concepts" include some more high level documentation
Looks a lot cleaner! I don't feel strongly either way, but in the future it might make sense to add "Operators" to "Reference" and to have "Concepts" include some more high level documentation. Knowing what each operator does _seems_ more referency to me. 

But because "Concepts" is only one doc for now, I think this structure works for the content we have.

_Originally posted by @jwitz in https://github.com/astronomer/astro-sdk/pull/692#pullrequestreview-1077794972_

rajaths010494:
Move  "Operators"  from "Guides" section to "Reference" section.
closes #694


pankajastro (commented):
Hey @rajaths010494 are you planning to add the concept docs for operators also here?

rajaths010494:
> Hey @rajaths010494 are you planning to add the concept docs for operators also here?

Is there any new concepts which would need to me to add in docs.

pankajastro:
> future it might make sense to add "Operators" to "Reference" and to have "Concepts" include some more high level documentation

ticket description says 
> future it might make sense to add "Operators" to "Reference" and to have "Concepts" include some more high level documentation

pankajastro:
Also, the PR description does not look correct.

sunank200:
> Also, the PR description does not look correct.

+1

sunank200:
Agree with @pankajastro on the comment he made on the issue. I am not very clear with issue description and PR description as well.

cc: @rajaths010494

pankajkoti (commented):
The description says Concepts to have some more high level documentation, do we plan to add anything in Concepts? If not can we update the description?  Also, then since it only partly addresses the ticket, I think we should change the hash tag from `closes` to `related` in the description.

pankajastro:
Response from user
> @pankajastro We can close this issue because my thoughts have evolved a bit since I wrote that message. Overall, I think the SDK guides should focus on "How to" guides for common actions/ pipeline structures, rather than for the operators themselves. So example, instead of a guide called drop table operator, we have a guide called "Drop a table"

Issue #1529, opened 2023-01-05
Change flow generate so it can receive the optional --output-dir argument


tatiana:
SQL CLI users need to clearly distinguish between the local environment (where they can use run and generate to play with things locally) from a remote deployment (when they intentionally decide what should be sent to a remote Airflow instance).

For this reason, we decided not to use `--airflow-dags-folder` to implement the SQL CLI deployment journey and to add a new optional parameter, `--output-dir` to `flow generate` to allow users to define where they want the DAGs to be exported to.

Issue #1520, opened 2023-01-04
Expose the deployment of SQL CLI workflows in the Astro CLI


tatiana:
**Context**

Using cobra, implement the command `astro flow deploy` in the Astro CLI. As a starting point, this command will work on Astro Cloud (ticket to support deploying to software: #1522), and it will deploy using the image-based approach (ticket to improve to use `--dags`: #1523).

For this first version of deployment, we are focused on deploying DAGs with expanded tasks (without using the `render` approach).

To deploy a single workflow:
```
astro flow deploy –all
```

To deploy multiple workflows:
```
astro flow deploy <workflow-name> --env dev
```
If the parameter `--env` is not defined, it should use the environment `default`.

**Behaviour**

1. Identify the SQL CLI deployment/environment configuration using #1517. If this task is not completed yet, have a function with the same interface and hard-coded values.

2. Identify the Python SDK availability/version compatibility using #1473. If this task is not completed, have a function with the same interface and hard-coded values.

3. If the folder `<project-dir>/.deploy/<env>/` does not exist, create it

4. Copy any dags from `<project-dir>/dags` into the folder (3) - so we don't delete them while deploying. Python DAGs are being run in Astro CLI. Perform this action silently.

5. Generate the SQL workflows dags of interest (one or all) into the folder (3). Perform this action silently.
    * #1529
    * #1528

6. (Future, post 0.3.0) check if the connections used in the SQL workflow are available in the target deployment environment. Notify users if they are not.
    * #1524
    * #1525
    * #1526

7. (Future, post 0.3.0) if the connections are defined locally, but no remotely, prompt users if they would like for them to be created.
    * #1527
 
8. Build the Docker image. Perform this action silently, reusing the existing logic in the `astro deploy` command.

9. Check if there are errors in the DAGs in (3), reusing the existing logic in the `astro deploy` command.

10. Push the Docker image to the container registry, reusing the existing logic in the `astro deploy` command.

11. Deploy reusing the existing logic in the `astro deploy` command.

12. Notify users about the success with the following message:

```
The deployment was successful!

The SQL workflows can be accessed using the following links:
* example_dag_basic: https://astronomer.astronomer.run/djhb6ppo/dags/example_dag_basic/graph
* example_dag_advanced: https://astronomer.astronomer.run/djhb6ppo/dags/example_dag_advanced/graph
```

tatiana:
I just extracted the following part, reducing the scope of this ticket:
[...]

PR #1533, opened 2023-01-05
Add output-dir option to flow generate

tatiana:
Fix: #1529

### DIFF
--- a/sql-cli/tests/test___main__.py
+++ b/sql-cli/tests/test___main__.py
@@ -247,6 +247,25 @@ def test_generate_invalid(workflow_name, message, initialised_project_with_tests
     assert message in result.stdout
 
 
+def test_generate_with_output_dir(initialised_project, tmp_path):
+    result = runner.invoke(
+        app,
+        [
+            "generate",
+            "example_basic_transform",
+            "--project-dir",
+            initialised_project.directory.as_posix(),
+            "--output-dir",
+            tmp_path.as_posix(),
+        ],
+    )
+    expected_dag_file_path = tmp_path / "example_basic_transform.py"
+    assert expected_dag_file_path.exists()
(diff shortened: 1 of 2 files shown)


tatiana:
I'm adding this tiny change so that we don't block #1520

sql-cli/tests/test___main__.py
```diff
@@ -245,6 +245,25 @@ def test_generate_invalid(workflow_name, message, initialised_project_with_tests
...
+            initialised_project.directory.as_posix(),
+            "--output-dir",
+            initialised_project.directory.as_posix(),
+        ],
+    )
+    expected_dag_file_path = initialised_project.directory / "example_basic_transform.py"
+    assert expected_dag_file_path.exists()
+    message = "has been successfully generated."
+    assert message in result.stdout
+    assert not result.exit_code
```
feluelle, line 264:
```suggestion
    assert result.exit_code == 0
```
I think this shows more clearly that you expect it to succeed.

sql-cli/tests/test___main__.py
```diff
@@ -247,6 +247,25 @@ def test_generate_invalid(workflow_name, message, initialised_project_with_tests
     assert message in result.stdout
 
 
+def test_generate_with_output_dir(initialised_project):
```
pankajkoti, line 250: how about we pass the tmp_path as a fixture here and use that as the output_dir ?

pankajkoti (approved):
LGTM, minor suggestion in the test if you feel can be checked.

sql-cli/tests/test___main__.py
tatiana: Great suggestion, @pankajkoti, particularly now that we are reusing the initialised project per test module!

feluelle (approved):

PR #1536, opened 2023-01-05
Fix pre-commit issues

feluelle:
This is related to https://github.com/astronomer/astro-sdk/commit/7d1a613d6b5620027d54e4d0da0c2112470f475c and tries to fix flaky pre-commit changes, specific to isort import ordering.

Additionally, it fixes a mypy issues:
```
python-sdk/tests/benchmark/analyse.py:133: error: Incompatible default for argument "output_filepath" (default has type "None", argument has type "str")  [assignment]
```

### DIFF
--- a/python-sdk/tests/benchmark/analyse.py
+++ b/python-sdk/tests/benchmark/analyse.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+from __future__ import annotations
 
 import argparse
 import json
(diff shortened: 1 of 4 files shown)


tatiana:
Thanks for checking this out, @feluelle ! I'm relieved to see the pre-commit bot didn't change the order of settings in the benchmark - but why/how?

pankajkoti (approved):

tatiana (approved):

Issue #1532, opened 2023-01-05
Change load_options to accept list


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, in the `load_file()` operator we are passing `load_options` params which accept LoadOptions as a single value. Since we can have a requirement to pass multiple value we need a list/dictionary of LoadOptions objects instead of single instance of LoadOptions.

**Describe the solution you'd like**
We can change the type of  `load_options` to list/dict and use it across the codebase.

**Are there any alternatives to this feature?**
Post discussion with the team, this is the best option at the moment.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1540, opened 2023-01-06
Change load_options to list[LoadOptions] from LoadOptions

utkarsharma2:
# Description
In load_file() we use multiple integrations with Pandas, and Databases at the same time.  But We currently just have one parameter to pass in integration-specific parameters. 

closes: https://github.com/astronomer/astro-sdk/issues/1532

## What is the new behavior?
We have changed load_options params to List[LoadOptions] so that we can pass multiple integration parameters in a list. 
Passing multiple values to `load_options` leads to the problem of mapping. How do we know that PandasLoadOptions need to be passed to dataframe specific methods?

We solved the mapping problem by having integration classes having `LOAD_OPTIONS_CLASS_NAME` parameter to store the `LoadOptions` subclass for that integration and initialize the File-type, File-location, and Database objects with the right LoadOption class.

No, since we never released the `load_options`

### Checklist


python-sdk/conftest.py
```diff
@@ -81,13 +81,7 @@ def database_table_fixture(request):
     database.populate_table_metadata(table)
     database.create_schema_if_needed(table.metadata.schema)
 
-    kwargs = {}
```
tatiana, line 84: Why did we change this?

python-sdk/src/astro/dataframes/load_options.py
```diff
@@ -0,0 +1,32 @@
...
+from astro.options import LoadOptions
+
+
+@define
+class PandasLoadOptions(LoadOptions):
+    pass
+
+
+@define
+class CsvLoadOption(PandasLoadOptions):
```
tatiana, line 15: I believe it is very misleading to have `CsvLoadOption` which is specific to Pandas. I think, for now, we should focus on having database/pandas-specific Options classes.

python-sdk/src/astro/dataframes/load_options.py
```diff
@@ -0,0 +1,32 @@
...
+    delimiter: str | None = None
+    dtype: DtypeArg | None = None
+
+
+@define
+class JsonLoadOption(PandasLoadOptions):
```
tatiana, line 21: I noticed some options we use `Options` and others `Option`. I suggest we stick to `Options`, since these classes will usually contain a group of configurations

python-sdk/src/astro/files/base.py
```diff
@@ -13,6 +13,7 @@
 from astro.files.locations import create_file_location
 from astro.files.locations.base import BaseFileLocation
 from astro.files.types import FileType, create_file_type
+from astro.options import LoadOptions, LoadOptionsList
```
tatiana, line 16: What is the benefit of having a class to represent a list of `LoadOptions`?

python-sdk/src/astro/files/types/base.py
```diff
@@ -5,19 +5,25 @@
...
 
+from astro.options import LoadOptions
+
 
 class FileType(ABC):
     """Abstract File type class, meant to be the interface to all client code for all supported file types"""
 
-    def __init__(self, path: str, normalize_config: dict | None = None):
+    def __init__(
+        self, path: str, normalize_config: dict | None = None, load_options: LoadOptions | None = None
```
tatiana, line 15: I think it would be clearer if the interface here explicitly had a list of `LoadOptions`

python-sdk/conftest.py
dimberman: Yeah I'd also be interested in why we changed this

python-sdk/src/astro/databases/base.py
```diff
@@ -68,9 +68,12 @@ class BaseDatabase(ABC):
     NATIVE_AUTODETECT_SCHEMA_CONFIG: Mapping[FileLocation, Mapping[str, list[FileType] | Callable]] = {}
     FILE_PATTERN_BASED_AUTODETECT_SCHEMA_SUPPORTED: set[FileLocation] = set()
 
-    def __init__(self, conn_id: str, table: BaseTable | None = None):  # skipcq: PYL-W0613
+    def __init__(
+        self, conn_id: str, table: BaseTable | None = None, load_options: LoadOptions | None = None
```
dimberman, line 72: Does it make sense to make load_options a database level member? Is there a scenario where we'd want to load with different options using the same database? cc: @tatiana @kaxil

python-sdk/src/astro/databricks/delta.py
```diff
@@ -24,11 +24,15 @@
...
+
+    LOAD_OPTIONS_CLASS_NAME = "DeltaLoadOptions"
+
     _create_table_statement: str = "CREATE TABLE IF NOT EXISTS {} USING DELTA AS {} "
 
-    def __init__(self, conn_id: str, table: BaseTable | None = None):
+    def __init__(self, conn_id: str, table: BaseTable | None = None, load_options: LoadOptions | None = None):
         super().__init__(conn_id)
         self.table = table
+        self.load_options = load_options or DeltaLoadOptions.get_default_delta_options()
```
dimberman, line 35: @tatiana I think this is why utkarsh was able to get rid of the lines we asked about here https://github.com/astronomer/astro-sdk/pull/1540/files#r1063468038.

python-sdk/src/astro/dataframes/load_options.py
dimberman: @tatiana maybe we have a `LoadFormat` class instead as a class member of the `PandasLoadOptions` class?

python-sdk/src/astro/files/base.py
dimberman: Yeah I'm against this, it should just be a list of LoadOptions.

dimberman (changes requested):
I'll probably have more comments later but for now lets' address the first round of comments from @tatiana and I

python-sdk/conftest.py
utkarsharma2: @dimberman @tatiana I have added the Load_option code as part of the Database and File/Location and File/Type classed because of which can pass the load_option as initialization of those classes. Hence we can remove it from this place and others. There are still places we can remove such code, I was thinking to do this in a separate PR.

python-sdk/src/astro/dataframes/load_options.py
utkarsharma2: cc: @pankajastro

utkarsharma2: I too like this suggestion, we should stick with `options`

python-sdk/src/astro/files/base.py
utkarsharma2: Only one benefit, since now we have a list of options we need to find the relevant options to an integration class. So I have added two helper methods to access relevant classes by passing the name of the class. 
example - 
 `get_by_class_name("DeltaLoadOptions")`

Also, this is not user-facing I intend to use it within the operator.

@dimberman I'm happy to change this, but we either would require to replicate code or create a helper method in utils to get the relevant option class. But regardless we do need a method for this, where do you suggest we keep it?

python-sdk/src/astro/files/types/base.py
utkarsharma2: @tatiana Wouldn't we need to update this every time we add a new file type? Would be better if we add base `FileLoadOptions` inheriting from `LoadOptions` and use it here?

python-sdk/src/astro/databases/base.py
utkarsharma2: So far we are not using multiple loading options per DB/File, I think it makes sense to put them at the initialization step. Anyway, we can still cater to the scenario of using different load options by providing helper methods to update those on the fly.

python-sdk/src/astro/databricks/delta.py
utkarsharma2: Yup, correct.

python-sdk/src/astro/dataframes/load_options.py
pankajastro: @tatiana for pandas options we call different methods of pandas lib based on the file type like `read_csv`, `read_parquet` etc and these methods accept different param. I added `CsvLoadOptions` because it would be easy to understand for a user to understand what they can pass if they are loading CSV file using pandas also if we will have a class instead of dict editor will also provide suggestions. But as @utkarsharma2 rightly pointed out the origin of this chnage is https://github.com/astronomer/astro-sdk/pull/1466 so we can take this discussion there.

python-sdk/src/astro/files/base.py
utkarsharma2: As per discussion with @tatiana and @pankajastro We need to change the implementation of LoadOptionsList to use dict instead of list to optimise the fetching of loadOption class.

python-sdk/src/astro/options.py
```diff
@@ -1,7 +1,41 @@
...
+        """
+        Get load_option by class name
+        :return:
+        """
+        if self._load_options is None:
+            return None
+        for option in self._load_options:
+            if type(option).__name__ == option_class_name:
+                return option
+        return None
```
tatiana, line 41: @utkarsharma2 since we are assuming that the relationship between `LoadOptions` class & `Database`/`FileFormat`/`FileLocation` is 1:1 and this class is only used internally, we can simplify the code by internally presenting the options given by users as a dictionary. My hope is that this would simplify the changes in methods such as `create_database`

python-sdk/src/astro/databases/__init__.py
```diff
@@ -37,5 +40,7 @@ def create_database(conn_id: str, table: BaseTable | None = None) -> BaseDatabas
     module_path = CONN_TYPE_TO_MODULE_PATH[conn_type]
     module = importlib.import_module(module_path)
     class_name = get_class_name(module_ref=module, suffix="Database")
-    database: BaseDatabase = getattr(module, class_name)(conn_id, table)
+    database: BaseDatabase = getattr(module, class_name)(
+        conn_id, table, load_options=load_options_list and load_options_list.get(getattr(module, class_name))
+    )
```
tatiana, line 45: I was wondering if we'd be able to simplify the code using something like this:
```suggestion
    database_class = getattr(module, class_name)
    load_options_list = class_to_load_options.get(database_class, load_options_list)
    database: BaseDatabase = database_class(
        conn_id, table, load_options=)
    )
```

dimberman (approved):

python-sdk/src/astro/databases/base.py
dimberman: Ok I think I'm ok with this since users never actually create database objects.

python-sdk/src/astro/files/types/base.py
utkarsharma2: As per our discussion load_options should be a object because we never use multiple load_option for a integration.

python-sdk/src/astro/databases/__init__.py
utkarsharma2: Added in a septate commit.

Issue #1447, opened 2022-12-16
Allow load and export from Azure Blob storage (WASB) using non-native path


tatiana:
**Please describe the feature you'd like to see**
Once we've introduced the Azure Blob Storage file location (WASB) #905 #1275 the next step is to make sure we can have load and export operations using the non-native path.

**Describe the solution you'd like**
Users should be able to load and export files using WASB using non-native path

**Sub Tasks**
- [ ] https://github.com/astronomer/astro-sdk/issues/1486
- [ ] https://github.com/astronomer/astro-sdk/issues/1485

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

tatiana:
Supporting @luanmorenomaciel and  @mateusoliveiraowshq to get this done :)

PR #1542, opened 2023-01-06
Load/export file from Azure blob storage using non-native path

tatiana:
Closes: #1447

Bonus:
- Improve and standardise `export_to_file` tests related to remote file locations so we have a consistent way of testing them (all file locations x all remote object stores)
- Fix the GCS and WASB location `exists` method

Pending
- Azure + Databricks support (separate ticket)

Co-authored-by: Luan Moreno <luan.moreno@owshq.com>
Co-authored-by: Mateus Oliveira <mateus.oliveira@owshq.com>

### DIFF
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -131,7 +131,9 @@ class File(LoggingMixin, Dataset):
     ) -> pd.DataFrame:
         """Read file from all supported location and convert them into dataframes."""
         mode = "rb" if self.is_binary() else "r"
-        with smart_open.open(self.location.smartopen_uri, mode=mode, transport_params=self.location.transport_params) as stream:
+        with smart_open.open(
+            self.location.smartopen_uri, mode=mode, transport_params=self.location.transport_params
+        ) as stream:
             return self.type.export_to_dataframe(stream, load_options, **kwargs)
 
     def _convert_remote_file_to_byte_stream(self) -> io.IOBase:


python-sdk/noxfile.py
```diff
@@ -60,7 +60,13 @@ def test(session: nox.Session, airflow) -> None:
     session.run("airflow", "db", "init", env=env)
 
     # Since pytest is not installed in the nox session directly, we need to set `external=true`.
-    session.run("pytest", *session.posargs, env=env, external=True)
+    session.run(
+        "pytest",
+        "-vv",
```
pankajastro, line 65: @tatiana why do you want to run test with `-vv` option?

pankajastro (approved):
LGTM. But look like CI has not triggered on the last commit.

python-sdk/noxfile.py
tatiana: @pankajastro I always find it useful to have a summary of the tests which are passing/failing while we run the tests - but we can remove it from this PR :)

tatiana:
@pankajastro thanks! There is something very odd going on with our Github actions; it is triggering the jobs but sometimes it's not deploying here :thinking: 
https://github.com/astronomer/astro-sdk/actions/runs/3874982912

pankajastro:
> @pankajastro thanks! There is something very odd going on with our Github actions; it is triggering the jobs but sometimes it's not deploying here 🤔
> https://github.com/astronomer/astro-sdk/actions/runs/3874982912

I did a force push then it showing up 
Screenshot 2023-01-09 at 9 44 59 PM

PR #1548, opened 2023-01-08
Use databricks one-time jobs instead of the jobs API

dimberman:
# Description
Currently, we find our jobs queue becoming filled with all of the databricks jobs we are submitting any time we run load_file. This has recently led to us reaching a quota error where databricks would not allow us to submit new jobs.

## What is the new behavior?
Using the RunsApi.submit_run function, we can submit a one-time job that does not persist in the jobs tab of databricks. This is more aligned with our current use-case as each run will be by definition unique.

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/databricks/api_utils.py
+++ b/python-sdk/src/astro/databricks/api_utils.py
@@ -8,7 +8,6 @@ from typing import Dict
 from airflow.configuration import conf
 from airflow.exceptions import AirflowException
 from databricks_cli.dbfs.api import DbfsApi, DbfsPath
-from databricks_cli.jobs.api import JobsApi
 from databricks_cli.runs.api import RunsApi
 from databricks_cli.sdk.api_client import ApiClient
 from databricks_cli.secrets.api import SecretApi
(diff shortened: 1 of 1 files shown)


python-sdk/src/astro/databricks/api_utils.py
```diff
@@ -158,16 +157,7 @@ def create_and_run_job(
...
-    run_id = jobs_api.run_now(
-        job_id=job["job_id"],
-        jar_params=None,
-        notebook_params=None,
-        python_params=None,
-        spark_submit_params=None,
-    )["run_id"]
+    run_id = runs_api.submit_run(json=json_info)["run_id"]
 
     runs_api = RunsApi(api_client)
```
pankajkoti, line 162:
```suggestion
```
We can remove this re-instantiation of RunsApi()

phanikumv (commented):
Do  we need to add in the documentation that we'll use the existing cluster id to submit jobs to  databricks api ?

dimberman:
@pankajkoti that is already mentioned in the documentation. Users can either offer a cluster id as an env variable/argument or create a new json if they want a custom cluster.

pankajkoti:
> @pankajkoti that is already mentioned in the documentation. Users can either offer a cluster id as an env variable/argument or create a new json if they want a custom cluster.

@dimberman , perhaps you meant to tag @phanikumv and not me. :)

dimberman:
oh oops yes, that's the case :). @pankajkoti while you're here can you approve this PR?

pankajkoti (approved):

phanikumv (approved):

PR #1552, opened 2023-01-09
Remove drop_table() statement

utkarsharma2:
Since we are using fixtures that takes care of the deletion of temp tables


tatiana (approved):

tatiana (commented):
Happy for this cleanup to be merged once it is rebase and the checks are passing

Issue #1535, opened 2023-01-05
Print error message in validate command


feluelle:
As mentioned in https://github.com/astronomer/astro-sdk/pull/1177#discussion_r1060660703, the "validate" command should output error messages. So that as a user I know what I can do to fix an invalid connection.

PR #1554, opened 2023-01-09
Add error message for validate connections

feluelle:
# Description

Currently, we are not giving any hints to the user about an invalid/failed connection.

closes: #1535

## What is the new behavior?

Print an error message when validating a connection.

Yes, kind of. We are not erroring out anymore when the file path internally could not get resolved in case of sqlite connections. This is so that an invalid sqlite connection does not block you from running any generate/validate/run commands. Previously it was raising an exception which led to an immediate exit.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/connections.py
+++ b/sql-cli/sql_cli/connections.py
@@ -33,8 +33,7 @@ def convert_to_connection(conn: dict[str, Any], data_dir: Path) -> Connection:
         resolved_host = data_dir / connection["host"]
         if not resolved_host.is_file():
             log.error(
-                "The relative file path %s was resolved into %s"
-                " but it does not exist.",
+                "The relative file path %s was resolved into %s" " but it does not exist.",
                 connection["host"],
                 resolved_host,
             )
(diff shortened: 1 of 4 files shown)


pankajkoti (approved):

sql-cli/sql_cli/connections.py
```diff
@@ -32,9 +32,11 @@ def convert_to_connection(conn: dict[str, Any], data_dir: Path) -> Connection:
         # Try resolving with data directory
         resolved_host = data_dir / connection["host"]
         if not resolved_host.is_file():
-            raise FileNotFoundError(
-                f"The relative file path {connection['host']} was resolved into {resolved_host}"
-                " but it's a failed resolution as the path does not exist."
+            log.error(
+                "The relative file path %s was resolved into %s"
+                " but it's a failed resolution as the path does not exist.",
```
tatiana, line 37:
```suggestion
                " but it does not exist.",
```

tatiana (approved):

sql-cli/sql_cli/connections.py
```diff
@@ -33,8 +33,7 @@ def convert_to_connection(conn: dict[str, Any], data_dir: Path) -> Connection:
         resolved_host = data_dir / connection["host"]
         if not resolved_host.is_file():
             log.error(
-                "The relative file path %s was resolved into %s"
-                " but it does not exist.",
+                "The relative file path %s was resolved into %s" " but it does not exist.",
```
feluelle, line 36:
```suggestion
                "The relative file path %s was resolved into %s but it does not exist.",
```

PR #1561, opened 2023-01-10
Add support to load from Azure blob storage into Databricks

tatiana:
Closes: #1250

### DIFF
--- a/python-sdk/src/astro/files/locations/azure/wasb.py
+++ b/python-sdk/src/astro/files/locations/azure/wasb.py
@@ -11,6 +11,10 @@ from astro.files.locations.base import BaseFileLocation
 from astro.options import contains_required_option
 
 
+class WASBLocationException(Exception):
+    pass
+
+
 class WASBLocation(BaseFileLocation):
     """Handler WASB object store operations"""
 
@@ -115,3 +119,27 @@ class WASBLocation(BaseFileLocation):
             netloc=f"{self.load_options.storage_account}.{self.AZURE_HOST}",  # type: ignore
         )
         return url.geturl()
+
+    def databricks_auth_settings(self) -> dict:
+        """
+        Required settings to transfer files in/to Databricks.
+
+        Reference:
+        https://docs.databricks.com/external-data/wasb-blob.html
+
+        :return: A dictionary of settings keys to settings values
+        """
+        parsed_url = urlparse(self.path)
+        container_name = parsed_url.netloc
+        account_name = self.hook.get_conn().account_name
+        try:
+            sas_token = self.hook.get_connection(conn_id=self.conn_id).extra_dejson["sas_token"]
+        except KeyError:
+            raise WASBLocationException(
+                "The connection extras must define `sas_token` for transfers from BlobStorage to Databricks using WASB"
+            )
+
+        cred_dict = {
+            f"fs.azure.sas.{container_name}.{account_name}.blob.core.windows.net": sas_token,
+        }
+        return cred_dict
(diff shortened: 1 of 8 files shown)


python-sdk/tests_integration/databricks_tests/test_load.py
```diff
@@ -84,6 +84,37 @@ def test_delta_load_file_gcs(database_table_fixture):
...
+    "database_table_fixture",
+    [
+        {
+            "database": Database.DELTA,
+        }
+    ],
+    indirect=True,
+    ids=["delta"],
+)
+def test_delta_load_file_azure_wasb(database_table_fixture, remote_files_fixture):
```
tatiana, line 107: It would be great if we unified the Databricks tests with the other databases' tests, where possible.

python-sdk/tests_integration/databricks_tests/test_load.py
```diff
@@ -84,6 +84,37 @@ def test_delta_load_file_gcs(database_table_fixture):
...
+    file_ = File(path=remote_files_fixture[0], conn_id="wasb_conn_with_sas_token")
+    database, table = database_table_fixture
+    database.load_file_to_table(
```
tatiana, line 110: At the moment, this test fails at `src/astro/databricks/load_file/load_file_job.py:88: in load_file_to_delta
    create_and_run_job` with the exception:
```
 E           airflow.exceptions.AirflowException: Databricks job failed. Job info ***'job_id': 764637050444463, 'run_id': 6507882, 'creator_user_name': 'daniel@astronomer.io', 'number_in_job': 6507882, 'state': ***'life_cycle_state': 'TERMINATED', 'result_state': 'FAILED', 'state_message': '', 'user_cancelled_or_timedout': False***, 'task': ***'spark_python_task': ***'python_file': 'dbfs:/mnt/pyscripts/load_file__tmp_fd003tmarfyucghqvccomivabj0fvuyzwq99b9a55bd01nndamj4p38gf.py'***, 'cluster_spec': ***'existing_cluster_id': '***'***, 'cluster_instance': ***'cluster_id': '***', 'spark_context_id': '7878723590138424515'***, 'start_time': 1673345570903, 'setup_duration': 1000, 'execution_duration': 667000, 'cleanup_duration': 0, 'end_time': 1673346238960, 'run_name': 'Untitled', 'run_page_url': 'https://dbc-9c390870-65ef.cloud.databricks.com/?o=4256138892007661#job/764637050444463/run/6507882', 'run_type': 'SUBMIT_RUN', 'attempt_number': 0, 'format': 'SINGLE_TASK'***
```
https://github.com/astronomer/astro-sdk/actions/runs/3882266043/jobs/6622222255

dimberman (changes requested):
Hey @tatiana this looks great! Also fantastic that we can add new sources by just adding one function :). Can you please add an example DAG and update the documentation to discuss any steps a user would need to take to load data from Azure?

python-sdk/src/astro/files/locations/azure/wasb.py
```diff
@@ -89,3 +93,27 @@ def openlineage_dataset_name(self) -> str:
...
+        Required settings to transfer files in/to Databricks.
+
+        Reference:
+        https://docs.databricks.com/external-data/wasb-blob.html
+
+        :return: A dictionary of settings keys to settings values
+        """
+        parsed_url = urlparse(self.path)
+        container_name = parsed_url.netloc
+        account_name = self.hook.get_conn().account_name
```
dimberman, line 108: Is this the only way to get an azure key or are there other ways a user might get these? I just want to understand if there are configurations a user might have that would break this.

dimberman (approved):

python-sdk/src/astro/files/locations/azure/wasb.py
tatiana: We explored a few paths to authenticate and agreed that we could start with one - and if the community is interested in more, we can extend the approach.

PR #1563, opened 2023-01-10
Update regex for `allowed_deserialization_classes`

kaxil:
This will not allow classes from packages like `astrox` etc but will allow `astro.`

This change only affects docs, tests and example dag and not the source code

### Checklist


phanikumv (commented):
We should add this change in `/docs/guides/xcom_backend.rst` and `dev/docker-compose.yml` s well.

tatiana (approved):

feluelle (approved):

phanikumv (approved):
Redshift test is failing , its unrelated though.

Issue #1559, opened 2023-01-10
Update document for load_file() operator


utkarsharma2:
**Please describe the feature you'd like to see**
Since we have added load_options as part of the interface of the load_file() operator, we should update the documents to use load_options instead of native_kwargs.

pankajastro:
closed by https://github.com/astronomer/astro-sdk/pull/1567

Issue #1558, opened 2023-01-10
Add example dag for list load_options


utkarsharma2:
**Please describe the feature you'd like to see**
We should add example dags demonstrating the use of load_options.

PR #1566, opened 2023-01-11
Update load_file op. to add load_options param in document

utkarsharma2:
# Description
Update document to add load_options parameter in load_file() operator. 

closes: #1559 #1558 

## What is the new behavior?
1. Updated example dag to showcase the use of the `load_options` parameter
2. Updated document to reflect the new `load_options` parameter.

Nope

### DIFF
--- a/python-sdk/docs/astro/sql/operators/load_file.rst
+++ b/python-sdk/docs/astro/sql/operators/load_file.rst
@@ -382,16 +382,20 @@ Load_options
 
  We can use this parameter to pass options that can be used by integrations SDK use internally. For example, there can be a simple case of passing a ``delimiter`` while loading CSV to pandas.read_csv() method.
 
-    .. literalinclude:: ../../../../example_dags/example_load_file.py
-       :language: python
-       :start-after: [START load_file_example_22]
-       :end-before: [END load_file_example_22]
+ .. note::
 
-    ``load_options`` is a list of different :py:class:`LoadOptions <astro.options.LoadOptions>` class objects. Please refer the list below for all valid options.
+    ``load_options`` will be replacing ``native_support_kwargs`` parameter
 
-    - :py:class:`PandasCsvLoadOptions <astro.dataframes.load_options.PandasCsvLoadOptions>`
-    - :py:class:`PandasJsonLoadOptions <astro.dataframes.load_options.PandasJsonLoadOptions>`
-    - :py:class:`PandasNdjsonLoadOptions <astro.dataframes.load_options.PandasNdjsonLoadOptions>`
-    - :py:class:`PandasParquetLoadOptions <astro.dataframes.load_options.PandasParquetLoadOptions>`
-    - :py:class:`DeltaLoadOptions <astro.databases.databricks.load_options.DeltaLoadOptions>`
-    - :py:class:`SnowflakeLoadOptions <astro.options.SnowflakeLoadOptions>`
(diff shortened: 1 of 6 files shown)


python-sdk/docs/astro/sql/operators/load_file.rst
```diff
@@ -76,6 +76,8 @@ Parameters to use when loading a file to a database table
        :start-after: [START load_file_example_3]
        :end-before: [END load_file_example_3]
 
+#. **load_options** - `load_options`_
```
tatiana, line 79: We probably want to give more details about what `load_options` is an which values it takes, referencing examples where possible

python-sdk/docs/astro/sql/operators/load_file.rst
```diff
@@ -116,6 +118,7 @@ Parameters to use when loading a file to a Pandas dataframe
        :start-after: [START load_file_example_6]
        :end-before: [END load_file_example_6]
 
+#. **load_options** - `load_options`_
```
tatiana, line 121: Same here; the description is not helping to understand what these options are. It is probably worth to also say that the `native_support_kwargs` will be replaced by this

python-sdk/docs/astro/sql/operators/load_file.rst
utkarsharma2: @tatiana Just checking, I have added this in one section at the end of this file and referenced it here. Below is the content of that section, do you feel we need more here?
```
We can use this parameter to pass options that can be used by integrations SDK use internally. For example, there can be a simple case of passing a ``delimiter`` while loading CSV to pandas.read_csv() method.

    .. literalinclude:: ../../../../example_dags/example_load_file.py
       :language: python
       :start-after: [START load_file_example_22]
       :end-before: [END load_file_example_22]

    ``load_options`` is a list of different :py:class:`LoadOptions <astro.options.LoadOptions>` class objects. Please refer the list below for all valid options.

    - :py:class:`PandasCsvLoadOptions <astro.dataframes.load_options.PandasCsvLoadOptions>`
    - :py:class:`PandasJsonLoadOptions <astro.dataframes.load_options.PandasJsonLoadOptions>`
    - :py:class:`PandasNdjsonLoadOptions <astro.dataframes.load_options.PandasNdjsonLoadOptions>`
    - :py:class:`PandasParquetLoadOptions <astro.dataframes.load_options.PandasParquetLoadOptions>`
    - :py:class:`DeltaLoadOptions <astro.databases.databricks.load_options.DeltaLoadOptions>`
    - :py:class:`SnowflakeLoadOptions <astro.options.SnowflakeLoadOptions>`
```

python-sdk/tests_integration/test_example_dags.py
```diff
@@ -88,8 +88,9 @@ def order(dag_id: str) -> int:
 
 @pytest.mark.parametrize("dag_id", sorted(dag_bag.dag_ids, key=order))
 def test_example_dag(session, dag_id: str):
-    dag = dag_bag.get_dag(dag_id)
-    wrapper_run_dag(dag)
+    if dag_id == "example_load_file":
+        dag = dag_bag.get_dag(dag_id)
+        wrapper_run_dag(dag)
```
feluelle, line 93: I assume this is only temporarily for local testing. But actually you could have just used `test_example_dag[example_load_file]`.

utkarsharma2:
Closing in favour of - https://github.com/astronomer/astro-sdk/pull/1567

PR #1565, opened 2023-01-11
Fix deepsource

pankajastro:
# Description
Address: https://github.com/astronomer/astro-sdk/pull/1555#issuecomment-1377189333
## What is the current behavior?
Fix some deepsource issues

## Does this introduce a breaking change?
No

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1568, opened 2023-01-11
Add tolerance for codecov patch

pankajastro:
# Description
currently, codecov patch fails many time because the current branch has coverage less than the main branch but code coverage pass because we have tolerance for it. here I'm trying to add tolerance for code patch also to allow contributors to miss some line of coverage if not required. 
here is is an example https://github.com/astronomer/astro-sdk/pull/1565/files#diff-9cb1f064d06ee36971715ef96955fbbd606d522bcb711a3a6fe3e48c401359ccR152 I just changed the existing code technically no new adding but it fail


codecov.yml
```diff
@@ -5,3 +5,5 @@ coverage:
         target: auto
         threshold: 5%
         only_pulls: true
+    patch:
+      threshold: 5%
```
pankajkoti, line 9: Should this line not go under a `default` key? Also, does this mean it will allow PRs with pull request having patch coverage upto 95% ?

pankajastro: yes, this should be inside the default. `threshold` mean x % drop is allowed from the main. I think the target is used for the coverage upto 95  https://docs.codecov.com/docs/commit-status#threshold https://docs.codecov.com/docs/commit-status#target
```
patch:
    default:
        threshold: 5%
```
You can use `curl -X POST --data-binary @codecov.yml https://codecov.io/validate` to validate codecov.yml

pankajastro: but I'm not sure we will merge this or not, most probably I'll close the PR

tatiana:
@pankajastro @pankajkoti, how are we proceeding with this PR?

pankajastro:
closing it, No plan to take it forward at the moment.

PR #1569, opened 2023-01-11
Disable logging for non-debug mode

feluelle:
# Description

Currently, we oftentimes print some logging statements eventhough we are not using the verbose nor debug flag.

## What is the new behavior?

We disable logging for non-debug mode. The only exception is when verbose mode is set it will selectively enable logging. If debug mode is set enable logging for a selective list of loggers.

Not really. There are really no logs being shown when non-debug and non-verbose mode is used.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -62,7 +62,7 @@ def set_debug_mode(debug: bool) -> None:
         for logger_name in EXT_LOGGER_NAMES:
             logging.getLogger(logger_name).setLevel(logging.DEBUG)
     else:
-        logging.disable()
+        logging.disable()  # skipcq PY-A6006
 
 
 @app.command(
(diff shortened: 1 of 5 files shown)


tatiana (approved):

sql-cli/sql_cli/__main__.py
```diff
@@ -60,13 +58,11 @@ def set_debug_mode(debug: bool) -> None:
...
-        logger.propagate = debug
-        if debug:
-            logger.setLevel(logging.DEBUG)
-        else:
-            logger.setLevel(logging.CRITICAL)
+    if debug:
+        for logger_name in EXT_LOGGER_NAMES:
+            logging.getLogger(logger_name).setLevel(logging.DEBUG)
+    else:
+        logging.disable()
```
feluelle, line 65:
```suggestion
        logging.disable()  # skipcq PY-A6006
```

PR #1584, opened 2023-01-13
add a results_format field to run_raw_sql

### LINKED ISSUE #1583: Add interface to get results in standard formats from run_raw_sql
**Please describe the feature you'd like to see**

When I use `aql.run_raw_sql`, I find myself writing handlers like:

```python
@aql.run_raw_sql(..., handler = lambda x: x.fetchall())
```

```python
@aql.run_raw_sql(..., handler = pd.DataFrame(x.fetchall(), columns = x.keys()))
```

It seems like there are a few natural results formats (list, dataframe) that you'd want to get results in, but the user is responsible for making these handlers. Additionally, I can't just use the same handler for both `SELECT` and `CREATE TABLE` functions because sqlalchemy's ResultProxy errors if there are no results and you call `x.fetchall()`

**Describe the solution you'd like**
Add two fields to the decorator/operator:
- `results_format`: let the user specify a common format they want their results (list, dataframe)
- `fail_on_empty`: let the user decide what the behavior should be if there are no results

**Are there any alternatives to this feature?**

We could also consider exporting common handlers instead of forcing the user to define them.

**Additional context**
Add any other context about the feature request here.

I'm considering this from the perspective of the Astro Cloud IDE, in which our requirements are:
- get the results as a DataFrame
- use the same decorator/code for all run_raw_sql's, regardless of wether it's a `SELECT` or `INSERT`
  - this is because we're generating a DAG and parsing the SQL code to do this dynamically is unreliable

**Acceptance Criteria**

- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

jlaneve:
# Description

related: #1583

This PR extends the `run_raw_sql` to make it easier to get results. Previously, you'd have to write a `handler` yourself. The idea behind this change is to abstract some of the common handlers (get the results as a list, dataframe).

Specifically, it:
- introduces two new fields to the run_raw_sql operator:
  - `results_format`: let the user specify a common format they want their results (list, dataframe)
  - `fail_on_empty`: let the user decide what the behavior should be if there are no results

No, everything is incremental

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -135,7 +135,8 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
             def handle_exceptions(result):
                 try:
                     return conversion_func(result)
-                except Exception:  # skipcq: PYL-W0703
+                except Exception as e:  # skipcq: PYL-W0703
+                    logging.info(f"Exception {e} handled 'fail_on_empty' parameter")
                     return None
 
             return handle_exceptions
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -136,7 +136,7 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
                 try:
                     return conversion_func(result)
                 except Exception as e:  # skipcq: PYL-W0703
-                    logging.info(f"Exception {e} handled 'fail_on_empty' parameter")
+                    logging.info(f"Exception {e} handled since 'fail_on_empty' parameter was passed")
                     return None
 
             return handle_exceptions
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -10,9 +10,8 @@ except ImportError:
     from airflow.decorators import _TaskDecorator as TaskDecorator
 
 import airflow
-from airflow.decorators.base import task_decorator_factory
-
 import pandas as pd
+from airflow.decorators.base import task_decorator_factory
 from sqlalchemy.engine import ResultProxy
 
 if airflow.__version__ >= "2.3":
@@ -37,7 +36,7 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
 
     def __init__(
         self,
-        results_format: Literal['list', 'dataframe'] | None = None,
+        results_format: Literal["list", "dataframe"] | None = None,
         fail_on_empty: bool = False,
         **kwargs: Any,
     ):
@@ -72,7 +71,7 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
                     return conversion_func(result)
                 except Exception:
                     return None
-            
+
             # otherwise, just return the result
             return conversion_func(result)
 
@@ -143,7 +142,7 @@ def run_raw_sql(
     database: str | None = None,
     schema: str | None = None,
     handler: Callable | None = None,
-    results_format: Literal['list', 'dataframe'] | None = None,
+    results_format: Literal["list", "dataframe"] | None = None,
     fail_on_empty: bool = False,
     response_size: int = settings.RAW_SQL_MAX_RESPONSE_SIZE,
     **kwargs: Any,
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -129,7 +129,7 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
         # if fail_on_empty is set to False, wrap in a try/except block
         # this is because the conversion function will fail if the result is empty
         # due to sqlalchemy failing when there are no results
-        if fail_on_empty:
+        if not fail_on_empty:
 
             def handle_exceptions(result):
                 try:
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -137,7 +137,7 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
                 try:
                     return conversion_func(result)
                 except Exception as e:  # skipcq: PYL-W0703
-                    logging.info(f"Exception {e} handled since 'fail_on_empty' parameter was passed")
+                    logging.info("Exception %s handled since 'fail_on_empty' parameter was passed", str(e))
                     return None
 
             return handle_exceptions
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -118,6 +118,7 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
         # note that it takes precedence over the handler
         if self.results_format:
             self.handler = self.get_results_format_handler(results_format=self.results_format)
+            logging.info("Provided 'handler' will be overridden when 'results_format' is given.")
         if self.handler:
             self.handler = self.get_wrapped_handler(
                 fail_on_empty=self.fail_on_empty, conversion_func=self.handler
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Iterable, Mapping
-from typing import Any, Callable, Literal, cast
+from typing import Any, Callable, List, cast
 
 try:
     from airflow.decorators.base import TaskDecorator
@@ -20,6 +20,7 @@ else:
     from sqlalchemy.engine.result import RowProxy as SQLAlcRow
 
 from astro import settings
+from astro.constants import RunRawSQLResultFormat
 from astro.exceptions import IllegalLoadToDatabaseException
 from astro.sql.operators.base_decorator import BaseSQLDecoratedOperator
 from astro.utils.compat.typing import Context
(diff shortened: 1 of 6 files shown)


python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -44,9 +58,30 @@ def execute(self, context: Context) -> Any:
...
+                    return None
+
+            # otherwise, just return the result
+            return conversion_func(result)
+
+        # if no results_format is specified, we will return the results in the format specified by the handler
         elif self.handler:
+            # if it's a delta table, we don't need to do any conversion
+            if self.database_impl.sql_type == "delta":
+                return result
```
kaxil, line 82: This should go to the DeltaDatabase class. So create a method/property on BaseDatabase called "format_result_handler_conversion" or something like that, which should probably just have `pass` in BaseDatabase. And then you override it in DeltaDatabase.

So then you can do:

```python
self.database_impl.format_result_handler_conversion(result)
```

So that we don't have "tech"-specific loginc in Operators

jlaneve: this was originally introduced as part of @dimberman's databricks work, so he agreed to address this comment as a follow on to this PR. filed an issue to track it [here](https://github.com/astronomer/astro-sdk/issues/1606)

kaxil: Sounds good

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -69,6 +96,52 @@ def make_row_serializable(rows: Any) -> Any:
...
+        # due to sqlalchemy failing when there are no results
+        if fail_on_empty:
+
+            def handle_exceptions(result):
+                try:
+                    return conversion_func(result)
+                except Exception:  # skipcq: PYL-W0703
+                    return None
+
+            return handle_exceptions
```
jlaneve, line 140: correct me if I'm wrong but shouldn't this be:

```
if not fail_on_empty:
```

Currently, if `fail_on_empty` is True (meaning a user wants things to fail/raise an exception), we _shouldn't_ wrap it in a try/except

utkarsharma2: Yup, correct. Changed it. :)

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -44,9 +59,20 @@ def execute(self, context: Context) -> Any:
                 "backend."
             )
 
-        if self.handler and self.database_impl.sql_type == "delta":
+        # ToDo: Currently, the handler param in run_sql() method is only used in databricks all other databases are
+        # not using it. Which leads to different response types since handler is processed within `run_sql()` for
+        # databricks and not for other databases. Also the signature of `run_sql()` in databricks deviates from base.
+        # We need to standardise and when we do, we can remove below check as well.
```
kaxil, line 65: Can you create a Github issue for it too so that @phanikumv and you can prioritize it in one of the future sprints

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -69,6 +96,52 @@ def make_row_serializable(rows: Any) -> Any:
...
+    def get_handler(self):
+        # if a user specifies a results_format, we will return the results in that format
+        # note that it takes precedence over the handler
+        if self.results_format:
+            self.handler = self.get_results_format_handler(results_format=self.results_format)
+        if self.handler:
+            self.handler = self.get_wrapped_handler(
+                fail_on_empty=self.fail_on_empty, conversion_func=self.handler
+            )
+        return self.handler
```
kaxil, line 125: We should document this behaviour in docstrings as well as changelog so users know about this.

cc @phanikumv

kaxil: Looks like that is documented in docstirgs in L208 already -- we should duplicate it for the Operator docstrings too and have explicit docs for it in https://astro-sdk-python.readthedocs.io/en/stable/astro/sql/operators/raw_sql.html

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -32,6 +35,18 @@ class RawSQLOperator(BaseSQLDecoratedOperator):
     and on the SQL statement/function declared by the user.
     """
 
+    def __init__(
+        self,
+        results_format: RunRawSQLResultFormat | None = None,
+        fail_on_empty: bool = False,
```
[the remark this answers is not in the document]
utkarsharma2: @kaxil I have added a comment on the parent ticket for the same. Will address this in a separate PR.
ref: https://github.com/astronomer/astro-sdk/issues/1583#issuecomment-1387103235

python-sdk/src/astro/sql/operators/raw_sql.py
utkarsharma2: Sure, will add this.

python-sdk/src/astro/sql/operators/raw_sql.py
utkarsharma2: I'll add a ticket for this as well, so that this doesn't slip.

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -69,6 +96,52 @@ def make_row_serializable(rows: Any) -> Any:
...
+        # due to sqlalchemy failing when there are no results
+        if not fail_on_empty:
+
+            def handle_exceptions(result):
+                try:
+                    return conversion_func(result)
+                except Exception:  # skipcq: PYL-W0703
+                    return None
+
+            return handle_exceptions
```
kaxil, line 140: We should atleast add some logging when it fails

python-sdk/src/astro/sql/operators/raw_sql.py
kaxil: We should also add a log line in the code so we users can know what took precedence by looking at logs

kaxil (approved):

kaxil (approved):
Please address my comments around docs, example DAGs and then feel free to merge

python-sdk/tests/sql/operators/test_run_raw_sql.py
```diff
@@ -16,3 +20,154 @@ def test_make_row_serializable(rows):
...
+    Test that run_sql should raise an exception when fail_on_empty=False
+    """
+    results_as_pandas_dataframe.return_value = []
+    return_value = [1, 2, 3]
+    run_sql.return_value = return_value
+
+    def raise_exception(result):
+        raise ValueError("dummy exception")
+
+    with pytest.raises(ValueError) as e:
```
kaxil, line 118:
```suggestion
    with pytest.raises(ValueError, match="dummy exception"):
```

https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

python-sdk/src/astro/sql/operators/raw_sql.py
```diff
@@ -69,6 +96,52 @@ def make_row_serializable(rows: Any) -> Any:
...
+        Convert the result of a SQL query to a list
+        """
+        return cast(List[Any], result.fetchall())
+
+    @staticmethod
+    def results_as_pandas_dataframe(result: ResultProxy) -> pd.DataFrame:
+        """
+        Convert the result of a SQL query to a pandas dataframe
+        """
+        return pd.DataFrame(result.fetchall(), columns=result.keys())
```
kaxil, line 111: we should use Astro's PandasDataframe so it is serializable with Airflow 2.5:

```
from astro.dataframes.pandas import PandasDataframe

return PandasDataframe(result.fetchall(), columns=result.keys())
```

python-sdk/src/astro/sql/operators/raw_sql.py
utkarsharma2: Added ticket to keep track -  https://github.com/astronomer/astro-sdk/issues/1664

python-sdk/src/astro/sql/operators/raw_sql.py
utkarsharma2: Added log.

python-sdk/src/astro/sql/operators/raw_sql.py
utkarsharma2: @kaxil Maybe we can add a log saying that the exception is handled because of `fail_on_empty`?

utkarsharma2: @kaxil I'm going ahead with the above suggestions, if you feel this is not right, I'll raise a separate PR for this.

python-sdk/tests/sql/operators/test_run_raw_sql.py
utkarsharma2: Added in a separate comment since there was one more line which we needed to remove.

python-sdk/src/astro/sql/operators/raw_sql.py
utkarsharma2: Done.

python-sdk/src/astro/sql/operators/raw_sql.py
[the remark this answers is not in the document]
utkarsharma2: Updated document.

PR #1586, opened 2023-01-13
refactor merge tests to use more unit tests and less integration tests

dimberman:
# Description
Currently our merge tests are some of our most time consuming CI tests as well as the most black-box. By mocking the running SQL we can both remove the need to run so many tests against the database AND allow external users to test the merge function without needing snowflake/bigquery credentials.

The new testing structure compares precise SQL strings, since the merge should be programmatically consistent if the column names remain the same.

No

### Checklist


dimberman:
@kaxil @tatiana this change cuts 5 minutes off of the total CI time. I'm gonna keep going through and finding the tests that take the longest to convert them to unit tests.

python-sdk/tests_integration/sql/operators/test_merge.py
```diff
@@ -261,3 +258,166 @@ def test_merge_with_different_schemas(database_table_fixture, multiple_tables_fi
...
+    "database_class,conn_id",
+    [
+        (base_database_class, "sqlite_conn"),
+        (base_database_class, "gcp_conn"),
+        (base_database_class, "snowflake_conn"),
+        (delta_database_class, "databricks_conn"),
+    ],
+    ids=["sqlite", "bigquery", "snowflake", "databricks"],
+)
+def test_merge_sql_generation(database_class, conn_id, merge_parameters):
```
kaxil, line 405: Should we move this to `python-sdk/tests/sql/operators/test_merge.py`?

dimberman: @kaxil the only thing here that's not "integration test" is that it does need a real connection ID to create the database. Are you ok with me moving it over considering that?

kaxil (approved):

PR #1587, opened 2023-01-14
Create unit tests for cleanup_table

dimberman:
# Description
Currenlty the `test_cleanup` file tests take 16 minutes to run. This is due to many external database calls and file loads/table creations. This is causing us unnecessary expenses and slowing down development as we wait for CI/CD to finish.

## What is the new behavior?
With this PR, we can use the unittest mocking library to mock out any external SQL calls. Since we already know that the SQL calls will work, we can focus purely on whether the astro library generates the correct SQL and deletes the correct tables. This has caused a 50% reduction in run-time (16 minutes to 8 minutes).

In the interest of not making the perfect the enemy of the good, I have opted not to mock out the mapped tasks tests, as those seem more complex and I believe there are lower hanging fruit we can address to reduce CI runtime.

No.

### Checklist


dimberman:
@tatiana @kaxil this reduces the runtime for CI[6] from 16 minutes to 8 minutes. I have a question on how we should handle databridcks since databricks uses a different `run_sql` function so I'm not sure if there's an easy way to handle via mock or if we need a separate function.

python-sdk/tests_integration/sql/operators/test_cleanup.py
```diff
@@ -28,164 +31,123 @@
...
-    SUPPORTED_DATABASES_OBJECTS_WITH_FILE,
-    indirect=True,
-    ids=SUPPORTED_DATABASES,
+    "temp_table",
+    [
+        Table(conn_id="sqlite_conn"),
+        Table(conn_id="snowflake_conn"),
+        Table(conn_id="bigquery"),
+    ],
+    ids=["sqlite", "snowflake", "bigquery"],
```
kaxil, line 44: Should we add Redshift, postgres and delta too?

kaxil:
> @tatiana @kaxil this reduces the runtime for CI[6] from 16 minutes to 8 minutes. I have a question on how we should handle databridcks since databricks uses a different `run_sql` function so I'm not sure if there's an easy way to handle via mock or if we need a separate function.

You could do something like the following:

```python
@pytest.mark.parametrize(
    "temp_table",
    [
        Table(conn_id="sqlite_conn"),
        Table(conn_id="snowflake_conn"),
        Table(conn_id="bigquery"),
        Table(conn_id="databricks_conn"),
    ],
    ids=["sqlite", "snowflake", "bigquery", "databricks"],
)
def test_cleanup_one_table(temp_table):
    module = create_database(temp_table.conn_id)
    with mock.patch(f"{module.__class__.__module__}.{module.__class__.__name__}.run_sql") as mock_run_sql:
        a = aql.cleanup([temp_table])
        a.execute({})
    mock_run_sql.assert_called_once_with(
        drop_table_statement.format("table_name", table_name=temp_table.name)
    )
```

python-sdk/tests_integration/sql/operators/test_cleanup.py
```diff
@@ -28,164 +31,140 @@
...
-    SUPPORTED_DATABASES_OBJECTS_WITH_FILE,
-    indirect=True,
-    ids=SUPPORTED_DATABASES,
+    "temp_table",
+    [
+        Table(conn_id="sqlite_conn"),
+        Table(conn_id="snowflake_conn"),
+        Table(conn_id="bigquery"),
+        Table(conn_id="databricks_conn"),
+        Table(conn_id="redshift_conn"),
```
pankajastro, line 44: look like Postgres is missing here. wondering why Postgres is missing in other tests too, is there any reason for this?

python-sdk/tests_integration/sql/operators/test_cleanup.py
dimberman: added all three

python-sdk/tests_integration/sql/operators/test_cleanup.py
dimberman: added postgres

kaxil (approved):

PR #1588, opened 2023-01-14
move load_file and example_dag tests into their own section

dimberman:
# Description

In trying to optimize the astro SDK CI system, I find that the load_file and example_dag sections are two sections that just can NOT be optimized using the mocking library. The load_file library can not be optimized because we are literally testing whether the files are being properly loaded into the external databases, and the example DAGs need to run exactly to spec so we are sure users can run them as-is.

To handle this limitation while still bringing down the total CI time, we separate the load_file tests and example_dags tests into their own section where we can split them using pytest-split and run them in parallel (where before they would run sequentially on a single worker).

We can now run all load_file tests within 10 minutes across three workers. and example_dag tests in 12 minutes.

### Checklist


python-sdk/.test_durations
```diff
@@ -0,0 +1,70 @@
+{
```
tatiana, line 1: @dimberman are we committing this file intentionally?

tatiana (commented):
LGTM, @dimberman , one minor comment

python-sdk/.test_durations
dimberman: @tatiana yes that file is used by pytest-split to properly balance the tests

python-sdk/tests_integration/sql/operators/test_load_file.py
```diff
@@ -515,7 +505,7 @@ def test_load_file_chunks(sample_dag, database_table_fixture):
         ),
     ],
     indirect=["database_table_fixture"],
-    ids=["Bigquery", "Redshift"],
+    ids=["Redshift"],
```
tatiana, line 508: Why did we remove BQ?

sunank200: +1

sunank200 (approved):

python-sdk/tests_integration/sql/operators/test_load_file.py
dimberman: @tatiana this one test took 5 minutes to complete and as far as I can tell it's not really testing anything that's not already tested (note that code coverage has not changes at all)

python-sdk/.test_durations
feluelle: Nice! ❤️

PR #1597, opened 2023-01-16
Add check to `load_file_to_table_natively` for `storage_integration`

utkarsharma2:
# Description
Currently, we did not check for getting the `storage_integration` param within the `load_file_to_table_natively` method.
Now that we have introduced the `load_options` for snowflake, we can pass `storage_integration` either in `native_support_kwargs` or `load_options`. 

## What is the new behavior?
1. Added condition check for `storage_integration` and added test cases.
2. Added another property `storage_integration` in `SnowflakeLoadOptions` to store `storage_integration`,  since `SnowflakeLoadOptions.copy_options` was intended to be used differently.

Nope

### Checklist


tatiana (approved):

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -618,9 +618,19 @@ def load_file_to_table_natively(
 
         """
         native_support_kwargs = native_support_kwargs or {}
-        storage_integration = native_support_kwargs.get("storage_integration")
         if not self.load_options:
             self.load_options = SnowflakeLoadOptions()
+
+        storage_integration = native_support_kwargs.get("storage_integration", None)
```
sunank200, line 624: We can create a deprecation warning in another PR.

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -165,3 +165,58 @@ def test_snowflake_load_options_empty():
...
+    with mock.patch("astro.databases.snowflake.SnowflakeDatabase.create_stage"), mock.patch(
+        "astro.databases.snowflake.SnowflakeDatabase._copy_into_table_from_stage"
+    ) as _copy_into_table_from_stage, mock.patch(
+        "astro.databases.snowflake.SnowflakeDatabase.evaluate_results"
+    ):
+        database.load_file_to_table_natively(
+            source_file=file,
+            target_table=table,
+            native_support_kwargs={"storage_integration": "some_integrations"},
+        )
```
sunank200, line 203: What are we checking as part of this test?

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -165,3 +165,58 @@ def test_snowflake_load_options_empty():
...
+        conn_id=table.conn_id,
+        load_options=SnowflakeLoadOptions(copy_options={"storage_integration": "some_integrations"}),
+    )
+
+    with mock.patch("astro.databases.snowflake.SnowflakeDatabase.create_stage"), mock.patch(
+        "astro.databases.snowflake.SnowflakeDatabase._copy_into_table_from_stage"
+    ) as _copy_into_table_from_stage, mock.patch(
+        "astro.databases.snowflake.SnowflakeDatabase.evaluate_results"
+    ):
+        database.load_file_to_table_natively(source_file=file, target_table=table)
```
sunank200, line 222: What are we checking as part of this test?

sunank200 (commented):
Overall looks good to me but in test we could assert to check the attributes.

python-sdk/tests/databases/test_snowflake.py
utkarsharma2: in the previous test `test_storage_integrations_params` we made sure that if the `storage_integration` is not passed the `load_file_to_table_natively` will raise the ValueError exception. Here we are making sure to pass `storage_integration` via native_support_kwargs don't.

python-sdk/tests/databases/test_snowflake.py
utkarsharma2: in the previous test `test_storage_integrations_params` we made sure that if the `storage_integration` is not passed the `load_file_to_table_natively` will raise the ValueError exception. Here we are making sure to pass `storage_integration` via `SnowflakeLoadOptions` don't.

sunank200 (approved):

python-sdk/tests/databases/test_snowflake.py
sunank200: Sure

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -165,3 +165,58 @@ def test_snowflake_load_options_empty():
...
+    `native_support_kwargs` or via `SnowflakeLoadOptions` it raises an exception.
+    """
+    table = Table(conn_id="fake-conn")
+    file = File(path="azure://data/homes_main.ndjson")
+    database = SnowflakeDatabase(conn_id=table.conn_id, load_options=SnowflakeLoadOptions())
+    with pytest.raises(DatabaseCustomError) as e:
+        database.load_file_to_table_natively(source_file=file, target_table=table)
+    assert e.match(
+        "Param 'storage_integration' is required. Please pass either in `native_support_kwargs` or"
+        " `SnowflakeLoadOptions.copy_options`"
```
pankajastro, line 182: this should be `SnowflakeLoadOptions.storage_integration` or `SnowflakeLoadOptions.copy_options`?

utkarsharma2: Yup, will change this. Thanks for pointing out

Issue #1592, opened 2023-01-16
Fix naming convention for Data validation


phanikumv:
**Describe the bug**
Naming conventions need to be updated for the data validations module
along with the open question around sql_check vs check_sql

**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Write the DAG  '...'
2. Create connection '....'
3. Run using  '....'

**Expected behavior**

**Screenshots**

**Additional context**

PR #1489, opened 2022-12-23
Change module name

utkarsharma2:
# Description
## What is the current behavior?
Module names were in camelCase
closes #1592 

## What is the new behavior?
Changed them to snake_case

## Does this introduce a breaking change?
Nope

PR #1608, opened 2023-01-17
Datavalidation module rename

utkarsharma2:
# Description
Changed module and operator names

closes: #1592 

### Checklist


python-sdk/src/astro/sql/__init__.py
```diff
@@ -4,13 +4,13 @@
 
 from astro.sql.operators.append import AppendOperator, append
 from astro.sql.operators.cleanup import CleanupOperator, cleanup
-from astro.sql.operators.data_validations.ColumnCheckOperator import (  # skipcq: PY-W2000
+from astro.sql.operators.data_validations.check_column import (  # skipcq: PY-W2000
```
pankajastro, line 7: you can remove `# skipcq` see  https://github.com/astronomer/astro-sdk/pull/1489#discussion_r1071811974

python-sdk/src/astro/sql/__init__.py
```diff
@@ -4,13 +4,13 @@
...
     ColumnCheckOperator,
     column_check,
 )
-from astro.sql.operators.data_validations.SQLCheckOperator import (  # skipcq: PY-W2000
+from astro.sql.operators.data_validations.check_table import (  # skipcq: PY-W2000
```
pankajastro, line 11: same here

python-sdk/src/astro/sql/__init__.py
```diff
@@ -4,13 +4,13 @@
...
     SQLCheckOperator,
-    sql_check,
+    check_table,
```
feluelle, line 13: Why are we calling this `check_table` and in `check_column` we call it `column_check`? Let's be consistent here.

python-sdk/src/astro/sql/__init__.py
utkarsharma2: Done

python-sdk/src/astro/sql/__init__.py
utkarsharma2: Done.

feluelle:
@utkarsharma2 could also update the naming of the test file, please?

feluelle (approved):

PR #1609, opened 2023-01-17
[Don't merge] Pin sphinx-autoapi version

pankajastro:
# Description
## What is the new behavior?

-
-
-

### Checklist

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -112,7 +112,7 @@ all = [
 doc = [
     "myst-parser>=0.17",
     "sphinx>=4.4.0",
-    "sphinx-autoapi>=2.0.0",
+    "sphinx-autoapi==2.0.0", # import pkg as pkg1 is broken in sphinx-autoapi 2.0.1
     "sphinx-rtd-theme"
 ]
 


feluelle:
@pankajastro to me this does not sound like an error. It is a warning which we treat as an error - and we should resolve this.

utkarsharma2:
@feluelle It makes sense to me when there are multiple references to FileType(in constant and file module), but it's raising this issue even when there is only one reference. I think's an issue with Sphinx right?

An alternative would be to update the codebase to use `FileType as FileTypeConstant` which I think is unnecessary when there is only one FileType in the namespace. WDYT?

feluelle:
Okay, can't we be more specific and help sphinx to find the right `FileType` somehow?

pankajastro:
> @pankajastro to me this does not sound like an error. It is a warning which we treat as an error - and we should resolve this.

Looks like some issue on sphinx-autoapi side though I have not investigated it thoroughly since we have simple case two class have same name and we are importing one of them.

feluelle:
[here](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/base.py#L28)
and [here](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/base.py#L35) we have `FileType` but in the latter one, we "import as" 🤔 Maybe indeed a bug?

Do we need to import both btw? I think we can remove one, no?

utkarsharma2:
> 

Even then we are having issues when there is only one import with FileType. I think it's an issue on `autoapi` side, we can fix it at our end but that will lead to multiple code changes on our end. I think for now maybe we can pin it to the older version and wait for a fix. @feluelle WDYT?

python-sdk/pyproject.toml
```diff
@@ -112,7 +112,7 @@ all = [
 doc = [
     "myst-parser>=0.17",
     "sphinx>=4.4.0",
-    "sphinx-autoapi>=2.0.0",
+    "sphinx-autoapi==2.0.0", # import pkg as pkg1 is broken in sphinx-autoapi 2.0.1
```
pankajastro, line 115:
```suggestion
    "sphinx-autoapi==2.0.0", # Multiple FileType class (same name) import is broken in sphinx-autoapi 2.0.1
```

dimberman (approved):

utkarsharma2 (approved):

PR #1610, opened 2023-01-17
Remove unused datasets

utkarsharma2:
# Description
We are not using this dataset for benchmarking, we should remove it. Also, there was a failure for the StackExchange dataset.


python-sdk/tests/benchmark/datasets.md
```diff
@@ -140,95 +140,3 @@ The [script](synthetic_ndjson_generator.py) is used to generate the NDJSON file.
 | five_gb    | 4.5 GB | NDJSON   | s3://astro-sdk/benchmark/synthetic-dataset/ndjson/five_gb.ndjson    |
 | five_gb    | 5 GB   | NDJSON   | s3://astro-sdk/benchmark/synthetic-dataset/ndjson/five_gb/          |
 | ten_gb     | 10 GB  | NDJSON   | s3://astro-sdk/benchmark/synthetic-dataset/ndjson/ten_gb/           |
-
-## Origin of the real datasets
```
sunank200, line 144: This dataset was originally used as a realistic dataset. Shouldn't we still keep it?

sunank200 (commented):
This dataset was originally used as a realistic dataset. Shouldn't we still keep it?

tatiana:
End-users use real datasets, which has allowed us to surface several issues with our stack. Much more than any synthetic dataset we have.

I believe we should have end-to-end tests that use these datasets instead of simply getting rid of them.

utkarsharma2:
I agree @tatiana , but currently, we are not using it in our code base anywhere. When do we start doing that we can add them again to the codebase, WDYT? we can maybe create a ticket(s) to add a test case with a real dataset?

python-sdk/tests/benchmark/datasets.md
utkarsharma2: We are not adding those numbers obtained by these datasets in benchmarking results either. right?

tatiana:
I'm happy for them to be removed and re-introduced as long as we have a task logged for us to have e2e tests using realistic datasets, referencing this PR :)

utkarsharma2:
@tatiana Added a ticket for the same - https://github.com/astronomer/astro-sdk/issues/1612

tatiana (approved):

pankajastro (commented):
Most of these data we have stored in gcs do you want to keep it or we should clean them also @utkarsharma2 ?

utkarsharma2:
> Most of these data we have stored in gcs do you want to keep it or we should clean them also example gs://astro-sdk/benchmark/trimmed/github/github_timeline_100mb.csv @utkarsharma2 ?

We should do that as well. I'll merge this PR for now. I'll create a separate ticket for this.

pankajkoti (approved):

pankajastro (approved):

PR #1620, opened 2023-01-18
Smart open to transfer datasets for universal transfer operator

sunank200:
# Description

Currently, all the files are being downloaded to temporary reference in the worker and uploaded back to the destination dataset. This is time-consuming and doesn't work well with database transfers. Hence, implemented the transfer for all file types using smart open.

## What is the new behavior?

- Add a feature to read the file using smart open and convert it into a buffer that can be streamed into other data structures.
- Run the example DAG for S3 and GCS transfers.
- Add class for `Filetype` and `PandasDataframe`.
- For filetypes CSV, json and ndjson, read as `io.StringIO()`. For other filetypes read as `io.BytesIO()`

No

### Checklist
- [x] Ran example DAG successfully.


sunank200:
Closing this as this branch is old one. The following PR: https://github.com/astronomer/astro-sdk/pull/1621 handles it.

Issue #1011, opened 2022-10-06
Uncomment kube pod teardown


utkarsharma2:
**Please describe the feature you'd like to see**
We have commented in python-sdk/tests/benchmark/wait_for_completion.sh
#kubectl delete job benchmark-"${GIT_HASH}" -n benchmark

This is done until we stabilize the benchmarking we need to look at logs.

**Describe the solution you'd like**
We need to uncomment it once we stabilize the benchmarking

sunank200:
This is old ticket not relevant any more

pankajastro:
this has yet not done PR https://github.com/astronomer/astro-sdk/pull/1623

PR #1623, opened 2023-01-19
Remove benchmark pod once job run is done

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1011

currently, a benchmark we do not delete and kept it for some debugging purpose  

## What is the new behavior?
let's delete the pod it helps in reusing cluster sometimes 

No

### Checklist

### DIFF
--- a/python-sdk/tests/benchmark/wait_for_completion.sh
+++ b/python-sdk/tests/benchmark/wait_for_completion.sh
@@ -10,7 +10,7 @@ while [[ $complete_status -ne 0 ]] && [[ $failed_status -ne 0 ]]; do
   complete_status=$?
 done
 
-#kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
+kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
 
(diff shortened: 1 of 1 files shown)


python-sdk/tests/benchmark/wait_for_completion.sh
```diff
@@ -10,7 +10,7 @@ while [[ $complete_status -ne 0 ]] && [[ $failed_status -ne 0 ]]; do
   complete_status=$?
 done
 
-#kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
+kubectl delete job benchmark-"${GIT_HASH}" -n benchmark
```
pankajkoti, line 13: Any idea why this was commented earlier?

pankajastro: yes, when we were working on running the benchmark in CI initially, we want to check the log of pods in case of some error. but now we even delete the cluster since the flow looks stable in the current setups and we run only once in a week.

utkarsharma2 (approved):

Issue #1591, opened 2023-01-16
Deprecate `native_support_kwargs` param in load_file


pankajastro:
Since we have introduced load_options so should consider to Deprecate or for now at least add a warning for `load_file` param `native_support_kwargs ` 

https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/load_file.py#L51

phanikumv:
Scheduled in 1.4.1

PR #1625, opened 2023-01-19
Add deprecation warning that native_support_kwargs in load file

sunank200:
# Description

Since we have introduced `load_options`  add a deprecation warning for `load_file` param `native_support_kwargs `

closes: #1591 

## What is the new behavior?

- Add deprecation warning that `load_options` will be replacing `native_support_kwargs` parameter
- Add the test for the same
- Changes in documentation is already part of last release: https://astro-sdk-python.readthedocs.io/en/stable/astro/sql/operators/load_file.html#loadoptions

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/load_file.py
+++ b/python-sdk/src/astro/sql/operators/load_file.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import Any
 
 import pandas as pd
@@ -62,6 +63,13 @@ class LoadFileOperator(AstroSQLBaseOperator):
                 output_datasets=output_table,
             )
         )
+        if native_support_kwargs:
+            warnings.warn(
+                """`load_options` will be replacing `native_support_kwargs` parameter. Please use `load_options`
+                parameter instead. And, `native_support_kwargs` will be removed in astro-sdk-python>=1.5.0.""",
+                DeprecationWarning,
+                stacklevel=2,
+            )
         self.output_table = output_table
         self.input_file = input_file
         self.input_file.load_options = load_options
@@ -82,6 +90,7 @@ class LoadFileOperator(AstroSQLBaseOperator):
         """
         if self.input_file.conn_id:
             check_if_connection_exists(self.input_file.conn_id)
+
         # TODO: remove pushing to XCom once we update the airflow version.
         if self.output_table:
             context["ti"].xcom_push(key="output_table_conn_id", value=str(self.output_table.conn_id))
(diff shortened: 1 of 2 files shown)


python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -165,3 +165,42 @@ def test_tables_creation_if_they_dont_exist(database_table_fixture, if_exists):
 
     database_df = db.export_table_to_pandas_dataframe(test_table)
     assert database_df.shape == (3, 9)
+
+
+# TODO: Remove this test in astro-sdk 1.5.0
```
pankajastro, line 170: For this PR scope, we are adding a warning and in 1.5.0 we will deprecate it. But we will remove it later maybe in 2.0.0 so I think this test will require even after 1.5.0 WDYT?

sunank200: As it would be deprecated in 1.5.0. So any version above that would not require this warning or test. Isn't it?

pankajastro: Deprecation means they can still pass it and use it, they won't be able to pass if we will remove it.

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -329,6 +338,14 @@ def load_file(
     # contain chars like - ?, * etc. Which are not acceptable as task id.
     task_id = task_id if task_id is not None else get_unique_task_id("load_file")
 
+    if native_support_kwargs:
+        warnings.warn(
+            """`load_options` will be replacing `native_support_kwargs` parameter. Please use `load_options`
+            parameter instead. And, `native_support_kwargs` will be removed in astro-sdk-python>=1.5.0.""",
```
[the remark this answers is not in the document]
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1625/commits/0ce76808b52dd3d534fa8d9c89c90480ab552329)

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -62,6 +63,13 @@ def __init__(
                 output_datasets=output_table,
             )
         )
+        if native_support_kwargs:
+            warnings.warn(
+                """`load_options` will be replacing `native_support_kwargs` parameter. Please use `load_options`
+                parameter instead. And, `native_support_kwargs` will be removed in astro-sdk-python>=1.5.0.""",
```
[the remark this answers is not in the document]
sunank200: [Done](https://github.com/astronomer/astro-sdk/pull/1625/commits/0ce76808b52dd3d534fa8d9c89c90480ab552329)

pankajastro (approved):
LGTM

pankajastro:
Merging it. CI fail is unrelated.

Issue #1574, opened 2023-01-12
Investigate some non-consistent databricks failure 


pankajastro:
**Describe the bug**
test_example_dag[example_load_file] 
```
self = <databricks.sql.thrift_backend.ThriftBackend object at 0x7f59fd46c160>
op_handle = TOperationHandle(operationId=THandleIdentifier(guid=b'\x01\xed\x92k=\x13\x13\x1c\x8b\xc7\xfa//\xc1\x9e\x85', secret=b'3\x8dR\x9d\x82rF\xeb\x84\x82\xcbA\x94f\x83\x9d'), operationType=0, hasResultSet=True, modifiedRowCount=None)
get_operations_resp = TGetOperationStatusResp(status=TStatus(statusCode=0, infoMessages=None, sqlState=None, errorCode=None, errorMessage=No...t org.apache.spark.sql.hive.HiveExternalCatalog.$anonfun$withClient$2(HiveExternalCatalog.scala:154)\n\t... 71 more\n')

    def _check_command_not_in_error_or_closed_state(self, op_handle, get_operations_resp):
        if get_operations_resp.operationState == ttypes.TOperationState.ERROR_STATE:
            if get_operations_resp.displayMessage:
>               raise ServerOperationError(
                    get_operations_resp.displayMessage, ***
                        "operation-id": op_handle and op_handle.operationId.guid,
                        "diagnostic-info": get_operations_resp.diagnosticInfo
                    ***)
E               databricks.sql.exc.ServerOperationError: org.apache.hadoop.hive.ql.metadata.HiveException: at least one column must be specified for the table

.nox/test-3-8-airflow-2-5-0/lib/python3.8/site-packages/databricks/sql/thrift_backend.py:359: ServerOperationError
```

test_create_table_from_select_statement[delta]
```
 if final_state == "FAILED":
>           raise AirflowException(f"Databricks job failed. Job info ***final_job_state***")
[...]

PR #1626, opened 2023-01-19
Add retry and delay in cleanup operator

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/1574

The databricks load example fails with an error
```
databricks.sql.exc.ServerOperationError: org.apache.hadoop.hive.ql.metadata.HiveException: at least one column must be specified for the table
```

## What is the new behavior?
Add retry and delay in cleanup operator

No

### Checklist

### DIFF
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -18,11 +18,9 @@ Pre-requisites for load_file_example_19:
 """
 import os
 import pathlib
-import time
 from datetime import datetime, timedelta
 
 import sqlalchemy
-from airflow.decorators import task
 from airflow.models import DAG
 
 from astro import sql as aql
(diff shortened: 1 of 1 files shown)


dimberman:
This solution feels really hacky, and it raises concerns that we are covering for a deeper issue. Are you sure that this delay is necessary?  Maybe there's something we need to do in the load operator or the cleanup operator (e.g. allowing the cleanup operator to perform multiple tries if the table is databricks). I would not be ok with advising users to add explicit delays in their DAGs.

pankajastro:
> This solution feels really hacky, and it raises concerns that we are covering for a deeper issue. Are you sure that this delay is necessary? Maybe there's something we need to do in the load operator or the cleanup operator (e.g. allowing the cleanup operator to perform multiple tries if the table is databricks). I would not be ok with advising users to add explicit delays in their DAGs.

- yeah, this hacky solution, I agree. 
- putting delay is just a hunch and trying to see if this work.
- look like the others Databricks user is also hitting a similar issue so I'm not sure if I can do much here. I have also suggested the team file some support tickets https://community.databricks.com/s/question/0D58Y00009Z6oHYSAZ/random-error-at-least-one-column-must-be-specified-for-the-table
- adding retry in cleanup is cleaner and looks a better solution.

dimberman:
@pankajastro what's the word on this? Do we still need this?

pankajastro:
> @pankajastro what's the word on this? Do we still need this?

The GH issue has closed, so maybe it has been fixed. closing it.

Issue #760, opened 2022-08-31
Support for Python 3.10 with Astro Python SDK


kaxil:
I think we do support Py 3.10 for Astro Python SDK -- but we should officially test it out and add a test in CI

pankajastro:
Look like currently we run tests in Python3.8 env only. I was thinking to enable test for other Python versions too but CI time will increase drastically.

pankajastro:
for now, as discussed we are putting this on hold cc: @phanikumv

pankajastro:
Since now we have some separation between unit and integration tests and units tests take relatively less time so I have started on it.  We will run unit tests for Python-{3.8,3.9,10}

PR #1629, opened 2023-01-19
Run unit tests for Python-3.10

pankajastro:
# Description
closes: https://github.com/astronomer/astro-sdk/issues/760

Currently, we run tests in CI for only Python-3.8

## What is the new behavior?
Run unit tests for `Python-{3.8,3.9,3.10}`

No

### Checklist

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -229,7 +229,7 @@ jobs:
           key: ${{ runner.os }}-2.5-${{ hashFiles('python-sdk/pyproject.toml') }}-${{ hashFiles('python-sdk/src/astro/__init__.py') }}
       - run: sqlite3 /tmp/sqlite_default.db "VACUUM;"
       - run: pip3 install nox
-      - run: nox -s "test-3.8(airflow='2.5.0')" -- tests/ --cov=src --cov-report=xml --cov-branch
+      - run: nox -s "test(airflow='2.5.0')" -- tests/ --cov=src --cov-report=xml --cov-branch
       - name: Upload coverage
         uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -194,6 +194,9 @@ jobs:
       AZURE_WASB_CONN_STRING: ${{ secrets.AZURE_WASB_CONN_STRING }}
 
   Run-Unit-tests-Airflow-2-5:
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


pankajkoti (commented):
Do we also need to run integration tests for Python 3.10 ? 
If yes, can we create a ticket for that? Otherwise the current ticket will get closed and we will miss running integrations tests for Python 3.10.

pankajastro:
> Do we also need to run integration tests for Python 3.10 ?
> If yes, can we create a ticket for that? Otherwise the current ticket will get closed and we will miss running integrations tests for Python 3.10.

I was anticipating that we will run unit tests only but @phanikumv @kaxil can add more to it.

kaxil (approved):

kaxil:
> > Do we also need to run integration tests for Python 3.10 ?
> > If yes, can we create a ticket for that? Otherwise the current ticket will get closed and we will miss running integrations tests for Python 3.10.
> 
> I was anticipating that we will run unit tests only but @phanikumv @kaxil can add more to it.

What I would like is to run longer term is:

1. Unit tests for multiple Python versions (3.8-3.10)
2. Full test suite (Unit + Integration) with the latest Airflow version (2.5 right now)
3. Selective Integration tests (just example DAGs) for the oldest Airflow version that is an LTS (2.2) from Astro Runtime perspective as described in https://docs.astronomer.io/astro/runtime-version-lifecycle-policy#long-term-support-lts-releases

pankajastro:
> > > Do we also need to run integration tests for Python 3.10 ?
> > > If yes, can we create a ticket for that? Otherwise the current ticket will get closed and we will miss running integrations tests for Python 3.10.
> > 
> > 
> > I was anticipating that we will run unit tests only but @phanikumv @kaxil can add more to it.
> 
> What I would like is to run longer term is:
> 
> 1. Unit tests for multiple Python versions (3.8-3.10)
> 2. Full test suite (Unit + Integration) with the latest Airflow version (2.5 right now)
> 3. Selective Integration tests (just example DAGs) for the oldest Airflow version that is an LTS (2.2) from Astro Runtime perspective as described in https://docs.astronomer.io/astro/runtime-version-lifecycle-policy#long-term-support-lts-releases

Ok, I have created https://github.com/astronomer/astro-sdk/issues/1632 to track it. It is easy to enable even integration tests for Python-{3.8,3.9,3.10} but I'm worried about a few flaky tests that would slow down development efficiency so I'll not enable them in this PR.

kaxil:
>It is easy to enable even integration tests for Python-{3.8,3.9,3.10}

No, we don't need to run all integration tests for all Python versions :)

Issue #1607, opened 2023-01-17
Add documentation for data validation


pankajastro:
@utkarsharma2  please add details

PR #1630, opened 2023-01-19
Fix for Check_Table operator

utkarsharma2:
# Description
Right now `check_table` operator is not able to process Xcom wrapped table object.

related: #1607 

## What is the new behavior?
Added a fix for the same.

Nope

### Checklist


dimberman (dismissed):

python-sdk/src/astro/sql/operators/data_validations/check_table.py
```diff
@@ -44,13 +44,17 @@ def __init__(
...
 
-        db = create_database(dataset.conn_id)
+        db = create_database(table.conn_id)
         super().__init__(
-            table=db.get_table_qualified_name(dataset),
+            table=db.get_table_qualified_name(table),
             checks=checks,
             partition_clause=partition_clause,
-            conn_id=dataset.conn_id,
+            conn_id=table.conn_id,
```
dimberman, line 57: @utkarsharma2 you shouldn't be creating databases in the init function of a class. This will run every single time that the airflow scheduler parses the DAG. You should override the execute function and add these elements before running "super.execute()".

dimberman (dismissed):

python-sdk/src/astro/sql/operators/data_validations/check_table.py
utkarsharma2: @dimberman I thought of doing that but then there was a risk since I'll be assuming the variable name used by the Parent operator like `self.table` and `self.conn_id` if the parent function changed in these variables it will cause a failure in check_table operator. 

Also, the query is generated in `__init__` of SQLTableCheckOperator. So we also have regenerate that in execute() of check_table().

utkarsharma2: Moved the logic to execute()

pankajastro (approved):
LGTM

PR #1631, opened 2023-01-19
Data validation operators documentation

utkarsharma2:
# Description
Right now we don't have documentation for the data validation operator. 

related: #1607

## What is the new behavior?
Added documentation for data validation operator. 

Nope

### Checklist


python-sdk/example_dags/example_check_table.py
```diff
@@ -0,0 +1,27 @@
...
+) as dag:
+    # [START data_validation__check_table]
+    imdb_movies = aql.load_file(
+        File("https://raw.githubusercontent.com/astronomer/astro-sdk/main/python-sdk/tests/data/homes.csv"),
+        output_table=Table(conn_id="sqlite_default"),
+    )
+    aql.check_table(
+        dataset=imdb_movies,
+        checks={
+            "sell_list": {"check_statement": "sell <= list"},
```
sunank200, line 23: What does this mean here - `sell <=list`?

sunank200 (commented):
Over all it looks good. But may be we should give more example of `check_column` and `check_table`. (may be with one realistic example DAG for a scenario) WDYT?

python-sdk/example_dags/example_check_table.py
utkarsharma2: Those a re column level checks and it's making sure all the values in `sell` col is less or equal to that `list` col

sunank200:
> But may be we should give more example of `check_column` and `check_table`. (may be with one realistic example DAG for a scenario) WDYT?

@utkarsharma2

pankajastro (approved):
LGTM

PR #1647, opened 2023-01-20
Support "s3" conn type for S3Location

kaxil:
Amazon S3 conn type was removed in Airflow in favor of AWS in https://github.com/apache/airflow/commit/cf73cb79d7aba49d3934864acdd9ce1599836d0f

In SDK, we should support both. This was identified in Cloud IDE

image


joedborg (approved):
Looks good to me

pankajastro (approved):

kaxil:
uff my bad -- looks like tests were failing, I will check if they were related or unrelated

PR #1648, opened 2023-01-20
Fix failing doc build because of Airflow 2.5.1 changes

kaxil:
The docs were refactored in Airflow 2.5.1


kaxil:
Merged it without approval as this fixes main branch and this is doc-only change

PR #1650, opened 2023-01-23
Bump astro-runtime version 7.2.0

pankajastro:
# Description
Bump astro-runtime version 7.2.0

Tested astro-sdk is pre-installed in runtime 7.2.0 image 

Screenshot 2023-01-23 at 12 54 18 PM

### DIFF
--- a/python-sdk/dev/Dockerfile
+++ b/python-sdk/dev/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/astronomer/astro-runtime:7.1.0-base
+FROM quay.io/astronomer/astro-runtime:7.2.0-base
 
 USER root
 RUN apt-get update -y && apt-get install -y git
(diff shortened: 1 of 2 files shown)


pankajkoti (commented):
This is just a regular image bump right? I mean the description says we tested runtime image has pre-installed Astro SDK, but in the Dockerfile we again have layers to install local Python SDK which I believe we would need to do so in the dev Dockerfile. Just that the description is confusing me if there was another reason for this PR too, and hence asking :)

pankajastro:
> This is just a regular image bump right? I mean the description says we tested runtime image has pre-installed Astro SDK, but in the Dockerfile we again have layers to install local Python SDK which I believe we would need to do so in the dev Dockerfile. Just that the description is confusing me if there was another reason for this PR too, and hence asking :)

astro-runtime 7.2.0 onward astro-sdk will be preinstalled in the runtime image so tested that too since this is the first time we have astro-sdk in the runtime image

feluelle (approved):

Issue #1652, opened 2023-01-23
Use json flag to return json output for version typer command in SQL CLI


pankajkoti:
Currently, the typer command prints rich text output for version command which make it difficult to parse in the integrations like Astro CLI. After discussions over Slack thread, we decided that any output that needs to be parsed should be returned as json output.

PR #1653, opened 2023-01-23
Support json output for version command

pankajkoti:
# Description

Currently we print the output of version typer command as rich text output. This makes it difficult to parse
and extract the version in integrations for further processing.

close: #1652 
## What is the new behavior?

Provide a json flag to the typer command which when provided in the command
prints a json output making it easier to parse for integraitons

No

### Checklist

### DIFF
--- a/sql-cli/pyproject.toml
+++ b/sql-cli/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "astro-sql-cli"
-version = "0.3.1a0"
+version = "0.3.0"
 description = "Empower analysts to build workflows to transform data using SQL"
 authors = [
     "Astronomer <humans@astronomer.io>",
--- a/sql-cli/pyproject.toml
+++ b/sql-cli/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "astro-sql-cli"
-version = "0.3.0"
+version = "0.3.1a0"
 description = "Empower analysts to build workflows to transform data using SQL"
 authors = [
     "Astronomer <humans@astronomer.io>",
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -88,8 +88,8 @@ def version(
     if as_json:
         output_dict = {"version": sql_cli.__version__}
         print(json.dumps(output_dict))
-        return
-    rprint("Astro SQL CLI", sql_cli.__version__)
+    else:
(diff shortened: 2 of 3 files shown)


sql-cli/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "astro-sql-cli"
-version = "0.3.0"
+version = "0.3.1a0"
```
feluelle, line 3: Isn't it worth having this separately? 🤔

sql-cli/sql_cli/__main__.py
```diff
@@ -79,7 +80,15 @@ def set_debug_mode(debug: bool) -> None:
...
+def version(
+    as_json: bool = typer.Option(
+        False, "--json", help="If the response should be in JSON format", show_default=True
+    ),
+) -> None:
+    if as_json:
+        output_dict = {"version": sql_cli.__version__}
+        print(json.dumps(output_dict))
+        return
     rprint("Astro SQL CLI", sql_cli.__version__)
```
feluelle, line 92:
```suggestion
    if as_json:
        output_dict = {"version": sql_cli.__version__}
        print(json.dumps(output_dict))
    else:
        rprint("Astro SQL CLI", sql_cli.__version__)
```

I think it is nicer to use if/else here just to make it more clear that it only prints one or the other.

sql-cli/tests/test___main__.py
```diff
@@ -114,6 +114,12 @@ def test_version():
     assert f"Astro SQL CLI {__version__}" in result.stdout
 
 
+def test_version_json_output():
+    result = runner.invoke(app, ["version", "--json"])
+    assert result.exit_code == 0
+    assert f'{{"version": "{__version__}"}}' in result.stdout
```
feluelle, line 120:
```suggestion
    assert f'"version": "{__version__}"' in result.stdout
```
Does it make sense to check for a particular key/value pair instead of the whole json object? That way it is more clear what you expect it to contain. I don't know 🤷 On the other side the entire stdout should be a valid json.

sql-cli/pyproject.toml
pankajkoti: I wanted to release from the current branch as it was needed for integration efforts. Will create a separate branch by cutting one from this branch and will keep rebasing it.

pankajkoti: Here is the other PR: https://github.com/astronomer/astro-sdk/pull/1658

feluelle (approved):

PR #1656, opened 2023-01-23
Use common sql 1.2.0 for airflow 2.2.5

utkarsharma2:
# Description
Right now this is not working for airflow 2.2.5

## What is the new behavior?
We need to select a different version of common-sql==1.2.0 provider

Nope

### Checklist


pankajastro (approved):
LGTM. We will more tests and compatibility in https://github.com/astronomer/astro-sdk/issues/1659

Issue #1651, opened 2023-01-23
Use workflow-name as an arg instead of flag in deploy command


pankajkoti:
As mentioned in https://github.com/astronomer/astro-sdk/issues/1520 , workflow-name should have been an arg but it is a flag currently.

tatiana:
This was released yesterday as part of SQL CLI 0.3.1  and a new Astro CLI 1.10 pre-release

PR #1660, opened 2023-01-23
Display workflow name as argument instead of flag in help

pankajkoti:
# Description

Currently, workflow name is used as a flag in the Astro CLI deploy command.

related: #1651 

## What is the new behavior?

As per the original specification, workflow name should be an argument and not a flag.
Same implementaiton changes are being done in the Astro CLI deploy cobra command, but
since Astro CLI relies on flow's help, we need to change the type of it here.

### Checklist

### DIFF
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -308,7 +308,7 @@ def init(
     help="""Deploy workflows to Astro Cloud.""",
 )
 def deploy(
-    workflow_name: str = typer.Option(  # skipcq: PYL-W0613
+    workflow_name: str = typer.Argument(  # skipcq: PYL-W0613
         default=None,
         show_default=False,
         help="name of the workflow directory within workflows directory.",
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -309,7 +309,7 @@ def init(
 )
 def deploy(
     workflow_name: str = typer.Argument(  # skipcq: PYL-W0613
-        default=...,
+        default=None,
         show_default=False,
         help="name of the workflow directory within workflows directory.",
     ),
--- a/sql-cli/sql_cli/__main__.py
+++ b/sql-cli/sql_cli/__main__.py
@@ -309,7 +309,7 @@ def init(
 )
 def deploy(
     workflow_name: str = typer.Argument(  # skipcq: PYL-W0613
-        default=None,
+        default=...,
         show_default=False,
         help="name of the workflow directory within workflows directory.",
     ),


feluelle (approved):

sql-cli/sql_cli/__main__.py
```diff
@@ -308,7 +308,7 @@ def init(
     help="""Deploy workflows to Astro Cloud.""",
 )
 def deploy(
-    workflow_name: str = typer.Option(  # skipcq: PYL-W0613
+    workflow_name: str = typer.Argument(  # skipcq: PYL-W0613
         default=None,
```
feluelle, line 312:
```suggestion
        default=...,
```

sql-cli/sql_cli/__main__.py
```diff
@@ -308,8 +308,8 @@ def init(
     help="""Deploy workflows to Astro Cloud.""",
 )
 def deploy(
-    workflow_name: str = typer.Option(  # skipcq: PYL-W0613
-        default=None,
+    workflow_name: str = typer.Argument(  # skipcq: PYL-W0613
+        default=...,
```
pankajkoti, line 312: @feluelle Why do we need to make it required? I was thinking that if no value is passed it means generate and deploy all workflows.

feluelle: Ah you are right 👍. But then we should already implement that here or?

feluelle: Let's do it separately. WDYT?

pankajkoti: No, if we keep it to None it accepts a parameter or None is fine too. If None then we generate all workflows and when a specific workflow is given it generates that specific. So my suggestion was we keep it as it is to None instead of making it required with `...`

pankajkoti, line 312:
```suggestion
        default=None,
```
@feluelle my suggestion we revert it back to this.

feluelle: Sorry 🤦 I thought we would do the generate in this function.

feluelle: Yes.

tatiana:
It's looking good, @pankajkoti , happy for us to merge this once it's rebased and tests are passing! I'll create a 0.3.1 milestone so we keep track of these changes

pankajkoti:
> It's looking good, @pankajkoti , happy for us to merge this once it's rebased and tests are passing! I'll create a 0.3.1 milestone so we keep track of these changes

I created the milestone and assigned it to the issues I am working on. Need your help in reviewing the milestone and setting the due date @tatiana 🙏🏽

PR #1671, opened 2023-01-24
Update compat matrix for 0.3.1

pankajkoti:
Runtime image 7.2.0 comes with Python SDK installed, so update that too in the matrix.

### DIFF
--- a/sql-cli/config/astro-cli.json
+++ b/sql-cli/config/astro-cli.json
@@ -15,15 +15,7 @@
   },
   "baseDockerImage": "quay.io/astronomer/astro-runtime:6.0.4-base",
   "compatibility": {
-    "0.3.0rc1": {
-      "astroRuntimeVersion": ">=7.2.0",
-      "astroSDKPythonVersion": ">=1.3.1"
-    },
     "0.3.1": {
-      "astroRuntimeVersion": ">=7.2.0",
-      "astroSDKPythonVersion": ">=1.3.1"
-    },
-    "0.3.1a0": {
       "astroRuntimeVersion": ">=7.2.0",
       "astroSDKPythonVersion": ">=1.3.1"
     },
--- a/sql-cli/config/astro-cli.json
+++ b/sql-cli/config/astro-cli.json
@@ -4,11 +4,11 @@
       "preRelease": true,
       "sqlCliVersion": ">0.2.2"
     },
-    "1.8": {
+    "1.8.0": {
       "preRelease": false,
       "sqlCliVersion": ">0.2, <0.3"
     },
-    "1.9": {
+    "1.9.0": {
       "preRelease": false,
       "sqlCliVersion": ">0.2, <0.3"
     }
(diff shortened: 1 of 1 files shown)


sql-cli/config/astro-cli.json
```diff
@@ -4,28 +4,32 @@
       "preRelease": true,
       "sqlCliVersion": ">0.2.2"
     },
-    "1.8": {
+    "1.8.0": {
       "preRelease": false,
       "sqlCliVersion": ">0.2, <0.3"
     },
-    "1.9": {
+    "1.9.0": {
```
tatiana, line 11: Why did we add the micro digit here? It would probably be less maintenance if we could have only major + minor (no micro) in the config. My concern is if we need to update this document for every release of the Astro CLI. We can have a dedicated ticket to implement this on the Astro CLI side in case the current implementation doesn't support this.

sql-cli/config/astro-cli.json
```diff
@@ -4,28 +4,32 @@
...
       "sqlCliVersion": ">0.2, <0.3"
     }
   },
   "baseDockerImage": "quay.io/astronomer/astro-runtime:6.0.4-base",
   "compatibility": {
     "0.3.0rc1": {
-      "astroRuntimeVersion": ">=8.0.0",
+      "astroRuntimeVersion": ">=7.2.0",
+      "astroSDKPythonVersion": ">=1.3.1"
+    },
```
tatiana, line 21: We can probably delete the pre-releases when they are no longer necessary (both "0.3.0rc1" and "0.3.1a0"), since the Astro CLI shouldn't be using those anymore, WDYT?

sql-cli/config/astro-cli.json
pankajkoti: Yes, we need a ticket to work on matching the minor versions. At the moment the integration checks for exact matches.

sql-cli/config/astro-cli.json
```diff
@@ -4,28 +4,32 @@
...
     },
-    "1.9": {
+    "1.9.0": {
       "preRelease": false,
       "sqlCliVersion": ">0.2, <0.3"
     }
   },
   "baseDockerImage": "quay.io/astronomer/astro-runtime:6.0.4-base",
   "compatibility": {
     "0.3.0rc1": {
```
pankajkoti, line 18:
```suggestion
```

sql-cli/config/astro-cli.json
pankajkoti, line 21:
```suggestion
```

sql-cli/config/astro-cli.json
```diff
@@ -4,28 +4,32 @@
...
-      "astroRuntimeVersion": ">=8.0.0",
+      "astroRuntimeVersion": ">=7.2.0",
+      "astroSDKPythonVersion": ">=1.3.1"
+    },
+    "0.3.1": {
+      "astroRuntimeVersion": ">=7.2.0",
       "astroSDKPythonVersion": ">=1.3.1"
     },
     "0.3.1a0": {
```
pankajkoti, line 26:
```suggestion
```

sql-cli/config/astro-cli.json
```diff
@@ -4,28 +4,32 @@
...
-      "astroRuntimeVersion": ">=8.0.0, <8.1.0",
-      "astroSDKPythonVersion": ">=1.3.0"
+      "astroRuntimeVersion": ">=7.2.0",
+      "astroSDKPythonVersion": ">=1.3.1"
     },
```
pankajkoti, line 29:
```suggestion
```

sql-cli/config/astro-cli.json
tatiana: https://github.com/astronomer/astro-sdk/issues/1673

tatiana (approved):

PR #1674, opened 2023-01-25
Improve tests

utkarsharma2:
# Description
Currently, if there is an issue with a dag, it won't be loaded in the DagBag and hence will never run, leading to wrong assumptions of the integration test working.

## What is the new behavior?
Moved the logic to run generate DagBag in the fixture. 

Nope

### Checklist


tatiana (approved):

python-sdk/tests_integration/test_example_dags.py
```diff
@@ -83,15 +85,14 @@ def order(dag_id: str) -> int:
...
+def example_dag_ids():
+    dag_bag: DagBag = get_dag_bag()
+    for dag_id in sorted(dag_bag.dag_ids, key=order):
+        yield dag_id, dag_bag
 
 
-@pytest.mark.parametrize("dag_id", sorted(dag_bag.dag_ids, key=order))
-def test_example_dag(session, dag_id: str):
+def test_example_dag(session, example_dag_ids):
+    dag_id, dag_bag = example_dag_ids
```
tatiana, line 96: Would it make sense to assert dag_bag import_errors is false?

python-sdk/tests_integration/test_example_dags.py
```diff
@@ -83,15 +85,14 @@ def order(dag_id: str) -> int:
     return -1
 
 
-dag_bag = get_dag_bag()
+@pytest.fixture()
+def example_dag_ids():
+    dag_bag: DagBag = get_dag_bag()
+    for dag_id in sorted(dag_bag.dag_ids, key=order):
+        yield dag_id, dag_bag
```
feluelle, line 92: I think we can make `dag_bag` its own fixture. WDYT?

python-sdk/tests_integration/test_example_dags.py
utkarsharma2: @tatiana We are doing this at the time of creation in the `get_dag_bag()` method.

python-sdk/tests_integration/test_example_dags.py
utkarsharma2: @feluelle We can, but are you also suggesting using it individually in some tests?

feluelle: Yes so that you don't have to return it there.

Instead, what you can do then is the following:
```python
@pytest.fixture(scope="module")
def dag_bag() -> DagBag:
    return get_dag_bag()


@pytest.fixture()
def example_dag_ids(dag_bag) -> str.
    for dag_id in sorted(dag_bag.dag_ids, key=order):
        yield dag_id


def test_example_dag(session, example_dag_ids, dag_bag):
    ...
```

utkarsharma2: Got it, thanks! 👍

utkarsharma2: @feluelle the existing code is raising an issue since there can only be one yield statement in a fixture. I have taken new approach to this, please take a look.

feluelle (approved):

python-sdk/tests_integration/test_example_dags.py
utkarsharma2: Resolved this on a call, and pushed agreed-upon changes.

Issue #1486, opened 2022-12-22
 Snowflake Native Path for Azure Blob storage (WASB) #1485 


utkarsharma2:
**Please describe the feature you'd like to see**
Add Snowflake Native Path for Azure Blob Storage

Ref: https://docs.snowflake.com/en/user-guide/data-load-azure-config.html

Issue #1448, opened 2022-12-16
Allow load and export from Azure Blob storage (WASB) using native path


tatiana:
**Please describe the feature you'd like to see**
Once we've introduced the Azure Blob Storage file location (WASB) #905 #1275 the next step is to make sure we can have load and export operations using the native path.

Relates to: #1447

**Describe the solution you'd like**
Users should be able to load and export files using WASB using the native path

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

phanikumv:
No native path is available for export from Azure as discussed in the sprint planning.

utkarsharma2:
As per my discussion with @tatiana we only native path for load_file and not export_file.
for load_file() there are two paths:
1. Snowflake-Azure - PR: https://github.com/astronomer/astro-sdk/pull/1675
2. Bigquery-Azure - Internal issue with Bigquey, which only bigquery can handle, we have raised a ticket with google for this. Ticket - https://github.com/astronomer/astro-sdk/issues/1485

PR #1675, opened 2023-01-25
Snowflake - azure native path

utkarsharma2:
# Description
Right now if we have to load data from Azure to snowflake we have to involve the worker node, which is time taking process.

closes: #1486 
closes: #1448

## What is the new behavior?
Add Snowflake - Azure native path to load data directly from Azure to Snowflake, without involving a worker node.

Nope

### Checklist


python-sdk/src/astro/files/locations/azure/wasb.py
```diff
@@ -91,3 +93,21 @@ def openlineage_dataset_name(self) -> str:
...
+         To bridge the gap we use this method
+        """
+        if not check_required_option(self.load_options, "storage_account"):
+            raise ValueError(
+                f"Required param missing 'storage_account', pass {self.LOAD_OPTIONS_CLASS_NAME}"
+                f"(storage_account=<account_name>) to load_options"
+            )
+        url = urlparse(self.path)
+        azure_host = "blob.core.windows.net"
+        return f"{FileLocation.AZURE}://{self.load_options.storage_account}.{azure_host}/{url.netloc}/"  # type: ignore
```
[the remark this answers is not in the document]
utkarsharma2: @tatiana We are already checking for `None` and `load_options` not having `storage_account` by function `check_required_option` and MyPy still raises a type error. Is there a better way to deal with this?

python-sdk/src/astro/files/locations/azure/wasb.py
```diff
@@ -91,3 +93,21 @@ def openlineage_dataset_name(self) -> str:
         https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
         """
         return urlparse(self.path).path
+
+    def get_new_path_for_snowflake_stage(self) -> str:
```
tatiana, line 97: Why not a property named `snowflake_stage_path`?

python-sdk/src/astro/files/locations/azure/wasb.py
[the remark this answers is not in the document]
tatiana: Would it be worth to use `urllib.parse.urlunparse` instead of pure string concatenation?

python-sdk/src/astro/options.py
```diff
@@ -65,3 +74,8 @@ class SnowflakeLoadOptions(LoadOptions):
 
     def empty(self):
         return not self.file_options and not self.copy_options
+
+
+@attr.define
+class WASBLocationLoadOptions(LoadOptions):
+    storage_account: str = attr.field(default=None)
```
tatiana, line 81: Shouldn't `storage_account` be part of the `SnowflakeOptions`, since it's not required by WASB in an isolated manner - one when the transfer is between any cloud storage and Snowflake?

python-sdk/src/astro/options.py
```diff
@@ -15,6 +15,15 @@ def to_dict(self) -> dict:
         return attr.asdict(self)
 
 
+def check_required_option(load_options: Optional[LoadOptions], option_name: str) -> bool:
```
tatiana, line 18:
```suggestion
def is_required_option(load_options: Optional[LoadOptions], option_name: str) -> bool:
```

python-sdk/src/astro/options.py
```diff
@@ -15,6 +15,15 @@ def to_dict(self) -> dict:
...
+    """
+    Check required options in load_option class
+    """
+    if load_options is None or getattr(load_options, option_name, None) is None:
+        return False
+    return True
```
feluelle, line 24:
```suggestion
    return load_options and getattr(load_options, option_name)
```

python-sdk/src/astro/files/locations/azure/wasb.py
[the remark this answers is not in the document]
tatiana: @utkarsharma2 did you try doing
```suggestion
        return f"{FileLocation.AZURE}://{str(self.load_options.storage_account)}.{azure_host}/{url.netloc}/"
```

python-sdk/src/astro/options.py
utkarsharma2: storage_account is specific to azure.

utkarsharma2: storage_account is specific to azure.

python-sdk/src/astro/options.py
utkarsharma2: With this method, we are checking if the required option is present or not.`is_required_option` to me gives an impression to check if a given option is required. I think it should be `check_required_option` WDYT?

python-sdk/src/astro/files/locations/azure/wasb.py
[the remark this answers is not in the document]
utkarsharma2: Using `str(self.load_options.storage_account)` will still raise the same exception right? Since the attribute may or not be present even for str() function

python-sdk/src/astro/files/locations/azure/wasb.py
utkarsharma2: Done.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -191,7 +197,8 @@ def set_url_from_file(self, file: File) -> None:
         """
         # the stage URL needs to be the folder where the files are
         # https://docs.snowflake.com/en/sql-reference/sql/create-stage.html#external-stage-parameters-externalstageparams
-        url = file.path[: file.path.rfind("/") + 1]
+        path = file.location.snowflake_stage_path
+        url = path[: path.rfind("/") + 1]
```
tatiana, line 201: @utkarsharma2, could we improve this by using stdlib `urlparse`?

python-sdk/src/astro/files/locations/azure/wasb.py
```diff
@@ -89,3 +91,31 @@ def openlineage_dataset_name(self) -> str:
...
+         "wasb://<container_name>/<filename>" or "wasbs://<container_name>/<filename>"
+         To bridge the gap we use this method
+        """
+        if not check_required_option(self.load_options, "storage_account"):
+            raise ValueError(
+                f"Required param missing 'storage_account', pass {self.LOAD_OPTIONS_CLASS_NAME}"
+                f"(storage_account=<account_name>) to load_options"
+            )
+        url = urlparse(self.path)
+        azure_host = "blob.core.windows.net"
```
tatiana, line 111: WDYT if this was a constant?

python-sdk/src/astro/options.py
tatiana: Hm, I see. What about `contains_required_option`? It would be more explicit!

utkarsharma2: Yup, this sounds much better. Updated it.

python-sdk/src/astro/files/locations/azure/wasb.py
utkarsharma2: Yup, this would be better. Done.

python-sdk/src/astro/databases/snowflake.py
utkarsharma2: @tatiana  Yup, done it. please check.

python-sdk/tests/files/locations/test_wasb.py
```diff
@@ -43,3 +45,26 @@ def test_wasb_smartopen_uri_wasbs_scheme():
...
     assert location.smartopen_uri == "azure://somepath"
+
+
+def test_snowflake_stage_path_raise_exception():
+    """
+    Test snowflake_stage_path raise exception when 'storage_account' is missing.
+    """
+    location = WASBLocation(path="azure://somepath")
+    error_message = f"Required param missing 'storage_account', pass {location.LOAD_OPTIONS_CLASS_NAME}"
+    "(storage_account=<account_name>) to load_options"
```
pankajastro, line 56:
```suggestion
    error_message = (
        f"Required param missing 'storage_account', pass {location.LOAD_OPTIONS_CLASS_NAME} "
        f"(storage_account=<account_name>) to load_options"
    )
```
look like f is missing here

pankajastro (approved):
LGTM

python-sdk/tests/files/locations/test_wasb.py
utkarsharma2: @pankajastro It's just an example with a placeholder `<account_name>`. I didn't wan this to be a fstring.

pankajastro: I see but it does not look right if we beak a long string in multiline half of them are f-string and half are not. even in this PR you have f-string without parameter https://github.com/astronomer/astro-sdk/pull/1675/files#diff-14125a15a8984cccdb2681309d0e24b1ae1800474ac05a3c319264a8446b36c0R107

tatiana (approved):

PR #1676, opened 2023-01-25
Refactor snowflake stage auth code

utkarsharma2:
# Description
Refactored code for getting the snowflake stage auth

## What is the new behavior?
Since that code was specific to snowflake and location, moved it to file-locations package

Nope

### Checklist

### DIFF
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -357,7 +357,7 @@ class SnowflakeDatabase(BaseDatabase):
         )
         if storage_integration is not None:
             return f"storage_integration = {storage_integration};"
-        return file.location.get_stage_auth_sub_statement()
+        return file.location.get_snowflake_stage_auth_sub_statement()
 
     def create_stage(
         self,
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -356,23 +356,8 @@ class SnowflakeDatabase(BaseDatabase):
             file.location.location_type
         )
         if storage_integration is not None:
-            auth = f"storage_integration = {storage_integration};"
-        else:
-            if file.location.location_type == FileLocation.GS:
-                raise DatabaseCustomError(
-                    "In order to create an stage for GCS, `storage_integration` is required."
-                )
-            elif file.location.location_type == FileLocation.S3:
-                aws = file.location.hook.get_credentials()
-                if aws.access_key and aws.secret_key:
-                    auth = f"credentials=(aws_key_id='{aws.access_key}' aws_secret_key='{aws.secret_key}');"
-                else:
-                    raise DatabaseCustomError(
-                        "In order to create an stage for S3, one of the following is required: "
-                        "* `storage_integration`"
-                        "* AWS_KEY_ID and SECRET_KEY_ID"
-                    )
-        return auth
+            return f"storage_integration = {storage_integration};"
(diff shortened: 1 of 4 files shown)


python-sdk/src/astro/databases/snowflake.py
```diff
@@ -356,23 +356,8 @@ def _create_stage_auth_sub_statement(file: File, storage_integration: str | None
             file.location.location_type
         )
         if storage_integration is not None:
-            auth = f"storage_integration = {storage_integration};"
```
tatiana, line 359: @utkarsharma2 I have mixed feelings about this refactoring. One of the goals of having the database packages/classes was for us to unify things specific to that database in a single place - so it would be easy to maintain and change if there were changes in the database spec. I know storage integration is a tricky concept because it is the glue between the database and the object storage - but I still think it is closer to the database since Snowflake is the only one which depends on this. My vote would be to keep this in the Snowflake class, and not move it to the file locations.

feluelle: I personally like that change, because it removes the check on `file.location`. With the refactored version we have reduced the dependencies from depending on N file locations to 1 file location at a time.

utkarsharma2: @tatiana I agree we initially decided to unify things specific to the database in a single place, however in this case I have two reasons why I think it can be beneficial if we move it.

1. Since the snowflake database is constant, all the possible changes that will happen in the file-location module and if we are making changes to the file-location those changes should not affect the snowflake database module.
2. Also, we are more likely to trip if this is part of the database module. When I'm adding a new file location I also have to remember this case in the Snowflake Database module VS if we move this to File-Location, then we are forced to deal with this case since it's a specific method in the module.

WDYT?

python-sdk/src/astro/files/locations/base.py
```diff
@@ -188,3 +189,6 @@ def get_stream(self):
         :param df: pandas dataframe
         """
         return smart_open.open(self.smartopen_uri, mode="wb", transport_params=self.transport_params)
+
+    def get_snowflake_stage_auth_sub_statement(self) -> str:
+        raise DatabaseCustomError("In order to create an stage, `storage_integration` is required.")
```
kaxil, line 194:
```suggestion
        raise DatabaseCustomError("In order to create a stage, `storage_integration` is required.")
```

kaxil (approved):

python-sdk/src/astro/databases/snowflake.py
kaxil: I like the change too

python-sdk/src/astro/files/locations/base.py
```diff
@@ -188,3 +189,6 @@ def get_stream(self):
         :param df: pandas dataframe
         """
         return smart_open.open(self.smartopen_uri, mode="wb", transport_params=self.transport_params)
+
+    def get_snowflake_stage_auth_sub_statement(self) -> str:
```
sunank200, line 193: Is this error relevant for every file location?

sunank200 (commented):
I like this change. I think the tests are missing. Added few comments

python-sdk/src/astro/files/locations/base.py
utkarsharma2: I think this is safe a default at the moment. Since this methods is only called for native transfer 
and  2/3 (Azure and GCS) needs to raise this error if the `storage_integration` is not passed.

sunank200 (approved):

Issue #581, opened 2022-07-27
Improve native Snowflake load_file to be verbose about ON_ERROR issues


tatiana:
Currently, the `load_file` operator can hide errors when doing native Snowflake transfers. We should raise an exception and display the errors to the users, not hiding this problem.

As it stands, a user may be loading 10k rows into a table, end up with five rows, and not realise there is an issue.

**Acceptance criteria**

* [x] By default: throw the error using Validation mode (so the task fails), explain user can pass ON_ERROR=CONTINUE
* [x] Allow users to pass ON_ERROR and other arguments in load_options... => how, still need to be detailed
* [x] What is the current behaviour when enable_fallback is True — and native load if tried which fails but loads 100 rows out of 500 — Will the pandas path load all the 500 rows --- which means in total 600 rows will be added with 100 duplicated
* [x] Update the documentation 

**References**
`VALIDATION_MODE` https://community.snowflake.com/s/question/0D50Z00008j0ftlSAA/is-there-a-way-to-use-both-onerrorcontinue-validationmode-returnallerrors-in-copy-into-statement-

tatiana:
@phanikumv @sunank200 @utkarsharma2 I believe this ticket is still valid, except that we should be doing this as part of `LoadOptions` - so we could rephrase it.

The problem is that with `ON_ERROR=CONTINUE`, we are hiding potential issues loading rows into Snowflake - we need to make sure that we give users the visibility of this and allow them to opt-in for this.

PR #1689, opened 2023-01-31
Add validation mode support in COPY INTO in snowflake

sunank200:
# Description

Currently, the load_file operator can hide errors when doing native Snowflake transfers. We should raise an exception and display the errors to the users, not hide this problem.

As it stands, a user may be loading 10k rows into a table, end up with five rows, and not realise there is an issue.

closes: #581

## What is the new behaviour?

- Add validation mode as part of the `COPY INTO` command. Specify the supported validation mode; `RETURN_n_ROWS` or `RETURN_ERRORS` or  `RETURN_ALL_ERRORS`. This instructs the COPY command to validate the data files instead of loading them into the specified table; i.e. the COPY command tests the files for errors but does not load them. Read more at:
        https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#optional-parameters
- Remove the `ON_ERROR=Continue` hardcoded for CSV file types
- Add an option for the user to pass `ON_ERROR=Continue` as part of `copy_options`

Example:
```
aql.load_file(
        input_file=File("s3://astro-sdk/python_sdk/example_dags/data/sample.csv", conn_id="aws_conn"),
        output_table=Table(
            conn_id=SNOWFLAKE_CONN_ID,
        ),
        load_options=[
            SnowflakeLoadOptions(
                file_options={"SKIP_HEADER": 1, "SKIP_BLANK_LINES": True},
                copy_options={"ON_ERROR": "CONTINUE"},
                validation_mode="RETURN_ALL_ERRORS",
            )
        ],
    )
```

Yes

### Checklist

### DIFF
--- a/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
+++ b/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
@@ -78,6 +78,7 @@ def example_amazon_s3_snowflake_transform():
         output_table=input_table_1,
         load_options=[
             SnowflakeLoadOptions(
+                file_options={"TYPE": "CSV", "TRIM_SPACE": True},
                 copy_options={"ON_ERROR": "CONTINUE"},
             )
         ],
@@ -87,6 +88,7 @@ def example_amazon_s3_snowflake_transform():
         output_table=input_table_2,
         load_options=[
             SnowflakeLoadOptions(
+                file_options={"TYPE": "CSV", "TRIM_SPACE": True},
                 copy_options={"ON_ERROR": "CONTINUE"},
             )
         ],
--- a/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
+++ b/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
@@ -9,6 +9,7 @@ from airflow.decorators import dag
 
 from astro import sql as aql
 from astro.files import File
+from astro.options import SnowflakeLoadOptions
 from astro.table import Metadata, Table
 
 
@@ -75,10 +76,22 @@ def example_amazon_s3_snowflake_transform():
     temp_table_1 = aql.load_file(
         input_file=File(path=f"{s3_bucket}/ADOPTION_CENTER_1_unquoted.csv"),
         output_table=input_table_1,
+        load_options=[
+            SnowflakeLoadOptions(
+                file_options={"SKIP_HEADER": 1, "SKIP_BLANK_LINES": True},
+                copy_options={"ON_ERROR": "CONTINUE"},
+            )
+        ],
     )
     temp_table_2 = aql.load_file(
         input_file=File(path=f"{s3_bucket}/ADOPTION_CENTER_2_unquoted.csv"),
         output_table=input_table_2,
+        load_options=[
+            SnowflakeLoadOptions(
+                file_options={"SKIP_HEADER": 1, "SKIP_BLANK_LINES": True},
+                copy_options={"ON_ERROR": "CONTINUE"},
+            )
+        ],
     )
 
     combined_data = combine_data(
--- a/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
+++ b/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
@@ -78,7 +78,6 @@ def example_amazon_s3_snowflake_transform():
         output_table=input_table_1,
         load_options=[
             SnowflakeLoadOptions(
-                file_options={"SKIP_HEADER": 1, "SKIP_BLANK_LINES": True},
                 copy_options={"ON_ERROR": "CONTINUE"},
             )
         ],
@@ -88,7 +87,6 @@ def example_amazon_s3_snowflake_transform():
         output_table=input_table_2,
         load_options=[
             SnowflakeLoadOptions(
-                file_options={"SKIP_HEADER": 1, "SKIP_BLANK_LINES": True},
                 copy_options={"ON_ERROR": "CONTINUE"},
             )
         ],
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -406,7 +406,7 @@ class SnowflakeDatabase(BaseDatabase):
                 auth,
             ]
         )
-
+        logging.debug("SQL statement executed: %s ", sql_statement)
         self.run_sql(sql_statement)
 
         return stage
@@ -639,21 +639,24 @@ class SnowflakeDatabase(BaseDatabase):
             self.drop_stage(stage)
         return rows
 
-    def _validate_before_copy_into(self, source_file, target_table, stage):
+    def _validate_before_copy_into(
+        self, source_file: File, target_table: BaseTable, stage: SnowflakeStage
+    ) -> None:
         """Validate COPY INTO command to tests the files for errors but does not load them."""
-        if self.load_options.validation_mode is None:
-            return
-        table_name = self.get_table_qualified_name(target_table)
-        file_path = os.path.basename(source_file.path) or ""
-        sql_statement = (
-            f"COPY INTO {table_name} FROM "
-            f"@{stage.qualified_name}/{file_path} VALIDATION_MODE='{self.load_options.validation_mode}'"
-        )
-        try:
-            self.hook.run(sql_statement, handler=lambda cur: cur.fetchall())
-        except ProgrammingError as load_exception:
-            self.drop_stage(stage)
-            raise load_exception
+        if self.load_options:
+            if self.load_options.validation_mode is None:
+                return
+            table_name = self.get_table_qualified_name(target_table)
+            file_path = os.path.basename(source_file.path) or ""
+            sql_statement = (
+                f"COPY INTO {table_name} FROM "
+                f"@{stage.qualified_name}/{file_path} VALIDATION_MODE='{self.load_options.validation_mode}'"
+            )
+            try:
+                self.hook.run(sql_statement, handler=lambda cur: cur.fetchall())
+            except ProgrammingError as load_exception:
+                self.drop_stage(stage)
+                raise load_exception
 
     @staticmethod
     def evaluate_results(rows):
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -619,10 +619,9 @@ class SnowflakeDatabase(BaseDatabase):
     def _copy_into_table_from_stage(self, source_file, target_table, stage):
         table_name = self.get_table_qualified_name(target_table)
         file_path = os.path.basename(source_file.path) or ""
-        sql_statement = (
-            f"COPY INTO {table_name} FROM "
-            f"@{stage.qualified_name}/{file_path} VALIDATION_MODE='{self.load_options.validation_mode}'"
-        )
+        sql_statement = f"COPY INTO {table_name} FROM @{stage.qualified_name}/{file_path}"
+
+        self._validate_before_copy_into(source_file, target_table, stage)
 
         # Below code is added due to breaking change in apache-airflow-providers-snowflake==3.2.0,
         # we need to pass handler param to get the rows. But in version apache-airflow-providers-snowflake==3.1.0
@@ -639,6 +638,22 @@ class SnowflakeDatabase(BaseDatabase):
         finally:
             self.drop_stage(stage)
         return rows
+
+    def _validate_before_copy_into(self, source_file, target_table, stage):
+        """Validate COPY INTO command to tests the files for errors but does not load them."""
+        if self.load_options.validation_mode is None:
+            return
+        table_name = self.get_table_qualified_name(target_table)
+        file_path = os.path.basename(source_file.path) or ""
+        sql_statement = (
+            f"COPY INTO {table_name} FROM "
+            f"@{stage.qualified_name}/{file_path} VALIDATION_MODE='{self.load_options.validation_mode}'"
+        )
+        try:
+            self.hook.run(sql_statement, handler=lambda cur: cur.fetchall())
+        except self.NATIVE_LOAD_EXCEPTIONS as load_exception:
+            self.drop_stage(stage)
+            raise load_exception
 
     @staticmethod
     def evaluate_results(rows):
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -393,7 +393,7 @@ class SnowflakeDatabase(BaseDatabase):
         file_options = [f"{k}={v}" for k, v in self.load_options.file_options.items()]
         file_options.extend([f"TYPE={fileformat}", "TRIM_SPACE=TRUE"])
         file_options_str = ", ".join(file_options)
-        copy_options_str = ", ".join(copy_options)  # type:ignore
+        copy_options_str = ", ".join(copy_options)
         sql_statement = "".join(
             [
                 f"CREATE OR REPLACE STAGE {stage.qualified_name} URL='{stage.url}' ",
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -392,7 +392,7 @@ class SnowflakeDatabase(BaseDatabase):
         stage.set_url_from_file(file)
 
         fileformat = ASTRO_SDK_TO_SNOWFLAKE_FILE_FORMAT_MAP[file.type.name]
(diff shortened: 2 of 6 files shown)


utkarsharma2:
@sunank200 Why we don't need ON_ERROR=CONTINUE for other filetypes - `NDJSON` or `PARQUET`??

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -393,12 +392,12 @@ def create_stage(
         stage.set_url_from_file(file)
 
         fileformat = ASTRO_SDK_TO_SNOWFLAKE_FILE_FORMAT_MAP[file.type.name]
-        copy_options = [COPY_OPTIONS[file.type.name]]
+        copy_options = [COPY_OPTIONS.get(file.type.name)] if COPY_OPTIONS.get(file.type.name) else []
```
feluelle, line 395: Won't we face the same issue with `fileformat`?

feluelle: Shouldn't we error out, if the file type is not supported?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -638,6 +639,22 @@ def _copy_into_table_from_stage(self, source_file, target_table, stage):
             self.drop_stage(stage)
         return rows
 
+    def _validate_before_copy_into(self, source_file, target_table, stage):
+        """Validate COPY INTO command to tests the files for errors but does not load them."""
+        if self.load_options.validation_mode is None:
+            return
```
feluelle, line 645: The if should be wrapped around the `_validate_before_copy_into` function call.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -393,12 +392,12 @@ def create_stage(
         stage.set_url_from_file(file)
 
         fileformat = ASTRO_SDK_TO_SNOWFLAKE_FILE_FORMAT_MAP[file.type.name]
-        copy_options = [COPY_OPTIONS[file.type.name]]
+        copy_options = [COPY_OPTIONS.get(file.type.name)] if COPY_OPTIONS.get(file.type.name) else []
         copy_options.extend([f"{k}={v}" for k, v in self.load_options.copy_options.items()])
```
utkarsharma2, line 396: I think we should log the final copy_options passed. Also, it would be nice if we can list the default options per-file basis in docs and/or log.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -393,12 +392,12 @@ def create_stage(
...
         file_options = [f"{k}={v}" for k, v in self.load_options.file_options.items()]
         file_options.extend([f"TYPE={fileformat}", "TRIM_SPACE=TRUE"])
         file_options_str = ", ".join(file_options)
-        copy_options_str = ", ".join(copy_options)
+        copy_options_str = ", ".join(copy_options)  # type:ignore
```
utkarsharma2, line 400: what was the issue here?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -638,6 +639,22 @@ def _copy_into_table_from_stage(self, source_file, target_table, stage):
             self.drop_stage(stage)
         return rows
 
+    def _validate_before_copy_into(self, source_file, target_table, stage):
```
utkarsharma2, line 642: can we add type hints here?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -638,6 +639,22 @@ def _copy_into_table_from_stage(self, source_file, target_table, stage):
...
+        if self.load_options.validation_mode is None:
+            return
+        table_name = self.get_table_qualified_name(target_table)
+        file_path = os.path.basename(source_file.path) or ""
+        sql_statement = (
+            f"COPY INTO {table_name} FROM "
+            f"@{stage.qualified_name}/{file_path} VALIDATION_MODE='{self.load_options.validation_mode}'"
+        )
+        try:
+            self.hook.run(sql_statement, handler=lambda cur: cur.fetchall())
```
utkarsharma2, line 653: What are we doing with the return value of the handler?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -638,6 +639,22 @@ def _copy_into_table_from_stage(self, source_file, target_table, stage):
             self.drop_stage(stage)
         return rows
 
+    def _validate_before_copy_into(self, source_file, target_table, stage):
+        """Validate COPY INTO command to tests the files for errors but does not load them."""
+        if self.load_options.validation_mode is None:
+            return
+        table_name = self.get_table_qualified_name(target_table)
+        file_path = os.path.basename(source_file.path) or ""
```
utkarsharma2, line 647: @sunank200  if the file_path is empty, are we loading all the files in a folder?

python-sdk/src/astro/options.py
```diff
@@ -57,10 +57,16 @@ class SnowflakeLoadOptions(LoadOptions):
...
+    :param storage_integration: Specify the previously created Snowflake storage integration
+    :param validation_mode: Specify the supported validation mode; RETURN_n_ROWS | RETURN_ERRORS | RETURN_ALL_ERRORS
+        This instructs the COPY command to validate the data files instead of loading them into the specified table;
+        i.e. the COPY command tests the files for errors but does not load them. Read more at:
+        https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#optional-parameters
     """
 
     file_options: dict = attr.field(init=True, factory=dict)
     copy_options: dict = attr.field(init=True, factory=dict)
+    validation_mode: Optional[str] = attr.field(default=None)
```
utkarsharma2, line 69: isn't this part of copy_options only?

python-sdk/tests/databases/test_snowflake.py
```diff
@@ -94,6 +94,10 @@ def test_load_file_to_table_natively_for_fallback_raises_exception_if_not_enable
         metadata=Metadata(database="SNOWFLAKE_DATABASE", schema="SNOWFLAKE_SCHEMA"),
     )
     database = SnowflakeDatabase()
+    database.load_options = SnowflakeLoadOptions(
+        copy_options={"ON_ERROR": "CONTINUE"},
```
utkarsharma2, line 98: I think ON_ERROR=CONTINUE and `validation_mode` are mutually exclusive, if that's the case wouldn't it make sense to keep ON_ERROR=CONTINUE as default? because anyone who has specified the `validation_mode` will anyway run into an issue when loading fails, right?

python-sdk/src/astro/databases/snowflake.py
tatiana: @feluelle I don't think this line's matter is of the filetype being supported, but we're defining default values for COPY OPTIONS for some of the filetypes.

@sunank200 Are we making it clear in the documentation that we are setting these values as defaults for these specific filetypes? It may be worth at least having a link to the code if we don't want to duplicate the information.

Do you remember why we added these defaults? I believe it was due to real datasets, but it may be worth adding some comments as well.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -638,6 +639,22 @@ def _copy_into_table_from_stage(self, source_file, target_table, stage):
...
+        sql_statement = (
+            f"COPY INTO {table_name} FROM "
+            f"@{stage.qualified_name}/{file_path} VALIDATION_MODE='{self.load_options.validation_mode}'"
+        )
+        try:
```
tatiana, line 652: @sunank200, how are we handling timeout for this operation? I remember we saw there are scenarios when it is very slow - it would be great if we handled this gracefully.

sunank200:
> @sunank200 Currently, why we don't need ON_ERROR=CONTINUE for other filetypes - `NDJSON` or `PARQUET`?? Post this change this option will be applied to all the file types, right?

@utkarsharma2 This option was even available previously for all file types. The user could previously pass `ON_ERROR=CONTINUE` as copy_options

python-sdk/src/astro/options.py
sunank200: No this is not part of `copy_options`. `validation_mode` is a separate parameter which has to be passed. Read more at: https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#optional-parameters

python-sdk/tests/databases/test_snowflake.py
sunank200: Not necessarily @utkarsharma2.  `validation_mode` is not only for errors. There are three modes: `RETURN_n_ROWS` | `RETURN_ERRORS` | `RETURN_ALL_ERRORS`. And this just validates the data it doesn't load it.

sunank200: Validation mode runs only when specified and it doesn't load the data. It just instructs the COPY command to validate the data files instead of loading them into the specified table; i.e. the COPY command tests the files for errors but does not load them.

python-sdk/src/astro/databases/snowflake.py
sunank200: `mypy` was complaining that this is increasing the complexity of the code. Any better way you could suggest @feluelle ?

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -52,7 +52,6 @@
 }
 
 COPY_OPTIONS = {
-    FileType.CSV: "ON_ERROR=CONTINUE",
```
tatiana, line 55: @sunank200, could you confirm if our benchmark tests still run without this?

python-sdk/src/astro/databases/snowflake.py
sunank200: Yes it loads all the file path in a folder.

python-sdk/src/astro/databases/snowflake.py
sunank200: @tatiana one option is to completely remove all the `copy_options` by default and the user can pass in values as required depending on the type of file as they do in the `COPY INTO` command. WDYT?. Currently, users cannot override `MATCH_BY_COLUMN_NAME=CASE_INSENSITIVE` for `NDJSON` and `PARQUET` files as its set as default values.

python-sdk/src/astro/databases/snowflake.py
sunank200: @tatiana so validation mode actually doesn't load the data to the table it just validates all the data and files and throws the relevant errors. It instructs the COPY command to validate the data files instead of loading them into the specified table; i.e. the COPY command tests the files for errors but does not load them. 

There are three modes possible: `RETURN_n_ROWS` | `RETURN_ERRORS` | `RETURN_ALL_ERRORS`
- `RETURN_15_ROWS` for example validates the specified 15 rows if no errors are encountered; otherwise, fails at the first error encountered in the rows.
- `RETURN_ERRORS` returns all errors (parsing, conversion, etc.) across all files specified in the COPY statement.
- `RETURN_ALL_ERRORS` returns all errors across all files specified in the COPY statement, including files with errors that were partially loaded during an earlier load because the ON_ERROR copy option was set to CONTINUE during the load.

I think even for large files if users want to perform validations using this mode, it would just take double the time - the first half to validate and another half to just load the file. If the user doesn't wish to validate, in that case, it would throw the correct error anyways. That was my understanding after multiple runs for various queries I tried yesterday. WDYT, do we still require a timeout? I felt its not necessary in this case.

python-sdk/src/astro/databases/snowflake.py
sunank200: I will conform but it would require this parameter to be passed as `copy_options`.

python-sdk/src/astro/databases/snowflake.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1689/commits/c4a6aec82c6a4e85d0bd1605c0fc77c0a0634447 done

python-sdk/src/astro/databases/snowflake.py
sunank200: added a debug log for whole  `COPY INTO` SQL statement

sunank200: https://github.com/astronomer/astro-sdk/pull/1689/commits/c4a6aec82c6a4e85d0bd1605c0fc77c0a0634447

python-sdk/src/astro/databases/snowflake.py
sunank200: https://github.com/astronomer/astro-sdk/pull/1689/commits/c4a6aec82c6a4e85d0bd1605c0fc77c0a0634447

python-sdk/src/astro/databases/snowflake.py
sunank200: Following is the mypy error
```
Argument 1 to "join" of "str" has incompatible type "List[Optional[str]]"; expected "Iterable[str]"
```

python-sdk/src/astro/databases/snowflake.py
tatiana: I wonder why our synthetic dataset would require us to set this option - ideally, it would have the correct number of rows, columns and values. But I'm also happy for this to be added as a follow-up ticket!

python-sdk/src/astro/databases/snowflake.py
tatiana: @sunank200 I think it is a good solution - not to have defaults and allow users to specify them. It feels wrong to decide on the user's behalf without allowing them to override.

There was a reason this was done in the past - probably due to real datasets. What happens to our tests if we remove these defaults? What happens to the benchmark script if we move them?

python-sdk/src/astro/databases/snowflake.py
tatiana: @sunank200 would it make sense to represent this argument as `Iterable[str]`?

python-sdk/src/astro/databases/snowflake.py
sunank200: I have removed all the default `copy_options`.

python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
```diff
@@ -75,10 +76,22 @@ def example_amazon_s3_snowflake_transform():
     temp_table_1 = aql.load_file(
         input_file=File(path=f"{s3_bucket}/ADOPTION_CENTER_1_unquoted.csv"),
         output_table=input_table_1,
+        load_options=[
```
phanikumv, line 79: could you add documentation for these changes?

```

- Add validation mode as part of the COPY INTO command. Specify the supported validation mode; RETURN_n_ROWS or RETURN_ERRORS or RETURN_ALL_ERRORS. This instructs the COPY command to validate the data files instead of loading them into the specified table; i.e. the COPY command tests the files for errors but does not load them. Read more at:
https://docs.snowflake.com/en/sql-reference/sql/copy-into-table.html#optional-parameters
- Remove the ON_ERROR=Continue hardcoded for CSV file types
- Add an option for the user to pass ON_ERROR=Continue as part of copy_options

```

python-sdk/src/astro/databases/snowflake.py
sunank200: I think in synthetic datasets, csv had the first rows as headers and we did not have the option to skip the header implemented then. Hence, ON_ERROR=CONTINUE was there as default. 

Created a separate issue for this to do the benchmarking: https://github.com/astronomer/astro-sdk/issues/1699

python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
sunank200: Screenshot 2023-02-02 at 8 52 41 PM
This is already added as part of the doc-string and is visible for `SnowflakeLoadOptions`

sunank200: https://github.com/astronomer/astro-sdk/pull/1689/commits/832a3a61e5166a5fe31bfe26f102f06f3951062c

python-sdk/src/astro/databases/snowflake.py
sunank200: All test work without default `copy_options`.

sunank200:
Screenshot 2023-02-02 at 9 09 18 PM
@tatiana added the documentation for `SnowflakeLoadOptions`.

Screenshot 2023-02-02 at 9 10 06 PM

tatiana (approved):
Thanks for addressing all the feedback, @sunank200!
It will be essential to add details about this change in the 1.5 changelogs.

sunank200:
> Thanks for addressing all the feedback, @sunank200! It will be essential to add details about this change in the 1.5 changelogs.

Sure @tatiana

Issue #1446, opened 2022-12-16
Improve the Azure location implementation


tatiana:
**Context**

There are a few points which could be improved from the PR #1275.

They are:
* Use the `wasb_default_conn` for WASB tests (both in the `conftest.py` and in `integration_tests/files/locations/test_wasb.py`. Stop using `AIRFLOW_CONN_WASB_DEFAULT`
* Improve the complexity of the fixture `remote_files_fixture`  and of the method `_upload_or_delete_remote_file`, available in the `integration_tests/conftest.py`
* Replace the usage of the `_get_blob_client` by a non "internal use method

utkarsharma2:
Closing this since all three PRs are merged.

PR #1694, opened 2023-02-01
Azure test cleanup

utkarsharma2:
# Description
Right now we are not using airflow's default connection for `wasb_default` leading to explicitly checking and passing of conn_id.

related: #1446 

## What is the new behavior?
Have a way to override conn_id in airflow's db with one specified in `test-connection.yaml`. With this, we can pass the required connection string needed to auth with azure.

Nope

### Checklist

### DIFF
--- a/.github/ci-test-connections.yaml
+++ b/.github/ci-test-connections.yaml
@@ -94,7 +94,7 @@ connections:
     password: $DATABRICKS_TOKEN
     extra:
       http_path: /sql/1.0/warehouses/cf414a2206dfb397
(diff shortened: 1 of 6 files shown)


tatiana (commented):
@utkarsharma2 ir may be worth to comment how people should set the envvar lovely for this to work

utkarsharma2:
@tatiana we are not picking connection from the env variable, but instead we are doing so from test-connection.yaml file. I personally think it's easier to configure than env variable.

PR #1696, opened 2023-02-01
Refactored remote_files_fixture

utkarsharma2:
# Description
Currently, the complexity of 'remote_files_fixture' is more than 6

related: #1446 

## What is the new behavior?
Refactored code

### Checklist


tatiana (approved):
That looks much better, thanks @utkarsharma2!

I'm happy for this to be merged once the tests/checks pass.

Moving forward, it would be great if we had a class similar to `database`, but for file management, which allowed us to delete/create these fixtures without having to define all the hooks in the `conftest.py` file.

PR #1698, opened 2023-02-02
Fix databricks broken tests

tatiana:
Databricks changed their error message when a table does not exist.
This leads our main branch to fail.

### DIFF
--- a/python-sdk/src/astro/databases/databricks/delta.py
+++ b/python-sdk/src/astro/databases/databricks/delta.py
@@ -232,7 +232,7 @@ class DeltaDatabase(BaseDatabase):
             )
             return True
         except ServerOperationError as s:
-            if "Table or view not found" in s.message:
+            if "TABLE_OR_VIEW_NOT_FOUND" in s.message:
                 return False
             else:
                 raise s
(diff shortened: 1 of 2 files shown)


sunank200 (approved):

python-sdk/src/astro/databases/databricks/delta.py
```diff
@@ -232,7 +232,7 @@ def table_exists(self, table: BaseTable) -> bool:
             )
             return True
         except ServerOperationError as s:
-            if "Table or view not found" in s.message:
+            if "TABLE_OR_VIEW_NOT_FOUND" in s.message:
```
pankajkoti, line 235: I think this qualifies to be moved to a constant and we re-use the constant in the tests so that we don't need to change this in 2 places next time there is a change, no?

feluelle (approved):

python-sdk/src/astro/databases/databricks/delta.py
tatiana: @pankajkoti I just noticed that the test should be using this same method - so we have a single occurrence of this string. Do you think it is still worth having a constant for it?

pankajkoti: No strict opinion. :) Marking the conservation as resolved.

Issue #1687, opened 2023-01-31
run_raw_sql() not creating different tables when passed dataframes


utkarsharma2:
**Describe the bug**
When we pass multiple dataframes to `run_raw_sql()` the expected behavior is that the dataframes should be loaded to different temp tables. Currently, that's not happening, and the second dataframe is loaded to the same temp table.
 
**To Reproduce**
Try running the following dag
```
    @aql.run_raw_sql(handler=lambda x: pd.DataFrame(x.fetchall(), columns=x.keys()))
    def raw_sql_query_1(input_table: BaseTable):
        return "SELECT * from {{input_table}}"

    @aql.run_raw_sql(handler=lambda x: pd.DataFrame(x.fetchall(), columns=x.keys()))
    def raw_sql_query_2(input_table: BaseTable):
        return "SELECT * from {{input_table}}"

    @aql.run_raw_sql(handler=lambda x: pd.DataFrame(x.fetchall(), columns=x.keys()), conn_id="sqlite_default")
    def raw_sql_query_3(table_1: BaseTable, table_2: BaseTable):
        assert table_1.name != table_2.name
        return "SELECT 1 + 1"

    with sample_dag:
        results_1 = raw_sql_query_1(input_table=test_table)
        results_2 = raw_sql_query_2(input_table=test_table)
        _ = raw_sql_query_3(table_1=results_1, table_2=results_2)
```

**Expected behavior**
Every dataframe should be loaded to a different temp table. 

**Additional context**

PR #1700, opened 2023-02-02
Fix run_raw_sql() operartor

utkarsharma2:
# Description
When we pass multiple dataframes to run_raw_sql() the expected behavior is that the dataframes should be loaded to different temp tables. Currently, that's not happening, and all the data frames are loaded to the same temp table. Later dataframe overrides a former one.

closes: #1687

## What is the new behavior?
Every dataframe is loaded to a different temp table.

Nope

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -88,14 +88,12 @@ class BaseSQLDecoratedOperator(UpstreamTaskMixin, DecoratedOperator):
         # Find and load dataframes from op_arg and op_kwarg into Table
         self.create_output_table_if_needed()
         self.op_args = load_op_arg_dataframes_into_sql(  # type: ignore
-            conn_id=self.conn_id,
-            op_args=self.op_args,  # type: ignore
-            target_table=self.output_table.create_similar_table(),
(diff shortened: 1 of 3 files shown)


pankajastro (commented):
@utkarsharma2 how does it will behave after this change if I'll pass multiple Dataframe and a non-temp table? Does it will accept a list of output tables??

utkarsharma2:
@pankajastro there is no impact for `output_table` it will still be the same. This change addresses an issue due to which if multiple dataframes were passed to function, we were not putting them in separate temp tables, instead, we were overriding/replacing the same temp table and populating it with a new data frame.

Example:
```
df_1 = Dataframe({"col1": [1, 2], "col2": [3, 4]})
df_2 = Dataframe({"col3": [1, 2], "col3": [3, 4]})

@aql.run_raw_sql()
def run(tbl_1, tbl_2):
   return "select col1 from {{tbl_1}}"   # <--- this will fail
```
Because the data of `df_2` is loaded to `temp_table` dropping the data of `df_1` so tbl_1 and tbl_2 are the same table.

tatiana (approved):

feluelle:
Good catch! 🚀

cc @tatiana: I think this is the PR you had meant yesterday 👍

PR #1705, opened 2023-02-03
Discrepancy in database's run_sql() method

rajaths010494:
Currently, the handler param in the run_sql() method is only used in data bricks all other databases are not using it, which leads to different response types since the handler is processed within run_sql() for data bricks and not for other databases. Also, the signature of run_sql() in data bricks deviates from the base. We need to standardize and when we do, we can remove the below check as well.

Describe the solution you'd like
Ideally, we should have the same signatures for all the run_sql() database implementations. This will prevent adding explicit conditions for data bricks in other parts of the code that should ideally be database agnostic.
example: https://github.com/astronomer/astro-sdk/blob/2f8c8d0ccbff3cc13af78685cbefa39f473991c3/python-sdk/src/astro/sql/operators/raw_sql.py#L47

### DIFF
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -107,6 +107,7 @@ class BaseDatabase(ABC):
         self,
         sql: str | ClauseElement = "",
         parameters: dict | None = None,
+        handler: Callable | None = None,
         **kwargs,
     ) -> CursorResult:
         """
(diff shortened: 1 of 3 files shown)


sunank200:
This PR is still a work in progress. This requires a return statement to return the cursor object so that its consistent across all the databases. @rajaths010494 mentioned he would pick this up after other tickets are done for 1.5.1.

rajaths010494:
https://github.com/astronomer/astro-sdk/pull/1773 PR has been raised to address this issue

Issue #1673, opened 2023-01-24
Update SQL CLI compatibility implementation so support matching without the micro digit


tatiana:
**Please describe the feature you'd like to see**
At the moment, as discussed in https://github.com/astronomer/astro-sdk/pull/1671#discussion_r1085368779, our compatibility keys are represented by `<major>.<minor>.<micro>` as:
```
{
  "astroCliCompatibility": {
    "1.10.0": {
      "preRelease": true,
      "sqlCliVersion": ">0.2.2"
    },
    "1.8.0": {
      "preRelease": false,
      "sqlCliVersion": ">0.2, <0.3"
    },
    "1.9.0": {
      "preRelease": false,
      "sqlCliVersion": ">0.2, <0.3"
    }
  },
  "baseDockerImage": "quay.io/astronomer/astro-runtime:6.0.4-base",
  "compatibility": {
    "0.3.1": {
      "astroRuntimeVersion": ">=7.2.0",
      "astroSDKPythonVersion": ">=1.3.1"
    },
    ">0.2.2": {
      "astroRuntimeVersion": ">=7.2.0",
      "astroSDKPythonVersion": ">=1.3.1"
    }
  }
}
```
Ideally, we'd also be able to use just `<major>.<minor>` versions as keys, so we don't need to update the file for every Astro CLI/SQL CLI release, unless necessary:
```
{
  "astroCliCompatibility": {
    "1.10.2": {
      "preRelease": true,
      "sqlCliVersion": "<0.3.2"
    },
    "1.10": {
      "preRelease": true,
      "sqlCliVersion": ">0.2.2"
    },
    "1.8": {
      "preRelease": false,
      "sqlCliVersion": ">0.2, <0.3"
    },
    "1": {
      "preRelease": false,
      "sqlCliVersion": ">0.2, <0.3"
    }
  },
  "baseDockerImage": "quay.io/astronomer/astro-runtime:6.0.4-base",
  "compatibility": {
    "0.3": {
      "astroRuntimeVersion": ">=7.2.0",
      "astroSDKPythonVersion": ">=1.3.1"
    },
  }
}
```
Also, we need to make sure that if there is a new (minor) Astro CLI release, we don't fail - we should fallback to the sqlCliVersion non-prelease version in those cases.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
[...]

PR #1706, opened 2023-02-03
Add minor version and default match keys to compatibility matrix

pankajkoti:
# Description

Currently, Astro CLI version does not allow minor version match keys and need an exact match of
the format <MAJOR_VERSION>.<MINOR_VERSION>.<PATCH_VERSION>

related: #1673 

## What is the new behavior?
The plan is that Astro CLI integration should allow version match as per the below steps:
1. Try to find an exact match in the compatibility matrix e.g. 1.10.0
2. If no exact match is found, trim the version to remove the patch version and match until the minor version e.g. 1.10
3. If still does not match, match a key named 'default' which is set to install the latest SQL CLI version.

For the same reason, we're adding the required keys in the compatibility matrix with this PR.

Yes

### Checklist

### DIFF
--- a/sql-cli/config/astro-cli.json
+++ b/sql-cli/config/astro-cli.json
@@ -4,6 +4,10 @@
       "preRelease": false,
       "sqlCliVersion": ">0.2.2"
     },
+    "1.10.0": {
+      "preRelease": false,
+      "sqlCliVersion": ">0.2.2"
+    },
     "1.8.0": {
       "preRelease": false,
       "sqlCliVersion": ">0.2, <0.3"
--- a/sql-cli/config/astro-cli.json
+++ b/sql-cli/config/astro-cli.json
@@ -1,6 +1,6 @@
 {
   "astroCliCompatibility": {
-    "1.10.0": {
+    "1.10": {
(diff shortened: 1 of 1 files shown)


pankajkoti:
We're keeping the `1.10.0` key in the compatibility matrix so that we do not break the Astro CLI integration in the release `1.10.0`

sql-cli/config/astro-cli.json
```diff
@@ -1,5 +1,9 @@
 {
   "astroCliCompatibility": {
+    "1.10": {
+      "preRelease": false,
+      "sqlCliVersion": ">0.2.2"
+    },
     "1.10.0": {
```
tatiana, line 7: :+1: it makes sense to keep this for the current Astro CLI 1.10.0 users since it does not do the full matching. I'm hopeful that from 1.10.1 onwards we'll be able to avoid declaring the versions using the three digits.

tatiana (approved):

sql-cli/config/astro-cli.json
pankajkoti: yes, right! :)

PR #1716, opened 2023-02-06
Fix Python SDK README quickstart

tatiana:
Previously, if someone tried to run the quickstart example described in the Python SDK README, it would not work out of the box, due to two problems:
1. The curl command was referencing a file that did not exist
2. The environment variables we were setting were not enough

This PR addresses these issues, also explaining the limitations of the proposed environment variable setup and pointing to how to setup in a non-local environment.


sunank200 (approved):

utkarsharma2:
I think it would be better if we include this dags in the example dag and refer here using permalinks

PR #1717, opened 2023-02-06
Generate Python3.10 constraints

pankajastro:
since we run some basic tests on Python3.10 so lets generate/release constraints file too

### DIFF
--- a/python-sdk/noxfile.py
+++ b/python-sdk/noxfile.py
@@ -150,7 +150,7 @@ def build_docs(session: nox.Session) -> None:
     session.run("make", "html")
 
 
-@nox.session(python=["3.7", "3.8", "3.9"])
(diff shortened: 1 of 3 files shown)


python-sdk/noxfile.py
```diff
@@ -150,7 +150,7 @@ def build_docs(session: nox.Session) -> None:
     session.run("make", "html")
 
 
-@nox.session(python=["3.7", "3.8", "3.9"])
+@nox.session(python=["3.7", "3.8", "3.9", "3.10"])
```
kaxil, line 153: We should update the following line too:

 https://github.com/astronomer/astro-sdk/blob/9a18c3392f90538c262b909859291071c5dad0cf/.github/workflows/ci-python-sdk.yaml#L621

pankajastro: updated

Issue #1719, opened 2023-02-07
load_options exposed to user in File - location/type 


sunank200:
**Describe the bug**
`load_options` exposed to the user in File location/type added in the following PR - https://github.com/astronomer/astro-sdk/pull/1540. Since it's a derived property and not something the user should set.

Concerns
load_options shouldn't be exposed to the end users in the File class.
load_options  is specific to load_file() and should not be part of File/Databases.

ref - https://astronomer.slack.com/archives/C03868KGF2Q/p1675690013077239

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1721, opened 2023-02-07
Move load_options out from the File object

utkarsharma2:
# Description
load_options exposed to the user in File location/type added in the following PR - https://github.com/astronomer/astro-sdk/pull/1540. Since it's a derived property and not something the user should set.

Concerns
load_options shouldn't be exposed to the end users in the File class.
load_options is specific to load_file() and should not be part of File/Databases.

closes: https://github.com/astronomer/astro-sdk/issues/1719

## What is the new behavior?
We added a getter/setter property on the File class such that the load_option interface is not exposed to the end user.

Yes

### Checklist

### DIFF
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -45,7 +45,7 @@ class File(LoggingMixin, Dataset):
 
     @property
     def load_options(self):
-        return self._load_options
+        return getattr(self, "_load_options", [])
 
     @load_options.setter
     def load_options(self, value):
--- a/python-sdk/src/astro/files/base.py
+++ b/python-sdk/src/astro/files/base.py
@@ -267,16 +267,15 @@ def resolve_file_path_pattern(
     location = create_file_location(path_pattern, conn_id)
     files = []
     for path in location.paths:
-        if path.endswith("/"):
-            continue
-        file = File(
-            path=path,
-            conn_id=conn_id,
-            filetype=filetype,
-            normalize_config=normalize_config,
-        )
-        file.load_options = load_options
-        files.append(file)
+        if not path.endswith("/"):
+            file = File(
+                path=path,
+                conn_id=conn_id,
+                filetype=filetype,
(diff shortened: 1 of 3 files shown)


python-sdk/src/astro/files/base.py
```diff
@@ -44,6 +43,14 @@ class File(LoggingMixin, Dataset):
         "conn_id",
     )
 
+    @property
+    def load_options(self):
+        return getattr(self, "_load_options", [])
+
+    @load_options.setter
+    def load_options(self, value):
+        self._load_options = value
```
kaxil, line 52: docstrings would be good

kaxil (approved):

python-sdk/src/astro/files/base.py
```diff
@@ -44,6 +43,14 @@ class File(LoggingMixin, Dataset):
         "conn_id",
     )
 
+    @property
+    def load_options(self):
+        return getattr(self, "_load_options", [])
+
+    @load_options.setter
+    def load_options(self, value):
```
tatiana, line 51: type hint?

python-sdk/src/astro/files/base.py
```diff
@@ -250,18 +257,18 @@ def resolve_file_path_pattern(
     :param normalize_config: parameters in dict format of pandas json_normalize() function
     """
     location = create_file_location(path_pattern, conn_id)
-
-    files = [
-        File(
+    files = []
+    for path in location.paths:
+        if path.endswith("/"):
+            continue
```
tatiana, line 263: Should we add a unit test to cover this?

python-sdk/src/astro/files/base.py
utkarsharma2: Added

python-sdk/src/astro/files/base.py
utkarsharma2: Changed condition to get better coverage.

Issue #1681, opened 2023-01-30
Standardize the pandas load options


utkarsharma2:
**Please describe the feature you'd like to see**
We currently have PandasCsvLoadOptions, PandasJsonLoadOptions, PandasNdjsonLoadOptions, and PandasParquetLoadOptions. Which can be confusing to users. 

**Describe the solution you'd like**
Instead of multiple classes, we can probably merge this in one single class PandasLoadOptions that can be superset for all the PandasLoadOptions.

**Are there any alternatives to this feature?**
Is there another way we could solve this problem or enable this use case?
Open to suggestions.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of the methods, classes, functions, and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

PR #1722, opened 2023-02-07
Combine PandasLoadOptions into single class

utkarsharma2:
# Description
We currently have PandasCsvLoadOptions, PandasJsonLoadOptions, PandasNdjsonLoadOptions, and PandasParquetLoadOptions. Which can be confusing to users.

closes: #1681 

## What is the new behavior?
Instead of multiple classes, we can probably merge this in one single class PandasLoadOptions that can be superset for all the PandasLoadOptions.

Yes

### Checklist


python-sdk/src/astro/dataframes/load_options.py
```diff
@@ -8,56 +8,30 @@
...
-
-
-@define
-class PandasCsvLoadOptions(PandasLoadOptions):
     """
     Pandas load options while reading and loading csv file.
 
-    :param delimiter: Delimiter to use. Defaults to None
-    :param dtype: Data type for data or columns. E.g. ``{"a": np.float64, "b": np.int32, "c": "Int64"}`` Use str or
+    :param delimiter: valid file type[CSV] - Delimiter to use. Defaults to None
```
sunank200, line 14: I think we need to have a way to pass all other supported parameters for Pandas as `kwargs` here. We should add this change as part of this PR.

sunank200 (changes requested):
I think the following changes would be required:
1. I think we need to have a way to pass all other supported parameters for Pandas as `kwargs` here. We should add this change as part of this PR. More at: https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html, https://pandas.pydata.org/docs/reference/api/pandas.read_json.html
2. Documentation has to be changed accordingly in `load_file.rst` for this.

pankajastro (commented):
In the earlier implementation, I thought a class per pandas API (`PandasJsonLoadOptions`  
 [read_json](https://pandas.pydata.org/docs/reference/api/pandas.read_json.html)) will be cleaner.
currently, you have put a comment like json/csv etc this comment does not look correct to me since `dtype` is a valid param for even read_json method and in future, we may like to add this param in `PandasJsonLoadOptions` class.
I feel we have different classes then the code editor also helps users to understand what parameters they can use, which would be confusing if we have a single class. 
So I would like to know the other thought on this PR

python-sdk/src/astro/dataframes/load_options.py
utkarsharma2: resolved.

utkarsharma2:
@pankajastro Thanks for pointing this out, I have updated the docstring. Now we are referring to the pandas doc for a superset of all the valid options.

Also, I think if we are anyway exposing to end users that we are using pandas in the backend to do stuff, I think we can also assume that they will be aware of which options need to be passed for a particular type of file. We have also included the pandas link to aid for this.

tatiana (approved):

sunank200 (approved):

Issue #1755, opened 2023-02-10
use_native_support param is not used while creating table


utkarsharma2:
**Describe the bug**
When creating a table, we are not using `use_native_support`; because of this, the flow is using a native approach and failing for `minio` conn.

ref: https://astronomer.slack.com/archives/C02B8SPT93K/p1675792250967869

**Version**
* Astro: 1.5.0

**To Reproduce**
Try running:
```
load_file(
   input_file=File("S3://astro-sdk/twitter_comments.parquet", conn_id="minio_conn"),
   output_table=test_table,
   use_native_support=False,
)
```

**Expected behavior**
Should not error out if the native approach is not available.

PR #1756, opened 2023-02-10
Fix issue - use_native_support should be used for table creation

utkarsharma2:
# Description
When creating a table, we are not using use_native_support; because of this, the flow is using a native approach and failing for minio conn.

closes: https://github.com/astronomer/astro-sdk/issues/1755

## What is the new behavior?
Used `use_native_support` for table creation too.

Nope

### Checklist


python-sdk/src/astro/databases/base.py
```diff
@@ -314,7 +315,7 @@ def create_table(
         """
         if table.columns:
             self.create_table_using_columns(table)
-        elif file and self.is_native_autodetect_schema_available(file):
+        elif use_native_support and file and self.is_native_autodetect_schema_available(file):
             self.create_table_using_native_schema_autodetection(table, file)
         else:
```
tatiana, line 320: What happens in the following case:

the user defines:
* `table.columns`
* `use_native_support`

And `is_native_autodetect_schema_available` is `True`.

Would we both call:
* `self.create_table_using_columns(table)`
* `self.create_table_using_native_schema_autodetection(table, file)`

dimberman: Yeah that makes sense. If a user wants to explicitly state the columns we shouldn't need to look anything up. If they are incorrect in their definition then it's on them to fix it.

dimberman (approved):
@utkarsharma2 this LGTM. Are there any edge cases we need to test before merging?

python-sdk/tests/sql/operators/test_load_file.py
```diff
@@ -204,3 +204,46 @@ def test_load_file_native_support_kwargs_warnings_message(database_table_fixture
...
+    indirect=True,
+    ids=["snowflake"],
+)
+@mock.patch("astro.databases.base.BaseDatabase.load_file_to_table_using_pandas")
+@mock.patch("astro.databases.base.resolve_file_path_pattern")
+@mock.patch("astro.databases.base.BaseDatabase.create_schema_if_needed")
+@mock.patch("astro.databases.base.BaseDatabase.drop_table")
+@mock.patch("astro.databases.snowflake.SnowflakeDatabase.create_table_using_schema_autodetection")
+def test_table_creation_and_population_done_via_pandas_path(
+    load_file_to_table_using_pandas,
```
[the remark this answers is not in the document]
utkarsharma2: Added an assert statement to verify inverse of this.

PR #1757, opened 2023-02-13
Add temporary dependencies

utkarsharma2:
# Description
**Added temporary dependencies**

**1. pandas-gbq**

It is a dependency of the [apache-airflow-providers-google](https://pypi.org/project/apache-airflow-providers-google/8.9.0/) and it does set an upper limit on it in version [apache-airflow-providers-google - 6.4.0](https://pypi.org/project/apache-airflow-providers-google/6.4.0/) of < 0.15.0, but [pandas - 1.5.3](https://pypi.org/project/pandas/1.5.3/) require a greater version of `pandas-gbq > 0.15.0` 
ref : https://github.com/pandas-dev/pandas/blame/0e12bfcd4b4d4e6f5ebe6cc6f99c0a5836df4478/pandas/compat/_optional.py#L31

So, we installed a greater version of apache-airflow-providers-google, but since [apache-airflow-providers-google](https://pypi.org/project/apache-airflow-providers-google/8.9.0/) don't have any version requirement of `pandas-gbq` we have to explicitly pass `pandas-gbq>0.15.0`

**2. protobuf**
While trying to install `protobuf` pip was trying to install version [2.5.0](https://pypi.org/project/protobuf/2.5.0/) which was released in 2013 and not compatible with the python version >= 3.0. 

ref: https://github.com/astronomer/astro-sdk/actions/runs/4173078408/jobs/7224985886

To prevent pip from trying that option we need to add a min dep `protobuf>2.5.0`

## What is the new behavior?
Added `protobuf>2.5.0` and `pandas-gbq > 0.15.0` dependencies but they are not our direct dependencies and should be moved to the provider.

Nope

### Checklist

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -56,9 +56,11 @@ tests = [
 ]
 google = [
     "protobuf<=3.20", # Google bigquery client require protobuf <= 3.20.0. We can remove the limitation when this limitation is removed
-    "apache-airflow-providers-google>=6.4.0",
+    "apache-airflow-providers-google>6.4.0", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
     "sqlalchemy-bigquery>=1.3.0",
-    "smart-open[gcs]>=5.2.1"
+    "smart-open[gcs]>=5.2.1",
+    "pandas-gbq>=0.15.0", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
+    "protobuf>2.5.0" # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
 ]
 snowflake = [
     "apache-airflow-providers-snowflake",
@@ -101,7 +103,7 @@ duckdb = [
 
 all = [
     "apache-airflow-providers-amazon",
-    "apache-airflow-providers-google>=6.4.0",
+    "apache-airflow-providers-google>6.4.0", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
     "apache-airflow-providers-ftp",
     "apache-airflow-providers-postgres",
     "apache-airflow-providers-snowflake",
@@ -119,6 +121,8 @@ all = [
     "azure-storage-blob",
     "apache-airflow-providers-microsoft-mssql>=3.2",
     "airflow-provider-duckdb>=0.0.2",
+    "pandas-gbq>=0.15.0", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
+    "protobuf>2.5.0" # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
 ]
 doc = [
     "myst-parser>=0.17",
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -59,7 +59,7 @@ google = [
     "apache-airflow-providers-google>6.4.0", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
     "sqlalchemy-bigquery>=1.3.0",
     "smart-open[gcs]>=5.2.1",
-    "pandas-gbq>=0.15.0", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
+    "pandas-gbq", # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
     "protobuf>2.5.0" # look at PR description for more information - https://github.com/astronomer/astro-sdk/pull/1757
 ]
 snowflake = [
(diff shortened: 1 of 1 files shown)


sunank200 (commented):
Tests seems to be failing:
```
FAILED tests_integration/sql/operators/test_dataframe.py::test_dataframe_replace_table_if_exist[bigquery] - ImportError: Pandas requires version '0.15.0' or newer of 'pandas_gbq' (version '0.13.2' currently installed).
FAILED tests_integration/sql/operators/test_dataframe.py::test_dataframe_append_table_if_exist[bigquery] - ImportError: Pandas requires version '0.15.0' or newer of 'pandas_gbq' (version '0.13.2' currently installed).
FAILED tests_integration/sql/operators/test_drop_table.py::test_drop_table_with_table_metadata[bigquery] - ImportError: Pandas requires version '0.15.0' or newer of 'pandas_gbq' (version '0.13.2' currently installed).
```

Lee-W:
I've tried a few different ways today, but not yet found a solution other than adding the constraint `protobuf>2.5.0`.  According to the [Dependency Resolution](https://pip.pypa.io/en/stable/topics/dependency-resolution/#possible-ways-to-reduce-backtracking) section in the pip document, this seems to be a reasonable solution.

I also tried to use [pipdeptree](https://pypi.org/project/pipdeptree/) to see whether `protobuf` is only used in `apache-airflow-providers-google` which turn out to be the case and make adding the constraint more reasonable.

## What I've tried
The following is what I tried. On the Github side, I forked the project and ran github action on my repo. On local machine, I tried to install them through different kinds of tool and see how it works

1. Run GitHub action job `Run-Unit-tests-Airflow-2-5` without `protobuf>2.5.0`
    * purpose: to reproduce the error
     * result: failed as expected ❌
2. Run GitHub action job `Run-Unit-tests-Airflow-2-5` without `protobuf>2.5.0`, but install `pip == 22.3.1`
    * purpose: to check whether pip version affects the resolution
    * result: failed ❌
3. Run `make local target=setup` locally
    * purpose: to reproduce the error locally
    * result: still installing 
4. Run `pip install apache-airflow-providers-google>6.4.0` with python 3.9, `pip == 23.0`
    * purpose: to reproduce the error locally with pure pip
    * result: succeeded ✅  (protobuf == 3.20.0 installed)
5. Run `pipenv install apache-airflow-providers-google>6.4.0` with python 3.9, `pipenv==2022.9.8`
    * purpose: to reproduce the error locally with pipenv
    * result: pipenv stuck
6. Run `poetry add apache-airflow-providers-google>6.4.0` with python 3.9, `poetry==1.3.2`
    * purpose: to reproduce the error locally with poetry
    * result: succeeded ✅  (protobuf == 3.20.0 installed)

## Other findings
I digged into the dependencies of `apache-airflow-providers-google` in method 4 and find out only the following 4 dependencies don't have a minimum contraint

```
  - google-cloud-dlp [required: >=0.11.0,<2.0.0, installed: 1.0.2]
    - protobuf [required: <4.0.0dev, installed: 3.20.0]
    ...

  - google-cloud-language [required: >=1.1.1,<2.0.0, installed: 1.3.2]
    - protobuf [required: <4.0.0dev, installed: 3.20.0]
    ...

 - google-cloud-texttospeech [required: >=0.4.0,<2.0.0, installed: 1.0.3]
    - protobuf [required: <4.0.0dev, installed: 3.20.0]
    ...

  - google-cloud-videointelligence [required: >=1.7.0,<2.0.0, installed: 1.16.3]
    - protobuf [required: <4.0.0dev, installed: 3.20.0]
    ...

  - google-cloud-vision [required: >=0.35.2,<2.0.0, installed: 1.0.2]
    - protobuf [required: <4.0.0dev, installed: 3.20.0]
   ...
```

python-sdk/pyproject.toml
```diff
@@ -56,9 +56,11 @@ tests = [
 ]
 google = [
     "protobuf<=3.20", # Google bigquery client require protobuf <= 3.20.0. We can remove the limitation when this limitation is removed
-    "apache-airflow-providers-google>=6.4.0",
+    "apache-airflow-providers-google>6.4.0",
```
utkarsharma2, line 59:
```suggestion
    "apache-airflow-providers-google==8.8.0",
```

tatiana: @utkarsharma2 as we discussed, we should try to have 8.8.x as an upper threshold - and remove the other pins. If this works, I suggest we report the issue and suggest to the Google provider team to pin the other dependencies...!

It does feel that the issue is within the Google provider's dependencies - and it would be better if we addressed the versions of 
```
    "pandas-gbq>=0.15.0",
    "protobuf>2.5.0"
```

tatiana (commented):
@utkarsharma2 it would be great if we could confirm if this is compatible with runtime as well

python-sdk/pyproject.toml
```diff
@@ -56,9 +56,11 @@ tests = [
...
     "sqlalchemy-bigquery>=1.3.0",
-    "smart-open[gcs]>=5.2.1"
+    "smart-open[gcs]>=5.2.1",
```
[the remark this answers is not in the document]
utkarsharma2: Nope.

python-sdk/pyproject.toml
utkarsharma2: Tried this, didn't work, still installing pandas-gbq==0.13.2 as there is no min dependency on the provider https://pypi.org/project/apache-airflow-providers-google/8.8.0/

python-sdk/pyproject.toml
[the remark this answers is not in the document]
utkarsharma2: I think we need to add dependencies for google and all section, right?

utkarsharma2:
Closing this in favor of - https://github.com/astronomer/astro-sdk/pull/1768

Issue #1302, opened 2022-11-23
Allow SQL CLI to run single task


tatiana:
**Please describe the feature you'd like to see**
At the moment `flow run` allows us to run full SQL workflows, but not individual tasks.

**Describe the solution you'd like**
`flow run <workflow> --task <sql/yml>`

**Additional context**
This feature has been requested by a few Astronomer customers (including Wall).

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

tatiana:
cc @mikeshwe please, feel free to update the ticket description :)

Issue #1764, opened 2023-02-15
Upgrade Python SDK version in the SQL CLI


tatiana:
Since Python SDK has started supporting DuckDB and MSSQL, the Astro CLI installation of the SQL CLI failed.

We will probably need to update the Astro Runtime version or add system dependencies to solve this issue and upgrade the SQL CLI to a newer Python SDK version.

PR #1766, opened 2023-02-15
Allow SQL CLI to run a single task for local run

pankajkoti:
# Description

Currently, we do not support running a single task of a SQL CLI workflow

related: #1302 

## What is the new behavior?

We now allow users to run a single task of a workflow by providing task_id.
Additionally, we also are exposing a boolen includer_upstream using which
the user could also trigger runs for upstream tasks of the task_id 

No

### Checklist


python-sdk/pyproject.toml
```diff
@@ -91,9 +91,9 @@ openlineage = ["openlineage-airflow>=0.17.0"]
 databricks = ["databricks-cli",
     "apache-airflow-providers-databricks"]
 
-mssql = [
-    "apache-airflow-providers-microsoft-mssql>=3.2",
-]
+#mssql = [
+#    "apache-airflow-providers-microsoft-mssql>=3.2",
+#]
```
dimberman, line 96: Why are we commenting this out? Are we no longer supporting mssql?

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -34,6 +34,10 @@ class LoadFileOperator(AstroSQLBaseOperator):
     :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
             in the resulting dataframe
     :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
+    :param skip_core_execution: Skips the core execution of the operator but run all ancillary operations. This is
+        useful for local run of tasks where subsequent tasks in the DAG might need the output of this operator like
+        XCOM results, but they do not want to actually run the core logic and hence not cause effects outside the
+        system.
```
dimberman, line 40: I'm not sure we should actually openly document this feature. This is not something I would want people using the python API to know about. A better option might be to keep it as a "hidden" feature and only run it if the argument happens to be in the kwargs.

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -330,6 +339,10 @@ def load_file(
     :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
         in the resulting dataframe
     :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
+    :param skip_core_execution: Skips the core execution of the operator but run all ancillary operations. This is
+        useful for local run of tasks where subsequent tasks in the DAG might need the output of this operator like
+        XCOM results, but they do not want to actually run the core logic and hence not cause effects outside the
+        system.
```
dimberman, line 345: Same as above

sql-cli/sql_cli/dag_runner.py
```diff
@@ -48,6 +48,39 @@ def _local_connections(self) -> dict[str, Connection]:
...
+
+
+def _get_tailored_dag(dag: DAG, task_id: str | None, include_upstream: bool) -> DAG:
+    """Gets subset of the dag based on whether a task ID is given and asked to include its upstream tasks in the dag."""
+    if task_id and include_upstream:
+        dag = dag.partial_subset(
+            task_ids_or_regex=task_id,
+            include_upstream=True,
+        )
+    return dag
```
dimberman, line 81: Can you explain what new functionality is happening here? I'd like to know why you broke these parts off.

dimberman (changes requested):
@pankajkoti this is a great start, let's chat a bit about it and get this in ASAP :)

python-sdk/pyproject.toml
pankajkoti: We're supporting, this was commented to test for local integration of Python SDK into SQL CLI. Mssql needs a few additional system dependencies to be installed which are not available yet in the runtime image. We have a ticket to address this. And I will make sure we uncomment this before merging the PR.

sql-cli/sql_cli/dag_runner.py
pankajkoti: We broke these into multiple methods with the addition of a few new conditionals like https://github.com/astronomer/astro-sdk/pull/1766/files#diff-468cfaa4c543746918f04d1238ff10da015a5ecdae4b58178f66150b5b6c65cdR148  because ruff was complaining that the earlier function was too complex.

python-sdk/src/astro/sql/operators/load_file.py
pankajkoti: Yes. Removed it as a docstring param and accepting it as a kwarg. But popping it from kwargs and assigning it as an instance variable as the base class BaseOperator complains that invalid argument received upon initialtisation
```
  File "/Users/pankajkoti/Library/Caches/pypoetry/virtualenvs/astro-sql-cli-r5JKmrdb-py3.8/lib/python3.8/site-packages/airflow/models/baseoperator.py", line 411, in apply_defaults
    result = func(self, **kwargs, default_args=default_args)
  File "/Users/pankajkoti/Library/Caches/pypoetry/virtualenvs/astro-sql-cli-r5JKmrdb-py3.8/lib/python3.8/site-packages/airflow/models/baseoperator.py", line 759, in __init__
    raise AirflowException(
airflow.exceptions.AirflowException: Invalid arguments were passed to LoadFileOperator (task_id: load_imdb_movies). Invalid arguments were:
**kwargs: {'skip_core_execution': True}
```

python-sdk/pyproject.toml
tatiana: https://github.com/astronomer/astro-sdk/issues/1764

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -56,6 +56,7 @@ def __init__(
         **kwargs,
     ) -> None:
         kwargs.setdefault("task_id", get_unique_task_id("load_file"))
+        self.skip_core_execution = kwargs.pop("skip_core_execution", None)
```
tatiana, line 59: We'll also need to have this logic at least in `transform`/`transform_file` if we are to focus on the operators we currently support in the SQL CLI. We could probably also benefit by having it in `export_file`.

sql-cli/sql_cli/__main__.py
```diff
@@ -218,28 +258,20 @@ def run(
     # TODO: Validate all DAG connections before running
 
     # Generate DAG before running
-    dag_file = _generate_dag(project=project, workflow_name=workflow_name, generate_tasks=generate_tasks)
+    tasks_to_execute = None
+    if task_id and not include_upstream:
+        tasks_to_execute = [task_id]
+    dag_file = _generate_dag(
```
tatiana, line 264: @pankajkoti, shouldn't we also pass `include_upstream` here to differentiate between running the single task - or also running its upstream dependencies?

sql-cli/sql_cli/dag_runner.py
```diff
@@ -80,6 +119,9 @@ def run_dag(
         session=session,
     )
     log.debug("Getting dagrun for dag %s", dag.dag_id)
+
+    dag = _get_tailored_dag(dag, task_id, include_upstream)
```
tatiana, line 123: I think the code would be easier to read if we had the following conditional here:
```
if task_id and include_upstream:
```
and only invoked the function to change the DAG if necessary.

sql-cli/sql_cli/__main__.py
```diff
@@ -203,11 +235,19 @@ def run(
         show_default=True,
     ),
     verbose: bool = typer.Option(False, help="Whether to show verbose output", show_default=True),
+    task_id: str = typer.Option(
+        None,
+        help="Id of the single task to run. If given, only this task of the DAG will run. Otherwise, all tasks of the "
+        "DAG will be run.",
+    ),
+    include_upstream: bool = typer.Option(
```
tatiana, line 243: if the user passes `--include_upstream`, we are potentially not showing all the tasks we are skipping. Perhaps we should document this.

sql-cli/sql_cli/dag_runner.py
```diff
@@ -135,7 +170,14 @@ def add_loghandler(ti: TaskInstance, verbose: bool) -> None:
     ti.log.addHandler(RichHandler(markup=True))
 
 
-def _run_task(ti: TaskInstance, session: Session) -> None:
+def _display_run_task_output(skip_task_execution: bool) -> None:
```
tatiana, line 173: Docstring..?

tatiana (commented):
@pankajkoti Let's see how others feel about us changing the Python SDK operators interface (by introducing `skip_core_execution` to `load_file` and `transform_file`, which probably does not make sense in other contexts than the SQL CLI).

The problem we were trying to solve was to populate `XComArgs` without necessarily running the previous tasks - it would be great if there were "cleaner" solutions to accomplish this

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -95,6 +96,11 @@ def execute(self, context: Context) -> BaseTable | File:  # skipcq: PYL-W0613
         if self.output_table:
             context["ti"].xcom_push(key="output_table_conn_id", value=str(self.output_table.conn_id))
             context["ti"].xcom_push(key="output_table_name", value=str(self.output_table.name))
+        if self.skip_core_execution:
```
utkarsharma2, line 99: @pankajkoti @tatiana I think we should not have this in the operators, there should be a better way of retrying a task.

utkarsharma2: Also, this change will be required for all operators, I presume.

tatiana: @utkarsharma2 this was not introduced for retries..! Its goal is to allow us to execute a task by only populating the XComArg (e.g. it does not load, it just populate the output of execute with the output table, so we can resolve references in other tasks).

tatiana: This change would probably not make sense in the `run_raw_sql` - and we wanted to collect feedback on the overall approach before changing any other operators :)

feluelle: I am wondering if we are creating a workaround, instead of reading the information (xcom) from the db. I think the problem we have in our local runner is that we are not storing and instead actively deleting the previous dag runs in `_get_or_create_dagrun`. If we would keep them, we would be able to retrieve the previous xcom without re-running it (I think). @dimberman would that be an option? I think that would be a much cleaner solution if we can handle potential collisions which you mentioned in the function's docstring.

feluelle: We currently cannot access the previous xcoms, because the deletion of dagruns causes a _cascade delete_, deleting xcoms created by the task instances of the dagrun.

utkarsharma2: @tatiana I see, assuming we have task `a -> b -> c` and now we want to rerun task `c`. We want to skip the `a` and `b` and just run `c` this is the use-case right?

tatiana: Exactly, @utkarsharma2 !

@feluelle I think it is not necessarily a workaround, I think there are a couple ways of addressing the problem - but let's try to see some pros and cons of each approach.

**1. Table name approach**

It doesn't rely on metadata from previous runs, but on the dynamic generation of table names.

Pros:
* Works even if the user didn't use the SQL CLI to populate the tables the task relies on
* Works across different users computers
* Could work even if the topology of the DAG changed

Cons:
* We had to change Python SDK operators -  this would not work for non-Python SDK operators
* This does not work for temporary tables

**2. Previous DAG Run approach**

Pros:
* We don't have to change the Python SDK operators - so it would be agnostic to the operator we use
* Works with temporary tables

Cons:
* Doesn't work if the table was populated outside of the SQL CLI
* Doesn't work cross-computers
* Would likely fail if the topology of the DAG changes

Would love to hear more pros and cons of each approach!

utkarsharma2: One Con of the `Table name approach` is that we may be able to handle this for SDK operators this way(because we control code), but it won't work for Airflow's operator if we decide to use them later on in CLI.

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -84,7 +84,7 @@ def __init__(
         self.enable_native_fallback = enable_native_fallback
         self.load_options_list = LoadOptionsList(load_options)
 
-    def execute(self, context: Context) -> BaseTable | File:  # skipcq: PYL-W0613
+    def execute(self, context: Context) -> BaseTable | File | None:  # skipcq: PYL-W0613
```
pankajkoti, line 87:
```suggestion
    def execute(self, context: Context) -> BaseTable | File:  # skipcq: PYL-W0613
```

python-sdk/src/astro/sql/operators/load_file.py
```diff
@@ -95,6 +95,7 @@ def execute(self, context: Context) -> BaseTable | File:  # skipcq: PYL-W0613
         if self.output_table:
             context["ti"].xcom_push(key="output_table_conn_id", value=str(self.output_table.conn_id))
             context["ti"].xcom_push(key="output_table_name", value=str(self.output_table.name))
+
```
pankajkoti, line 98:
```suggestion
```

sql-cli/sql_cli/dag_runner.py
pankajkoti: Unfortunately, adding more conditionals is increasing the code complexity; hence, I have moved that conditional to another method.

python-sdk/src/astro/sql/operators/load_file.py
pankajkoti: As suggested by @tatiana and @feluelle, we're now overriding the Python SDK operators in SQL CLI. So, there is no change in Python SDK operators now.

sql-cli/pyproject.toml
```diff
@@ -37,8 +37,10 @@ black = { version = "*", allow-prereleases = true }
 python-frontmatter = "^1.0.0"
 PyYAML = ">=5.4.1"
 python-dotenv = "^0.21.0"
+boto3 = ">=1.21.29, <=1.26.72"
+botocore = "==1.29.72"
 # Set the minimum version to v1.3.1 to include a fix for https://github.com/astronomer/astro-sdk/issues/1351
-astro-sdk-python = {extras = ["all"], version = ">=1.3.1,<1.5"}
+astro-sdk-python = {path = "../python-sdk", extras = ["all"]}
```
pankajkoti, line 43:
```suggestion
astro-sdk-python = {extras = ["all"], version = ">=1.3.1,<1.5"}
```

sql-cli/pyproject.toml
```diff
@@ -37,8 +37,10 @@ black = { version = "*", allow-prereleases = true }
 python-frontmatter = "^1.0.0"
 PyYAML = ">=5.4.1"
 python-dotenv = "^0.21.0"
+boto3 = ">=1.21.29, <=1.26.72"
+botocore = "==1.29.72"
```
pankajkoti, line 41:
```suggestion
```

tatiana (approved):
Great work, @pankajkoti !

I logged a follow up ticket, so we can review the approach - as we agreed with @feluelle and @utkarsharma2 last Friday:
https://github.com/astronomer/astro-sdk/issues/1782

Issue #1746, opened 2023-02-09
Minio support with Snowflake not working


utkarsharma2:
**Describe the bug**
Slack thread - https://astronomer.slack.com/archives/C02B8SPT93K/p1675792250967869

utkarsharma2:
We need to treat the connection for `minio` differently from the one for S3 and add a new file location for `minio`. So that we can avoid going into the native path for S3 and databases.

sunank200:
@sunank200 to sync up with Michael Gregory and understand the issue better.

sunank200:
@sunank200 to sync with @utkarsharma2 for the approach and discussion on slack thread.

PR #1767, opened 2023-02-15
Check if connection is minio then set use_native_support=False

sunank200:
# Description

When the user passes the s3 url but has minio as the setup, for minio to any database transfers by default it uses the native path and fails as there is no native path. There is no direct way to check if the connection is of type minio. Introducing a parameter `minio` in extras as part of connections would help to recognize the connection type

closes: #1746

## What is the new behavior?

- Check if the AWS connection has `minio` in `extra` then set use_native_support=False
- Example minio conn:
```
  - conn_id: minio_conn
    conn_type: aws
    description: null
    extra:
      aws_access_key_id: "dummy"
      aws_secret_access_key: "dummy"
      endpoint_url: "http://127.0.0.1:9000"
      minio: True
```

No

### Checklist

### DIFF
--- a/.github/ci-test-connections.yaml
+++ b/.github/ci-test-connections.yaml
@@ -139,4 +139,3 @@ connections:
       aws_access_key_id: "ROOTNAME"
       aws_secret_access_key: "CHANGEME123"
       endpoint_url: "http://127.0.0.1:9000"
-      minio: True
--- a/.github/ci-test-connections.yaml
+++ b/.github/ci-test-connections.yaml
@@ -139,3 +139,4 @@ connections:
       aws_access_key_id: "ROOTNAME"
       aws_secret_access_key: "CHANGEME123"
       endpoint_url: "http://127.0.0.1:9000"
+      minio: True
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -462,6 +462,8 @@ class BaseDatabase(ABC):
         :param enable_native_fallback: Use enable_native_fallback=True to fall back to default transfer
         """
         normalize_config = normalize_config or {}
+        if self.check_for_minio_connection(input_file=input_file):
+            use_native_support = False
 
         self.create_schema_and_table_if_needed(
             file=input_file,
@@ -471,8 +473,6 @@ class BaseDatabase(ABC):
             normalize_config=normalize_config,
             use_native_support=use_native_support,
         )
-        if self.check_for_minio_connection(input_file=input_file):
-            use_native_support = False
 
         if use_native_support and self.is_native_load_file_available(
             source_file=input_file, target_table=output_table
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -14,8 +14,7 @@ from sqlalchemy import column, insert, select
 if TYPE_CHECKING:  # pragma: no cover
     from sqlalchemy.engine.cursor import CursorResult
 
-from airflow.exceptions import AirflowNotFoundException
-
+import requests
 from sqlalchemy.sql import ClauseElement
 from sqlalchemy.sql.elements import ColumnClause
 from sqlalchemy.sql.schema import Table as SqlaTable
@@ -427,15 +426,16 @@ class BaseDatabase(ABC):
 
     @staticmethod
     def check_for_minio_connection(input_file: File) -> bool:
-        """Returns True if minio is passed in extras in connections."""
+        """Automatically check if the connection is minio or S3"""
+        is_minio = False
         if input_file.location.location_type == FileLocation.S3:
+            conn = input_file.location.hook.get_connection(input_file.conn_id)
             try:
-                conn = input_file.location.hook.get_connection(input_file.conn_id)
-            except AirflowNotFoundException as exe:
-                logging.warning(exe)
-                return False
-            return bool(conn.extra_dejson.get("minio"))
-        return False
+                endpoint_url = conn.extra_dejson["endpoint_url"]
+                is_minio = requests.get(f"{endpoint_url}/minio/health/live").ok
+            except (requests.exceptions.RequestException, KeyError):
+                pass
+        return is_minio
 
     def load_file_to_table(
         self,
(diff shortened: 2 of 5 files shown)


dimberman:
@sunank200 can you please offer more context in the PR template as to what this PR is doing and why we are doing it? Maybe a bit on how it solves this bug?

sunank200:
> @sunank200 can you please offer more context in the PR template as to what this PR is doing and why we are doing it? Maybe a bit on how it solves this bug?

sure @dimberman This PR is still in draft but i will update the description

.github/ci-test-connections.yaml
```diff
@@ -139,3 +139,4 @@ connections:
       aws_access_key_id: "ROOTNAME"
       aws_secret_access_key: "CHANGEME123"
       endpoint_url: "http://127.0.0.1:9000"
+      minio: True
```
tatiana, line 142: @sunank200 is this a convention we are creating or is this an standard in Airflow? It may be worth to document it, if the first

python-sdk/src/astro/databases/base.py
```diff
@@ -422,6 +423,18 @@ def fetch_all_rows(self, table: BaseTable, row_limit: int = -1) -> list:
...
+    @staticmethod
+    def check_for_minio_connection(input_file: File) -> bool:
+        """Returns True if minio is passed in extras in connections."""
+        if input_file.location.location_type == FileLocation.S3:
+            try:
+                conn = input_file.location.hook.get_connection(input_file.conn_id)
+            except AirflowNotFoundException as exe:
+                logging.warning(exe)
+                return False
+            return bool(conn.extra_dejson.get("minio"))
```
tatiana, line 435: What is the advantage of this approach as opposed to identifying automagically if the object store is minio, for instance, using something like:
```
requests.get(https://minio.example.net:9000/minio/health/live)
```
https://min.io/docs/minio/linux/operations/monitoring/healthcheck-probe.html

sunank200: @tatiana I am okay with doing this change as well but I feel that this is an external URL which can be changed by minio. And there can be the scenarios where health check URL can be publicly pointing to a different port (or using reverse proxy depending on client setup).

With the current approach, the user only would need to add `minio=True` as part of the connection.

.github/ci-test-connections.yaml
sunank200: Yes @tatiana we are introducing a parameter `minio=True/False` which user can pass as part of connection.

sunank200: Added documentation here: https://github.com/astronomer/astro-sdk/pull/1767/commits/6646843d1866ef796fbccfab3559b8bda71fae7a

python-sdk/src/astro/databases/base.py
tatiana: @ sunank200 since this URL path directly affects production deployment (e.g. health checks and probes), changes are unlikely.

Regarding the port / reverse proxy - wouldn't those affect the overall Airflow connection endpoint the same way?

With the current approach (`minio=True`), we are adding overhead to end-users (who could not realise they needed to add this property to the connection, leading to the Python SDK incorrectly trying to use a native S3 transfer, when it shouldn't).

feluelle: I agree with @tatiana, we should not add another extra to the connection. The goal of tools such as MinIO and localstack is to provide the same interface. If we really need to check whether it is s3 or not, we should check the endpoint url as this is (and should be) the only differentiation.

For example, take a look at this guide.

More specifically, in the "Configuring Logs" section:
> In Extras, let's set the URL to our local MinIO deployment with the following syntax
> { "endpoint_url": "http://<ip>:9000" }

The endpoint url is what aws let us overwrite. localstack is using the same approach.

**So I would suggest we know that it is not native s3 if the conn type is aws s3 but the endpoint URL is defined** or the default s3 endpoint url is used if there is any.

utkarsharma2: @tatiana Made changes according to healtcheck approach.

python-sdk/tests/databases/test_base_database.py
```diff
@@ -96,3 +98,44 @@ def test_subclass_missing_append_table_raises_exception():
     target_table = Table()
     with pytest.raises(NotImplementedError):
         db.append_table(source_table, target_table, source_to_target_columns_map={})
+
+
+def test_database_with_check_for_minio_connection():
```
tatiana, line 103: It feels like we could split this test into two and use the `patch` decorator as we've done in most python SDK tests

tatiana (approved):

python-sdk/src/astro/databases/base.py
feluelle: @utkarsharma2 why do we need to get the health state? Even if minio is not up & running, we can detect if it is aws s3 or something else. Please see my comment above.

utkarsharma2: @feluelle What will happen if the user runs minio server on some other port?

utkarsharma2: @feluelle Should we just check for the presence of a port in the endpoint_url? because the user can host it locally(or on any server) and on any port, right?

python-sdk/tests/databases/test_base_database.py
utkarsharma2: Yup, done

python-sdk/src/astro/databases/base.py
feluelle: We check whether the endpoint url is set at all. Why would a user set the endpoint url if they are using native s3?

utkarsharma2: There is a use case for this - https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteEndpoints.html

feluelle: That is the wrong "endpoints", see [this](https://docs.aws.amazon.com/general/latest/gr/rande.html).

The endpoint URL usually will be created from the information you provide in your aws connfig (or airflow connection). The only time (I can think of) you actually need it, is when you use a different service such as MinIO or localstack.

However, if our users are reporting issues about that (using custom endpoint which are targeting real aws services), we can think about it again. WDYT?

utkarsharma2: @feluelle Make sense, updated code accordingly.

PR #1773, opened 2023-02-17
Use handler in run_sql

### LINKED ISSUE #1664: Discrepancy in database's run_sql() method
**Please describe the feature you'd like to see**
Currently, the handler param in the run_sql() method is only used in data bricks all other databases are not using it, which leads to different response types since the handler is processed within `run_sql()` for data bricks and not for other databases. Also, the signature of `run_sql()` in data bricks deviates from the base. We need to standardize and when we do, we can remove the below check as well.

**Describe the solution you'd like**
Ideally, we should have the same signatures for all the run_sql() database implementations. This will prevent adding explicit conditions for data bricks in other parts of the code that should ideally be database agnostic. 
example: https://github.com/astronomer/astro-sdk/blob/2f8c8d0ccbff3cc13af78685cbefa39f473991c3/python-sdk/src/astro/sql/operators/raw_sql.py#L47

**Acceptance Criteria**
- [ ] Where ever we have used run_sql(), we shouldn't be having any database-specific checks.

rajaths010494:
**Please describe the feature you'd like to see**
Currently, the handler param in the run_sql() method is only used in data bricks all other databases are not using it, which leads to different response types since the handler is processed within `run_sql()` for data bricks and not for other databases. Also, the signature of `run_sql()` in data bricks deviates from the base. We need to standardize and when we do, we can remove the below check as well.

**Describe the solution you'd like**
Ideally, we should have the same signatures for all the run_sql() database implementations. This will prevent adding explicit conditions for data bricks in other parts of the code that should ideally be database agnostic. 
example: https://github.com/astronomer/astro-sdk/blob/2f8c8d0ccbff3cc13af78685cbefa39f473991c3/python-sdk/src/astro/sql/operators/raw_sql.py#L47

**Acceptance Criteria**
- [ ] Where ever we have used run_sql(), we shouldn't be having any database-specific checks.

closes #1664

### DIFF
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -424,7 +424,7 @@ class BaseDatabase(ABC):
         statement = f"SELECT * FROM {self.get_table_qualified_name(table)}"
         if row_limit > -1:
             statement = statement + f" LIMIT {row_limit}"
-        response = self.run_sql(statement, handler=lambda x: x.fetchall())  # type: ignore
+        response: list = self.run_sql(statement, handler=lambda x: x.fetchall())
         return response
 
     def load_file_to_table(
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -14,7 +14,7 @@ from sqlalchemy import column, insert, select
 from astro.dataframes.pandas import PandasDataframe
 
 if TYPE_CHECKING:  # pragma: no cover
-    from sqlalchemy.engine.cursor import CursorResult
+    pass
 
 from sqlalchemy.sql import ClauseElement
 from sqlalchemy.sql.elements import ColumnClause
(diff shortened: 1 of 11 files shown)


tatiana (approved):
Nice to see an alignment of the interface cross-databases, @rajaths010494 ! I'm happy for this change to be merged once the checks are passing.

Since this is a breaking change, we just need to remember to add this to the release notes. WDYT if we already added to the changelog?

rajaths010494:
> Nice to see an alignment of the interface cross-databases, @rajaths010494 ! I'm happy for this change to be merged once the checks are passing.
> 
> Since this is a breaking change, we just need to remember to add this to the release notes. WDYT if we already added to the changelog?

The changelog PR is not yet created will make sure to add this change to that.

dimberman (approved):
Great job @rajaths010494! This all LGTM

PR #1777, opened 2023-02-17
Add the DatabaseDataProvider and SqliteDataProvider 

sunank200:
# Description

- Add `Dataprovider` for Databases
- Add `DataProvider` for Sqlite
- Add non-native transfer implementation for s3 to Sqlite
- Add non-native transfer implementation for GCS to Sqlite
- Add example DAG

closes: #1731

## What is the new behavior?

- Add `Dataprovider` for Databases
- Add `DataProvider` for Sqlite
- Add non-native transfer implementation for s3 to Sqlite
- Add non-native transfer implementation for GCS to Sqlite
- Add example DAG

Following is the recording of example DAG running: https://astronomer.slack.com/archives/C03868KGF2Q/p1676662728169199

No

### Checklist


universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -0,0 +1,614 @@
...
+    _drop_table_statement: str = "DROP TABLE IF EXISTS {}"
+    _create_table_statement: str = "CREATE TABLE IF NOT EXISTS {} AS {}"
+    # Used to normalize the ndjson when appending fields in nested fields.
+    # Example -
+    #   ndjson - {'a': {'b': 'val'}}
+    #   the col names generated is 'a.b'. char '.' maybe an illegal char in some db's col name.
+    # Contains the illegal char and there replacement, where the value in
+    # illegal_column_name_chars[0] will be replaced by value in illegal_column_name_chars_replacement[0]
+    illegal_column_name_chars: list[str] = []
+    illegal_column_name_chars_replacement: list[str] = []
```
dimberman, line 48: When are these values set? Does somebody create the class and then do "provider. illegal_column_name_chars = []"? If so it might be cleaner to turn these into properties.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -0,0 +1,614 @@
...
+        :param columns_names_capitalization:  determines whether to convert all columns to lowercase/uppercase
+        :param if_exists:  Overwrite file if exists
+        :param use_native_support: Use native support for data transfer if available on the destination
+        """
+        is_schema_autodetection_supported = self.check_schema_autodetection_is_supported(source_file=file)
+        is_file_pattern_based_schema_autodetection_supported = (
+            self.check_file_pattern_based_schema_autodetection_is_supported(source_file=file)
+        )
+        if if_exists == "replace":
+            self.drop_table(table)
```
dimberman, line 425: This seems somewhat unsafe. Like if I run a "replace" and it drops the table but fo some reason fails to create a new one, that would lead to the table no longer existing. @kaxil @tatiana WDYT? Should we try to create a temp table and then swap or is that overoptimization for this stage?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+        if not path.endswith("/"):
+            file = File(
+                path=path,
+                conn_id=file.conn_id,
+                filetype=filetype,
+                normalize_config=normalize_config,
+            )
+            files.append(file)
+    if len(files) == 0:
+        raise FileNotFoundError(f"File(s) not found for path/pattern '{file.path}'")
```
dimberman, line 43: Should this always return an error? Is there no scenario where someone might run against an empty directory (with the idea that sometimes there are files and sometimes there aren't)

dimberman (changes requested):
Looking great so far! A few comments I'd love to discuss first though

sunank200:
Following is the recording of example DAG running: https://astronomer.slack.com/archives/C03868KGF2Q/p1676662728169199

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -30,6 +30,22 @@
         ),
     )
 
+    transfer_non_native_s3_to_sqlite = UniversalTransferOperator(
+        task_id="transfer_non_native_s3_to_sqlite",
+        source_dataset=File(
+            path="s3://astro-sdk-test/uto/csv_files/", conn_id="aws_default", filetype=FileType.CSV
```
tatiana, line 36: Having a configurable variable/constant for the bucket name would be great, so non-Astronomer developers can also run the test suite.

We followed this pattern for the Python SDK:
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/example_dags/example_amazon_s3_postgres.py#L11

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -30,6 +30,22 @@
...
+        ),
+        destination_dataset=Table(name="uto_s3_table", conn_id="sqlite_default"),
+    )
+
+    transfer_non_native_gs_to_sqlite = UniversalTransferOperator(
+        task_id="transfer_non_native_gs_to_sqlite",
+        source_dataset=File(
+            path="gs://uto-test/uto/csv_files/", conn_id="google_cloud_default", filetype=FileType.CSV
```
tatiana, line 44: The same applies here

universal_transfer_operator/src/universal_transfer_operator/data_providers/__init__.py
```diff
@@ -5,12 +5,24 @@
...
 
-DATASET_CONN_ID_TO_DATAPROVIDER_MAPPING = dict.fromkeys(
-    ["s3", "aws"], "universal_transfer_operator.data_providers.filesystem.aws.s3"
-) | dict.fromkeys(
-    ["gs", "google_cloud_platform"], "universal_transfer_operator.data_providers.filesystem.google.cloud.gcs"
+DATASET_CONN_ID_TO_DATAPROVIDER_MAPPING = (
+    dict.fromkeys(
+        [("s3", File), ("aws", File)], "universal_transfer_operator.data_providers.filesystem.aws.s3"
+    )
+    | dict.fromkeys(
```
tatiana, line 16: Are we only supporting >= Python 3.9? IIR, it is only possible to merge dictionaries from that version onwards.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/__init__.py
tatiana: It is a valid scenario that a user may want to run against an empty directory in the case of incremental updates. If users transfer deltas using a pattern (e.g. `YYYY/MM/DD/HH`), there could be circumstances when the directory is empty.

It may make sense to expose a flag in the operator to raise an exception if the directory is empty, if the user wants this behaviour, but perhaps that could be a separate task/PR.

tatiana (commented):
I think we're close to getting this PR merged. I left some minor comments inline, and my primary concern is the absence of tests. We need tests covering the changes - and I'd like to see the coverage.

universal_transfer_operator/src/universal_transfer_operator/data_providers/__init__.py
utkarsharma2: We need to change this since we will be supporting >= 3.7

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
utkarsharma2: Done.

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
utkarsharma2: Done.

universal_transfer_operator/src/universal_transfer_operator/data_providers/__init__.py
utkarsharma2: Done.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/__init__.py
utkarsharma2: Hmmm ya, I too think this is a valid scenario. But I think in some scenarios user might want the operator to fail in case there are no files. So would it make sense to expose a parameter to the end user - `fail_if_no_files_found` or something else?

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
utkarsharma2: @dimberman I think it's a good idea to be Atomic. But this exactly what we are doing in the SDK's load_file() should we follow same there as well?

dimberman: Hmm I guess if we haven't run into an issue yet this might be ok. WDYT @tatiana ?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/__init__.py
utkarsharma2: default `fail_if_source_empty = False`

utkarsharma2: I'll add a separate ticket for the this.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
tatiana: @dimberman Are we worried about the user losing the original table schema, its previous data or both?

Any replacement operation has the inherent risk of losing the original table schema/data. Given that we are supporting both non-native & native transfers for a growing number of databases, @utkarsharma2, it feels unlike we'd manage to have a generic atomic database operation to delete/create/load all the supported databases.

As suggested by @dimberman, we could have the "backup" approach before wiping things down - but I think this should be a separate ticket. A downside of the backup approach is if the user is trying to override a huge table. We could have a flag to control this, WDYT?

Probably some of the native transfers will support this operation in a neater way (replacement) - we could also have another ticket to look into this.

tatiana (changes requested):
We're missing unit/integration tests in this PR - several lines are not covered with tests. I think it is very important we address this before merging this PR.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
utkarsharma2: @dimberman They were added in PythonSDK to flatten NDJSON files,  currently we are using them for - 
1. https://github.com/astronomer/astro-sdk/blob/9e94cb59b062060f758f142a82e4060c1a41c45f/python-sdk/src/astro/databases/aws/redshift.py#L75
2. https://github.com/astronomer/astro-sdk/blob/9e94cb59b062060f758f142a82e4060c1a41c45f/python-sdk/src/astro/databases/postgres.py#L30
3. https://github.com/astronomer/astro-sdk/blob/9e94cb59b062060f758f142a82e4060c1a41c45f/python-sdk/src/astro/databases/google/bigquery.py#L87

They are meant to hold data and we don't expect user of the class to change this on the fly.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
utkarsharma2: @tatiana I think we should explore this possibility, would be a nice addition to existing features. I'll create a separate ticket for the same.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/__init__.py
utkarsharma2: Ticket - https://github.com/astronomer/astro-sdk/issues/1816

utkarsharma2: Ticket - https://github.com/astronomer/astro-sdk/issues/1817

utkarsharma2:
Some of the test cases which refer to a local file are not added as of now, because the notion of LocalFile is present here it is introduced in another PR - https://github.com/astronomer/astro-sdk/pull/1794, post we have merged this PR we will go ahead and add test case which refers local files.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/sqlite.py
```diff
@@ -0,0 +1,139 @@
...
+    def read(self):
+        """ ""Read the dataset and write to local reference location"""
+        raise NotImplementedError
+
+    def write(self, source_ref: FileStream):
+        """Write the data from local reference location to the dataset"""
+        return self.load_file_to_table(
+            input_file=source_ref.actual_file,
+            output_table=self.dataset,
+        )
```
tatiana, line 71: @ utkarsharma2 I would like to have simple tests for all these functions - are these the ones you said are not needed in the context of UTO? If they are not necessary, perhaps we should just remove them for now.

tatiana (commented):
@utkarsharma2, thanks a lot for all the effort in adding tests. It improved significantly.

I noticed we still have coverage issues in a couple of files, including:
1. `data_providers/base.py`
2. `data_providers/database/base.py`
3. `data_providers/database/sqlite.py`
4. `data_providers/filesystem/__init__.py`
5. `datasets/file/base.py`

Not all of them are related to local file. I understand this PR is blocking others from being merged, but I'd love to have a follow-up ticket to ensure we address this before any UTO release.

tatiana (approved):
I just spoke to @utkarsharma2, and the ticket https://github.com/astronomer/astro-sdk/issues/1788 will address the coverage issues. So the UTO work is not blocked, I'm merging this PR.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/sqlite.py
utkarsharma2: @tatiana I meant the signature of the methods might change since we already keep a reference to a dataset in the Dataset provider object.

PR #1807, opened 2023-02-27
Skip Redshift Integration test cases

rajaths010494:
# Description
We are running integration test cases on astro sdk CI currently.
To make cost optimisation we need to run the redshift-based test cases on scheduled based to run when the redshift cluster is available

### DIFF
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -31,9 +31,6 @@ from astro.files import File
 from astro.options import SnowflakeLoadOptions
 from astro.table import Metadata, Table
 
-# To create IAM role with needed permissions,
-# refer: https://www.dataliftoff.com/iam-roles-for-loading-data-from-s3-into-redshift/
-REDSHIFT_NATIVE_LOAD_IAM_ROLE_ARN = os.getenv("REDSHIFT_NATIVE_LOAD_IAM_ROLE_ARN")
 SNOWFLAKE_CONN_ID = "snowflake_conn"
 DATABRICKS_CONN_ID = "databricks_conn"
 MSSQL_CONN_ID = "mssql_conn"
@@ -176,30 +173,6 @@ with dag:
     )
     # [END load_file_example_12]
 
-    # [START load_file_example_13]
-    # aql.load_file(
-    #     input_file=File(
-    #         "s3://astro-sdk/sample_pattern",
-    #         conn_id=AWS_CONN_ID,
-    #         filetype=FileType.CSV,
-    #     ),
-    #     output_table=Table(conn_id="redshift_conn", metadata=Metadata(schema="astro")),
-    #     use_native_support=False,
-    # )
-    # [END load_file_example_13]
-
-    # [START load_file_example_14]
-    # aql.load_file(
-    #     input_file=File(
-    #         "gs://astro-sdk/workspace/sample_pattern.csv",
-    #         conn_id="bigquery",
-    #         filetype=FileType.CSV,
-    #     ),
-    #     output_table=Table(conn_id="redshift_conn", metadata=Metadata(schema="astro")),
-    #     use_native_support=False,
-    # )
-    # [END load_file_example_14]
-
     # [START load_file_example_15]
     aql.load_file(
         input_file=File(path=str(CWD.parent) + "/tests/data/homes*", filetype=FileType.CSV),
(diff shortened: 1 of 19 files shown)


python-sdk/example_dags/example_load_file.py
```diff
@@ -177,27 +177,27 @@
...
-    aql.load_file(
-        input_file=File(
-            "s3://astro-sdk/sample_pattern",
-            conn_id=AWS_CONN_ID,
-            filetype=FileType.CSV,
-        ),
-        output_table=Table(conn_id="redshift_conn", metadata=Metadata(schema="astro")),
-        use_native_support=False,
-    )
+    # aql.load_file(
```
tatiana, line 180: Do we have other alternatives as opposed to commenting out these examples?
Perhaps have them in a separate folder - and update the documentation to reference them...?

python-sdk/tests_integration/databases/test_all_databases.py
```diff
@@ -56,7 +56,7 @@
         },
     ],
     indirect=True,
-    ids=["bigquery", "postgres", "redshift", "snowflake", "sqlite", "delta", "duckdb"],
+    ids=["bigquery", "postgres", "snowflake", "sqlite", "delta", "duckdb"],
```
tatiana, line 59: An alternative to remove redshift/comment it out would be to use
```
-k "not redshift"
```

and having a redshift marker:
https://docs.pytest.org/en/7.1.x/example/markers.html#marking-test-functions-and-selecting-them-for-a-run

python-sdk/tests_integration/databases/test_all_databases.py
```diff
@@ -118,15 +118,15 @@ def test_export_table_to_pandas_dataframe_mssql(
     [
         {"database": Database.BIGQUERY},
         {"database": Database.POSTGRES},
-        {"database": Database.REDSHIFT},
+        # {"database": Database.REDSHIFT},
```
tatiana, line 121: See https://github.com/astronomer/astro-sdk/pull/1807/files#r1118616324

python-sdk/tests_integration/databases/test_all_databases.py
rajaths010494: will checkout that out and try it.

rajaths010494: Added the `-k "not redshift"` for integration test cases.

python-sdk/tests_integration/databases/test_all_databases.py
rajaths010494: Added `-k not redshift` for integration test cases.

python-sdk/example_dags/example_load_file.py
rajaths010494: We can move it to a different separate file and then add a file name with redshift so it skips the test also we can update the docs to point to the new file.

rajaths010494: moved it

utkarsharma2:
@rajaths010494 can you also make sure the documents are building correctly.

python-sdk/example_dags/example_load_file_redshift.py
```diff
@@ -0,0 +1,70 @@
...
+    start_date=datetime(2019, 1, 1),
+    max_active_runs=1,
+    schedule_interval=None,
+    default_args=default_args,
+    catchup=False,
+)
+
+
+with dag:
+    # [START load_file_redshift_example_1]
```
utkarsharma2, line 33: @rajaths010494 no changes in documents?

python-sdk/tests_integration/sql/operators/test_drop_table.py
```diff
@@ -133,7 +133,7 @@ def validate_table_exists(table: Table):
         },
         {
             "database": Database.DUCKDB,
-            "table": Table(conn_id="redshift_conn"),
```
utkarsharma2, line 136: @rajaths010494 this should have failed earlier, right?

python-sdk/example_dags/example_load_file_redshift.py
rajaths010494: Yes, I checked for the previous place where the redshift `load_file_example` was used but couldn't find it anywhere in docs.

python-sdk/tests_integration/sql/operators/test_drop_table.py
rajaths010494: Yes, it should have ideally failed.

rajaths010494:
> @rajaths010494 can you also make sure the documents are building correctly.

Will validate the docs once again. If it has broken links due to redshift changes.

phanikumv (commented):
Let's park this PR for now, as the cost spike was an outlier due to usage of redshift serverless by someone outside DAG authoring team.

phanikumv:
Not needed for now

PR #1815, opened 2023-03-02
Fix CodeCov

utkarsharma2:
# Description
Currently, the codecov is broken because we don't generate the codecoverage.xml file for the entire set of test cases available, we generate it for only a subset of it like - PythonSDK or UTO or SQL CLI and codecov expects coverage file for all the tests cases, without which it leads to thinking the code coverage went down. 

There is a feature of flags and carry-forward flags - https://docs.codecov.com/docs/carryforward-flags and https://docs.codecov.com/docs/flags Which can help is carry forwarding the right coverage report from the correct commit if we generate the coverage report for only a subset of test cases.

## What is the new behavior?
codecov has a feature of flags to handle the mono repo use case - `Flags`.

Nope

### Checklist


.github/workflows/ci-uto.yaml
```diff
@@ -4,6 +4,7 @@ on:
     branches: [ 'main', 'release-**' ]
     paths:
       - 'universal_transfer_operator/**'
+      - '.github/workflows/ci-uto.yaml'
```
pankajkoti, line 7: This corresponding path for tracking the changes in `ci-sql-cli.yaml` file and triggering the CI is missing in `.github/workflows/ci-sql-cli.yaml`. Can we please also add that in this PR ?

codecov.yml
```diff
@@ -5,3 +5,16 @@ coverage:
...
+      - universal_transfer_operator/src
+    carryforward: true
+  PythonSDK:
+    paths:
+      - python-sdk/src
+    carryforward: true
+  SQL-CLI:
+    paths:
+      - sql-cli/sql_cli
+    carryforward: false
```
pankajkoti, line 20: Why is carryforward false for SQL-CLI?

pankajkoti (commented):
Can we add more description to the PR, please? By saying Codecov is broken, what do we mean exactly? And reading at Flags documentation: https://docs.codecov.com/docs/flags , it says it just groups the coverage based on the paths/projects, is that what we intend to solve here or something else is broken too?

codecov.yml
tatiana: I wondered the same!

utkarsharma2: Typo, fixed

.github/workflows/ci-uto.yaml
utkarsharma2: Done.

utkarsharma2:
Didn't work, closing it.

utkarsharma2:
@pankajkoti added complete description, please check.

kaxil (approved):

Issue #1826, opened 2023-03-03
Add argument: only cleanup if all tasks succeeded


scottleechua:
**Please describe the feature you'd like to see**
Right now, `astro.sql.cleanup` deletes temporary tables once upstream tasks are done. However, it does not currently distinguish between upstream tasks finishing because of success vs failure.

I would like `cleanup` to have an optional argument that prevents cleanup if any upstream task fails. This would be useful during DAG development, where I may not want to keep regenerating all temporary tables while bugfixing failing tasks.

**Describe the solution you'd like**
An optional boolean argument to `cleanup` that instructs no cleanup in case of any upstream failure (e.g., `pass_on_failure`).

tatiana:
Thanks, @scottleechua ! I agree the flag `skip_on_failure` would be an excellent addition.

Do you have interest and availability to implement this feature? We could support you in getting up to speed if you'd like to!

Otherwise, we'll add it to a milestone and implement it as soon as possible.

scottleechua:
@tatiana Yep, I'd be glad to contribute! If you could let me know where to look in the codebase, I'll have a look.

(And if you have any guidelines for contributors, too!)

Many thanks 😄

tatiana:
Hi @scottleechua! Thank you very much.

The cleanup operator implementation is mostly in [python-sdk/src/astro/sql/operators/cleanup.py](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/cleanup.py#L96-L109)

We'd be introducing `skip_on_failure` in a [similar place](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/sql/operators/cleanup.py#L87) to `run_sync_mode`:

Tests are available at:
* [unit tests](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/sql/operators/test_cleanup.py): `python-sdk/tests/sql/operators/test_cleanup.py`
* [integration tests](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests_integration/sql/operators/test_cleanup.py): `tests_integration/sql/operators/test_cleanup.py`

There's a [development guide](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/docs/development/DEVELOPMENT.md) and the [contribution guide](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/docs/development/CONTRIBUTING.md).

[...]

PR #1837, opened 2023-03-10
Add argument `skip_on_failure` to `CleanupOperator`

scottleechua:
# Description
Currently, `astro.sql.cleanup` deletes temporary tables once upstream tasks are done, whether or not the DAG succeeded. 

While usually desirable, more detailed control is sometimes useful. For example, during DAG development, I may not want to keep regenerating all temporary tables while bugfixing failing tasks.

## What is the new behavior?

- `CleanupOperator` has a new optional argument `skip_on_failure` that prevents table cleanup if any upstream task fails.
- To mimic current behavior, `skip_on_failure=False` by default.

This PR closes issue #1826. 

No

### Checklist
[x ] Created tests which fail without the change (if possible)
[ ] Extended the README / documentation, if necessary

### DIFF
--- a/python-sdk/src/astro/sql/operators/cleanup.py
+++ b/python-sdk/src/astro/sql/operators/cleanup.py
@@ -104,8 +104,10 @@ class CleanupOperator(AstroSQLBaseOperator):
             if not self.run_immediately:
                 self.wait_for_dag_to_finish(context)
             self.tables_to_cleanup = self.get_all_task_outputs(context=context)
-        if self.skip_on_failure and self._is_task_failed(context=context):
-            pass
+        if self.skip_on_failure:
+            task_instances = context["dag_run"].get_task_instances()
+            if self._is_task_failed(task_instances=task_instances):
+                pass
         temp_tables = filter_for_temp_tables(self.tables_to_cleanup)
         self.log.info(
             "Tables found for cleanup: %s",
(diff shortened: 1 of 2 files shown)


phanikumv (approved):
looks good to me

python-sdk/src/astro/sql/operators/cleanup.py
```diff
@@ -113,6 +121,27 @@ def drop_table(self, table: BaseTable) -> None:
...
+        :param task_instances:
+        :return: boolean if at least one task besides this one has failed
+        """
+        failed_tasks = [
+            (ti.task_id, ti.state)
+            for ti in task_instances
+            if ti.task_id != self.task_id and ti.state == State.FAILED
+        ]
+        if failed_tasks:
+            self.log.info(
```
tatiana, line 137: This log is  better placed in the `execute` method as part of the `self.skip_on_failure` - since that's where we'll have the skipping behaviour

tatiana (approved):
Excited to see your first contribution, @scottleechua ! Thank you very much! Two minor comments - I believe we are close to merge this.

python-sdk/src/astro/sql/operators/cleanup.py
```diff
@@ -113,6 +121,27 @@ def drop_table(self, table: BaseTable) -> None:
         self.log.info("Dropping table %s", table.name)
         db.drop_table(table)
 
+    def _is_task_failed(self, task_instances: list[TaskInstance]) -> bool:
```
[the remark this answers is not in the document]
scottleechua: Implemented!

python-sdk/src/astro/sql/operators/cleanup.py
scottleechua: I may push back, as moving the log to `execute` means that `has_task_failed` would have to return the a list of failed tasks instead of a simple boolean;

Also, other cleanup functions `drop_table()` and `_is_dag_running()` have logging behavior inside them.

scottleechua:
Resolved conversation and updated code!

scottleechua:
(Sorry, I think I pressed something in the reviewer section!)

Issue #1825, opened 2023-03-03
Unable to deploy SQL workflows to Astro Cloud


tatiana:
**Describe the bug**
We're no longer able to deploy SQL workflows to Astro Cloud

**Version**
* Astro CLI: 1.11
* SQL CLI: 0.4.0

**To Reproduce**

Try to deploy any DAG to Astronomer cloud, e.g:
```
$ astro flow deploy example_load_file_s3_to_snowflake

Generating the DAG file from workflow example_load_file_s3_to_snowflake for 
default environment..

The DAG file 
/home/tati/Demo/devex/dags/.astro-sql-deploy/default/example_load_file_s3_to_sno
wflake.py has been successfully generated. 🎉
Authenticated to Astro 

Building image...
Sending build context to Docker daemon  1.292MB
Step 1/1 : FROM quay.io/astronomer/astro-runtime:7.3.0
# Executing 7 build triggers
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> c54bf78d4bd3
Successfully built c54bf78d4bd3
Successfully tagged frozen-sunspot-3603/airflow:latest
Runtime Version: 7.3.0
testing frozen-sunspot-3603/airflow:latest
Testing image...

Checking your DAGs for errors,
this might take a minute if you haven't run this command before…
============================= test session starts ==============================
platform linux -- Python 3.9.16, pytest-7.2.1, pluggy-1.0.0
rootdir: /usr/local/airflow
plugins: anyio-3.6.2
collected 4 items

.astro/test_dag_integrity_default.py .F.F

=================================== FAILURES ===================================
_ test_file_imports[dags/.astro-sql-deploy/default/example_load_file_s3_to_snowflake.py] _

rel_path = 'dags/.astro-sql-deploy/default/example_load_file_s3_to_snowflake.py'
rv = 'Traceback (most recent call last):\n  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed\n ...py", line 10, in <module>\n    from sql_cli import operators as naql\nModuleNotFoundError: No module named \'sql_cli\''

    @pytest.mark.parametrize(
        "rel_path,rv", get_import_errors(), ids=[x[0] for x in get_import_errors()]
    )
    def test_file_imports(rel_path, rv):
        """Test for import errors on a file"""
        if rel_path and rv:  # Make sure our no op test doesn't raise an error
[...]

PR #1841, opened 2023-03-13
Fix astro flow deploy integration by having a conditional sql_cli import

pankajkoti:
# Description

The `astro flow deploy` command fails currently as it cannot find the sql_cli library installed

closes: #1825 

## What is the new behavior?
This library `sql_cli` is required only for local runs when running a single task. So, to fix the deploy command, we identify if it is for a local run using the `task_to_execute` list in the context populated only for single task runs. This list will be empty when deploying DAGs that are supposed to generate and run wholly.

No.

### Checklist


dimberman (approved):
@pankajkoti do we need any tests around this? +1ing because it's a small fix but please do add tests if necessary :)

pankajkoti:
> @pankajkoti do we need any tests around this? +1ing because it's a small fix but please do add tests if necessary :)

Thank you Daniel, yes, I have added the missing test to check the generated DAGs for single task runs include `naql` import and nominal operators for tasks to not run core logic.

PR #1845, opened 2023-03-14
Add example dags testcase as part of CI

utkarsharma2:
# Description
Add example dags as part of integration tests.

### DIFF
--- a/universal_transfer_operator/example_dags/example_dag_fivetran.py
+++ b/universal_transfer_operator/example_dags/example_dag_fivetran.py
@@ -1,0 +1,81 @@
+import os
+from datetime import datetime
+
+from airflow import DAG
+
+from universal_transfer_operator.constants import TransferMode
+from universal_transfer_operator.datasets.file.base import File
+from universal_transfer_operator.datasets.table import Metadata, Table
+from universal_transfer_operator.integrations.fivetran import Connector, Destination, FiveTranOptions, Group
+from universal_transfer_operator.universal_transfer_operator import UniversalTransferOperator
+
+s3_bucket = os.getenv("S3_BUCKET", "s3://astro-sdk-test")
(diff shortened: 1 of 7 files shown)


universal_transfer_operator/example_dags/example_dag_fivetran.py
```diff
@@ -0,0 +1,81 @@
...
+                    "skip_before": 0,
+                    "role_arn": "arn::your_role_arn",
+                    "file_type": "csv",
+                    "delimiter": "",
+                    "is_public": False,
+                    "on_error": "fail",
+                    "compression": "bz2",
+                    "table": "fivetran_test",
+                    "archive_pattern": "regex_pattern",
+                },
```
dimberman, line 65: Might be easier to read to move this dictionary into its own variable so users that are looking at the DAG can focus on DAG structure.

universal_transfer_operator/example_dags/example_dag_fivetran.py
```diff
@@ -0,0 +1,81 @@
...
+                time_zone_offset="-5",
+                region="GCP_US_EAST4",
+                config={
+                    "host": "your-account.snowflakecomputing.com",
+                    "port": 443,
+                    "database": "fivetran",
+                    "auth": "PASSWORD",
+                    "user": "fivetran_user",
+                    "password": "123456",
+                },
```
dimberman, line 78: Same here.

dimberman (approved):
Two minor comments but overal LGTM

universal_transfer_operator/example_dags/example_dag_fivetran.py
utkarsharma2: Done.

universal_transfer_operator/example_dags/example_dag_fivetran.py
utkarsharma2: Done.

PR #1846, opened 2023-03-14
Fix S3provider issue

utkarsharma2:
# Description
Fix s3 provider issue, needs {} instead of None in`transfer_config_args` config.


pankajastro (approved):
LGTM

Issue #1787, opened 2023-02-22
Address TypeCheck


utkarsharma2:
**Please describe the feature you'd like to see**
Currently, there are type-checks that are failing in CI for UTO

**Describe the solution you'd like**
We should Fix it, asap.

PR #1849, opened 2023-03-16
Fix mypy type checks

utkarsharma2:
# Description
Currently, type checks are not passing, fixing it in this PR.

closes: #1787


universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -1,16 +1,20 @@
...
 import attr
 from airflow.hooks.base import BaseHook
 
 from universal_transfer_operator.constants import Location
-from universal_transfer_operator.datasets.base import Dataset
+from universal_transfer_operator.datasets.file.base import File
+from universal_transfer_operator.datasets.table import Table
 from universal_transfer_operator.utils import TransferParameters, get_dataset_connection_type
 
+T = TypeVar("T", File, Table)
```
tatiana, line 14: Why was this necessary? Perhaps we could give it a more meaningful name (e.g. `DatasetType` or something similar

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -61,7 +65,7 @@ def write(self, source_ref):
         """Write the data from local reference location to the dataset"""
         raise NotImplementedError
 
-    def load_data_from_source_natively(self, source_dataset: Dataset, destination_dataset: Dataset) -> None:
+    def load_data_from_source_natively(self, source_dataset: T, destination_dataset: T) -> Any:
```
tatiana, line 68: What type of output would this method usually return?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
```diff
@@ -129,15 +134,15 @@ def download_file(self, file) -> TempFile:
...
 
     @property
-    def transfer_config_args(self) -> dict | None:
+    def transfer_config_args(self) -> Any:
         return self.dataset.extra.get("transfer_config_args", {})
 
     @property
-    def s3_extra_args(self) -> dict | None:
+    def s3_extra_args(self) -> Any:
         return self.dataset.extra.get("s3_extra_args", {})
```
tatiana, line 146: Feeling a bit uneasy about these `Any` return values :see_no_evil: Why didn't it work to use a more specific type?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
```diff
@@ -151,19 +156,19 @@ def s3_key(self) -> str:
...
         return self.dataset.extra.get("prefix", None)
 
     @property
-    def keep_directory_structure(self) -> bool:
+    def keep_directory_structure(self) -> Any:
         return self.dataset.extra.get("keep_directory_structure", False)
 
     @property
-    def delimiter(self) -> str | None:
+    def delimiter(self) -> Any:
```
tatiana, line 171: Same here

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
utkarsharma2: No reason, I'll change it.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
utkarsharma2: I too don't like it, but since `self.dataset.extra` is of type dict, which is actually dict[Any, Any] we can't be specific here and the user can actually populate any values for the self.dataset.extra.get("s3_extra_args"). 
If we have to narrow it down, we need to change it to `self.dataset.extra` but I'm not sure what all types of values are valid for this at the moment, I'll more that happy to narrow it down if we can figure this out somehow,

utkarsharma2: And this is the reasoning for other methods having `Any`

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
tatiana: WDYT if we just removed it, while we don't need it - and then we can look into this again when we have native support?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/aws/s3.py
tatiana: Hm, I see - we'll start like this and review in future :crossed_fingers:

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
utkarsharma2: Done.

tatiana (approved):
Thanks a lot for addressing the feedback and improving the code/usage of type hints, @utkarsharma2 ! I believe it's good enough for now - in future we can review the possibility of extras not being a `dict[Any, Any]` or alternatives to the current return any methods.

PR #1859, opened 2023-03-20
Fix database to file issue

utkarsharma2:
# Description
Currently, when loading data from a database we are saving data in a parquet file. Due to which the data when transferring data from the database to the filesystem, we were limited to the parquet file type.

## What is the new behavior?
Now, we have to change the interface of read() also to give an output of data frame and write() to consume data frame for csv, JSON, ndjson, and parquet file type. This gives us the freedom to convert the data in any format we prefer csv, Json, Ndjson or Parquet. 

Nope

### Checklist

### DIFF
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import io
 from abc import ABC
-from typing import Generic, TypeVar
+from pathlib import Path
+from typing import Generic, Iterator, TypeVar
 
 import attr
+import pandas as pd
 from airflow.hooks.base import BaseHook
 
 from universal_transfer_operator.constants import Location
@@ -12,6 +15,13 @@ from universal_transfer_operator.datasets.table import Table
 from universal_transfer_operator.utils import TransferParameters, get_dataset_connection_type
 
 DatasetType = TypeVar("DatasetType", File, Table)
+
+
+@attr.define
+class FileStream:
+    remote_obj_buffer: io.IOBase
+    actual_filename: Path
+    actual_file: File
 
 
 class DataProviders(ABC, Generic[DatasetType]):
@@ -57,11 +67,11 @@ class DataProviders(ABC, Generic[DatasetType]):
         source_connection_type = get_dataset_connection_type(source_dataset)
         return Location(source_connection_type) in self.transfer_mapping
 
-    def read(self):
+    def read(self) -> Iterator[pd.DataFrame] | Iterator[FileStream]:
         """Read the dataset and write to local reference location"""
         raise NotImplementedError
 
-    def write(self, source_ref):
+    def write(self, source_ref) -> str:  # type: ignore
         """Write the data from local reference location to the dataset"""
         raise NotImplementedError
 
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
@@ -68,11 +68,14 @@ class DataProviders(ABC, Generic[DatasetType]):
         return Location(source_connection_type) in self.transfer_mapping
 
     def read(self) -> Iterator[pd.DataFrame] | Iterator[FileStream]:
-        """Read the dataset and write to local reference location"""
+        """Read from filesystem dataset or databases dataset and write to local reference locations or dataframes"""
         raise NotImplementedError
 
-    def write(self, source_ref) -> str:  # type: ignore
-        """Write the data from local reference location to the dataset"""
+    def write(self, source_ref: pd.DataFrame | FileStream) -> str:  # type: ignore
+        """Write the data from local reference location or a dataframe to the database dataset or filesystem dataset
+
+        :param source_ref: Stream of data to be loaded into output table or a pandas dataframe.
+        """
         raise NotImplementedError
 
     @property
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
@@ -18,7 +18,7 @@ DatasetType = TypeVar("DatasetType", File, Table)
 
 
 @attr.define
-class FileStream:
+class DataStream:
     remote_obj_buffer: io.IOBase
     actual_filename: Path
     actual_file: File
@@ -67,11 +67,11 @@ class DataProviders(ABC, Generic[DatasetType]):
         source_connection_type = get_dataset_connection_type(source_dataset)
         return Location(source_connection_type) in self.transfer_mapping
 
-    def read(self) -> Iterator[pd.DataFrame] | Iterator[FileStream]:
+    def read(self) -> Iterator[pd.DataFrame] | Iterator[DataStream]:
         """Read from filesystem dataset or databases dataset and write to local reference locations or dataframes"""
         raise NotImplementedError
 
-    def write(self, source_ref: pd.DataFrame | FileStream) -> str:  # type: ignore
+    def write(self, source_ref: pd.DataFrame | DataStream) -> str:  # type: ignore
         """Write the data from local reference location or a dataframe to the database dataset or filesystem dataset
 
         :param source_ref: Stream of data to be loaded into output table or a pandas dataframe.
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
@@ -464,7 +464,7 @@ class DatabaseDataProvider(DataProviders[Table]):
         use_native_support: bool = True,
     ):
         """
-        Checks if the autodetect schema exists for native support else creates the schema and table
+        Creates the schema and table from dataframe
         :param table: Table to create
         :param file: File path and conn_id for object stores
         :param normalize_config: pandas json_normalize params config
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
@@ -193,6 +193,10 @@ class DatabaseDataProvider(DataProviders[Table]):
         Write the data from local reference location to the dataset.
         :param source_ref: Stream of data to be loaded into output table.
         """
+        # `source_ref` can be a dataframe for all the filetypes we can create a dataframe for like -
+        # CSV, JSON, NDJSON, and Parquet or SQL Tables. This gives us the option to perform various
+        # functions on the data on the fly, like filtering or changing the file format altogether. For other
+        # files whose content cannot be converted to dataframe like - zip or image, we get a Filestream object.
         if isinstance(source_ref, FileStream):
             return self.load_file_to_table(input_file=source_ref.actual_file, output_table=self.dataset)
         return self.load_dataframe_to_table(input_dataframe=source_ref, output_table=self.dataset)
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
@@ -23,9 +23,8 @@ from universal_transfer_operator.constants import (
     Location,
     TransferMode,
 )
-from universal_transfer_operator.data_providers.base import DataProviders
+from universal_transfer_operator.data_providers.base import DataProviders, FileStream
 from universal_transfer_operator.data_providers.filesystem import resolve_file_path_pattern
-from universal_transfer_operator.data_providers.filesystem.base import FileStream
 from universal_transfer_operator.datasets.dataframe.pandas import PandasDataframe
 from universal_transfer_operator.datasets.file.base import File
 from universal_transfer_operator.datasets.table import Metadata, Table
@@ -189,15 +188,14 @@ class DatabaseDataProvider(DataProviders[Table]):
         """Read the dataset and write to local reference location"""
         yield self.export_table_to_pandas_dataframe()
 
-    def write(self, source_ref: FileStream | pd.DataFrame):
+    def write(self, source_ref: FileStream | pd.DataFrame) -> str:
         """
         Write the data from local reference location to the dataset.
         :param source_ref: Stream of data to be loaded into output table.
         """
         if isinstance(source_ref, FileStream):
             return self.load_file_to_table(input_file=source_ref.actual_file, output_table=self.dataset)
-        elif isinstance(source_ref, pd.DataFrame):
-            return self.load_dataframe_to_table(input_dataframe=source_ref, output_table=self.dataset)
+        return self.load_dataframe_to_table(input_dataframe=source_ref, output_table=self.dataset)
 
     @property
     def openlineage_dataset_namespace(self) -> str:
@@ -507,7 +505,7 @@ class DatabaseDataProvider(DataProviders[Table]):
         chunk_size: int = DEFAULT_CHUNK_SIZE,
         columns_names_capitalization: ColumnCapitalization = "original",
         **kwargs,
-    ):
+    ) -> str:
         """
         Load content of multiple files in output_table.
         Multiple files are sourced from the file path, which can also be path pattern.
@@ -539,6 +537,7 @@ class DatabaseDataProvider(DataProviders[Table]):
             if_exists="append",
             chunk_size=chunk_size,
         )
+        return self.get_table_qualified_name(output_table)
 
     def load_dataframe_to_table(
         self,
@@ -547,7 +546,7 @@ class DatabaseDataProvider(DataProviders[Table]):
         if_exists: LoadExistStrategy = "replace",
         chunk_size: int = DEFAULT_CHUNK_SIZE,
         columns_names_capitalization: ColumnCapitalization = "original",
-    ):
+    ) -> str:
         """
         Load content of dataframe in output_table.
 
@@ -572,6 +571,7 @@ class DatabaseDataProvider(DataProviders[Table]):
             chunk_size=chunk_size,
             if_exists=if_exists,
         )
+        return self.get_table_qualified_name(output_table)
 
     def load_file_to_table_using_pandas(
         self,
--- a/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
+++ b/universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
@@ -185,7 +185,7 @@ class DatabaseDataProvider(DataProviders[Table]):
         source_connection_type = get_dataset_connection_type(source_dataset)
         return Location(source_connection_type) in self.transfer_mapping
 
-    def read(self):
+    def read(self) -> pd.DataFrame:
         """Read the dataset and write to local reference location"""
         yield self.export_table_to_pandas_dataframe()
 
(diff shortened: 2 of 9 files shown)


universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -186,24 +185,19 @@ def check_if_transfer_supported(self, source_dataset: Table) -> bool:
...
+        yield self.export_table_to_pandas_dataframe()
 
-    def write(self, source_ref: FileStream):
+    def write(self, source_ref: FileStream | pd.DataFrame):
         """
         Write the data from local reference location to the dataset.
         :param source_ref: Stream of data to be loaded into output table.
         """
-        return self.load_file_to_table(input_file=source_ref.actual_file, output_table=self.dataset)
+        if isinstance(source_ref, FileStream):
```
sunank200, line 197: We should comment on the logic of why we are doing the check for both `Filestream` and pandas data frame. Filestream is needed as well for data in formats like zip, images etc and pandas dataframe is relevant for filestypes like csv, json, parquet etc.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py
```diff
@@ -90,17 +92,17 @@ def check_if_transfer_supported(self, source_dataset: File) -> bool:
...
 
-    def read_using_smart_open(self):
+    def read_using_smart_open(self) -> Iterator[FileStream]:
         """Read the file dataset using smart open returns i/o buffer"""
         files = self.paths
         for file in files:
             yield FileStream(
                 remote_obj_buffer=self._convert_remote_file_to_byte_stream(file),
-                actual_filename=file,
+                actual_filename=Path(file),
```
sunank200, line 105: Is this tested with this change for all transfers in example DAGs? Are the files getting uploaded properly with correct name?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py
```diff
@@ -118,19 +120,22 @@ def _convert_remote_file_to_byte_stream(self, file: str) -> io.IOBase:
...
 
-    def write_using_smart_open(self, source_ref: FileStream):
+    def write_using_smart_open(self, source_ref: FileStream | pd.DataFrame):
         """Write the source data from remote object i/o buffer to the dataset using smart open"""
-        mode = "wb" if self.read_as_binary(source_ref.actual_file.path) else "w"
+        mode = "wb" if self.read_as_binary(self.dataset.path) else "w"
         destination_file = self.dataset.path
         with smart_open.open(destination_file, mode=mode, transport_params=self.transport_params) as stream:
-            stream.write(source_ref.remote_obj_buffer.read())
+            if isinstance(source_ref, FileStream):
```
sunank200, line 135: Logic for why we check for filestream and dataframe can be documented here

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
```diff
@@ -114,7 +115,13 @@ def get_complete_url(self, dst_url: str, src_url: str) -> str:
 
         return urlunparse(final_url)
 
-    def write_using_smart_open(self, source_ref: FileStream) -> str:
+    def write_using_smart_open(self, source_ref: FileStream | pd.DataFrame):
```
sunank200, line 118: Doc-strings are missing

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
```diff
@@ -125,6 +132,17 @@ def write_using_smart_open(self, source_ref: FileStream) -> str:
             stream.write(source_ref.remote_obj_buffer.read())
         return complete_url
 
+    def write_dataframe_using_smart_open(self, source_ref: pd.DataFrame) -> str:
```
sunank200, line 135: Have we tested this?

sunank200 (commented):
Overall looks good but have added comments

Return type should be added in base of dataprovider as well and the logic for why it can return Filestream or Pandas dataframe should be added here: https://github.com/astronomer/astro-sdk/blob/main/universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py#L60

Filestream is needed for scenarios when its file-to-file transfers (Eg, transfer of zip files, images etcs). Pandas dataframe would be relevant when file is of type csv, parquet, json etc.

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -57,11 +67,11 @@ def check_if_transfer_supported(self, source_dataset: DatasetType) -> bool:
         source_connection_type = get_dataset_connection_type(source_dataset)
         return Location(source_connection_type) in self.transfer_mapping
 
-    def read(self):
+    def read(self) -> Iterator[pd.DataFrame] | Iterator[FileStream]:
         """Read the dataset and write to local reference location"""
```
tatiana, line 71: We could improve this description. Do we read the dataset potentially from any data source (e.g. object store) and copy it to a local file system?

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -57,11 +67,11 @@ def check_if_transfer_supported(self, source_dataset: DatasetType) -> bool:
...
         raise NotImplementedError
 
-    def write(self, source_ref):
+    def write(self, source_ref) -> str:  # type: ignore
         """Write the data from local reference location to the dataset"""
```
tatiana, line 75: Same here, are we moving a file located in the local file system to a target destination (file, table..?)

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
utkarsharma2: Yup, added it.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -186,24 +184,18 @@ def check_if_transfer_supported(self, source_dataset: Table) -> bool:
         source_connection_type = get_dataset_connection_type(source_dataset)
         return Location(source_connection_type) in self.transfer_mapping
 
-    def read(self):
+    def read(self) -> Iterator[pd.DataFrame]:
         """Read the dataset and write to local reference location"""
```
tatiana, line 188: This docstring can be improved - read from SQL table into a pandas dataframe

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py
utkarsharma2: Yup, this test case checks for everything - https://github.com/astronomer/astro-sdk/blob/8dc5da8d7ccdc5ecd665bba70434e2fc4c8b831f/universal_transfer_operator/tests_integration/test_data_provider/test_databases/test_base.py#L75 
Do you have anything besides this you want to test?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
utkarsharma2: Yup, this test case checks for everything - https://github.com/astronomer/astro-sdk/blob/8dc5da8d7ccdc5ecd665bba70434e2fc4c8b831f/universal_transfer_operator/tests_integration/test_data_provider/test_databases/test_base.py#L75 
Do you have anything besides this you want to test?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
```diff
@@ -114,7 +116,14 @@ def get_complete_url(self, dst_url: str, src_url: str) -> str:
 
         return urlunparse(final_url)
 
-    def write_using_smart_open(self, source_ref: FileStream) -> str:
+    def write_using_smart_open(self, source_ref: FileStream | pd.DataFrame):
+        """Write the source data from remote object i/o buffer to the dataset using smart open"""
+        if isinstance(source_ref, FileStream):
+            return self.write_file_using_smart_open(source_ref=source_ref)
+        elif isinstance(source_ref, pd.DataFrame):
+            return self.write_dataframe_using_smart_open(source_ref=source_ref)
```
tatiana, line 124: Don't we want to return `source_ref` in this case?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
```diff
@@ -125,6 +134,17 @@ def write_using_smart_open(self, source_ref: FileStream) -> str:
             stream.write(source_ref.remote_obj_buffer.read())
         return complete_url
 
+    def write_dataframe_using_smart_open(self, source_ref: pd.DataFrame) -> str:
+        """Write the source data from dataframe to the dataset using smart open
```
tatiana, line 138: We can improve the docstrings significantly.

Does this function write from a Pandas DataFame into a  sFTP file? Let's be explicit about it.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
```diff
@@ -125,6 +134,17 @@ def write_using_smart_open(self, source_ref: FileStream) -> str:
             stream.write(source_ref.remote_obj_buffer.read())
         return complete_url
 
+    def write_dataframe_using_smart_open(self, source_ref: pd.DataFrame) -> str:
```
tatiana, line 137: How important is it to have in the function name how it is implemented? Do we have a `write_dataframe` that does not use smart open?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/local.py
```diff
@@ -63,12 +65,14 @@ def check_if_exists(self) -> bool:
         """Return true if the dataset exists"""
         return exists(self.dataset.path)
 
-    def write_using_smart_open(self, source_ref: FileStream):
+    def write_using_smart_open(self, source_ref: FileStream | pd.DataFrame) -> str:
```
tatiana, line 68:
```suggestion
    def write_using_stream(self, source_ref: FileStream | pd.DataFrame) -> str:
```

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
utkarsharma2: It will be covered as part of the this integration test - https://github.com/astronomer/astro-sdk/blob/8dc5da8d7ccdc5ecd665bba70434e2fc4c8b831f/universal_transfer_operator/tests_integration/test_data_provider/test_databases/test_base.py#L75
But I have also added test to check specifically that method.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -461,6 +457,34 @@ def create_schema_and_table_if_needed(
...
+        self,
+        table: Table,
+        dataframe: pd.DataFrame,
+        columns_names_capitalization: ColumnCapitalization = "original",
+        if_exists: LoadExistStrategy = "replace",
+        use_native_support: bool = True,
+    ):
+        """
+        Creates the schema and table from dataframe
+        :param table: Table to create
```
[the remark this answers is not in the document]
phanikumv: need to have a carriage return between docstring and param for docs to be rendered correctly

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -517,6 +541,41 @@ def load_file_to_table(
...
+        :param input_dataframe: dataframe
+        :param output_table: Table to create
+        :param if_exists: Overwrite file if exists
+        :param chunk_size: Specify the number of records in each batch to be written at a time
+        :param normalize_config: pandas json_normalize params config
+        :param columns_names_capitalization: determines whether to convert all columns to lowercase/uppercase
+            in the resulting dataframe
+        """
+
+        self.create_schema_and_table_if_needed_from_dataframe(
```
phanikumv, line 566:
```suggestion
        self.create_schema_and_table_if_needed(
```

thats a big method name IMO, WDYT abt the suggestion?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py
```diff
@@ -118,19 +113,26 @@ def _convert_remote_file_to_byte_stream(self, file: str) -> io.IOBase:
             remote_obj_buffer.seek(0)
             return remote_obj_buffer
 
-    def write(self, source_ref: FileStream):
+    def write(self, source_ref: FileStream | pd.DataFrame) -> str:
         """
         Write the data from local reference location to the dataset
         :param source_ref: Source FileStream object which will be used to read data
```
phanikumv, line 119:
```suggestion

        :param source_ref: Source FileStream object which will be used to read data
```

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
phanikumv, line 137:
```suggestion
    def write_dataframe(self, source_ref: pd.DataFrame) -> str:
```

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
utkarsharma2: I think we should return `self.dataset.path` which is the destination value. source_ref will contain the value where data is being read from. It can be local worker path as well in some cases, which is not useful to end user. 

Also for being consistent with existing code we are giving the value of destination, which user already have, maybe we can provide something like `bytes transferred` or `rows affected` instead? but this has to be a separate ticket, WDYT?

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
utkarsharma2: Not important, maybe we can change it to - `write_from_dataframe()` and `write_from_stream()`

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/local.py
utkarsharma2: @tatiana  I like the existing name in comparison because of the following reasons:
1. We don't always use a stream to write in some cases we also use dataframe. 
2. Also, we have different ways to write like write_using_hooks() and write_using_smart_open(), and these methods are called in general write() method.
3. I think it's clear from the typing of params what is used as source for the write operator. 

WDYT?

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
utkarsharma2: @phanikumv We have two ways of creating the schema via file and dataframe, that's why this name and there is already a method which has this name which uses files. But I'm open to other suggestions.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
utkarsharma2: Added a response here - https://github.com/astronomer/astro-sdk/pull/1859#discussion_r1146089560

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py
utkarsharma2: Done.

universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/sftp.py
utkarsharma2: Done.

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
utkarsharma2: Done.

sunank200 (commented):
Added small comment on base class for data provider. Rest of it looks good to me.

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -14,6 +17,13 @@
 DatasetType = TypeVar("DatasetType", File, Table)
 
 
+@attr.define
+class FileStream:
```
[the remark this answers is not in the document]
utkarsharma2: Updated it.

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -57,12 +67,15 @@ def check_if_transfer_supported(self, source_dataset: DatasetType) -> bool:
         source_connection_type = get_dataset_connection_type(source_dataset)
         return Location(source_connection_type) in self.transfer_mapping
 
-    def read(self):
-        """Read the dataset and write to local reference location"""
+    def read(self) -> Iterator[pd.DataFrame] | Iterator[FileStream]:
```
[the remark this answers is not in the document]
utkarsharma2: I think it is important that we specify the superset of return types.

universal_transfer_operator/src/universal_transfer_operator/data_providers/database/base.py
```diff
@@ -186,24 +184,23 @@ def check_if_transfer_supported(self, source_dataset: Table) -> bool:
...
-            df = self.export_table_to_pandas_dataframe()
-            df.to_parquet(tmp_file.name)
-            local_temp_file = FileStream(
-                remote_obj_buffer=tmp_file.file,
-                actual_filename=tmp_file.name,
-                actual_file=File(path=tmp_file.name),
-            )
-            yield local_temp_file
+    def read(self) -> Iterator[pd.DataFrame]:
+        """Read database dataset and convert them to dataframes"""
```
tatiana, line 188:
```suggestion
        """Convert a Table to a Pandas DataFrame"""
```

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
```diff
@@ -14,6 +17,13 @@
 DatasetType = TypeVar("DatasetType", File, Table)
 
 
+@attr.define
+class DataStream:
+    remote_obj_buffer: io.IOBase
+    actual_filename: Path
+    actual_file: File
```
tatiana, line 24: What are `actual_filename` and `actual_file`? Are they the local representation of a remote file?

sunank200: `actual_filename` is the name of source dataset which is to be transferred to destination. `actual_file` is local representation of remote file

sunank200 (approved):
Overall looks good to me

universal_transfer_operator/src/universal_transfer_operator/data_providers/base.py
tatiana: @sunank200 I think we can improve the naming in a separate PR, WDYT?

sunank200: Sure @tatiana I can create an issue for this in new repo.

PR #1866, opened 2023-03-23
Add example dag for SFTP and Local 

utkarsharma2:
# Description
Missing example dag for SFTP and Local data provides.

## What is the new behavior?
Added example dag for SFTP and Local data provides

Nope

### Checklist


sunank200 (approved):

rajaths010494 (approved):

universal_transfer_operator/example_dags/example_universal_transfer_operator.py
```diff
@@ -134,3 +138,9 @@
         destination_dataset=Table(name="uto_bigquery_to_sqlite_table", conn_id="sqlite_default"),
     )
     # [END transfer_non_native_bigquery_to_sqlite]
+
+    transfer_non_native_local_to_sftp = UniversalTransferOperator(
+        task_id="transfer_non_native_local_to_sftp",
+        source_dataset=File(path=f"{DATA_DIR}sample.csv", filetype=FileType.CSV),
```
tatiana, line 144: Don't we support `pathlib.Path` here? It would be great if we did, it feels a bit clumsy to use strings to concatenate local file paths - and to have to convert a local file path to a string

phanikumv: I think we need to do this all across the repo...everywhere we have string paths

utkarsharma2: Ya, I think it will be a good idea, I'll create a ticket for the same.

Issue #1839, opened 2023-03-13
Dataframes being stored on local machines without warning


dimberman:
**Describe the bug**

When using the aql.dataframe function in Airflow, the resulting dataframe is stored on the local worker instead of being accessible by future tasks. This behavior was reported by a user who also noticed that there was no warning or exception indicating the need to define a connection. The user had to search through the code to find the relevant information.

Steps to Replicate:

Expected behavior:
When using the aql.dataframe function, the user should receive a warning or exception if a connection is not defined, and the resulting dataframe should be stored in a way that allows it to be accessed by future tasks.
**Version**
* Astro: [e.g. 0.6.0]
* OS: [eg. Debian]

**To Reproduce**
1. Create an Airflow instance using Astro CLI.
2. Write a DAG that uses the aql.dataframe function.
3. Create a large dataframe that cannot be stored in memory.
4. Run the DAG on an Astro Cloud instance with more than one worker (alternatively, you can confirm that the dataframe has been written into the local machine by checking logs/seeing that no error is passed even though we do not define an S3 backent)

**Expected behavior**
When using the aql.dataframe function, the user should receive a warning or exception if a connection is not defined AND they do not explicitly set, and the resulting dataframe should be stored in a way that allows it to be accessed by future tasks.

**Additional context**
When we were using a custom XCom backend we had the error explicitly named [here](https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/custom_backend/astro_custom_backend.py#L34-L40). I have a theory that since we are now using Airflow's Serializer, that this is no longer the case. Looking at the [pandas class](https://github.com/astronomer/astro-sdk/blob/ecf2d6101ce9efa635418716e5e2a9168cb099a4/python-sdk/src/astro/dataframes/pandas.py#L18-L34) and the [convert_dataframe_to_file](https://github.com/astronomer/astro-sdk/blob/fe2bbe3db8cfd41664f7251a6169b84a8f699390/universal_transfer_operator/src/universal_transfer_operator/datasets/dataframe/pandas.py#L20) function. I do not see anywhere that creates this warning.

PR #1869, opened 2023-03-23
Raise error when remote storage connection is unset for large dataframes

pankajkoti:
# Description
With aql.dataframe, in case of large dataframes that cannot be stored in Airflow metadata DB, they are silently 
stored on local machines without warning and are unavailable for subsequent tasks via XCOM.

closes: #1839 

## What is the new behavior?

Astro SDK allows setting an Airflow config variable AIRFLOW__ASTRO_SDK__XCOM_STORAGE_CONN_ID
to allow dataframes to be stored in a remote location using the connection and hence making it possible
for subsequent tasks to use that remotely stored dataframe for it's inputs. However, if this connection ID is
not set, we were not alerting the user to set it and were instead silently storing the dataframe on the local machine.
We now raise an exception in case the dataframe is large enough that it cannot be stored in Airflow's Metadata DB and 
that the remote storage connection ID is unset.

No.

### Checklist

### DIFF
--- a/python-sdk/src/astro/dataframes/pandas.py
+++ b/python-sdk/src/astro/dataframes/pandas.py
@@ -22,14 +22,7 @@ class PandasDataframe(DataFrame):
         if df_size < (settings.MAX_DATAFRAME_MEMORY_FOR_XCOM_DB * 1024):
             logger.info("Dataframe size: %s bytes. Storing it in Airflow's metadata DB", df_size)
             return {"data": self.to_json()}
-        else:
-            if settings.DATAFRAME_STORAGE_CONN_ID is None:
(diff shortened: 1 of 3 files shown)


python-sdk/src/astro/dataframes/pandas.py
```diff
@@ -22,6 +23,13 @@ def serialize(self):
             logger.info("Dataframe size: %s bytes. Storing it in Airflow's metadata DB", df_size)
             return {"data": self.to_json()}
         else:
+            if settings.DATAFRAME_STORAGE_CONN_ID is None:
```
tatiana, line 26: Would it make more sense to have an `elif` here?

Issue #1870, opened 2023-03-23
Folder or Directory not being handled


sunank200:
**Describe the bug**
As part of this [PR 1829](https://github.com/astronomer/astro-sdk/pull/1829), a bug was introduced. This is not handling the case when the File path is a file pattern or directory.
<img width="1680" alt="Screenshot 2023-03-24 at 12 23 45 AM" src="https://user-images.githubusercontent.com/8670962/227315288-b11bc232-65a0-458f-a217-b74c9b882b07.png">

In case the destination dataset is a folder it doesn't pass the filename from the source at all here: https://github.com/astronomer/astro-sdk/blame/1b35bb9443d57ef7834259a3cd3c26695ea5fcf8/universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py#L131 instead of https://github.com/astronomer/astro-sdk/blame/bf189de90e4ad9f2db77c7ac46a7193bde9a9c92/universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py#L123

It should check if the provided file is a directory or file pattern and then add the actual file name from the source dataset from `FileStream`

```
destination_file = self.dataset.path
# check if destination dataset is folder or file pattern
if self.dataset.is_pattern():
    destination_file = os.path.join(self.dataset.path, os.path.basename(source_ref.actual_filename))
```

**Expected behavior**
File transfer should work for all the directories as well as for individual files.

PR #1871, opened 2023-03-23
Fix bug to handle file pattern or directory

sunank200:
# Description

As part of this [PR 1829](https://github.com/astronomer/astro-sdk/pull/1829), a bug was introduced. This is not handling the case when the File path is a file pattern or directory.
Screenshot 2023-03-24 at 12 23 45 AM

In case the destination dataset is a folder it doesn't pass the filename from the source at all here: https://github.com/astronomer/astro-sdk/blame/1b35bb9443d57ef7834259a3cd3c26695ea5fcf8/universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py#L131 instead of https://github.com/astronomer/astro-sdk/blame/bf189de90e4ad9f2db77c7ac46a7193bde9a9c92/universal_transfer_operator/src/universal_transfer_operator/data_providers/filesystem/base.py#L123

closes: #1870

## What is the new behavior?

- It should check if the provided file is a directory or file pattern and then add the actual file name from the source dataset from `FileStream`

```
destination_file = self.dataset.path
# check if destination dataset is folder or file pattern
if self.dataset.is_pattern():
    destination_file = os.path.join(self.dataset.path, os.path.basename(source_ref.actual_filename))
```
- Add the test

Screenshot 2023-03-24 at 12 44 01 AM

No

### Checklist


phanikumv (approved):

universal_transfer_operator/tests/conftest.py
```diff
@@ -157,26 +162,47 @@ def populate_file(src_file_path: str, dataset_provider: BaseFilesystemProviders,
...
+            stream.write(src_file_object.read())
+            stream.flush()
+
+
+def is_dir(path: str) -> bool:
+    """Check if a given path is dir or an absolute path to a file. This is different from os.is_dir()
+    because we just look at the str
+    """
+    end_str: str = path.split("/")[-1]
+    return end_str.find(".") == -1
```
sunank200, line 202:
```suggestion
     return not pathlib.PosixPath(path).suffix
```

utkarsharma2 (approved):

PR #1879, opened 2023-03-31
Allow `aql.transform` to receive `sql` filepath (fix breaking change)

tatiana:
Fix `aql.transform` breaking change (pass a `.sql` file to the `sql` argument).

This PR contains:
1) Test case which illustrates data team's usage of `aql.transform`, which raises an exception in 1.5.3 (and used to work in 1.3.3)
2) Bug fix that allows users to continue using the Python SDK as they used with Python SDK 1.3.3.

Astronomer's Data team raised this issue in an internal slack: https://astronomer.slack.com/archives/C02B.8SPT93K/p1680202480707669

The alternative to this solution is to use `aql.transform_file`. However, since this feature was working before and stopped working between minor releases, we opted to fix it as well.


utkarsharma2:
@tatiana IMO, we should be backward compatible and should only introduce a breaking changes in 2.0.

utkarsharma2 (approved):
LGTM!

PR #1893, opened 2023-04-10
testing new release of azure 5.3.1rc1

utkarsharma2:
testing new release of azure 5.3.1rc1

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -77,7 +77,7 @@ amazon = [
     "smart-open[s3]>=5.2.1",
 ]
 azure = [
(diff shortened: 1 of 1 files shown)


pankajkoti (commented):
Looks like the tests are passing well. But wanted to discuss that maybe we could add all RCs in a single PR as well, no? Like that we need to create and close only 1 PR instead of maintaining multiple PRs and running multiple CI jobs.

utkarsharma2:
@pankajkoti  We can, but it's much easier to isolate issues(if any) in separate PR. IMO

pankajkoti:
Should we close the PR @utkarsharma2 ?

Issue #1820, opened 2023-03-03
Support for query tags


dimberman:
**Please describe the feature you'd like to see**
Snowflake users use query tags to add necessary lineage and governance data to queries. Currently there’s no easy way to use query tags in the astro SDK except to create a unique connection per query-tag. A customer told me that to do this via connections he would need to create ~10 unique connections per team which is not sustainable. 

**Describe the solution you'd like**
We should implement a `query_tags` parameter to the `aql.dataframe`, `aql.transform`, and `aql.run_raw_sql` functions.

```
@aql.dataframe(conn_id='foo', query_tags=['fizz', 'buzz'])
def foo...
```

We can also make this a runtime argument

```
@aql.dataframe(conn_id='foo')
def foo...

with dag:
    foo(query_tags=['fizz','buzz'])
```

**Are there any alternatives to this feature?**
As mentioned the query tag alternative is to create connections for each set of tags.

**Additional context**
Add any other context about the feature request here.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

tatiana:
@dimberman I think having tags related to queries is a useful feature. Given that the Python SDK tries to expose generic interface cross-databases, do the other databases supported by the Python SDK have a similar concept to `query_tags`?

If more databases have this concept, I believe the proposed interface would be fine (BTW: why would we change `dataframe` and not `transform_file`?)

If Snowflake is the only currently supported database with this concept, it may be worth exposing it in another way.

PR #1898, opened 2023-04-12
Add a Session class that will allow users to create pre-queries and post-queries

dimberman:
# Description
Currently there is no way to include "pre-queries" and "post-queries" to queries generated by transform and raw_sql function. This issue is notably an issue in https://github.com/astronomer/astro-sdk/issues/1820 where the user needs to create a new connection every time they want to add a query tag to a query. This creates a lot of clutter as a single team can need 10 unique connections dependent on the query tags needed.

closes: https://github.com/astronomer/astro-sdk/issues/1820

## What is the new behavior?
The new behavior is to open up "pre_queries" and "post_queries" to the user. This will allow users to generalize a single "Session" object that can be passed to multiple functions without cluttering the returned SQL statement. This session object can also be modified based on Airflow variables.

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/session.py
+++ b/python-sdk/src/astro/session.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+from airflow.exceptions import AirflowException
+from sqlalchemy.sql import ClauseElement
+
+
+class Session:
+    """
+    A series of options that will allow users to determine session variables.
+    These session variables will relate to queries
+    that can happen before or after the query is run.
+    """
+
+    def __init__(self, pre_queries: list[str] | None = None, post_queries: list[str] | None = None):
+        self.pre_queries = pre_queries or []
+        self.post_queries = post_queries or []
+
+    def merge_pre_and_post_queries(self, sql: str | ClauseElement):
+        if isinstance(sql, str):
+            return ";".join(self.pre_queries + [sql] + self.post_queries)
+        elif isinstance(sql, ClauseElement):
+            if self.pre_queries or self.post_queries:
+                raise AirflowException(
+                    "pre-and-post queries are currently only supported for "
+                    "string SQL statements."
+                    "Please turn your SQL statement into a text statement."
+                )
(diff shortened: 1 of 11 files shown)


python-sdk/src/astro/session.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+from airflow.exceptions import AirflowException
+from sqlalchemy.sql import ClauseElement
+
+
+class Session:
```
utkarsharma2, line 7: @dimberman I think the name is a little generic, can we change it to something that indicates the SQL link?

utkarsharma2:
This looks like a cool feature! :) maybe we can document it as well?

python-sdk/tests/sql/operators/test_run_raw_sql.py
```diff
@@ -56,6 +56,27 @@ def dummy_method():
     results_as_pandas_dataframe.assert_called_with([])
 
 
+@mock.patch("astro.databases.base.BaseDatabase.connection")
+def test_run_sql_calls_with_query_tag(run_sql, sample_dag):
```
utkarsharma2, line 60: I think we can just add a unit testcase here for `Session` class. WDYT?

dimberman:
@utkarsharma2 yes there will be documentation, sorry for some reason thought this was a draft, set it back to draft :)

tatiana:
@dimberman I remember when we discussed the query tag problem, that we agreed to have the session modifier and have a `SnowflakeSession` modifier that would expose query tags.

Why did the approach change?

It feels that if users just need to prefix or suffix statements to their queries, they could manage this with a decorator function on their end, without the need of us implementing this feature in the SDK. Wouldn't this work?

tatiana (approved):
I discussed with @jlaneve and we agreed the following:
- to rename this argument to be query_modifier
- to rename the class to be QueryModifier

Once these changes and the following is done:
- once the tests and checks are passing
- test coverage is kept

I'm happy for the PR to be merged. To avoid delays due to timezones difference, I'm pre-approving it.

dimberman:
Thank you @tatiana ,

yeah discussing with @jlaneve the reason we went in this direction is to ultimately give users the ability to have flexibility. Rather than us needing to support a bunch of different query modifiers, we just give users to create arbitrary pre-and-post queries. I've made the changes requested and will merge.

Issue #812, opened 2022-09-09
Easier access to the Airflow context


rossturk:
**Please describe the feature you'd like to see**

For getting Airflow context, `get_current_context()` is supported, and it seems to work fine. However, I had to search through the Airflow docs to find it and the resulting text is a bit verbose:

```python
@aql.dataframe
def do_stuff():
    context = get_current_context()
    data_interval_start = context["data_interval_start"]
    data_interval_end = context["data_interval_end"]
	  ...
```

I think this is something that pipeline builders will need to do on every pipeline that can backfill, which should be a lot of 'em.

**Describe the solution you'd like**

Could context be available via `kwargs` like they are in TaskFlow?

```python
@aql.dataframe
def do_stuff(data_interval_start=None,data_interval_end=None):
    ...
```

**Are there any alternatives to this feature?**

We could document `get_current_context()` as the correct way to do this in the Astro SDK docs.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

utkarsharma2:
@feluelle ^^

utkarsharma2:
@rossturk Thanks for creating this ticket, we'll take a look at it. :)

phanikumv:
@kaxil any thoughts on how critical this is ? Considering the backlog we have at the moment, I dont think we can work on this atleast till next sprint

kaxil:
@phanikumv - low

PR #1900, opened 2023-04-19
Add the ability to access context variables via function parameters in sql and pandas functions

dimberman:

Currently, users cannot access Airflow context variables directly via function parameters. 
To access context variables, users must rely on the `get_current_context()` method.

CLOSES: #812 
## What is the new behavior?

This PR enables users to access Airflow context variables directly via function parameters, making it more consistent with the Taskflow API. Users can now pass context variables as function parameters, as shown in the example below:

```python
@aql.dataframe
def validate_file_and_context(df: pandas.DataFrame, ds=None, execution_date=None):  # skipcq: PY-D0003
    context = get_current_context()
    assert ds == context["ds"]
    assert execution_date == context["execution_date"]
```

No

### Checklist

### DIFF
--- a/python-sdk/src/astro/sql/operators/base_decorator.py
+++ b/python-sdk/src/astro/sql/operators/base_decorator.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import Any, Callable, Sequence, cast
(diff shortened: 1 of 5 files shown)


tatiana:
@dimberman this is a great feature, what are we missing to take it from the draft to be ready for code review?

Issue #1911, opened 2023-04-28
Unable to use dynamic task mapping to export list of dataframes


tatiana:
**Describe the bug**
Astro SDK raises an exception when trying to export a list of dataframes.

**Version**
* Astro: 1.5.3

**To Reproduce**

1. Inside the `python-sdk` folder, run:

```
make local target=setup 
source ~/.virtualenvs/astro-sdk/bin/activate
AIRFLOW_HOME=`pwd` airflow db init 
```

2. Write the  following DAG to the file `calculate_popular_movies_by_genre.py`, inside the `example_dags` folder:

```
from datetime import datetime

import pandas as pd
from airflow import DAG

from astro import sql as aql
from astro.files import File
from astro.table import Table
from astro.sql.operators.export_to_file import ExportToFileOperator

@aql.dataframe(columns_names_capitalization="original")
def top_movies_by_genre(input_df: pd.DataFrame):
    top_movies = []
    for genre, genre_df in input_df.groupby("genre1"):
        genre_df.sort_values(by="rating", ascending=False)[["title", "rating", "genre1"]].head(5)
        top_movies.append(genre_df)
    return top_movies


with DAG(
    "calculate_popular_movies_by_genre",
    schedule_interval=None,
    start_date=datetime(2022, 1, 1),
    catchup=False,
) as dag:
    imdb_movies = aql.load_file(
        File("https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb_v2.csv"),
        output_table=Table(conn_id="sqlite_default"),
    )
    movies_by_genre = top_movies_by_genre(input_df=imdb_movies)
    ExportToFileOperator.partial(output_file=File("/tmp/output.json"), task_id="export_movies").expand(input_data=movies_by_genre)
```

3. Try to run the test:

```
AIRFLOW_HOME=`pwd` pytest 'tests_integration/test_example_dags.py::test_example_dag[calculate_popular_movies_by_genre]'
```

4. See the error:

```
tests_integration/test_example_dags.py:31: in wrapper_run_dag
    test_utils.run_dag(dag)
tests_integration/sql/operators/utils.py:69: in run_dag
    return test_dag(dag=dag)
../../../.virtualenvs/astro-sdk/lib/python3.9/site-packages/airflow/utils/session.py:75: in wrapper
    return func(*args, session=session, **kwargs)
tests_integration/sql/operators/utils.py:142: in test_dag
    _run_task(ti, session=session)
[...]

PR #1912, opened 2023-04-28
Add an example DAG for using dynamic task with dataframe

tatiana:
Add a test case for issue #1911 

Disclaimer: this test should fail until the issue is fixed in the upstream Airflow: https://github.com/apache/airflow/pull/30943


sql-cli/CONTRIBUTING.md
```diff
@@ -52,7 +52,6 @@ Internally, what happens is that we build a command that we pass on to a docker
 Additionally, the `flow` command requires some more functionality like mounting files from the host to the container to be able to use the created sql-cli project outside of the container after the command exits. Furthermore, we resolve directory paths before mounting so that we have the exact same directory structure outside and inside the container.
 _You can find the code for that in `sql` directory._
 
-_You can find the user documentation of the "flow" command [here](https://docs.astronomer.io/astro/cli/sql-cli)._
```
pankajkoti, line 55: I guess we no longer need this here considering https://github.com/astronomer/astro-sdk/pull/1917

sql-cli/README.md
```diff
@@ -2,7 +2,6 @@
 
 Empower analysts to build workflows to transform data using SQL.
 
-Find out more in the [docs](https://docs.astronomer.io/astro/cli/sql-cli).
```
pankajkoti, line 5: same suggestion here WRT https://github.com/astronomer/astro-sdk/pull/1917

python-sdk/tests_integration/test_example_dags.py
```diff
@@ -45,6 +45,7 @@ def session():
 MIN_VER_DAG_FILE: dict[str, list[str]] = {
     "2.3": ["example_dynamic_task_template.py", "example_bigquery_dynamic_map_task.py"],
     "2.4": ["example_datasets.py"],
+    "2.6.1": ["calculate_popular_movies_by_genre.py"],
```
pankajkoti, line 48: I am not sure what's the best way, but we might need to reach out to the release manager to request them to tag the release for the PR and include https://github.com/apache/airflow/pull/30943 in `2.6.1`. WDYT?

pankajkoti (approved):
Looks good to me. Would be nice if we could also get eyes from @utkarsharma2 , @pankajastro or @sunank200

python-sdk/tests_integration/test_example_dags.py
pankajkoti: Usually, I see Ephraim helping and taking care of releases. Might make sense to check with him once.

python-sdk/example_dags/calculate_popular_movies_by_genre.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+with DAG(
+    "calculate_popular_movies_by_genre",
+    schedule_interval=None,
+    start_date=datetime(2022, 1, 1),
+    catchup=False,
+) as dag:
+    imdb_movies = aql.load_file(
+        File("https://raw.githubusercontent.com/astronomer/astro-sdk/main/tests/data/imdb_v2.csv"),
```
sunank200, line 29: We can use S3 URL instead, as HTTP URL sometimes makes the tests flaky.

sunank200 (approved):
Overall LGTM. Added a comment

python-sdk/example_dags/calculate_popular_movies_by_genre.py
tatiana: @sunank200, could you point out when the HTTP tests were flaky? This is the first time I have heard about this! HTTP tests are useful because users can run these DAGs without additional providers or permissions.

python-sdk/tests_integration/test_example_dags.py
tatiana: Thanks a lot, @pankajkoti , he confirmed it will be included on 2.6.1 🎉

pankajastro (approved):

PR #1914, opened 2023-05-02
Update tests and docs regarding availability of the argument query_modifier

tatiana:
Although we `transform` and `transform_file` supported the `query_modifier` argument, we didn't have tests nor documentation describing this. This PR adds tests and docs for these use cases.


pankajkoti (approved):
Looks great to me! 

Should we also mention somewhere that the argument for after queries is called `post_queries` or it's reflected in some other docs already?

tatiana:
> Looks great to me!
> 
> Should we also mention somewhere that the argument for after queries is called `post_queries` or it's reflected in some other docs already?

@pankajkoti I was about to document `post_queries` and decided to review the feature code again. We'll need to review the implementation in the next iteration, and I suggest we don't document it for now.

 I think there are two issues with the current feature:

1.  The statements are not executed as part of the same transaction, which may mislead users, particularly if they want to use `post_queries` for audit purposes, for instance.

2. If the user uses `post_queries`, that would potentially affect the expected output. Since it is run together with the main SQL statement, the `post_queries` query result would override any query result from the main SQL statement, changing the originally expected behaviour of `transform`, `transform_file` or `run_raw_sql`.

Since a customer is waiting on the support for `query tags`, I suggest we don't document `post_queries` and remove this feature altogether in the next release. This was not a requirement for the `query_tags` and can lead to unexpected behaviours for the user.

@jlaneve please, let me know if you'd like to discuss this further.

PR #1925, opened 2023-05-05
Refactor `assume_schema_exists` and expose it in `aql.transform`

tatiana:
This is a follow-up for #1922. In that PR we allowed users to skip schema check & creation for `aql.load_file`, but we missed the fact that `aql.transform` and `aql.transform_file` had the same issue. This PR aims to address this limitation.

Changes included in this PR:
* Rename config `load_table_schema_exists` to `assume_schema_exists`
* Rename (`load_file`) argument `schema_exists` to `assume_schema_exists`
* Refactor where the check for `assume_schema_exists` happens. Before, it happened only inside the `load_file_to_table`. Now, it is part of `create_schema_if_applicable`. This makes this feature available in the `aql.transform` task as well
* Rename `Database.create_schema_if_needed` to `Database.create_schema_if_applicable`
* Expose `assume_schema_exists` in `aql.transform`
* Release 1.7.0a2

### DIFF
--- a/python-sdk/docs/configurations.rst
+++ b/python-sdk/docs/configurations.rst
@@ -48,11 +48,11 @@ or by updating Airflow's configuration
 Configuring if schemas existence should be checked and if the SDK should create them
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By default, during ``aql.load_file``, the SDK checks if the schema of the target table exists, and if not, it tries to create it. This type of check can be costly.
+By default, during ``aql.load_file`` and ``aql.transform``, the SDK checks if the schema of the target table exists, and if not, it tries to create it. This type of check can be costly.
 
-The configuration ``AIRFLOW__ASTRO_SDK__LOAD_TABLE_SCHEMA_EXISTS`` allows users to inform the SDK that the schema already exists, skipping this check for all ``load_file`` tasks.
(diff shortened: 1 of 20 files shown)


python-sdk/docs/configurations.rst
```diff
@@ -48,11 +48,11 @@ or by updating Airflow's configuration
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By default, during ``aql.load_file``, the SDK checks if the schema of the target table exists, and if not, it tries to create it. This type of check can be costly.
+By default, during ``aql.load_file`` and ``aql.transform``, the SDK checks if the schema of the target table exists, and if not, it tries to create it. This type of check can be costly.
 
-The configuration ``AIRFLOW__ASTRO_SDK__LOAD_TABLE_SCHEMA_EXISTS`` allows users to inform the SDK that the schema already exists, skipping this check for all ``load_file`` tasks.
+The configuration ``AIRFLOW__ASTRO_SDK__ASSUME_SCHEMA_EXISTS`` allows users to inform the SDK that the schema already exists, skipping this check for all ``load_file`` and ``transform`` tasks.
 
-The user can also have a more granular control, by defining the ``load_file`` argument ``schema_exists`` on a per-task basis :ref:load_file.
+The user can also have a more granular control, by defining the ``load_file`` argument ``assume_schema_exists`` on a per-task basis :ref:load_file.
```
kaxil, line 55: You will need to change L68 and L61 too

also instead of ``assume_schema_exists``, I would suggest `schema_exists_check`

tatiana: Thanks, @kaxil , I updated the parts of the doc which needed to be updated!

Calling it `schema_exists_check` may be misleading because we're not only checking if it exists but also creating the schema.

kaxil: makes sense

kaxil (approved):

PR #1932, opened 2023-05-10
Add `mysql` dependences  in dev Dockerfile

phanikumv:
The command `make container target=build-run` is failing due to below error on local dev environment.
This PR resolves it by adding the dependency to the dev Dockerfile

#0 27.87         File "/tmp/pip-install-dbzietu6/mysqlclient_5cc72f3ff0094519a046c7f1ebeebee2/setup_posix.py", line 31, in mysql_config
#0 27.87           raise OSError("{} not found".format(_mysql_config_path))
#0 27.87       OSError: mysql_config not found

### DIFF
--- a/python-sdk/dev/Dockerfile
+++ b/python-sdk/dev/Dockerfile
@@ -10,7 +10,8 @@ RUN apt-get install -y --no-install-recommends \
         libsasl2-modules \
         freetds-dev \
         libssl-dev \
-        libkrb5-dev
+        libkrb5-dev \
+        libmariadb-dev
 
 ENV SETUPTOOLS_USE_DISTUTILS=stdlib
 


python-sdk/dev/Dockerfile
```diff
@@ -10,7 +10,8 @@ RUN apt-get install -y --no-install-recommends \
         libsasl2-modules \
         freetds-dev \
         libssl-dev \
-        libkrb5-dev
+        libkrb5-dev \
+        libmariadb-dev
```
tatiana, line 14: @phanikumv thanks a lot for looking into this! 
What is the benefit of installing the maria DB package as opposed to `libmysqlclient-dev`?

phanikumv: libmysqlclient-dev is not available, hence used the alternative.


```
#0 0.536 Package libmysqlclient-dev is not available, but is referred to by another package.
#0 0.536 This may mean that the package is missing, has been obsoleted, or
#0 0.536 is only available from another source
#0 0.536 However the following packages replace it:
#0 0.536   libmariadb-dev-compat libmariadb-dev
#0 0.536 
#0 0.583 E: Package 'libmysqlclient-dev' has no installation candidate

```

tatiana (approved):
Thanks for fixing this, @phanikumv !

PR #1933, opened 2023-05-10
Fix deploy script

utkarsharma2:
# Description
CI deploy job was failing - https://github.com/astronomer/astro-sdk/pull/1923

## What is the new behavior?
Fixed it in this PR and moved the Astro-deploy job out into its own yml file.
ref - https://github.com/astronomer/astro-sdk/actions/runs/4937329398/jobs/8825922048?pr=1933

Nope

### Checklist


python-sdk/tests_integration/astro_deploy/Dockerfile
```diff
@@ -1,5 +1,18 @@
 FROM quay.io/astronomer/astro-runtime:7.4.2-base
```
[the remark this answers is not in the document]
utkarsharma2: We can I don't see it as a problem we have done testing with Airflow-2.6.0, but it should be a separate PR IMO.

python-sdk/tests_integration/astro_deploy/Dockerfile.single_worker
```diff
@@ -1,5 +1,19 @@
 FROM quay.io/astronomer/astro-runtime:7.3.0-base
```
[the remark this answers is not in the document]
utkarsharma2: Answered here - https://github.com/astronomer/astro-sdk/pull/1933#discussion_r1189901463

python-sdk/tests_integration/astro_deploy/Dockerfile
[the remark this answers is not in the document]
utkarsharma2: @phanikumv @tatiana thoughts?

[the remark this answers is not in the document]
pankajastro: separate PR should be ok IMO

.github/workflows/ci-benchmark.yaml
```diff
@@ -65,20 +65,3 @@ jobs:
           ( echo "cat <<EOF >python-sdk/test-connections.yaml"; cat .github/ci-test-connections.yaml; ) >python-sdk/test-connections.yaml && . python-sdk/test-connections.yaml
           python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
           cd python-sdk/tests/benchmark && make
-
-  Astro-deploy:
```
pankajastro, line 69: can we remove `- cron: '0 0 * * *'` line 8??

pankajastro: also on L18 do we need `github.event.schedule != '0 0 * * *'`  check now?

python-sdk/tests_integration/astro_deploy/deploy.sh
```diff
@@ -66,6 +66,11 @@ function deploy(){
 
     # Build image and deploy
     BUILD_NUMBER=$(awk 'BEGIN {srand(); print srand()}')
+
+    # Enforce registry name to be in lowercase
+    docker_registry_astro=$(echo $docker_registry_astro | tr '[:upper:]' '[:lower:]')
```
pankajastro, line 71: do we need to enforce this from our side?

.github/workflows/ci-astro-deploy.yml
```diff
@@ -0,0 +1,20 @@
...
+      ASTRO_DEPLOYMENT_ID:  ${{ secrets.ASTRO_DEPLOYMENT_ID }}
+      ASTRO_KEY_ID: ${{ secrets.ASTRO_KEY_ID }}
+      ASTRO_KEY_SECRET: ${{ secrets.ASTRO_KEY_SECRET }}
+      ASTRO_DEPLOYMENT_ID_SINGLE_WORKER: ${{ secrets.ASTRO_DEPLOYMENT_ID_SINGLE_WORKER }}
+      ASTRO_KEY_ID_SINGLE_WORKER: ${{ secrets.ASTRO_KEY_ID_SINGLE_WORKER }}
+      ASTRO_KEY_SECRET_SINGLE_WORKER: ${{ secrets.ASTRO_KEY_SECRET_SINGLE_WORKER }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: cd python-sdk/tests_integration/astro_deploy && /bin/bash deploy.sh $ASTRO_DOCKER_REGISTRY $ASTRO_ORGANIZATION_ID $ASTRO_DEPLOYMENT_ID $ASTRO_KEY_ID $ASTRO_KEY_SECRET $ASTRO_DEPLOYMENT_ID_SINGLE_WORKER $ASTRO_KEY_ID_SINGLE_WORKER $ASTRO_KEY_SECRET_SINGLE_WORKER
```
pankajastro, line 20: can we do `bash deploy.sh  ...` instead of `/bin/bash deploy.sh`?

.github/workflows/ci-astro-deploy.yml
```diff
@@ -0,0 +1,20 @@
+name: "Astro Deploy"
```
tatiana, line 1: How could we give more visibility if this job fails? Would it be an option to notify Slack..?

Could it also be one of the steps of the default CI pipeline so that we ensure things are working and see them failing when we make changes to the SDK?

.github/workflows/ci-astro-deploy.yml
utkarsharma2: okay, I'll change.

utkarsharma2: done.

python-sdk/tests_integration/astro_deploy/deploy.sh
pankajastro: ok, I was thinking that docker_registry_astro, deployment_id, deployment_id should have been generated in the right format in Astro, if we are changing to a lowercase here by ourselves then there might be the value which we are setting is not correct. But if we are able to deploy to both (multi/single worker) instances with these changes I'm ok with it. Do we have some job which has deployed it successfully??

pankajastro: or if have tested locally, that should also be fine?

.github/workflows/ci-astro-deploy.yml
pankajastro: separate PR?? maybe we should create a GH issue

pankajastro (approved):

PR #1947, opened 2023-05-25
Release 1.6.1

tatiana:


utkarsharma2 (approved):
LGTM

phanikumv (commented):
@tatiana what about the compatibility with redshift on latest amazon provider- i think we need to fix that as well as part of this release

PR #1952, opened 2023-06-02
Automate Provider RC - Add dependencies replacing script

utkarsharma2:
Add a script to replace dependencies in pyproject.toml file.


pankajastro (approved):

python-sdk/dev/scripts/replace_dependencies.py
```diff
@@ -0,0 +1,110 @@
...
+    return h2_soup.a.get("href")
+
+
+def parse_providers_release_testing_gh_issue(issue_url: str) -> list[str]:
+    """Parse the pinned packages from The URL of the github issue that announce provider testing
+    (e.g., https://github.com/apache/airflow/issues/31322)
+    :param issue_url: The URL of the github issue that announce provider testing
+    """
+    req = requests.get(issue_url)
+    soup = BeautifulSoup(req.text, "html.parser")
```
tatiana, line 58: Did we consider the possibility of using the Gihtub HTTP API to retrieve the information we need?

tatiana (commented):
@utkarsharma2 I know this PR was already merged, and I was a bit late, but I left some comments. Whenever I see HTML parsing in this type of code, I feel a bit worried - because if either Github or PyPI change their HTML markup, our logic will be broken. Have we considered using their APIs (more comments in line)?

PR #1957, opened 2023-06-06
Sftp ftp ec2

utkarsharma2:
This PR removed dependency from the old SFTP and FTP servers(probably locally hosted)
1. Start SFTP and FTP servers in aws ec2 instance
2. Create connections based on the servers IP address

### DIFF
--- a/python-sdk/tests_integration/astro_deploy/master_dag.py
+++ b/python-sdk/tests_integration/astro_deploy/master_dag.py
@@ -149,7 +149,7 @@ def create_sftp_ftp_airflow_connection(task_instance: Any) -> None:
     sftp_conn = Connection(
         conn_id="sftp_default",
         conn_type="sftp",
-        host=task_instance.xcom_pull(key=INSTANCE_PUBLIC_IP, task_ids=["create_ec2_instance"]),
+        host=task_instance.xcom_pull(key=INSTANCE_PUBLIC_IP, task_ids=["start_sftp_ftp_services"]),
         login=SFTP_USERNAME,
         password=SFTP_PASSWORD
     )  # create a connection object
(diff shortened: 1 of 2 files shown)


python-sdk/dev/scripts/test.py
tatiana, line 1: What is this file..?

python-sdk/tests_integration/astro_deploy/master_dag.py
```diff
@@ -94,6 +107,92 @@ def prepare_dag_dependency(task_info, execution_time):
...
+        if connection is None:
+            logging.info("Connection %s doesn't exist.", str(conn.conn_id))
+        else:
+            session.delete(connection)
+            session.commit()
+            logging.info("Connection %s deleted.", str(conn.conn_id))
+
+        session.add(conn)
+        session.commit()  # it will insert the connection object programmatically.
+        logging.info(f"Connection {conn.conn_id} is created")
```
tatiana, line 178: Why are we managing the creation of this connection differently when compared with all the other connections we use?

python-sdk/dev/scripts/test.py
utkarsharma2: Removed it, as it was not required.

python-sdk/tests_integration/astro_deploy/master_dag.py
utkarsharma2: This is done to because we want SFTP/FTP servers while running master dag in astro cloud. Unlike Github actions, we don't have access to services in the local container. 

Also, we are creating ec2 instances on the fly which don't have a static IP, because of which we need to get the host dynamically and create the connections.

python-sdk/tests_integration/astro_deploy/master_dag.py
```diff
@@ -237,3 +358,4 @@ def prepare_dag_dependency(task_info, execution_time):
 
     last_task >> end  # skipcq PYL-W0104
     last_task >> report  # skipcq PYL-W0104
+    last_task >> terminate_ec2_instance
```
utkarsharma2, line 361:
```suggestion
    last_task >> terminate_ec2_instance  # skipcq PYL-W0104
```

PR #1959, opened 2023-06-07
Dummy change

utkarsharma2:
# Description
## What is the new behavior?

-
-
-

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1962, opened 2023-06-08
Fix Snowflake QueryModifier issue

tatiana:
Add a DAG to illustrate how users can set `query_tags` in Snowflake using the Python SDK.

Before this PR, it would fail with:
```
E       sqlalchemy.exc.ProgrammingError: (snowflake.connector.errors.ProgrammingError) 000008 (0A000): 01acd6d5-0607-a92d-0000-68213eeeebda: Actual statement count 2 did not match the desired statement count 1.
E       [SQL: ALTER SESSION SET query_tag='not_guinea_pig';;CREATE TABLE IF NOT EXISTS SANDBOX.ASTROFLOW_CI._tmp_ksimu2ab9s9kbatbtrexodw6q9alcskhfcumxpxuvi2ir60ew81a3st3k AS SELECT *
E           FROM IDENTIFIER(%(input_table)s) WHERE type NOT LIKE 'Guinea Pig'
E           ]
E       [parameters: ***'input_table': 'SANDBOX.ASTROFLOW_CI._tmp_tddzk0vrcf3lbt1nxucnv3ejyx1ds3su84fvdqpoa4p69yguysppti2mo'***]
E       (Background on this error at: https://sqlalche.me/e/14/f405)
```

Note: the failing Redshift tests are unrelated to the change introduced in this PR. They are being investigated by PR #1959 . So far, they don't happen when we use a newer version of Redshift, they don't happen, but the costs are higher. Therefore, I suggest we open an exception and merge this PR disregarding them.

### DIFF
--- a/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
+++ b/python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
@@ -10,6 +10,7 @@ from airflow.decorators import dag
 from astro import sql as aql
 from astro.files import File
 from astro.options import SnowflakeLoadOptions
+from astro.query_modifier import QueryModifier
 from astro.table import Metadata, Table
 
 
@@ -19,7 +20,10 @@ def combine_data(center_1: Table, center_2: Table):
     UNION SELECT * FROM {{center_2}}"""
 
 
-@aql.transform(assume_schema_exists=True)
+@aql.transform(
+    assume_schema_exists=True,
+    query_modifier=QueryModifier(pre_queries=["ALTER SESSION SET query_tag='not_guinea_pig';"]),
+)
 def clean_data(input_table: Table):
     return """SELECT *
     FROM {{input_table}} WHERE type NOT LIKE 'Guinea Pig'
--- a/python-sdk/src/astro/databases/base.py
+++ b/python-sdk/src/astro/databases/base.py
@@ -103,6 +103,30 @@ class BaseDatabase(ABC):
         """Return Sqlalchemy engine."""
         return self.hook.get_sqlalchemy_engine()  # type: ignore[no-any-return]
 
+    def run_single_sql_query(
+        self,
+        sql: str | ClauseElement = "",
+        parameters: dict | None = None,
+    ) -> Any:
+        """
+        Return the results to running a SQL statement.
+
+        Whenever possible, this method should be implemented using Airflow Hooks,
+        since this will simplify the integration with Async operators.
+
+        :param sql: Contains SQL query to be run against database
+        :param parameters: Optional parameters to be used to render the query
+        """
+        # We need to autocommit=True to make sure the query runs. This is done exclusively for SnowflakeDatabase's
+        # truncate method to reflect changes.
+        if isinstance(sql, str):
+            result = self.connection.execute(
+                sqlalchemy.text(sql).execution_options(autocommit=True), parameters
+            )
+        else:
+            result = self.connection.execute(sql, parameters)
+        return result
+
     def run_sql(
         self,
         sql: str | ClauseElement = "",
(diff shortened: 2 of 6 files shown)


jlaneve (approved):

python-sdk/src/astro/databases/base.py
```diff
@@ -134,15 +158,15 @@ def run_sql(
...
-        # truncate method to reflect changes.
-        if isinstance(sql, str):
-            result = self.connection.execute(
-                sqlalchemy.text(sql).execution_options(autocommit=True), parameters
-            )
-        else:
-            result = self.connection.execute(sql, parameters)
+
+        for sql_query in query_modifier.pre_queries:
+            _ = self.run_single_sql_query(sql_query, parameters)
```
utkarsharma2, line 163: @tatiana AFAIK query tags can also be used to track the usage of a particular query. Executing them separately will make the tracking difficult right?

utkarsharma2: ref - https://docs.vividcortex.com/general-reference/query-tags/

tatiana: @utkarsharma2 That's a very good question. Our query tags support was designed focused on Snowflake, and the expectation is that the users would give `ALTER SESSION` statements which would be executed before the main `transform` / `run_raw_query`.

If we only had one `ALTER SESSION` statement and ran the query of interest, this approach would be safe since it would be only for that broader query. However, since ATM, the user can potentially give multiple queries to be run before, and after the main statement, all of them would be labelled the same way, which can be prone to errors.

We can also look into how other databases support query tags - I wonder if it already works for Postgres and MySQL since they rely only on comments. I'll try it out.

For now, I suggest we focus on the customer issue of running the feature as it was originally designed - and review the overall approach in a separate PR/ticket, how do you feel about this?

pankajastro: I was thinking if it is worth mentioning in the docs string how we run the query and if you have multiple queries then how metrics will be impacted wdyt?

pankajastro (approved):
LGTM

python-sdk/src/astro/databases/base.py
utkarsharma2: @tatiana I understand to the naming is misleading since `QueryModifier` means something else generally. Would it make sense to change it or isolate it Snowflake?

I won't consider this as a blocker since it's a pressing issue.

pankajastro (approved):

python-sdk/example_dags/example_amazon_s3_snowflake_transform.py
```diff
@@ -19,7 +20,10 @@ def combine_data(center_1: Table, center_2: Table):
     UNION SELECT * FROM {{center_2}}"""
 
 
-@aql.transform(assume_schema_exists=True)
+@aql.transform(
+    assume_schema_exists=True,
+    query_modifier=QueryModifier(pre_queries=["ALTER SESSION SET query_tag='not_guinea_pig';"]),
```
pankajastro, line 25: @tatiana We ran this DAG on Astro but like are still getting this error 
```
Actual statement count 2 did not match the desired statement count 1.
```

tatiana: @pankajastro, as we spoke in the call, the issue seems that the Astro CI is pointing to an old version of the Python SDK. The stack trace had:
```
  File "/home/astro/.local/lib/python3.10/site-packages/astro/databases/base.py", line 141, in run_sql
    result = self.connection.execute(
```

Which does not match this version (or the main branch):
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/src/astro/databases/base.py#L106-L128

Please, let me know if that's not the issue and we can investigate further#1962

PR #1965, opened 2023-06-12
Add note for redshift issue

utkarsharma2:
Add a note for the redshift issue with the gen2 instance issue.


python-sdk/README.md
```diff
@@ -175,3 +175,6 @@ Contributors and maintainers should abide by the [Contributor Code of Conduct](C
 ## License
 
 [Apache Licence 2.0](LICENSE)
+
+## Notes
+Redshift support is having an issue with dc2.large(gen2) instances. You can still use it with ra3.xlplus(gen 3). We have registered a case with AWS for the same.
```
phanikumv, line 180:
```suggestion
We are currently experiencing issues with dc2.large(gen2) instances on Redshift. We have registered a case with AWS for the same.You can still use Astro SDK with ra3.xlplus(gen 3) instances. 
```

PR #1969, opened 2023-06-16
Add git job for the creating/deleting redshift cluster

utkarsharma2:
Still making sure this is working please don't review :)


utkarsharma2:
Moving this to draft as we are still facing issues for the `dc2` instance created from snapshot and `ra3`.

PR #1972, opened 2023-06-21
Master DAG slack report changes

vatsrahul1001:
# Description
Currently, we have a lot of noise in Slack report, recently for astronomer-providers, we formatted the report to only show failures DAG details and publish the count of Total DAGS, Passed DAG, Failed DAG

Below is the format we updated for providers

image

### DIFF
--- a/python-sdk/tests_integration/astro_deploy/master_dag.py
+++ b/python-sdk/tests_integration/astro_deploy/master_dag.py
@@ -71,7 +71,6 @@ def get_report(dag_run_ids: List[str], **context: Any) -> None:  # noqa: C901
                 task_code = ":black_circle: "
                 if ti.task_id not in ["end", "get_report"]:
                     if ti.state == "success":
-                        if ti.state == "success":
                             continue
                     elif ti.state == "failed":
                         task_code = ":red_circle: "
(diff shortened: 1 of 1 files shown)


python-sdk/tests_integration/astro_deploy/master_dag.py
```diff
@@ -52,30 +52,59 @@ def get_report(dag_run_ids: List[str], **context: Any) -> None:  # noqa: C901
...
             dr_status = f" *{dr.dag_id} : {dr.get_state()}* \n"
-            message_list.append(dr_status)
+            dag_count += 1
+            failed_tasks = []
             for ti in dr.get_task_instances():
                 task_code = ":black_circle: "
                 if ti.task_id not in ["end", "get_report"]:
                     if ti.state == "success":
-                        task_code = ":large_green_circle: "
+                        if ti.state == "success":
```
tatiana, line 74: Why do we have two lines checking for the same thing?

tatiana (commented):
This will be very helpful, @vatsrahul1001 , thanks for improving the report. It will help to spot issues!
A minor comment is inline.

python-sdk/tests_integration/astro_deploy/master_dag.py
vatsrahul1001: @tatiana rightly pointed, will fix this

vatsrahul1001:
@tatiana Now all the test are passed, can we merge this?

tatiana (approved):

PR #1975, opened 2023-06-23
Skip redshift test cases in CI

utkarsharma2:
There is an upstream issue for redshift cluster type - `dc2`
Error:
```
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf0 in position 122: unexpected end of data` Can you please explain why is that the case?
```
ref - https://github.com/astronomer/astro-sdk/actions/runs/5355001227/jobs/9712737921?pr=1974 

Case created with AWS: https://support.console.aws.amazon.com/support/home?region=us-east-2#/case/?displayId=12988936661&language=en

Awaiting response till then we are skipping the tests


tatiana:
Thank you very much, @utkarsharma2 , for looking into this!

tatiana (commented):
@utkarsharma2, do we have a follow-up Github ticket for us to undo these changes once we have a reply from Redshift?

utkarsharma2:
There you go -  https://github.com/astronomer/astro-sdk/issues/1976

tatiana (approved):

Issue #1956, opened 2023-06-04
Add an Excel FileType


skolchin:
**Please describe the feature you'd like to see**
I believe it'd be really nice to add support of native Excel files loading and saving.

**Describe the solution you'd like**
This would require creation of new `ExcelFileType` class (it would be very similar to `CSVFileType`) and some small modifications in base `File` and `FileType` classes. Also, some Excel processing packages (e.g. `openpyxl`) has to be added to dependencies.

**Are there any alternatives to this feature?**
I have implemented an Excel Airflow hook myself (didn't publish it though), but it's kinda old-style and could be nicely replaced with Astro-SDK Files.

**Additional context**
I have a  particular DWH project where a lot of data exchange is going to be through the Excel files generated by some kind of legacy reporting system. Having native Excel support would definetely help in this case

**Acceptance Criteria**

- [x] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

pankajastro:
Hi @skolchin would you like to draft a PR since you already have some implementation, we would love to collaborate and get it done.  Thank you!

skolchin:
Yeah, I would try!

pankajastro:
closing it since https://github.com/astronomer/astro-sdk/pull/1978 has merged

PR #1978, opened 2023-07-02
ExcelFileType class added

skolchin:
New file type class (ExcelFileType) added to support operations on Excel files (see https://github.com/astronomer/astro-sdk/issues/1956)


python-sdk/src/astro/constants.py
```diff
@@ -39,6 +39,7 @@ class FileType(Enum):
     JSON = "json"
     NDJSON = "ndjson"
     PARQUET = "parquet"
+    EXCEL = "xls,xlsx"
```
tatiana, line 42: Since Excel implements [many different formats](https://support.microsoft.com/en-au/office/file-formats-that-are-supported-in-excel-0943ff2c-6014-4e8d-aaea-b83d51d46247), apparently 14 MS-specific file formats to date, it is better to name the file format by the extension, e.g.:

```suggestion
    XLS = "xls"
    XLSX = "xlsx"
```

python-sdk/dev/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN apt-get install -y --no-install-recommends \
         freetds-dev \
         libssl-dev \
         libkrb5-dev \
-        libmariadb-dev
+        libmariadb-dev \
+        pkg-config
```
tatiana, line 15: Why are we introducing this dependency?

tatiana (commented):
Thank you very much for the contribution, @skolchin. It's looking great, including code, docs, and tests.

Minor comments in-line. I just enabled the tests, so we can confirm all the tests pass in the CI. Please, could you rebase this PR based on the latest `main`?

python-sdk/src/astro/constants.py
skolchin: Thanks, but in the case of 2 or more types it would require separate FileType's descendants for each of them, like XlsFileType and XlsxFileType. Currently you could just have all of them in one line - as you might have noticed, get_filetype() now can handle multiple extensions per type.
But I don't mind, if you wish, I'll do it.

python-sdk/dev/Dockerfile
skolchin: I couldn't build a docker without this package, it seems to be a dependency of mariadb: https://stackoverflow.com/questions/76533384/docker-alpine-build-fails-on-mysqlclient-installation-with-error-exception-can

python-sdk/src/astro/constants.py
tatiana: Why would we need different classes (`XlsFileType` and `XlsFileType`)..?

In `types/__init__.py`, we could have two entries pointing to the same class, for instance:
```
filetype_to_class: dict[FileTypeConstants, type[FileType]] = {
        FileTypeConstants.XLS: ExcelFileType,
        FileTypeConstants.XLSX: ExcelFileType
}
```
This way, if we decide to support other Excel formats, let's say `xml` or any new format MS decides to come up with, the user interface wouldn't change.

tatiana: @skolchin, please, if you could address this, we can merge this valuable contribution!

There is no such thing as one EXCEL file type. There are 14 file types. By replacing the generic `EXCEL` with specific ones, we are being future-proof if MS introduces more file formats or if we decide to support other Excel file formats.

python-sdk/tests/test_constants.py
```diff
@@ -7,7 +7,7 @@ def test_supported_file_locations():
 
 
 def test_supported_file_types():
-    expected = {"csv", "json", "ndjson", "parquet"}
+    expected = {"csv", "json", "ndjson", "parquet", "xls,xlsx"}
```
tatiana, line 10: Let's split this into two lines. We designed to have a file extension mapping 1:1 a `FileType` - this could break other parts of the code.

python-sdk/src/astro/constants.py
skolchin: ok, will redo it this way

skolchin: @tatiana done

pankajastro (commented):
LGTM. But look like there is a conflict and would be great if we can add a test here if possible https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/sql/operators/test_load_file.py

skolchin:
> LGTM. But look like there is a conflict and would be great if we can add a test here if possible https://github.com/astronomer/astro-sdk/blob/main/python-sdk/tests/sql/operators/test_load_file.py

Hi, when I run `pytest -s -v -x -k load_file` in `tests\sql`, I'm getting an error `astro.exceptions.NonExistentTableException: The table _tmp_xxx does not exist`. Is this a problem you've mentioned?

pankajastro:
> @pankajastro , when I run `pytest -s -v -x -k load_file` in `tests\sql`, I'm getting an error `astro.exceptions.NonExistentTableException: The table _tmp_xxx does not exist`. Is this a problem you've mentioned?

Hi @skolchin, from log you have shared it is difficult to say why it happening. but test work on the local machine. We have some integration tests that require more setup like credentials etc.

 I would suggest, maybe adding an airflow task here https://github.com/astronomer/astro-sdk/blob/main/python-sdk/example_dags/example_load_file.py would be great. We have everything in CI to run this

```
my_homes_table = aql.load_file(
        task_id="local_to_postgress_glob",
        input_file=File(path=str(CWD.parent) + "/tests/data/example. xlsx*", filetype=FileType. XLSX),
        output_table=Table(
            conn_id=ASTRO_POSTGRESS_CONN_ID,
        ),
    )
```

skolchin:
> @skolchin do let us know if you need any support on this, looks like we are very close to merging it

@pankajastro Hi, sorry, was buzy yesterday. Test added and it seems it works

pankajastro:
> @pankajastro Hi, sorry, was buzy yesterday. Test added and it seems it works

Thanks, @skolchin, I'm looking into CI when it green I'll merge it

pankajastro:
Look like some issue with the Github action.

pankajastro (approved):
LGTM

python-sdk/pyproject.toml
```diff
@@ -87,8 +87,11 @@ ftp = [
 ]
 openlineage = ["openlineage-airflow>=0.17.0"]
 
-databricks = ["databricks-cli",
-    "apache-airflow-providers-databricks"]
+databricks = [
+    "databricks-cli",
+    "apache-airflow-providers-databricks",
+    "databricks-sql-connector<2.9.0",
```
pankajastro, line 93: This temporary requirement since https://github.com/databricks/databricks-sql-python/releases/tag/v2.9.0 is breaking our test. they have PR https://github.com/databricks/databricks-sql-python/pull/191 to fix it but we will have to wait for release

utkarsharma2: Could have been a separate PR, might be easy to revert :)

python-sdk/src/astro/files/types/excel.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+import io
```
utkarsharma2, line 3: Looking good. Just one thing - can we change the dir structure? if it's not too much trouble?
it can be something like:
 python-sdk/src/astro/files/types/excel
python-sdk/src/astro/files/types/excel/base.py
python-sdk/src/astro/files/types/excel/xls.py
python-sdk/src/astro/files/types/excel/xlsx.py

WDYT?

python-sdk/tests_integration/conftest.py
```diff
@@ -301,13 +301,19 @@ def method_map_fixture(method, base_path, classes, get):
...
+    ]
     base_path = ("astro.files.types",)
-    suffix = "FileType"
 
     yield method_map_fixture(
         method=method,
         classes=classes,
         base_path=base_path,
-        get=lambda x: FileType(x.rstrip(suffix).lower()),
+        get=lambda x: FileType(x[0:-8].lower()),  # remove FileType suffix
```
utkarsharma2, line 318:
```
suffix = "FileType"
yield method_map_fixture(
        method=method,
        classes=classes,
        base_path=base_path,
        get=lambda x: FileType(x[0:-len(suffix)].lower()),  # remove FileType suffix
```
Something like this would be better, compared to `0: -8`. WDYT?

python-sdk/src/astro/files/types/excel.py
pankajastro: I thought about it but some test refactoring requires

tatiana (approved):
Thanks for addressing all the feedback, @skolchin ! Looks great!

python-sdk/pyproject.toml
pankajastro: done https://github.com/astronomer/astro-sdk/pull/2013

python-sdk/src/astro/files/types/excel.py
pankajastro: maybe we can do it as separate PR if needed

pankajastro:
This is awesome, thanks @skolchin 👏 for creating the PR. Appreciate it.

PR #1981, opened 2023-07-05
Limit Pydantic < 2.0.0 until Airflow resolves incompatibilities

tatiana:
Before this change, the project's `main` branch tests failed.

Example of error:
```
    module = import_module(module_path)
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/runner/work/astro-sdk/astro-sdk/python-sdk/.nox/test_examples_by_dependency-extras-sqlite-include-sqlite/lib/python3.8/site-packages/airflow/models/skipmixin.py", line 25, in <module>
    from airflow.serialization.pydantic.dag_run import DagRunPydantic
  File "/home/runner/work/astro-sdk/astro-sdk/python-sdk/.nox/test_examples_by_dependency-extras-sqlite-include-sqlite/lib/python3.8/site-packages/airflow/serialization/pydantic/dag_run.py", line 24, in <module>
    from airflow.serialization.pydantic.dataset import DatasetEventPydantic
  File "/home/runner/work/astro-sdk/astro-sdk/python-sdk/.nox/test_examples_by_dependency-extras-sqlite-include-sqlite/lib/python3.8/site-packages/airflow/serialization/pydantic/dataset.py", line 40, in <module>
    class TaskOutletDatasetReferencePydantic(BaseModelPydantic):
  File "/home/runner/work/astro-sdk/astro-sdk/python-sdk/.nox/test_examples_by_dependency-extras-sqlite-include-sqlite/lib/python3.8/site-packages/pydantic/_internal/_model_construction.py", line 95, in __new__
    private_attributes = inspect_namespace(
  File "/home/runner/work/astro-sdk/astro-sdk/python-sdk/.nox/test_examples_by_dependency-extras-sqlite-include-sqlite/lib/python3.8/site-packages/pydantic/_internal/_model_construction.py", line 328, in inspect_namespace
    raise PydanticUserError(
[...]

PR #1991, opened 2023-08-05
Revert "Limit Pydantic < 2.0.0 until Airflow resolves incompatibilities"

pankajastro:
Reverts astronomer/astro-sdk#1981

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -319,7 +319,7 @@ jobs:
       - run: python -c 'import os; print(os.getenv("GOOGLE_APPLICATION_CREDENTIALS_JSON", "").strip())' > ${{ env.GOOGLE_APPLICATION_CREDENTIALS }}
       - run: sqlite3 /tmp/sqlite_default.db "VACUUM;"
       - run: pip3 install nox
-      - run: nox -s "test-3.8(airflow='2.6.0')" -- tests_integration/ -k "test_load_file.py and not redshift" --splits 3 --group ${{ matrix.group }} --store-durations --durations-path /tmp/durations-${{ matrix.group }} --cov=src --cov-report=xml --cov-branch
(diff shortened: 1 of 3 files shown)


tatiana (approved):
@pankajastro it looks great, thanks!
What do you think of removing the micro version and pinning only on the minor one? This way we can test the latest 2.6 releases

pankajastro:
> What do you think of removing the micro version and pinning only on the minor one? This way we can test the latest 2.6 releases

sorry @tatiana , I'm a bit confused are you talking about pinning a minor version of Airflow not micro for example pin 2.6 and not 2.6.3?

pankajastro:
> > What do you think of removing the micro version and pinning only on the minor one? This way we can test the latest 2.6 releases
> 
> sorry @tatiana , I'm a bit confused are you talking about pinning a minor version of Airflow not micro for example pin 2.6 and not 2.6.3?

I run `pip install apache-airflow==2.6` in a clean Python virtual env and it installed 2.6.0 and not the latest (2.6.3).
I selected the last patch because I feel it most stable for a given minor version

pankajastro:
> > > What do you think of removing the micro version and pinning only on the minor one? This way we can test the latest 2.6 releases
> > 
> > 
> > sorry @tatiana , I'm a bit confused are you talking about pinning a minor version of Airflow not micro for example pin 2.6 and not 2.6.3?
> 
> I run `pip install apache-airflow==2.6` in a clean Python virtual env and it installed 2.6.0 and not the latest (2.6.3). I selected the last patch because I feel it most stable for a given minor version

I tried to use `2.6.*` but in this case, it uses cache if available so maybe it would not install latest until we pass no-cache param

pankajastro:
I address the CI config changes in a separate PR since the original purpose of this PR was to revert https://github.com/astronomer/astro-sdk/pull/1981 and the ci config came into the picture since tests were failing. thank you @tatiana for the clarification.

Issue #1950, opened 2023-05-26
Automate Airflow Providers RC testing with AstroSDK


utkarsharma2:
Similar to what we have in providers can be implemented in SDK. Sync with @Lee-W for more details.

PR #1995, opened 2023-08-08
Add workflow to test Airflow RC

pankajastro:
closes: https://github.com/astronomer/astro-sdk/issues/1950
mostly taken from https://github.com/astronomer/astronomer-providers

### DIFF
--- a/python-sdk/dev/integration_test_scripts/replace_dependencies.py
+++ b/python-sdk/dev/integration_test_scripts/replace_dependencies.py
@@ -31,13 +31,13 @@ def _parse_pypi_url_from_h2_title(h2_soup: BeautifulSoup) -> str:
     return h2_soup.a.get("href")
 
 
-def parse_providers_release_testing_gh_issue(issue_url: str) -> list[str]:
+def parse_providers_release_testing_gh_issue(gh_issue_url: str) -> list[str]:
     """Parse the pinned packages from The URL of the github issue that announce provider testing
     (e.g., https://github.com/apache/airflow/issues/31322)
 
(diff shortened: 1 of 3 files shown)


tatiana (approved):
This will help us avoid issues like we had with the recent MS provider release, thanks @pankajastro !

utkarsharma2 (approved):

Lee-W (approved):
I'm good with it now. But I think we should make it reusable later. Let me see how we can do that. (Also found a typo by me 🫠 but I think that's something we should fix on the provider side when we make it resuable)

python-sdk/dev/integration_test_scripts/replace_dependencies.py
```diff
@@ -0,0 +1,100 @@
...
+        if requirement.specifier:
+            pinned_package = f"{package_name_to_search}{requirement.specifier}"
+        elif requirement.url:
+            pinned_package = f"{package_name_to_search} @{requirement.url}"
+        else:
+            raise Exception(
+                f"Invalid package {package} provided. It needs to be pinned to a specific version."
+            )
+
+        with fileinput.FileInput("pyproject.toml", inplace=True) as setup_file:
```
pankajkoti, line 68: Were we able to test this workflow on a RC url? I am wondering if the path works here or it needs to be `python_sdk/pyproject.toml`

pankajastro: thanks for pointing it out. I have added `working-directory: python-sdk` hopefully should work

PR #2006, opened 2023-08-10
Fix astro deploy job

pankajastro:
currently, Astro deploy job fails with an error mentioned in PR https://github.com/astronomer/astro-sdk/pull/2003


tatiana (approved):
Thanks, @pankajastro , one character can make tons of difference

PR #2008, opened 2023-08-11
use bot's personal access token for pushing rc testing branch and triggering ci checking workflows

Lee-W:
# Description

Even if we trigger the rc testing GitHub actions workflows and it creates a testing branch as expected, it won't trigger ci testing workflows due to https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow. Thus, we'll need to use personal access token for it.

## What is the new behavior?

automatically trigger other ci testing workflows when rc testing branch is triggered

no

### Checklist

### DIFF
--- a/.github/workflows/ci-rc-test.yaml
+++ b/.github/workflows/ci-rc-test.yaml
@@ -119,6 +119,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ inputs.base_git_rev }}
+          token: ${{ secrets.BOT_ACCESS_TOKEN }}
 
       - name: Install dev dependency
         run: |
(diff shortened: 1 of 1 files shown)


Lee-W:
Will need @phanikumv 's help setting up `BOT_TOKEN`

pankajastro (approved):
Thanks, @Lee-W, for looking into it.

tatiana (commented):
Thanks, @Lee-W , I can setup `BOT_ACCESS_TOKEN` - if possible, send it over via DM

Lee-W:
as https://github.com/astronomer/astro-sdk/pull/2009 is merged, I think we can close this one

PR #2009, opened 2023-08-11
reuse reusable workflows for rc testing from astronomer-providers

Lee-W:
# Description

Reuse the workflow from https://github.com/astronomer/astronomer-providers/pull/1295 (Will need to wait for its merge)

## What is the new behavior?

nothing should be changed

no

### Checklist


Lee-W:
In addition to `BOT_ACCESS_TOKEN`, we might need to check the permission of running these workflow in setting as well.

tatiana (approved):
Hey @Lee-W, thanks. This looks much cleaner!

Please let me know if you'd like help setting `BOT_ACCESS_TOKEN` to the repo, I can do this for you (just need you to send me the desired token)

.github/workflows/ci-rc-test.yaml
```diff
@@ -30,192 +30,15 @@ defaults:
...
-          if [ "${{ inputs.rc_testing_branch }}" == "" ]; then
-            rc_testing_branch=${{ needs.create-branch-for-testing-rc-release.outputs.rc_testing_branch }}
-          else
-            rc_testing_branch=${{ inputs.rc_testing_branch }}
-          fi
-          echo "rc_testing_branch=$rc_testing_branch" >> $GITHUB_OUTPUT
-    outputs:
-      rc_testing_branch: ${{ steps.export-rc-testing-branch-name-step.outputs.rc_testing_branch }}
+  check-airflow-provider-rc-release:
+    uses: astronomer/astronomer-providers/.github/workflows/reuse-wf-check-rc-release.yaml@main
```
pankajastro, line 34: wondering if we need to update https://github.com/astronomer/astronomer-providers/blob/main/dev/integration_test_scripts/replace_dependencies.py too since in astronomer-provider we update the dependency in setup.cfg but here we update in pyproject.toml?

Lee-W: nope, it'll still use the dev/integration_test_scripts/replace_dependencies.py in astro-sdk. I think there's some room for deduplicate further. But I'm occupied by other tasks. Maybe we can get back to this later.

pankajastro (approved):

PR #2021, opened 2023-08-26
[DO NOT MERGE] Test RC provider packages for https://github.com/apach…

vatsrahul1001:
[DO NOT MERGE] Test RC provider packages for https://github.com/apache/airflow/issues/33751

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -55,12 +55,12 @@ tests = [
 ]
 google = [
     "protobuf",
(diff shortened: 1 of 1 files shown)


utkarsharma2:
@vatsrahul1001 we also need to update the constraint file.

pankajastro:
> @vatsrahul1001 we also need to update the constraint file.

why constraint file? this PR is to run a test against the new provider rc. we use constraint file only to run test for airflow 2.2.5

Issue #1982, opened 2023-07-08
Include Snowflake staged file metadata in LoadFileOperator


Andrew-Wichmann:
**Please describe the feature you'd like to see**
I would like to be able to include [the Snowflake metadata for staged data](https://docs.snowflake.com/en/user-guide/querying-metadata) in the table that I'm loading data into using the LoadFileOperator.

**Describe the solution you'd like**
```python
    new_table = aql.load_file(
        input_file=File(path="s3://astro-sdk/python_sdk/example_dags/data/sample.csv"),
        output_table=Table(
            conn_id="snowflake_conn",
        ),
        # here, I would like an option to include the METADATA$XYZ columns in the output table
        load_options=[
            SnowflakeMetadataOptions(
                include=["METADATA$FILENAME", "METADATA$FILE_ROW_NUMBER", ...]
            )
        ],
    )
 ```

**Are there any alternatives to this feature?**
I could (and will if instructed to by senior management) use a SnowflakeOperator to perform the `COPY INTO` command exactly how I want it instead of this nice shiny astro-sdk.

**Additional context**
Nope. I'm really enjoying the toolchain so far, and I hope the ease of use continues. Great work.

**Acceptance Criteria**

- [ ] All checks and tests in the CI should pass
- [ ] Unit tests (90% code coverage or more, [once available](https://github.com/astronomer/astro-sdk/issues/191))
- [ ] Integration tests (if the feature relates to a new database or external service)
- [ ] Example DAG
- [ ] Docstrings in [reStructuredText](https://peps.python.org/pep-0287/) for each of  methods, classes, functions and module-level attributes (including Example DAG on how it should be used)
- [ ] Exception handling in case of errors
- [ ] Logging (are we exposing useful information to the user? e.g. source and destination)
- [ ] Improve the documentation (README, Sphinx, and any other relevant)
- [ ] How to use Guide for the feature ([example](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html))

phanikumv:
@pankajkoti adding it to your backlog

pankajkoti:
Okay @phanikumv 👍🏾

pankajkoti:
hi @Andrew-Wichmann 

[...]

PR #2023, opened 2023-08-28
Support loading metadata columns from stage into table for Snowflake

pankajkoti:
Adds support to load metadata columns like METADATA$FILENAME, 
METADATA$FILE_ROW_NUMBER, etc. from stage into target table 
while natively loading files into Snowflake tables. Read more at: 
https://docs.snowflake.com/en/user-guide/querying-metadata#example-3-loading-metadata-columns-into-a-table

Note that, you cannot specify both `validation_mode` and 
`metadata_columns` together in the Snowflake load options 
because when we need to load `metadata_columns`, we need 
to explicitly name the metadata columns in the `COPY INTO` 
sql statement and such a transformed SQL statement does not 
allow specifying `VALIDATION_MODE` with it. It's a limitation 
for Snowflake queries.
The transformed SQL appears like in the snippet in the following link: 
https://docs.snowflake.com/en/user-guide/querying-metadata#example-3-loading-metadata-columns-into-a-table

closes: #1982

### DIFF
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -647,7 +647,7 @@ class SnowflakeDatabase(BaseDatabase):
 
     def _get_copy_into_with_metadata_sql_statement(
         self, file_path: str, target_table: BaseTable, stage: SnowflakeStage
-    ):
+    ) -> str:
         """Return the sql statement for copy into with metadata columns."""
         if self.load_options is None or not self.load_options.metadata_columns:
             raise ValueError("Error: Requires metadata columns to be set in load options")
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -652,10 +652,21 @@ class SnowflakeDatabase(BaseDatabase):
         if self.load_options is None or not self.load_options.metadata_columns:
             raise ValueError("Error: Requires metadata columns to be set in load options")
         table_name = target_table.name.upper()
-        sql_statement = "SELECT count(*) from INFORMATION_SCHEMA.columns where table_name=%(table_name)s;"
-        table_columns_count = self.hook.run(
-            sql_statement, parameters={"table_name": table_name}, handler=lambda cur: cur.fetchone()
-        )[0]
+        sql_statement = (
+            "SELECT count(*) COLUMN_COUNT from INFORMATION_SCHEMA.columns where table_name=%(table_name)s"
+        )
+        try:
+            table_columns_count = self.hook.run(
+                sql_statement, parameters={"table_name": table_name}, handler=lambda cur: cur.fetchone()
+            )[0]
+        except AttributeError:
+            # For apache-airflow-providers-snowflake <3.2.0.
+            try:
+                table_columns_count = self.hook.run(sql_statement, parameters={"table_name": table_name})[0][
+                    "COLUMN_COUNT"
+                ]
+            except ValueError as exe:
+                raise DatabaseCustomError from exe
         non_metadata_columns_count = table_columns_count - len(self.load_options.metadata_columns)
         select_non_metadata_columns = [f"${col}" for col in range(1, non_metadata_columns_count + 1)]
         select_columns_list = select_non_metadata_columns + self.load_options.metadata_columns
--- a/python-sdk/src/astro/databases/snowflake.py
+++ b/python-sdk/src/astro/databases/snowflake.py
@@ -659,7 +659,7 @@ class SnowflakeDatabase(BaseDatabase):
             table_columns_count = self.hook.run(
                 sql_statement, parameters={"table_name": table_name}, handler=lambda cur: cur.fetchone()
             )[0]
-        except AttributeError:
+        except AttributeError:  # pragma: no cover
             # For apache-airflow-providers-snowflake <3.2.0.
             try:
                 table_columns_count = self.hook.run(sql_statement, parameters={"table_name": table_name})[0][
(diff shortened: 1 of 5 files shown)


pankajkoti:
The transformed COPY INTO command executed is like the below when metadata_columns are specified:
```
'COPY INTO <TABLE_NAME> FROM (SELECT $1,$2,METADATA$FILENAME,METADATA$FILE_ROW_NUMBER,METADATA$FILE_CONTENT_KEY,METADATA$FILE_LAST_MODIFIED,METADATA$START_SCAN_TIME FROM @SANDBOX.<schema_name>.<stage_name>/sample.csv) '
```

pankajkoti:
Sample load output with metadata columns

Screenshot 2023-08-28 at 1 09 30 PM

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -460,6 +468,21 @@ def is_native_autodetect_schema_available(  # skipcq: PYL-R0201
...
+            return
+        table_name = self.get_table_qualified_name(table)
+        metadata_columns = ",".join(
+            [
+                "IF NOT EXISTS " + f"{col} {self.METADATA_COLUMNS_DATATYPE[col]}"
+                for col in self.load_options.metadata_columns
+            ]
+        )
+        sql_statement = f"ALTER TABLE {table_name} ADD COLUMN {metadata_columns};"
+        self.hook.run(sql_statement, autocommit=True)
```
utkarsharma2, line 484: @pankajkoti Just curious Why are we altering a table? Do you think we can do the required changes at creation time only? might save us one extra query execution.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -622,13 +645,35 @@ def load_file_to_table_natively(
         )
         self.evaluate_results(rows)
 
+    def _get_copy_into_with_metadata_sql_statement(
+        self, file_path: str, target_table: BaseTable, stage: SnowflakeStage
+    ):
```
utkarsharma2, line 650:
```suggestion
    ) -> str:
```

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -622,13 +645,48 @@ def load_file_to_table_natively(
         )
         self.evaluate_results(rows)
 
+    def _get_copy_into_with_metadata_sql_statement(
```
utkarsharma2, line 648: Not a blocker, but would be nice to split this into multiple functions.

python-sdk/src/astro/databases/snowflake.py
sunank200: Have we tested this for the following scenarios? The table is created using pandas and the same table is used to load metadata. Would it work for this?

python-sdk/src/astro/databases/snowflake.py
sunank200: +1

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -460,6 +468,21 @@ def is_native_autodetect_schema_available(  # skipcq: PYL-R0201
         )
         return is_file_type_supported and is_file_location_supported
 
+    def create_table(self, table: BaseTable, *args, **kwargs):
+        """Override create_table to add metadata columns to the table if specified in load_options"""
```
sunank200, line 472: We should add documentation for loading metadata from Snowflake.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -622,13 +645,48 @@ def load_file_to_table_natively(
...
+        table_name = target_table.name.upper()
+        sql_statement = (
+            "SELECT count(*) COLUMN_COUNT from INFORMATION_SCHEMA.columns where table_name=%(table_name)s"
+        )
+        try:
+            table_columns_count = self.hook.run(
+                sql_statement, parameters={"table_name": table_name}, handler=lambda cur: cur.fetchone()
+            )[0]
+        except AttributeError:  # pragma: no cover
+            # For apache-airflow-providers-snowflake <3.2.0.
```
sunank200, line 658: Any reason why it's less than 3.2.0? We should document this.

python-sdk/src/astro/databases/snowflake.py
```diff
@@ -622,13 +645,53 @@ def load_file_to_table_natively(
...
+                raise DatabaseCustomError from exe
+        return table_columns_count
+
+    def _get_copy_into_with_metadata_sql_statement(
+        self, file_path: str, target_table: BaseTable, stage: SnowflakeStage
+    ) -> str:
+        """Return the sql statement for copy into with metadata columns."""
+        if self.load_options is None or not self.load_options.metadata_columns:
+            raise ValueError("Error: Requires metadata columns to be set in load options")
+        table_name = target_table.name.upper()
```
sunank200, line 673: Why are we changing this to upper case?

pankajkoti: Snowflake tables have upper case names.

python-sdk/src/astro/databases/snowflake.py
pankajkoti: Yes, tested this. This alter statement is after the create table, so this is applied to both whether the table is created by pandas or via infer schema and hence we are able to load metadata as those columns are created.

python-sdk/src/astro/databases/snowflake.py
pankajkoti: We have added it https://github.com/astronomer/astro-sdk/pull/2023/files#diff-6f3049267c42087321bf6e68905f412cc9ccac24a2efdbd0fda0e4967e625d76R74

python-sdk/src/astro/databases/snowflake.py
pankajkoti: It the same reason as is for other places in Snowflake in the same file. Also added a detailed note here too.

sunank200 (approved):

Andrew-Wichmann:
Nice! Thank you!

PR #2034, opened 2023-09-20
Enable GH actions job for daily deploy and Master DAG Trigger

vatsrahul1001:
Currently we are not running Astro SDK Example DAG daily, This PR enable Actions job to deploy everyday and trigger example Master Dag for celery and Kubernetes deployment


Lee-W (approved):
Overall, LGTM. we probably would want to extract the common logic somewhere (in another repo for astronomer-providers, astro-sdk and other project to reference?), but I think that can be improvements for the next time

PR #2043, opened 2023-09-27
ci(scripts): databricks script for bring up and terminate cluster

Lee-W:
# Description
## What is the current behavior?

run databricks testing on an existing cluster

## What is the new behavior?

- bring up a databricks cluster
- run the databricks tests
- terminate cluster

## Does this introduce a breaking change?
no

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #2044, opened 2023-10-04
test(snowflake): add missing ORDER in column default

Lee-W:
# Description

`python-sdk/tests_integration/databases/test_snowflake.py::test_snowflake_create_table_with_columns` fails

During testing on https://github.com/astronomer/astro-sdk/pull/2043, I found out there's this snowflake failure in integration test

## What is the new behavior?

- `python-sdk/tests_integration/databases/test_snowflake.py::test_snowflake_create_table_with_columns` passes

no

### Checklist

### DIFF
--- a/python-sdk/tests_integration/databases/test_snowflake.py
+++ b/python-sdk/tests_integration/databases/test_snowflake.py
@@ -87,7 +87,7 @@ def test_snowflake_create_table_with_columns(database_table_fixture):
         "NUMBER(38,0)",
         "COLUMN",
         "N",
-        "IDENTITY START 1 INCREMENT 1",
+        "IDENTITY START 1 INCREMENT 1 ORDER",
         "Y",
         "N",
         None,


tatiana (approved):
@Lee-W, thanks for working on this!

Do you have any idea why "Build and test astro Python SDK / Build-Docs" is breaking? The change is not related to this change, so I'm approving it. It would be great for us to address that in a separate PR

Lee-W:
> @Lee-W, thanks for working on this!
> 
> Do you have any idea why "Build and test astro Python SDK / Build-Docs" is breaking? The change is not related to this change, so I'm approving it. It would be great for us to address that in a separate PR

Not quite sure 🤔 Let me rerun it and see how it works

Lee-W:
seems to fail. might need to handle that in another pr

tatiana:
@Lee-W it seems there is a conflict ATM - when you have a chance, please, take a look, so we can merge this

Lee-W:
> @Lee-W it seems there is a conflict ATM - when you have a chance, please, take a look, so we can merge this

I just rebased from main and found out it's no longer needed

PR #2060, opened 2023-12-09
Pin databricks provider

pankajastro:


pankajastro:
closing since 5.0.1 databricks provider has yanked

Issue #2070, opened 2023-12-21
investigate astro-sdk tests


pankajastro:
see https://astronomer.slack.com/archives/C04Q58877PV/p1703168223011419

pankajastro:
maybe duplicate of https://github.com/astronomer/astro-sdk/issues/2068

Issue #2068, opened 2023-12-18
Debug failure reason for s3_to_databricks DAG


vatsrahul1001:
Currently, our 's3_to_databricks' DAG is failing in the integration test. We should investigate this issue, as frequent failures may result in the leakage of a valid bug

[Deployment link](https://cll0nj92h00iu01j51htnafwh.astronomer-stage.run/dkvggim6/dags/example_load_file/grid?search=example_load_file&dag_run_id=manual__2023-12-18T06%3A39%3A48.226103%2B00%3A00&task_id=s3_to_databricks&tab=logs)

**Logs**

[2023-12-18, 07:09:33 UTC] {base.py:83} INFO - Using connection ID 'aws_conn' for task execution.
[2023-12-18, 07:09:34 UTC] {adapter.py:122} WARNING - Failed to emit OpenLineage event of id ebf1b197-3ca4-3186-bc1e-1c5c56202b89
[2023-12-18, 07:09:34 UTC] {taskinstance.py:2699} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 433, in _execute_task
    result = execute_callable(context=context, **execute_callable_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/astro/sql/operators/load_file.py", line 131, in execute
    return self.load_data(input_file=self.input_file, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/astro/sql/operators/load_file.py", line 136, in load_data
    return self.load_data_to_table(input_file, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/astro/sql/operators/load_file.py", line 155, in load_data_to_table
    database.load_file_to_table(
  File "/usr/local/lib/python3.11/site-packages/astro/databases/databricks/delta.py", line 151, in load_file_to_table
    load_file_to_delta(
  File "/usr/local/lib/python3.11/site-packages/astro/databases/databricks/load_file/load_file_job.py", line 88, in load_file_to_delta
[...]

PR #2071, opened 2023-12-21
Set load_secret True in in Databricks S3 load file DAG

pankajastro:
closes: https://github.com/astronomer/astro-sdk/issues/2070
closes: https://github.com/astronomer/astro-sdk/issues/2068

The default value for load_secret is false 
and because of that Databricks cluster was 
not able to access the AWS S3 file. 
I have changed that to True in this PR and 
the Airflow task is green now.
but not sure if this is because of some recent changes in Databricks.

### DIFF
--- a/python-sdk/example_dags/example_load_file.py
+++ b/python-sdk/example_dags/example_load_file.py
@@ -339,6 +339,7 @@ with dag:
             DeltaLoadOptions(
                 copy_into_format_options={"header": "true", "inferSchema": "true"},
                 copy_into_copy_options={"mergeSchema": "true"},
+                load_secrets=True,
             )
         ],
     )


utkarsharma2 (approved):
LGTM

pankajkoti (approved):

python-sdk/example_dags/example_load_file.py
```diff
@@ -339,6 +339,7 @@
             DeltaLoadOptions(
                 copy_into_format_options={"header": "true", "inferSchema": "true"},
                 copy_into_copy_options={"mergeSchema": "true"},
+                load_secrets=True,
```
phanikumv, line 342: why do we need to set this to `True`, does this mean users need to make this change for their DAGs to work?
Somehow I dont feel comfortable with this solution.

pankajastro: Since the S3 file is located remotely for Databricks, it needs a way to read it by setting load_secrets 
to true we are setting secret in Databricks secret store so that it can read S3 file. https://docs.databricks.com/en/connect/storage/amazon-s3.html#global-configuration

pankajkoti: The issue here is that Spark configuration for the compute job is missing credentials to access the S3 bucket and hence it gives a forbidden error when it tries to access the file. load_secrets is a programmatic way to provide S3 access creds from Astro SDK to the spark configuration. If the user has already established auth within Databricks for S3 access they would not need to set load_secrets. There would be a way to establish a relationship from Databricks compute to AWS S3, but this is a research activity and I think we can create a follow-up ticket to explor that infra setting.

phanikumv: Okay, nice.  This explanation should have been mentioned in the PR description for reviewers to understand the context better

phanikumv (approved):

Issue #2075, opened 2023-12-22
Check why latest runtime is not deployed on Astro 


pankajastro:
Job: https://github.com/astronomer/astro-sdk/actions/runs/7297552472
history run: https://github.com/astronomer/astro-sdk/actions/workflows/ci-astro-deploy.yml

Job to update runtime image tag: https://github.com/astronomer/astro-sdk/actions/workflows/runtime-image-update.yaml

PR #2080, opened 2023-12-22
Fix job to update the runtime tag in dockerfile

pankajastro:
closes: https://github.com/astronomer/astro-sdk/issues/2075
Sample PR: https://github.com/astronomer/astro-sdk/pull/2081

This PR fixes the CI job to update runtime image tag in dockerfiles

### DIFF
--- a/.github/scripts/get_latest_runtime_image_tag.py
+++ b/.github/scripts/get_latest_runtime_image_tag.py
@@ -1,0 +1,38 @@
+"""This script fetches the latest runtime image tag from the provided Quay.io repository URL."""
+from __future__ import annotations
+
+import sys
+
+import requests
+from semantic_version import Version
+
+
+def get_latest_tag(repository: str) -> str:
+    """Get the latest semantic version tag from a Quay.io repository."""
+    url = f"https://quay.io/api/v1/repository/{repository}/tag/"
+    response = requests.get(url)
+    response.raise_for_status()
+    data = response.json()
+    tags = data["tags"]
+    valid_tags = []
+    for tag in tags:
+        try:
+            if not tag["name"].endswith("-base"):
+                continue
+            version = Version(tag["name"])
+            valid_tags.append(version)
+        except ValueError:
+            continue
+    if valid_tags:
+        _latest_tag = max(valid_tags)
+        return str(_latest_tag)
+    else:
+        sys.exit("No valid semantic version tags found.")
+
+
+if __name__ == "__main__":
+    _repository = "astronomer/astro-runtime"
+    if len(sys.argv) == 2 and sys.argv[1]:
+        _repository = sys.argv[1]
+    latest_tag = get_latest_tag(repository=_repository)
+    print(latest_tag)
--- a/.github/workflows/runtime-image-update.yaml
+++ b/.github/workflows/runtime-image-update.yaml
@@ -42,7 +42,7 @@ jobs:
         id: get_latest_tag
         # yamllint disable rule:line-length
         run: |
-          LATEST_TAG=$(python ../.github/scripts/get_latest_runtime_image_tag.py ${{ github.event.inputs.quay_repo }})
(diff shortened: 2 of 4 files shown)


.github/scripts/get_latest_runtime_image_tag.py
```diff
@@ -0,0 +1,38 @@
...
+    for tag in tags:
+        try:
+            if not tag["name"].endswith("-base"):
+                continue
+            version = Version(tag["name"])
+            valid_tags.append(version)
+        except ValueError:
+            continue
+    if valid_tags:
+        _latest_tag = max(valid_tags)
```
utkarsharma2, line 27: I think max() will fail to compare versions like `1.2.2` < `2.1.1`. How about using `distutils.version.StrictVersion` to get the max version instead?

pankajkoti: This won't fail. Do you have different example in mind that could fail?
Screenshot 2023-12-26 at 12 57 01 PM

pankajkoti: Here, we already convert it to Version object https://github.com/astronomer/astro-sdk/pull/2080/files#diff-4381c7dd83cf01b553505437a870a4fb0da096c814049616aaf8b8d2750d4c06R22

python-sdk/tests_integration/astro_deploy/Dockerfile.single_worker
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/astronomer/astro-runtime:8.1.0-base
+FROM quay.io/astronomer/astro-runtime:10.0.0-base
```
pankajkoti, line 1: Did we manually update this? 
I see the version was updated for the other Dockerfile in PR https://github.com/astronomer/astro-sdk/pull/2081/, but not for this file. Are we missing something?

.github/scripts/get_latest_runtime_image_tag.py
pankajastro: right this file is from astronomer-provider repo

.github/scripts/get_latest_runtime_image_tag.py
```diff
@@ -0,0 +1,38 @@
...
+        return str(_latest_tag)
+    else:
+        sys.exit("No valid semantic version tags found.")
```
Lee-W, line 30:
```suggestion

    sys.exit("No valid semantic version tags found.")
```

python-sdk/tests_integration/astro_deploy/Dockerfile.single_worker
pankajastro: the PR https://github.com/astronomer/astro-sdk/pull/2081 was created by this job but it ran only on schedule intervals I updated it to run on push to test the changes. This line in the current PR is a manual change so that all docker file will point to same version once the PR will merge and going forward the job will create PR

.github/workflows/runtime-image-update.yaml
```diff
@@ -39,7 +42,7 @@ jobs:
         id: get_latest_tag
         # yamllint disable rule:line-length
         run: |
-          LATEST_TAG=$(python ./.github/scripts/get_latest_runtime_image_tag.py ${{ github.event.inputs.quay_repo }})
+          LATEST_TAG=$(python ../.github/scripts/get_latest_runtime_image_tag.py ${{ github.event.inputs.quay_repo }})
```
Lee-W, line 45: Look like we only use this file in python-sdk. Should we move it there?

python-sdk/tests_integration/astro_deploy/Dockerfile.single_worker
pankajkoti: okay, should we remove this manual change and re-run the job once to verify it works fine? Will leave it to you to decide.

pankajkoti (approved):
LGTM

.github/workflows/runtime-image-update.yaml
pankajastro: moved it at `python-sdk/dev/integration_test_scripts`

PR #2093, opened 2023-12-26
Fix Generate-Constraints job

pankajastro:
Recently, I upgraded the Airflow version in CI https://github.com/astronomer/astro-sdk/pull/2084 after that the Generate-Constraints job failed. This PR will restrict to generate of the constraints for the min and max versions of Python and Airflow version which OSS support

### DIFF
--- a/.github/workflows/ci-python-sdk.yaml
+++ b/.github/workflows/ci-python-sdk.yaml
@@ -626,8 +626,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: [ '3.8', '3.9', '3.10' '3.11']
-        airflow: [ '2.6', '2.7', '2.8']
+        python: [ '3.8', '3.9', '3.10' '3.11' ]
+        airflow: [ '2.6', '2.7', '2.8' ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-python-sdk.yaml
```diff
@@ -626,8 +626,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: [ '3.7', '3.8', '3.9', '3.10' ]
-        airflow: [ '2.2.5', '2.3', '2.4', '2.5', '2.6', '2.7', '2.8']
+        python: [ '3.8', '3.9', '3.10', '3.11' ]
+        airflow: [ '2.6', '2.7', '2.8' ]
```
pankajkoti, line 630: In the upcoming Astro SDK release should we bump the min version for the Airflow dependency here https://github.com/astronomer/astro-sdk/blob/main/python-sdk/pyproject.toml#L20 as we would not test lower versions of Airflow?

pankajastro: Yes, I agree that we should have continuously upgraded the Python and Airflow versions. However, this may require some discussion with @phanikumv. I have similar thoughts in the provider repo https://github.com/astronomer/astronomer-providers/issues/775 here I was thinking restrict so that CI will generate Constraints for these versions and not fail

pankajkoti: IMO, it could be reasonably fine to bump minor versions of depedencies in minor releases of Astro SDK. Yes, let's wait for Phani and team's opinion on this.

pankajastro:
not require since we have updated job

PR #2103, opened 2024-01-24
Use `apache-airflow-providers-openlineage`, bump min Airflow version to 2.7 and Python version to 3.8

pankajastro:
Fixes issue reported where deployments don't contain the `openlineage-airflow`
library installed but instead contain `apache-airflow-providers-openlineage`.
To fix the issue we transform the usages from `openlineage-airflow` to 
corresponding available source code in the `apache-airflow-providers-openlineage`
Airflow OSS provider.

Since the apache-airflow-providers-openlineage depends on Airflow >= 2.7,
we also correspondingly bump the Airflow version.
Airflow 2.7.0 is also the first release that drops support for end-of-life Python 3.7,
and hence, we bump up the min Python version to 3.8

related: https://github.com/astronomer/issues-airflow/issues/503

### DIFF
--- a/python-sdk/pyproject.toml
+++ b/python-sdk/pyproject.toml
@@ -87,7 +87,7 @@ ftp = [
     "apache-airflow-providers-ftp>=3.0.0",
     "smart-open>=5.2.1",
 ]
-openlineage = ["openlineage-airflow>=0.17.0"]
+openlineage = ["apache-airflow-providers-openlineage>=1.4.0"]
 
 databricks = [
     "databricks-cli",
@@ -125,7 +125,7 @@ all = [
     "databricks-sql-connector<2.9.0",
     "s3fs",
     "protobuf",
-    "openlineage-airflow>=0.17.0",
+    "apache-airflow-providers-openlineage>=1.4.0",
     "apache-airflow-providers-microsoft-azure",
     "azure-storage-blob",
     "apache-airflow-providers-microsoft-mssql>=3.2",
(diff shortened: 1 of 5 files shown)


python-sdk/pyproject.toml
```diff
@@ -87,7 +87,7 @@ ftp = [
     "apache-airflow-providers-ftp>=3.0.0",
     "smart-open>=5.2.1",
 ]
-openlineage = ["openlineage-airflow>=0.17.0"]
+openlineage = ["apache-airflow-providers-openlineage>=1.4.0"]
```
JDarDagran, line 90: Note that the OpenLineage provider only supports Airflow versions greater than or equal to 2.7.0.
For compatibility with older Airflow versions, wouldn't there be dependency conflict on strict requirement with [openlineage] extra?

pankajkoti: yes, thanks a lot @JDarDagran for the suggesiton

Bumping up min Airflow version to 2.7 and min Python version to 3.8

pankajastro: most of the airflow provider has `apache-airflow>=2.6.0` as a dependency. hopping bumping 2.7.0  will not impact users

pankajkoti: We will release a minor version of Astro SDK and it should not cause a problem IMHO. If users would like to use the new minor version, they would need to have Airflow 2.7. For earlier versions, they can continue using prior versions of Astro SDK.

Similar fashion is followed in Airflow OSS providers when bumping versions for dependencies

pankajkoti: Besides we cannot allow usage of `apache-airflow-providers-openlineage` without bumping Airflow to 2.7

pankajkoti (approved):

phanikumv (approved):

PR #2105, opened 2024-01-24
Fix CI failures

pankajkoti:
Fix CI failures.

Limit pandas to <2.2.0 as it breaks our tests. 2.2.0 was released on Jan 20, 2024 and there is an open issue reported for this https://github.com/pandas-dev/pandas/issues/57053. We are getting the same error as reported in the issue.


pankajkoti:
All CI failures that were observed in #2103 have been fixed here except for 1 failure in CI. The 1 failure we're observing is wrt to Bigquery failures wherein we're being rate limited due to multiple CI runs today.

Screenshot 2024-01-24 at 7 39 48 PM

Thanks to @utkarsharma2 for helping with fixing other CI failures :)

utkarsharma2 (approved):
LGTM

PR #2114, opened 2024-01-25
Use apache-airflow-providers-openlineage in debug log

pankajastro:
since now import from `apache-airflow-providers-openlineage` so let's update the log accordingly

### DIFF
--- a/python-sdk/src/astro/lineage/__init__.py
+++ b/python-sdk/src/astro/lineage/__init__.py
@@ -16,4 +16,4 @@ try:
     )
     from openlineage.client.run import Dataset as OpenlineageDataset
 except ImportError:
-    logging.debug("openlineage-airflow python dependency is missing")
+    logging.debug("apache-airflow-providers-openlineage python dependency is missing")


pankajkoti (commented):
I am thinking here to check if openlineage is enabled then we could log this error with Logging level ERROR. Otherwise this debug log won't be much helpful. WDYT??

python-sdk/src/astro/lineage/__init__.py
```diff
@@ -1,7 +1,25 @@
...
+    return (
+        conf.getboolean("openlineage", "disabled", fallback=False)
+        or os.getenv("OPENLINEAGE_DISABLED", "false").lower() == "true"
+        or (
+            conf.get("openlineage", "transport", fallback="") == ""
+            and conf.get("openlineage", "config_path", fallback="") == ""
+            and os.getenv("OPENLINEAGE_URL", "") == ""
+            and os.getenv("OPENLINEAGE_CONFIG", "") == ""
+        )
+    )
```
pankajastro, line 19: This code is copy from https://github.com/apache/airflow/blob/390eacb01a710239e69f71d0c52134b7e87d17d4/airflow/providers/openlineage/plugins/openlineage.py#L27-L37

pankajastro:
> I am thinking here to check if openlineage is enabled then we could log this error with Logging level ERROR. Otherwise this debug log won't be much helpful. WDYT??

Added a check PTAL

PR #2123, opened 2024-02-27
Bump upstream provider versions

pankajastro:
Bump the upstream provider version to make it consistent with astronomer-providers dependency


tatiana (commented):
@pankajastro It seems this broke a few tests - are you looking into this?

pankajastro:
> @pankajastro It seems this broke a few tests - are you looking into this?

Not actively, But sometime next week will try to fix it

vatsrahul1001 (approved):

PR #2126, opened 2024-03-07
Fix missing regex var allowed_deserialization_classes

pankajastro:


tatiana (approved):
Thanks for fixing this, @pankajastro !

PR #2130, opened 2024-03-21
Disable daily run of master dag for integration tests

pankajkoti:
Since we're not actively developing here, it makes sense to turn off 
the daily schedule of the master DAG (integration tests) to save up 
on the cloud resources that get spunned off from the master DAG.

We do have example DAG run in CI for each new PR, and additionally 
mechanism to tests RCs to guarantee catching of regressions.


tatiana (approved):
It makes sense, thanks for improving this, @pankajkoti !

PR #2146, opened 2024-05-02
Remove condition import for SQLAlcRow

pankajastro:
since airflow is pin to >2.7.0 remove conditional import 
https://github.com/astronomer/astro-sdk/blob/main/python-sdk/pyproject.toml#L20

### DIFF
--- a/python-sdk/src/astro/sql/operators/raw_sql.py
+++ b/python-sdk/src/astro/sql/operators/raw_sql.py
@@ -9,15 +9,11 @@ try:
 except ImportError:  # pragma: no cover
     from airflow.decorators import _TaskDecorator as TaskDecorator  # type: ignore[attr-defined]
 
-import airflow
 import pandas as pd
 from airflow.decorators.base import task_decorator_factory
 from sqlalchemy.engine import ResultProxy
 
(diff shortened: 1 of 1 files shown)


Lee-W:
Looks like the CI is failing. Just rerun it. Let's see how it works

pankajkoti (approved):
LGTM. Some tests seems to be failing. Should we fix them in this PR or a subsequent PR is better? Leaving the call to you to decide @pankajastro

Lee-W:
> Looks like the CI is failing. Just rerun it. Let's see how it works

It still fails. I think we might need to check it 🤔

pankajastro:
> It still fails. I think we might need to check it 🤔

Yes, it is broken since 2.9 I'm fixing in https://github.com/astronomer/astro-sdk/pull/2149

pankajastro:
> GTM. Some tests seems to be failing. Should we fix them in this PR or a subsequent PR is better? Leaving the call to you to decide @pankajastro

yeah, the failure is unrelated to this change, I'll merge this one to fix the import error in example DAG and will fix CI in https://github.com/astronomer/astro-sdk/pull/2149

PR #2155, opened 2024-05-07
Fix pre-commit: UP031 Use format specifiers instead of percent format

pankajastro:
Fix pre-commit failure post merge of PR https://github.com/astronomer/astro-sdk/pull/2137


pankajastro:
depend on https://github.com/astronomer/astro-sdk/pull/2149

